PR #505, opened 2015-09-27
Fix GBC movies showing as incorrect platform

adituv:
Previously, GBC movies showed their SystemId as "GB" and were
highlighted as a platform mismatch because of it.  As the emulator
infrastucture lumps GB and GBC together everywhere, this is just a GUI
hack to change the displayed platform to "GB(C)" and not highlight.

Unfortunately, Gambatte's reported system id often differs to the game
database: for both GBC and color-enabled GB games, it reports "GB"
while the database reports "GBC"; the database therefore cannot be
used for this without significant changes throughout the emulator.

Such changes may be a better future fix to look into, if desired.


adituv:
I think I'm going to redo this properly, taking on the Gambatte change and its knock-on effects.  As such, I'm closing this pull request for now and will reopen it when it's redone.

PR #524, opened 2015-10-18
External custom gametool feature

Hathor86:
As I introduced in [this](http://tasvideos.org/forum/viewtopic.php?t=17257) topic, here's changes I made in order to call a tool from a dll compiled outside of the Bizhawk project.

Hope it'll be ok for you.


Hathor86:
As discussed on IRC with adelikat, it works but it has to be improved (not limited to a specific game for example).
I'll work on it

PR #560, opened 2016-01-24
General improvements to the C64 Core.

SaxxonPike:
- Convert internal math to use `int` for performance and to reduce casts.
- Restructure folders to make more sense.
- Add support for digitized sound using $D418.
- Fix right column visual bug.
- This PR breaks C64 savestate compatibility.


SaxxonPike:
Will move these to a branch that doesn't have continuous commits. (closing)

vadosnaprimer:
Hmmm, why are others' commits also there? Maybe you forgot to rebase? Or it was **because** of rebase

PR #569, opened 2016-02-09
Feature/cheat comparison types

AdamSturge:
The feature branch adds the ability to specify a comparison type when setting a compare value. Previously the only option was to compare through strict equality. Now users can specify ==, > , >=, < ,<=, !=


AdamSturge:
I found a bug to do with empty comparison types. Will fix and recreate pull request

PR #678, opened 2016-08-08
Add address mask setting for breakpoints

parasyte:


vadosnaprimer:
Just wow! (haven't checked yet) Any work on debugger is greatly appreciated!

parasyte:
I'm no stranger to [debuggers](http://www.fceux.com/web/version.html). :)

> In 2002, Parasyte modified the then-current version (0.81.3) of FCE Ultra and added a Nesten-style debugger, along with several other features, and named it "FCEUD" (FCE Ultra Debugger).

vadosnaprimer:
> FCEUD

Well, I think it was the only fceu with a debugger for quite some time, does it mean that the same tool was basically merged into fceux in the end?

parasyte:
Yes, much of the original code that I wrote is essentially untouched in FCEUX. They have made some significant modifications, but I still see plenty of legacy bad decisions I made that remain to this day. ☺️

vadosnaprimer:
Yeah, regardless, fceux has that kind of a debugger that can be considered ideal for most cases, especially seeing how other emus use to get it wrong (unless one is Nemesis).

Anyway, I hope to see more of your fixes, and some day we'll fix breakpoint implementation to be aware what kind of a breakpoint it is, then some serious work could be started.

Recently, we added debugger/tracer to gpgx, the core that doesn't disassemble! Thankfully, the unfinished Genesis core by vecna had all the disasm features, and gpgx had read/write/execute callbacks, so we simply tied those together and magically got it to work. So basically, any external disasm for a core that can't do it can be added.

All in all, the tool just needs experienced users that know what to add to it.

parasyte:
I have mostly been using the debugger with the SNES core, and have encountered some annoyances along the way. (I'll document them here for giggles and either submit further PRs, or create some tickets so they can be tracked for fixing by someone else later on.)

* This is probably the easiest of the batch to fix; The "To PC" button doesn't seem to do anything useful, it just refreshes the data in the disassembler (if it's currently pointing at a RAM address).
* The "Seek to Address" feature relies on breakpoints. This combined with the above issue makes it a real pain to disassemble from arbitrary addresses.
* Scrolling in the disassembler is uncomfortable, providing the ability to only scroll one line at a time, with a little bit of wiggle room (about 4 lines). It attempts to make intelligent guesses about instruction length while scrolling, but that leads to the total inability to disassemble from an address that falls in the middle of an instruction, according to the disassembler. And that is exacerbated by: 
* In classic style, 65c816 uses variable-length instructions, and takes it even a step further by making some instruction lengths variable *based on processor flags at runtime*. This complicates static disassembly, since you will rarely get correct results when `sep` and `rep` instructions are present. A tracing disassembler, or human-assisted "flag hints" per instruction would save the day.
* Breakpoints only work under some currently-unknown conditions. When multiple breakpoints are enabled and we expect them all to be encountered within a single frame, the debugger pauses only on one of them. (Presumably on the first or last encountered.) I have *not* looked at any of the code, but based on the behavior, I suspect that the core is not capable of pausing on any given instruction, and instead receives a pause request from the debugger at some point in time long after the event has been triggered.
* Stepping single instructions is not supported by the core. If the above assumption is correct, that explains why this feature is unavailable.

So far, I've found that the trace logger is a little bit friendlier to use (when writing to a file); it provides the kind of information I'm interested in, and it's not possible to completely lose "breakpoint hits" with it (the instruction and data accesses are guaranteed to be in the log somewhere... it's just a matter of knowing what to look for). Also, none of these annoyances are insurmountable. It just needs some TLC.

vadosnaprimer:
I think it needs to work with code-data-logger to mark code/data that would allow scrolling independently. Anything that hasn't yet been executed will just look like bytes (or shorts for 68k for example).

Issue #690, opened 2016-08-27
Why in the world are you using IPC in the SNES core?


meepdarknessmeep:
Why are you even using two threads? You are literally making the threads do nothing but idle the times where the other thread is doing stuff. You don't need two threads.

SwadicalRag:
:+1:

Velkon:
![yesplz](https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png)

oplexz:
👍

Techbot121:
![bigthumb](http://techbot.xyz/ShareX/screenshots/2016/08/27/asdf.png)

vadosnaprimer:
Sorry Mario, Byuu is in the other castle.

zeromus:
libco and .net do not play well together. They're fundamentally incompatible, because .net depends on a proper win32 callstack to fire GC any time C++ calls back into .net, and libco tears up the stack. We tried fixing libco to reconstruct the stack well enough for .net, but failed.

Our bsnes fork calls into .net from c++ for two main reasons: 1) for callbacks such as acquiring input and delivering audio samples (could be worked around by re-engineering insides to buffer whatever's needed); and 2) for lua and trace events (can't be worked around, in principle). Since reason (2) can't be worked around, we didn't bother with reason (1).

We originally solved this whole problem by putting bsnes in _another process entirely_ so that .net and c++ never existed in the same stack. Thus the IPC mechanism was needed.

Later we realized we could move bsnes back into the main bizhawk process if we were very careful to make sure that .net and c++ code never touched the same stack. This entailed using the bulky IPC framework we already had to marshal the calls between c# and c++ without actually directly calling on a foreign stack.  

The performance benefits of moving bsnes back into emuhawk.exe are essentially only a reduction of the bulky IPC overhead.  This would not really be greatly appreciated except in abstruse scenarios. I don't know how much improvement it made for casual gaming, but it certainly did not make anything worse. 

It's possible that the current mechanisms for running the IPC (there are two; if you've used bizhawk more than 3 minutes you may have found the ring buffer option) are not as efficient as could be given the new reality of it running as a thread in the bizhawk process. We haven't re-assessed that since the process->thread move.  

But that finally gets me to the answer to your question. The threads idle because it increased the performance of the communication between the processes, which was running with such direly bad performance that no cost (a wasted CPU core) was too high for any tiny speedup through reduced synchronization overhead.


[...]

PR #692, opened 2016-08-29
rework snes core to not use pipes

meepdarknessmeep:
references: #690


meepdarknessmeep:
just a head up - don't merge this until it's ready - it's only for people to look into for bugs and such

meepdarknessmeep:
@vadosnaprimer @zeromus I have pretty much optimized this all I could and made it work perfectly - any thoughts, suggestions, worries?

vadosnaprimer:
@parasyte Can you try this out on the matter of tracing/debugging?

parasyte:
First, this is a *really nice* PR! The IPC stuff was really a pain to deal with when trying to make sense of the code.

The patch looks very reasonable from a cursory glance. It's probably a lot faster when the debugger is enabled, too (I haven't verified this assumption).

I'll check it out on my Windows machine in a bit and report any feedback here.

zeromus:
I have some nitpicks of varying size but nothing worth blocking a PR with this much work done in it and nothing that cant be cleaned up later.

parasyte:
Performance is roughly the same with and without this patch; CDL, TraceLogger, and Graphics Debugger all have pretty much the same stuttering problem when running.

Letting the TraceLogger run for 5-15 seconds causes a crash. The crash is 100% reproducible, but VS2015 debugger isn't able to handle it:

```
FatalExecutionEngineError occurred
Message: Managed Debugging Assistant 'FatalExecutionEngineError' has detected a problem in 'C:\Users\parasyte\Documents\other-projects\BizHawk\output\EmuHawk.exe'.
Additional information: The runtime has encountered a fatal error. The address of the error was at 0x61dd7594, on thread 0x214c. The error code is 0xc0000005. This error may be a bug in the CLR or in the unsafe or non-verifiable portions of user code. Common sources of this bug include user marshaling errors for COM-interop or PInvoke, which may corrupt the stack.
```

```
System.ExecutionEngineException was unhandled
Message: An unhandled exception of type 'System.ExecutionEngineException' occurred in Unknown Module.
```

Sometimes there are other crashes just by opening the TraceLogger window, and once the VS debugger showed me that a crash occurred here: https://github.com/TASVideos/BizHawk/blob/144f08cfc3fb1df5a5f6301acfad7ab2b69a786d/BizHawk.Client.EmuHawk/Input/Input.cs#L381 (heh, "analyze moose" ...)

zeromus:
I'm having a bit of a hard time combing through this code since it was not generally done in the simplest way possible, but a couple of points.

1. these are typical of the crashes you get due to libco+.net
2. The code relevant to sending the trace calls back to the main thread is missing. The managed code is called from one of bsnes's coroutines. This is precisely the problem we were originally trying to solve. You've purged too much code.

parasyte:
Speaking of libco ... https://github.com/TASVideos/BizHawk/blob/37e6f683703e1a3a2ea6e936f2d3b4a3e1b518dd/libsnes/bsnes/libco/amd64.c#L22-L36

dude-wtf

zeromus:
I feel like I saw the code for that once somewhere. Ours could be kind of old. But it's a novelly-disturbed creative solution to portable assembly....

meepdarknessmeep:
> these are typical of the crashes you get due to libco+.net
ah i see that it uses it inside the core code as well, i could probably fix this with creating a csharp coroutine and switching from/to it on entry/exit of vital points in the wrapper.

And yes, that's inline assembly opcodes for each individual architecture it supports.

meepdarknessmeep:
Actually, does libco store the stack and recreate it on reactivation? or does it intertwine stacks?

zeromus:
It looks like CMD_run is setup just about right for what you need. You simply need to handle a ton of other "exit reasons". Previously it only handled exit reasons known to bsnes, and we had other "exit reasons" (for instance, opcode trace) which were managed through another system (that you deleted). You could extend the bsnes exit reason to have 'custom' and set another custom type elsewhere as a kind of simple way to start rebuilding the old system that was deleted. But that's just a guess. If I wanted to think this through completely, I would just do it.

I don't think libco does either of those things. Instead it has a stack for each cothread (a completely new stack for each one, other than the 'main' cothread) and as part of restoring the state for each cothread, it restores the relevant registers for establishing its stack. The magic part of libco is in setting up the cothreads and their stacks, really. There isn't much to it though.

Parasyte, you shouldnt notice much performance improvement over the last release unless youre doing a heavily-loaded lua script.  Tracing ought to be faster, if it wasnt crashing. For ordinary gaming I can't believe there's any speedup to any of this --there wasnt THAT MUCH traffic over the IPC system.

meepdarknessmeep:
zeromus is right about the performance. It shouldn't be a huge speedup in regular code, all that would be optimized with regular gaming is for cpu usage

i have an idea i am looking into that should satisfy .net

parasyte:
> Parasyte, you shouldnt notice much performance improvement over the last release unless youre doing a heavily-loaded lua script. Tracing ought to be faster, if it wasnt crashing. For ordinary gaming I can't believe there's any speedup to any of this --there wasnt THAT MUCH traffic over the IPC system.

I'm aware. I was testing it with the Code-Data Logger, Trace Logger, and SNES Graphics Debugger. Each of which have a lot of IPC traffic.

zeromus:
I'm working on this now: fbd7601b773f6b2c4ed1ab67b42aafbe0da86766

General gameplay performance is not yet improved (due I think to a problem with a change I had to make to libco) but the code is considerably less byzantine and it's far easier to see where inefficiencies would be (if there are any) -- and the IPC overhead is vanished.

QUERY functions are still 'marshaled' (through a much slimmer mechanism) solely for orthogonality with the other type of messages. That should not be costing a noticeable amount.

meepdarknessmeep:
Glad to hear! The main reason I did this in the first place was because the lua api took so long to do anything since of all this back-and-forth calls. I am going to close this now.

parasyte:
@zeromus :+1:

vadosnaprimer:
@parasyte should I read this as "all tests I ran do indeed work"? :)

In case you missed my IRC message, I can elaborate.

parasyte:
@vadosnaprimer I haven't had a chance to try it yet. I meant to respond back to you on IRC, but it was late and you were already offline.

```
[06:43:10]  <feos>	hi. we need some more of your tests here https://github.com/TASVideos/BizHawk/commits/master
[06:43:54]  <feos>	zero applied the proper version of that no-pipe pr and fixed what was broken due to its former approach
[06:45:17]  <feos>	I will test a bunch of options regarding it, but I might miss the erroneous behavior if it results in tiny errors here and there
[06:45:41]  <feos>	since I never worked with snes stuff
[22:55:17]  <Parasyte>	woah cool
[22:55:40]  <Parasyte>	I'll check this out, probably tomorrow
[22:56:01]  <Parasyte>	There are a few things I can play with right away. Will let you know.
```

The last SNES game I remember trying to debug with BizHawk, I was planning to use lua to draw all of the collision boxes. I suspect these changes probably won't resolve the issues that prevented progress. But this sounds like a good start. FWIW, here's the list of issues I faced at the time: https://github.com/TASVideos/BizHawk/pull/678#issuecomment-239086515

In any case, will definitely check it out!

vadosnaprimer:
@zeromus it all seems to be working, aside from some minor bugs that might've been there beforehand too. If something's still wrong, I think users will find it, otherwise looks good.

PR #693, opened 2016-08-29
Arcade pit

ampillion:
[NesCarts.zip](https://github.com/TASVideos/BizHawk/files/443218/NesCarts.zip)

Minor tweak that adds 3-D World Runner to the list of games needing to have a forced core.


AdmiralCurtiss:
Wrong repository, man!

adelikat:
This is the right repo...

However, there's somethings I don't want to pull from this, and some things I do

AdmiralCurtiss:
This is a branch for our weekly gameshow on Twitch to make it go smoother. @ampillion was supposed to make a PR to that branch, not to PR the branch to this repo. If you want to cherry-pick some of those commits (or tell me which parts you're interested in) feel free to do so, but be wary that some of it is a bit hacky to fit our specific needs!

adelikat:
Yeah, there's some stuff in here I want to take.  I think I will just manually port the relevant changesets manually.  Thanks for PRing the work you did :)

zeromus:
* Implement forced core gamedb setting. (for Snes9x only right now)
* Set a few games to use the Snes9x core.
* Built-in Snes9x has problems, load libretro dll instead.
* Fix Nullref exception when loading SNES ROMs not in database. 
* Assume Libretro is a SNES
This is going to require much more detailed design and elaborate implementation. it relates intimately with bizhawk core mission. But this helps us see how you'd like to use bizhawk, for sure.

* Libretro core used should not be part of the game name.
* Libretro saves/states should not have a subdirectory, but instead handle similar to the multiple-cores-for-same-system name logic. This is so we can have Snes9x and BSNES saves in the same directory.
* Autoloaded Libretro games should keep their canonical names.
I'll have to analyze all this libretro user experience stuff independently. Some of this may relate to the libretro-masquerading-as-a-first-class-snes9x-bizhawk-core stuff

* Let the title bar show that we're a ArcadePit specific build. 
Irrelevant

* Keep aspect ratio even when forcing a window size. (sorta hacky but whatever) 
Unacceptable breakage to design. You can get what you want by manually sizing window, or editing ini file. 
I get that this might be annoying; alternative would be a popup box to change window to demanded client size, or another disp manager option to change the definition of 1x. 
We need to think about this some more or hear your thoughts.

* Let user change the string in the emulator title bar via a file.
Decent idea. Decent implementation. I may need to implement it with a different file path. Sorry, if I change this it will conflict with your patch.

* Extract NesCarts.xml for easier modification.
This is a decent idea. You made us rethink how we do this. We're gonna change it so we commit the xml, and then archive it during dist process. 
We get better git diffs then, and you get easier editing
However, you won't find any use in this, because...

* Allow forcing core on NES games via the NesCarts.xml
We intended gamedb to be used for this. We're going to make sure gamedb can do that (for neshawk/quicknes anyway)
* Allow forcing of NES games before they reach QuickNes. 
Oh, it looks like you did that. I'll either take this patch or rework it if needed.

* Option to enable/disable the core forcing. 
Good idea. Need to think about UI (I didn't inspect it in detail but it was slim enough I'll probably take as-is for now)

* Disable rewind OSD messages. 
Obviously we can't take this, but we'll register this as feedback that we need granularity over how much noise people see.
I have a concept of keying OSD messages with an enum, and having an UI to disable them individually.
Additionally, we can tag the keys to let them be toggled by group. This is later work.

* Add the Buck Punch-Out hack to the gamedb. 
I think this needs to be labeled as a hack. Otherwise it's OK. We'll give you a section.

-----games----
* Force NesHawk for California Games
Why? We should incorporate that into quicknes blacklist
* Force 3D World Runner to NesHawk
Why?
* Add the Buck Punch-Out hack to the gamedb. 
Why does this one need forced neshawk? Do you know something about punch-out in quicknes that we don't?

AdmiralCurtiss:
Re games:

- California Games has major graphical bugs in QuickNes in the BMX event.
- In 3D World Runner the floor graphics mess up in QuickNes when approaching pits.
- The Punch-Out hack probably does not need NesHawk, but we already had savestates for NesHawk so it was easier to just keep them.

AdmiralCurtiss:
Re resolution/aspect ratio, due to the way we run ArcadePit we needed an option that keeps the emulator client size always the same while switching between systems, so you can set up OBS to capture that client size once and not mess with it when switching to, for example, GBA. The previous options always either resized the client window, or resized the game to a non-native aspect ratio, so I quickly hack-converted one of the settings to not mess with the aspect ratio. Obviously you should not do that in an actual BizHawk release build, but I would appreciate an official 'fixed client size without aspect ratio changes' option.

AdmiralCurtiss:
You should probably close this so you don't get notified when I rebase changes on top of the newest release version.

zeromus:
Will reopen if there's any more useful contributions

PR #731, opened 2016-11-19
m64p: Use built-in debugger

Wyst3r:
The current N64 breakpoint implementation in BizHawk causes major slowdowns whenever a breakpoint has been set. The cause of this is the fact that a callback accross the managed/native boundary is being done on every instruction, which for N64 is simply too much. To fix this issue, I've modified the code to instead use the mupen64plus lib's built-in debugger features. 

I had hoped to include features such as mid-frame pausing and step-into in this PR. However, my final tests revealed some fundamental issues with my approach to pausing, and it's now clear that some bigger modifications to StepRunLoop_Core()/FrameAdvance() is required for it to work in all corner cases. The changes in this patch therefor doesn't aim to change any functionality, but instead simply improve performance by using native debugging.

There's one feature that is no longer supported though,, address masks. The reason for this is that there's no simple way to implement this in the native code. You'd have to find a way to check if any address within an arbitary address range matches the breakpoint address when masked. I haven't managed to find a formula for this, despite many attempts. It might be possible to brute force it by looping through all addresses, but I don't know what the performance impact would be. This implementation simply assumes a mask of 0xFFFFFFFF.

Otherwise this should work very similarly  to the current implementation, and even fix some issues I noticed with the old breakpoint system (some read/write breakpoints failed to trigger). Some bugs in the old implementation remain in this one, mainly related to Lua (don't try to load/save states while in a breakpoint callback!). I might look into the mid-frame pausing stuff again in the near future, but if so, it'll be in a separate PR.


vadosnaprimer:
> some fundamental issues with my approach to pausing

Which ones?

Wyst3r:
Well the main issue comes from the fact that you need to return control to the GUI when hitting a breakpoint, i.e. you need to return from the frame_advance() function. This means that BizHawk will think the frame has completed, even though it's not, and update stuff that shouldn't be updated (Frame count, AVI dump, Lua callbacks etc...). There's also the issue of how to set the emulator to "paused" when hitting a Lua breakpoint, since the GUI classes have no knowledge of this event (a callback from MemoryCallbackSystem could solve that, but it's a hacky solution). And then you have to handle save/load state while in mid-frame paused states, which mupen64plus doesn't like very much (at least, you can't be paused mid-frame when loading a state. Saving seems to work). There might be other cases i haven't considered as well.

The easiest solution is definately to fix the main loop.

vadosnaprimer:
But how does A2600 debugger handle step into? Doesn't it keep the core paused and the client somewhat running?

Wyst3r:
Does that core actually pause itself when hitting a breakpoint though? From what i can tell, FrameAdvance() seems to always finish the frame?

vadosnaprimer:
I think it doesn't. I hoped its Step Into approach could be reapplied, why not?

Wyst3r:
Well maybe it could, but then you could only step from the start of the frame, which is kind of useless for N64, unless you want to step through 100000 instructions manually :p

vadosnaprimer:
What I was asking is how does it pause the core at particular point and keep it paused, while the client is available for some actions. I guess **pausing** the core somewhere might be the same, but what it does when you **unpause** (or resume emulation for a known (one step into) or unknown (till the next breakpoint) amount of instructions) is what makes it not so helpful for us?

Anyway, as I said earlier, I could try looking into all of this myself when I'm not so busy.

Wyst3r:
GUI control is not a problem when you only support StepInto from frame start. You just press the Step button, execute one instruction, then return control.

As for the frame handling, it does basically the same thing that I had in mind for the main loop, it divides it into before/after functions, which are only triggered on the first/last instruction of the frame (except, it only does so within the emulator class).  So when you press StepInto the first time, it will do the Before stuff, then execute one CPU cycle, then return to the GUI. If you later unpause, then the FrameAdvance() function will skip the before step, and simply execute instructions until the frame end condition is fulfilled, and then call the after function. This way, you don't run into the issues of setting/updating variables too early or too many times.

zeromus:
Oh my god, we've discussed this a billion times.
Bizhawk CANNOT debug properly until a MASSIVE rewrite happens.
You CANNOT fix it.

vadosnaprimer:
This PR is not aiming to "fix" the "hacky" debugger, it just speeds up the callbacks.

vadosnaprimer:
@Wyst3r can you resolve the conflict? I'm about to merge it.

Wyst3r:
Oh sorry, haven't checked this thread in a while. The conflict should be resolved now at least.

vadosnaprimer:
@Wyst3r does it build for you? I think appveyor was giving a false success notice.

PR #792, opened 2017-02-05
Added togglepause to the list of lua commands

chozandrias76:
@zeromus I probably don't know what I am doing here but, I would like to be able to toggle pausing so that I can pause the emulator after I am done running my script and before I exit out of it. This way if I run my script for 1000 iterations, I don't have to sit around to see when the frame is that it stops or be required to set marker and hope the emulator doesn't crash by the time I get back to it.


zeromus:
client.togglepause

PR #793, opened 2017-02-07
Added the ability to toggle recording mode in LUA

chozandrias76:


chozandrias76:
This PR is not in a mergeable state yet. I am able to set recording mode once but, after that I mess up the state of recording mode. I think I am making a shallow copy of the recording mode or don't update the global state properly. I am gonna need some help fixing this problem probably.

vadosnaprimer:
It shouldn't be done like this.

Since https://github.com/TASVideos/BizHawk/commit/230dcc2356f202ed8f87c4c6dbf3356605b1f6b4 you can ask movie.getreadonly() as well as movie.mode() whether tastudio is recordiong or not.

And to set its recording mode we could use movie.setreadonly(), which can be made to actually toggle the mode if a new value is coming in. But it should be discussed with @adelikat how to do it, as movie lua lib is unaware of MainForm (where ToggleReadOnly() is located).

This has been discussed yesterday with zeromus:
https://github.com/TASVideos/BizHawk/commit/d87a2b27a25d4258d4b124a2e46de191baaeec62#commitcomment-20748189

vadosnaprimer:
We figured out the proper way to do this, so I'm closing this, as I'll implement it soon (after the release though).

PR #880, opened 2017-06-14
Possible fix for Mugen Senshi Valis

alyosha-tas:
Motivation:

This game is failing on a frame where screen size as specified by VDC exceeds the physical screen size. As such, the VEC automatically clamps it with a VSYNC signal.

The game expects to see a VBlank interrupt without the VBlank flag being set. I believe this is occuring when the VEC throws the VDC out of its normal execution, so I delay the flag being set by 1 scanline in this case.

This is purely speculative, I have no reason to believe it is true aside from the fact that it works. It also doesn't fix or break any other relevent PCE issues (i.e. sound in 4-in-1.)

There doesn't seem to be any other obvious fix for this game, but this does feel a bit hackish.

Thoughts?


zeromus:
I don't know anything about PCE. I suggest you ask vecna or debug it in pcejin to see what's happening.

alyosha-tas:
Nevermind, I debugged it in Mednafen, it's not even close to what's happening. Looks like a pretty bad tiing bug.

PR #882, opened 2017-06-15
LuaCanvas: Add drawing features

Trivial-Man:
Brings canvas up to feature parity with EmuLuaLibrary.Gui.

The features added here are slightly altered versions of those already in EmuLuaLibrary.Gui. The only difference in behavior should be that when a file is not found using DrawImage or DrawIcon no error message is printed.


zeromus:
OK, I havent tested this, but ill take responsibility for it if there's problems.
Can you please consider making an example lua script to test all the canvas drawing?

Trivial-Man:
You can find a script to test nearly every canvas drawing feature here:
https://github.com/Trivial-Man/LuaCanvasDrawingExamples

zeromus:
Please work this out with @TASEditor instead of doing the same thing slightly differently

PR #883, opened 2017-06-16
Canvas: Cleanup

TASeditor:
- DefaultTextBackground implemented
- Default Foreground Color changed to black (for a white canvas better)
- Aligment for DrawText implemented
- Default Coloring for DrawArc and DrawRectangle
- DrawString implemented


Trivial-Man:
As long as you're cleaning up my mess, I accidentally left a reference to the gui version of the ClearImageCache function in the method attributes when I copied it. That should be fixed.
Also in the LuaMethodAttributes many of the function names (that I copied) begin with a lowercase letter when it should be uppercase. DrawPixel for example is "drawPixel" in there. And the setters for foreground and background colors are completely misnamed in there.

Thank you so much for this. Sorry I was too hasty in getting features in to check them thoroughly.

PR #887, opened 2017-06-24
Waterbox2: add bios loading option for Gambatte

alyosha-tas:
both GB and GBC

@nattthebear : I'm not sure if/where/how this interacts with anything you are doing with Pizza or anything else waterbox related, let me know if something needs to change.


nattthebear:
This has no effect one way or the other on Pizza; I have no immediate plans to release Pizza, nor to waterbox Gambatte, nor to make any other modifications to Gambatte.  If the issues outlined above are resolved, I would like to see this merged.

alyosha-tas:
Oh yeah, for some reason VS2015 is saying that max is not a member of std, but fmax is. I couldn't find any way to fix it, am I missing something simple?

nattthebear:
No idea.  Reference says you need #include <algorithm> for max.  If VS15 needs fmax, I'm OK with leaving that as is.

Warepire:
Calling `delete[]` on a nullptr is fine, unless you went and replaced the delete-operator with some dumb code.

Also [std::max](http://en.cppreference.com/w/cpp/algorithm/max) is in the `<algorithm>` header, you may need to define `NOMINMAX` in the project settings for this to work. (disables microsofts macros)

alyosha-tas:
Ok, I addressed everything mentioned and also fixed broken savestates.
I tested out with TAStudio and everything seems fine.

PR #909, opened 2017-07-08
case sensitivity on linux

morningpee:
These are some things I had to change when getting BizHawk to compile on Linux. It shouldn't affect anything for case-insensitive volumes.


Sappharad:
You don't want master, it doesn't have any of the other changes you need.

This guy already took care of the changes needed to run on Linux:
https://github.com/FelipeLopes/BizHawk/tree/mono-portable

morningpee:
@Sappharad, thank you for that link!

PR #937, opened 2017-07-24
Add header and add list of features

superusercode:
The header makes this document compatible with cmark. The list of features has not been updated since 2015, and because it is so short I have moved it onto the README to help it gain attention.


zeromus:
The features list is pathetic. We should delete it. The "supported systems" list isn't even up to date. We'll discuss this internally and decide which information we should have, whether it should be updated, and where we should put it. I don't think we should be putting it here, duplicate copies is how things get up to date. I agree that "not updated since 2015" is a good rationale for something being copyable; however, it's out of date in any event. It's an equally good rationale for deleting it in this case.

PR #964, opened 2017-08-18
Timed SaveRAM flush to disk

shanehoman:
Following on the menu item and hotkey to save SaveRAM, I created a menu to choose how often you want your save to be saved during play.  Manually flushing SaveRAM resets the counter, automatic saving in the core frame loop is done after the movie is handled that frame.

Fun fact: a 1 frame flush timer is workable on a RAM disk, depending on the speed of various computer subsystems.  Writing 60 files per second to flash (SSD, SD, etc) lags the SNES core a little bit, and could be mildly abusive of the storage hardware.  Trying it on a hard drive was not fun.

### DIFF
--- a/BizHawk.Client.Common/PathManager.cs
+++ b/BizHawk.Client.Common/PathManager.cs
@@ -266,6 +266,12 @@ namespace BizHawk.Client.Common
 
 			return Path.Combine(MakeAbsolutePath(pathEntry.Path, game.System), name) + ".SaveRAM";
 		}
+		
+		public static string AutoSaveRamPath(GameInfo game)
+		{
+			var path = SaveRamPath(game);
+			return path.Insert(path.Length - 8, ".AutoSaveRAM");
+		}
 
 		public static string RetroSaveRAMDirectory(GameInfo game)
 		{
--- a/BizHawk.Client.Common/config/Config.cs
+++ b/BizHawk.Client.Common/config/Config.cs
@@ -134,7 +134,10 @@ namespace BizHawk.Client.Common
 		public string Update_IgnoreVersion = "";
 		public bool CDLAutoSave = true, CDLAutoStart = true;
 
-		public int FlushSaveRamFrames = 0;
+		private int _flushSaveRAMFrames;
+		public int FlushSaveRamFrames { get { return _flushSaveRAMFrames; } set { _flushSaveRAMFrames = value; } }
+		private bool _autosaveSaveRAM;
+		public bool AutosaveSaveRAM { get { return _autosaveSaveRAM; } set { _autosaveSaveRAM = value; } }
 		//check CurrentDomain_AssemblyResolve if you change the defaults or name of this key
 		public bool UseNLua = true; // Whether or not to use a good, reliable, memory-leak-free lua interface that is slower than the original luainterface
 
--- a/BizHawk.Client.Common/config/Config.cs
+++ b/BizHawk.Client.Common/config/Config.cs
@@ -114,6 +114,8 @@ namespace BizHawk.Client.Common
 		public bool ShowLogWindow = false;
 		public bool BackupSavestates = true;
 		public bool BackupSaveram = true;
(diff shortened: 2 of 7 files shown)


zeromus:
Please explain why you think this is a valuable feature

shanehoman:
1. Computer crashes mid session
2. Automatically updating out of emulator HUDs, for example https://zarby89.github.io/ZeldaHud/ for Legend of Zelda: A Link to the Past.

Snes9x also has this feature, but I just don't like using that emulator.  Adding it to EmuHawk would add the backup feature to multiple systems as well, not just SNES emulation

zeromus:
This is a bad feature. If the computer crashes mid-session, then your save data can be corrupted, in case the timed autosave happened while the game was in the middle of saving. This can only be prevented by adding additional code which detects when the game is no longer writing to the save memory--which isnt practical for every platform, and isnt practical for bizhawk.

Anyone promoting this as backup security hasn't analyzed the situation correctly.

As far as the HUD goes, this can be fixed possibly by using a lua script that exports the save ram periodically. 

I'm stubborn about this because eventually I'm going to remove auto-flushing from the emulator completely and put the user in charge of it. 

If you put your options instead in a new tab page in "customization options" called "Save RAM", move the advanced "backup saveram" option in there, and add a new option for "flush saveram on rom close" which can be used in combination with your option, then we might have something we can move forward with.

shanehoman:
As bizhawk is, saveram only gets flushed to disk when the ROM is closed (or trigger with the hotkey/lua/menu).  If anything crashes, computer, or emulation core, then you lose everything since the rom was opened.  Its not backup security, just autosaving the saveRAM.  I agree that it would be impractical to make sure that bizhawk isn't saving in the middle of the rom's sram write routine, this should probably save to a secondary {rom}.autosave.SaveRAM file.

Savestates are a pain playing with a controller on a long cord that doesn't have extra buttons to map to save.  Having a separate autosave of SaveRAM might actually help players that accidentally load an old save state, as well.

Not everyone writes Lua, I actually wrote this in C# so I didn't have to deal with it any more after the first script.

I can move the option into a configuration dialog next to the existing backup saveram option.

zeromus:
I know. If a timed autosave happens when something crashes then you may get a CORRUPTED save, not just an old one.

Your naming concept to prevent that is a good one. of course, the user experience from yours or snes9x's point of view will be bad. That {rom}.autosave.SaveRAM file won't get used. It's going to only be there as a (possibly corrupted) backup in case something crashes. Which is fine with me.

There are better ways of helping people that accidentally load an old savestate but it will be irrelevant once I make my change to default to no automatic flushing.

shanehoman:
bizhawk could check for an autosave saveram when you load a game, alongside the regular saveram, and if the date is newer than the latest real save offer to load the autosave instead, similar to office autorecovery.  Either way its better than playing for hours and then crashing and saving nothing the whole time.

> There are better ways of helping people that accidentally load an old savestate but it will be irrelevant once I make my change to default to no automatic flushing.

Curious about the first part (save the current state to a backup file when you load state would be good, too, and more "live") but can you elaborate more on the second? By no automatic flushing, do you mean to add "Save your SRAM? Yes/No/Cancel" dialogs like most other applications use?

zeromus:
1. yes, we could do that. I prefer to fall back on a clear manual recovery process though. think carefully about how we name things so we never lose anything, and the user can sort it out himself. I'm calling timed autosave an advanced feature (it will not be enabled by default). An advanced feature for an advanced user--advanced user, if this is for data recovery then let the user recover it through detailed manual steps.

2. I will add that dialog, but it will have a checkbox to never see it again. But yes, `editing a document` is the model I want to move towards. It's stupid for us to have this strange unusual (traditional) way of managing save data for games which does not conform to the way business applications work, or even how GAMES work in most cases.  However I'm grateful to you for making me realize we should have busines applicationish "autorecovery" etc. type stuff running also to bail people out in emergencies.

PR #987, opened 2017-08-31
Adds Sega Master System paddle support

### LINKED ISSUE #950: [request] SMSHawk: paddle controller emulation
Some SMS games (like Alex Kidd BMX Trial) require paddle controller and cannot be played without it.

Tastyfish:
To switch controllers, go to _SMS_ -> _Controller Type_ -> _Paddle_.
Uses left analog stick on a game pad, by default; can be remapped of course.
I would not recommend the alternative digital left/right inputs, although they are technically usable in Alex Kidd. Afterall, the games are designed around fine-grained & quick analog control.

## Tested On
* Alex Kidd BMX Trial
* Galactic Protector
* Megumi Rescue

## Weird Notes
It works on both _Export_ and _Japan_ regions. Being noted because for some reason the game code for virtually every paddle game detects it and the controller is expected to have a different I/O interface, despite it only ever being sold in Japan.
Hopefully there isn't some weird game that throws a wrench in the works regarding that.

I wanted to also add light gun support too, but it appears the following frame is just solid gray instead on the light gun flash frame, so there's something odd happening with the video code there..

Resolves #950

### DIFF
--- a/Assets/defctrl.json
+++ b/Assets/defctrl.json
@@ -743,6 +743,13 @@
       "P2 B1": "",
       "P2 B2": ""
     },
+    "SMS Paddle Controller": {
+      "P1 Left": "LeftArrow, J1 POV1L",
+      "P1 Right": "RightArrow, J1 POV1R",
+      "P1 B1": "Z, J1 B1, X1 X",
+      "Reset": "J1 B9, X1 Back",
+      "Pause": "J1 B10, X1 Start",
+    },
     "GG Controller": {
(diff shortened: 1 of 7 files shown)


alyosha-tas:
Looks good to me, nice work on this!

Tastyfish:
Delayed merge because technically Player 2's paddle is not working correctly, but most of these games don't even support 2 player, and it's possible that Galactic Defender is just doing something non-standard in that mode, so I'll just merge this now and make another PR if I figure that out.

alyosha-tas:
Ok cool, I'll test things out on my end too.

PR #989, opened 2017-09-01
Fixed SpeedThrottle()'s (relatively) high CPU usage

danebou:
SpeedThrottle() was using a large portion of computation (~10 -15% of the entire program). When the sleep time is less than zero, the while true loop tags over. For some reason this appears to be occuring for more than a single. The solution is to always sleep for at least 1 ms if throttling is needed.

(For future) Look into using Threading.Timer. Might provide a better overall architecture.

### DIFF
--- a/BizHawk.Client.EmuHawk/Throttle.cs
+++ b/BizHawk.Client.EmuHawk/Throttle.cs
@@ -338,28 +338,21 @@ namespace BizHawk.Client.EmuHawk
 				}
 
 				int sleepTime = (int)((timePerFrame - elapsedTime) * 1000 / afsfreq);
-				if (sleepTime >= 2 || paused)
-				{
-#if WINDOWS
-					// Assuming a timer period of 1 ms (i.e. timeBeginPeriod(1)): The actual sleep time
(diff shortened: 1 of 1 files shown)


zeromus:
no, the solution is to find out why it's malfunctioning. if the sleep time is less than zero, then `if (elapsedTime >= timePerFrame)` will trigger next time through. That's the clear intent. Why isn't it happening?

danebou:
I agree that also needs to be fixed. But even if it's not malfunctioning, we should always have it sleep to avoid high usage. The while(true) loop is dangerous to performance (without sleeping).

zeromus:
It is not dangerous to performance. It is only dangerous to performance if it is stealing time needed for emulating. If the frontend is actually being throttled, it doesn't need to be emulating.

zeromus:
There's a bunch of wrecked optimization flags, I'm going to analyze them all.

danebou:
Alright. I'm going to close this PR and reopen it as an issue

danebou:
#989

danebou:
#989

PR #1020, opened 2017-10-13
DiscSystem - CDFS classes now consume CD-I discs

Asnivor:
(Green Book)

* Volume descriptors and root tree are both parsed
* Uses existing ISO9660 class structures (with a few modifications)


Asnivor:
The most lurid example I could find:

image

Asnivor:
Last update commit is a selfish one tbh. I don't want to have to traverse the directory tree in my client app to get all the files.

PR #1033, opened 2017-10-23
Retroizer Shader + Time uniform in custom shaders

peeweek:
Integrates a old and used CRT Shader using scanline, distortion, ghosting, bloom and animated off-sync scanline effect similar to NES Classic Mini.

base repo was here (https://github.com/peeweek/bizhawk-shaders)

Added also Time uniform to the pipeline for the animated scanline off sync.

rockman2
sf2
kururin

### DIFF
--- a/Assets/Shaders/retroizer/retroizer.cg
+++ b/Assets/Shaders/retroizer/retroizer.cg
@@ -1,0 +1,205 @@
+
+////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////
+////
+////    VERTEX SHADER : STOLEN FROM NEIGHBOURING SHADERS
+////
+////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////
+
+struct input
+{
+   float2 video_size;
+   float2 texture_size;
+   float2 output_size;
+   
+};
+
+void main_vertex
+(
+   float4 position	: POSITION,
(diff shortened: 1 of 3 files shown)


zeromus:
So, just so you know, the shader system needs a lot of help. Most of retroarch's shaders can't survive the CG conversion (theyre illegal CG files or noncomformant to the retroarch specs in ways that break the CG files even worse, or CG generates nonsense output) so we need to follow their approach of using GLSL and HLSL shaders

peeweek:
Thanks, indeed if the CG cross-compiler does a bad work it should be more affordable to write all the HLSL/GLSL code by hand to ensure everything works well. Do bizhawk follow all these specs : https://github.com/libretro/common-shaders or are there some missing implementation atm?

zeromus:
Some functionalities are missing from the "specifications"

We dont support glslp/glsl files. But....... we need to. 

Since you're around as a power user, maybe I can prioritize fixing this. But it will be at least a week before I can even really think about it.  I don't think it will be incredibly hard.

One thing we'd really need a power user for is combing through the shaders and curating which set works on bizhawk, if only a subset does. Other than that, just testing them and figuring out whats going on will be a big help.

zeromus:
well, im worried someone is going to get epilepsy while watching this, but i salute you for making the shader you wanted work in bizhawk. feel free to work on any others that excite you

Issue #1030, opened 2017-10-20
[Bug][Saturn]  Blazing Heroes,  BizHawk thinks it is a PCE game... :/


MelchiorGaspar:
aka "Mystaria  --  The Realms of Lore (U)"
in bin/cue format.. fails to load..
BizHawk thinks it is a PCE game..
and wants the PC-Engine BIOS file.. which I do not have..
I have the Sega Saturn BIOS file....
noy sure what is going on..

alyosha-tas:
Try the Dev Build. We just had a pretty big PR merged involving CD identification, so maybe this is fixed there.

Asnivor:
Still won't work in the current dev build. That particular image is prevalent on the internet and is missing the requisite Saturn identifier string. In fact it contains very little of the sega header that should be expected:

![image](https://user-images.githubusercontent.com/7911038/32052524-6bb0efe6-ba50-11e7-9bdd-624f5597316b.png)

It is possible that whoever dumped that version managed to zero out the first 80 bytes at LBA0. Why or how, I have no idea.

The redump version *should* have the header intact:

http://redump.org/disc/18148/

I'll know for sure when archive.org is finished pretending to be hosted on a 56k modem.

MelchiorGaspar:
ooh well that explains that lol ;-p

Asnivor:
Regardless, the last CD-id PR should make PC-Engine CDs detectable, so I might take a look at how the RomLoader is working (as it appears that it is defaulting to PCE when the disc is not detected).

Unless i'm mistaken, it would be nice if it gave you the option to choose a system if the disc is undetected. Although I'm not a bizhawk dev, so there might be a very good reason for it not working in this way...

MelchiorGaspar:
hmm.. yeah oh well I should have my original copy that worked... before I bothered.. with this new copy...
meh.. oh we;; its not I was probably ever gonna play it again anyway lol :p ;-)

Asnivor:
Yup (as nearly always) the redump version is fine:

![image](https://user-images.githubusercontent.com/7911038/32054216-9e9c3c58-ba55-11e7-9d69-e6c4f736cf81.png)

MelchiorGaspar:
ahh.. normally an emu just does it buy extension doesn't it?
apparently not.. lol ;-)

alyosha-tas:
Resolved in 11b6bb6

PR #1037, opened 2017-10-26
RomLoader - better disc load failures

Asnivor:
* RomLoader no longer returns type PCE when DiscType is unknown
* PCECD is explicitly detected now
* New 'NoAvailableCore' exception is thrown when a system is detected that bizhawk cannot emulate

Issue #1030


Asnivor:
AutoBuild failure was due to previous failure from this: https://github.com/TASVideos/BizHawk/commit/7b8a61b8e4d25a4be039aca316e3671c67afe22b

Should build OK now.

alyosha-tas:
My bad breaking the autobuild!
Thanks for the nice work cleaning up disc ID stuff though.

PR #1052, opened 2017-11-10
Add display type column (resolves #683)

### LINKED ISSUE #683: RAM watch : A "Size" columns
Currently there is only one way(to my knowledge) to check the size of an address: Editing the address. It would be neat if the size was already displayed in a column like the Value, Address,Notes, Prev or the Changes columns.

YoshiRulz:


alyosha-tas:
Tested and works well. Pretty nice convenience.

PR #1065, opened 2017-11-23
Fixes debugger tool for SMSHawk

Tastyfish:
The Z80 disassembler reported all of the instructions as 0 bytes long so it would just print the first instruction down the entire window.

Also, breakpoints weren't setup to work at all in `SMSHawk`.


Tastyfish:
Also just realized that the `SP` in the debugger was actually `IY`.

PR #1066, opened 2017-11-24
Keyboard support for SG-1000/Mark III

### LINKED ISSUE #1059: SG-1000: Implement SK-1100 keyboard
Needed for some games.

I'm not sure how best to do this one. 

It might be worth it to seperate SG-1000 into it's own core so this can be integrated more naturally. It would be easier to do this after the graphics and sound chips are made into independent components though.

Tastyfish:
I couldn't find a game that used more than 10 keys or so, since most just use _the arrow keys_, _Insert_, and _Home_, basically pretending they make up a controller, so I can't guarantee that all the keys actually work.

SG-1000 games tested:
- Girl's Garden
- Sega-Galaga
- Champion Boxing
  - Actually uses more keys if you select 2 player!

Mark III Games tested:
- F-16 Fighting Falcon
  - I barely have any clue what's happening in this game, but I managed to shoot at a plane so I assume it all works.
- Alex Kidd no Miracle World*
  - This one is weird, it detects the keyboard, and you can start the game with it, but once it launches into the first level, it zero-outs the flags byte that includes the detected controllers and just forcibly uses the basic controller in-game.

Resolves #1059

### DIFF
--- a/Assets/defctrl.json
+++ b/Assets/defctrl.json
@@ -833,8 +833,8 @@
       "Key PI": "RightShift",
       "Key Down Arrow": "DownArrow",
       
-      "Key Graph": "LeftControl",
-      "Key Kana": "LeftAlt",
+      "Key Graph": "PageUp",
+      "Key Kana": "PageDown",
       "Key Space": "Space",
       "Key Home/Clear": "Home",
       "Key Insert/Delete": "Insert",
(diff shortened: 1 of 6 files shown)


alyosha-tas:
Awesome!
Tested Champion Tennis (which originally prompted this request) and it works fine.

Tastyfish:
Thanks, I wasn't sure which game it was that prompted it.

PR #1073, opened 2017-12-03
Update Vic.VideoProvider.cs

Mad-Matty:
Updated C64 Colour Palette Values using the new (Feb 2017) Pepto Colodore values.  This is an update to Pepto's original palette currently in use.

Values taken from Winvice Colodore.vpl) Note that the vpl also has Dither values which I guess are not used by this emu.

http://www.pepto.de/projects/colorvic/


vadosnaprimer:
Can you post some comparison pics for the curious?

Mad-Matty:
bizhawkc64

A quick comparison at the Basic screen

Mad-Matty:
bizhawkc64-batmantitle

Batman Title Screen.  Another option would be to have selectable palette files like there is for Gameboy and nes. Compatible with Winvice palette files.

alyosha-tas:
It looks good to me. I mean, I've never even seen a Commodore 64 in real life but the research on that linked page seemed pretty thorough.

It looks like this is all for PAL. Is NTSC different at all? Probably being able to pick a palette will be a good idea at some point, but this looks like a good update for now.

vadosnaprimer:
Wow, this new palette is SO much better! I don't think anyone will ever want the shitty old colors as an option.

PR #1098, opened 2018-01-19
tastudio lua: setinput functions, onbranch callbacks

TASeditor:
lua functions for setting bool and float states in tastudio.
I'm not sure if it wouldn't be better to use the logic TAS Editor uses.

### DIFF
--- a/BizHawk.Client.Common/movie/tasproj/TasStateManager.cs
+++ b/BizHawk.Client.Common/movie/tasproj/TasStateManager.cs
@@ -26,27 +26,6 @@ namespace BizHawk.Client.Common
 			InvalidateCallback?.Invoke(index);
 		}
 
-		public Action<int> LoadCallback { get; set; }
-
-		private void CallLoadCallback(int index)
-		{
-			LoadCallback?.Invoke(index);
-		}
-
-		public Action<int> SaveCallback { get; set; }
-
-		private void CallSaveCallback(int index)
-		{
-			SaveCallback?.Invoke(index);
-		}
-
-		public Action<int> RemoveCallback { get; set; }
-
-		private void CallRemoveCallback(int index)
-		{
-			RemoveCallback?.Invoke(index);
-		}
-
 		private readonly List<StateManagerState> _lowPriorityStates = new List<StateManagerState>();
 		internal NDBDatabase NdbDatabase { get; set; }
 		private Guid _guid = Guid.NewGuid();
@@ -952,7 +931,6 @@ namespace BizHawk.Client.Common
 				// We aren't creating any new states, just adding a reference to an already-existing one; so Used doesn't need to be updated.
 				stateList.Add(branchHash, kvp.Value);
 			}
-			CallSaveCallback(_movie.BranchCount - 1);
 		}
 
 		public void RemoveBranch(int index)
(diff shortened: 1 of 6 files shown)


TASeditor:
Using the TAS Editor method is probably better. I get slowdowns with the method I wrote when changing input of multiple frames at once.

vadosnaprimer:
You mean you are setting it one by one and cause autorestore and all the logic every time?

TASeditor:
Yes.

vadosnaprimer:
TASEditor was designed wisely enough to allow borrowing majority of its approaches. There's rarely need to reinvent the wheel and see yourself fail.

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Tastudio.cs
```diff
@@ -23,6 +23,32 @@ public TastudioLuaLibrary(Lua lua, Action<string> logOutputCallback)
...
+			INPUTCHANGE,
+			INSERTFRAMES,
+			DELETEFRAMES,
+		};
+
+		public enum INPUT_CHANGE_TYPES
+		{
+			BOOL,
+			FLOAT,
+		};
```
vadosnaprimer: Use CamelCase for enums (names and members) instead of caps. Otherwise, should be fine: this is unlikely to break anything, and once issues are found, we'll fix them.

BizHawk.Client.Common/movie/tasproj/TasStateManager.cs
```diff
@@ -26,6 +26,27 @@ private void CallInvalidateCallback(int index)
...
+		{
+			SaveCallback?.Invoke(index);
+		}
+
+		public Action<int> RemoveCallback { get; set; }
+
+		private void CallRemoveCallback(int index)
+		{
+			RemoveCallback?.Invoke(index);
+		}
```
vadosnaprimer: All this logic should be in https://github.com/TASVideos/BizHawk/blob/master/BizHawk.Client.EmuHawk/tools/TAStudio/BookmarksBranchesBox.cs

StateManager is only managing savestates related to branches, bot branch operations themselves, which are a few levels above, and separated from Common.

TASeditor: I'm not sure how to do that. The functions in BookmarksBranchesBox don't have branch index

vadosnaprimer: You can do `Movie.Branches.IndexOf(branch);` there.

TASeditor:
@vadosnaprimer please help with the stuff in TAStudio.cs

vadosnaprimer:
What happened there?

TASeditor:
Line 536 and 593.

vadosnaprimer:
I see it, I mean explain the problem.

TASeditor:
I don't know how to refer to BookmarksBranches' LoadedCallback ect.

vadosnaprimer:
https://github.com/TASeditor/BizHawk/blob/e401956a98e1ad144b71fe27acdb243ee3da24f1/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs#L121

PR #1125, opened 2018-03-03
Tastudio: drawing stuff, selection, patterns, add column lua function

TASeditor:
- Selection drawing now only happens in FrameColumn or lua generated columns. Same with Shift and Control clicks.
- Selection modification with Control. Ctrl+Left/Right: Add/Remove selection at last selected cell. Ctrl+Shift+Left/Right: Add/Remove selection at first selected cell.
- Playback courser stuff happens with Shift.  Shift+Up/Down: Rewind/Advance one frame. Shift + PageUp/PageDown: Go to previous/next marker with playback courser.
- Draw input from first selected row to current cell if shift is held.
- Draw pattern if Alt is held.
- Apply pattern to selected rows when clicking on column header and Alt is held.
- tastudio.addcolumn() lua function.

TODOs:
- extraFloatRows can no longer be added by clicking on Float columns, users must use FrameColumn. Revert it.
- Shift+Down holding repeatedly plays the current frame.
- Don't add lua columns to tasproj file.
- Display lua columns to ColumnsMenu.
- Alt+Shift drawing pattern from first selected row to current cell.
- Don't change focus to menu when clicking Alt.


vadosnaprimer:
I think it's fine, but in future, create a separate PR for each feature.

PR #1132, opened 2018-03-10
Tastudio: fix selecting outside bounds

TASeditor:


vadosnaprimer:
Does it fix both of [these](http://tasvideos.org/forum/viewtopic.php?p=466588#466588)?

> 1. shift+up: when holding down it skips multiple frames (because it reacts so fast). 
>  shift+down: only works when not holding. If I hold it down, it only jumps back and forth between 2 frames 
>  2. ctrl+left right, ctrl+shift left right: you can go out of bounds (for example behind 0th frame)

Issue #1075, opened 2017-12-09
Rom status icon doesn't work on Saturn


adelikat:
Open a saturn game and click the rom status icon in the staus bar, nothing happens.

This is also true of any core that takes a list of roms (for some there's both a single rom and multi-rom constructors and it only happens for multi-rom).

Either we need the client to take a stab at implementing RomAnnotationDetails for multi-roms or each core affected needs to implement that logic

Asnivor:
see #1152

Asnivor:
This now works as of a few months ago.

The last thing is that disc-based 'ROMS' just show their hashes as 'N/A' (because thats what I did at the time).

Can we have a discussion on what we should be showing here in this instance? 

* cue hashes are pointless
* bin/img/iso hashes can be expensive I would imagine
* iirc PSX has some blackmagic internal hash thing that @zeromus mentioned to me a while back. Do we think about using this?
* is there such a thing as a 'known good dump' when it comes to disc images?
* do we just not bother....

zeromus:
cue is the same situation as bin/img/iso. It's the mounted disc which is hashed. All of the disc images have a black magic hashing method (although some like PSX are far blacker than others). I did generate a known good list of PSX roms by running my blacker magic hashing method on all of redump. It's in our gamedb.

We should be able to show a disc "hash" in the rom status. These are useful for identification purposes. Thats what the movie tools should be doing to confirm the "correct" game is _probably_ being used. Of course, it can't be known for 100% sure unless the entire disc's contents are hashed. For now, a weaker identifying hash is all we can do.

For all discs we should (maybe we arent, I dont know) using a hash based on the TOC, plus the first  several sectors. First several sectors was chosen by vecna originally. I determined how many sectors were required to uniquely identify all the PSX games (part of blacker magic) although it would be possible to trick it. 

A more robust identification would involve parsing the filesystem, hashing that, AND identifying the executable, and hashing that. The goal is to hash as much as we can, except for the last bit of hashing everything. While keeping in mind romhacks are likely to only change the EXE, and bad rips are likely to have different TOCs before anything else is wrong.

However, we're probably not changing any hashing methods now. I'm just telling you what my grand plan was.

[...]

PR #1152, opened 2018-03-22
Partial fix for issue #1075

Asnivor:
@adelikat I thought i'd have a go at implementing your first suggestion, "take a stab at implementing RomAnnotationDetails for multi-roms".

Firstly, this PR handles xml files generated through the multidisk bundler:

image

It also makes the RomStatus icon clickable for single disc based 'roms' (detected if loader.Rom is null):

image

The reason that this is a partial fix is because the RomStatus window only shows the game name (or names in the case of an xml) when the Rom is a disc (or discs). SHA1 and MD5 are just written as N/A:

image

 I'm just not sure what else to put here with regard to discs. cue/ccd hashes? iso/bin/img hashes? What is actually useful here?

 I also couldn't see a way to get this info without making another DiscSystem instantiation call which would probably be expensive. Maybe I'm missing something?

Anyways, hopefully this is a start/step in the right direction (if the code is acceptable).

Thanks,

-Asni


adelikat:
I like this change and it greatly improves the situaiton :)

PR #1170, opened 2018-03-31
Fix GPGX issues mentioned in #1167 by moving certain settings into the SyncSettings

### LINKED ISSUE #1167: Various GPGX settings break savestates despite not being syncsettings
Various settings in the GPGX core break savestates despite them not being flagged as SyncSettings

```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.InvalidOperationException: Memory consistency check failed.  Is this savestate from different SyncSettings?
```

The ones that appear to affect it are:
Audio Filter
Low Pass Range
Three band low cutoff
Three band high cutoff
Three band low gain
Three band mid gain
Three band high gain
BackdropColor

"Use custom backdrop color" doesn't affect the loading of savestates, but it appears to be saved inside the state regardless (turn it on, savestate, turn it off, loadstate, it'll be marked as "disabled" while drawing the custom bg color)

NarryG:
Feos closed #1167  due to what appears to have been a misunderstanding of the issue. This fixes the issue by moving the aforementioned settings into the SyncSettings

This doesn't touch "Use Custom Backdrop Color" as I'm unsure if that persisting through savestates is just a result of how the feature works internally (if the override background color is baked into the savestate), and that setting doesn't appear to break savestate consistency

### DIFF
--- a/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.ISettable.cs
+++ b/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.ISettable.cs
@@ -198,6 +198,63 @@ namespace BizHawk.Emulation.Cores.Consoles.Sega.gpgx
 			[DefaultValue(false)]
 			public bool PadScreen320 { get { return _PadScreen320; } set { _PadScreen320 = value; } }
 
+
+			[DeepEqualsIgnore]
+			[JsonIgnore]
+			private bool _Backdrop;
+
+			[DisplayName("Use custom backdrop color")]
+			[Description("Filler when layers are off")]
+			[DefaultValue((bool)false)]
+			public bool Backdrop { get { return _Backdrop; } set { _Backdrop = value; } }
+
+
+			public GPGXSettings()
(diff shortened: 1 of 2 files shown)


nattthebear:
The reason why use custom backdrop doesn't trip the consistency checker is that the consistency state is taken before it's ever set.  If you follow it in the C# side, you'll see that it's only set by gpgx_set_draw_mask, which is called by PutSettings, which is never called until after _elf.Seal has been called.

This brings up another way to get around this issue; if a syncsetting doesn't need to be set at global init/construct time at all, it can just be done after that.  The downside with this method (as opposed to moving the sync setting out of savestates entirely) is that because the sync setting is still in the savestate, loading a non-matching savestate will actually change the syncsetting.

NarryG:
On that note, should I move custom backdrop into syncsettings, or is it fine as is?

nattthebear:
Since it does work where it is, I'm OK to leave it as is.  The only thing this PR is waiting on is some time for others to look at, if they care.

PR #1172, opened 2018-04-02
Gambatte x lua and fix #1159

### LINKED ISSUE #1159: MemoryCallbackSystem ActiveChanged only invoked on first Add
See: [MemoryCallbackSystem.cs](https://github.com/TASVideos/BizHawk/blob/519e14aa1ad7a9d6df2edc7808c5ed687dfee046/BizHawk.Emulation.Common/Base%20Implementations/MemoryCallbackSystem.cs#L69-L74)

This is an issue for cores that rely on the event handler to pass callbacks in:

https://github.com/TASVideos/BizHawk/blob/519e14aa1ad7a9d6df2edc7808c5ed687dfee046/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IDebuggable.cs#L74-L84

For example, if onmemoryexecute is called from lua, it will run RefreshMemoryCallbacks with only HasExecutes as true. Any calls to onmemorywrite or onmemoryread will still register the functions on the lua end, but the function pointer for those collections will never be passed to the core.

gifvex:
Implements TotalExecutedCycles for Gambatte after a general cleanup of its MemoryCallbacks. Gambatte uses its internal cycle counter as an arbitrary timer, so the callbacks use relative values to accurately represent executed cycles.

Included is a fix for #1159 (cores unable to receive multiple callback types). UpdateHasVariables now returns a bool indicating if anything changed.


alyosha-tas:
One other thing I noticed with Gambatte is that the scanline register is only intermittently updated when looking at trace logs. Do you think it would be possible (and easy) to continuously update this?

It's fine if not, I just happen to remember it and thought it would be a useful thing to include in this pr.

PR #1180, opened 2018-04-16
Lua: Add optional position parameters to `DrawPolygon`

AntyMew:
In some cases, a user may want to draw a polygon at a different position without needing to update every point. The proposed changes optimize this task by applying the translation while populating the array of points.

Note that this change breaks backwards compatibility of `gui.drawPolygon`, `forms.drawPolygon`, and `LuaCanvas.DrawPolygon` by placing the new `x` and `y` parameters before `line` and `background`. This was done to maintain the convention of color arguments going last.

### DIFF
--- a/BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Forms.cs
+++ b/BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Forms.cs
@@ -1043,7 +1043,7 @@ namespace BizHawk.Client.EmuHawk
 					{
 						if (control is LuaPictureBox)
 						{
-							(control as LuaPictureBox).DrawPolygon(points, line, x, y, background);
+							(control as LuaPictureBox).DrawPolygon(points, x, y, line, background);
 						}
 					}
 				}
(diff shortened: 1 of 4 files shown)


vadosnaprimer:
Nice! Can you post an example script? Also, does it build for you? Built bot dies here:
https://ci.appveyor.com/project/zeromus/bizhawk-udexo/build/0.0.0.2142#L470

AntyMew:
Gotcha! This should demonstrate the new behavior effectively...

```lua
triangle = {{0, 19}, {19, 19}, {19, 0}}

while true do
    gui.drawPolygon(triangle, emu.framecount() % 240, 0)
    emu.frameadvance()
end
```

AntyMew:
...I have no idea what this CI is complaining about. Maybe the breaking change has to do with it?

AntyMew:
PFFFFT, I missed that typo rereading it about 10 times. Whoops...

vadosnaprimer:
Doesn't work for me if these offsets are omitted: `gui.drawPolygon({{100, 119}, {119, 119}, {119, 100}})` does nothing after this change, but works before it. I think those variables need to be nullable type to become truly optional? Also, I think it'd be better to call them offset rather than position, because position is already included in the polygon table.

adelikat:
Going to merge this into a branch and then take a shot at the suggestions feos had

PR #1189, opened 2018-05-03
Use generic interface type on MOS 6052X for talking to the emulator core

Scepheo:
This PR creates an interface (`IMOS6052XLink`) that has all the methods required for the CPU to talk to the emulator core. The `MOS6052X` then takes a type implementing this interface as a generic type parameter.

The `MOS6052X` constructor takes a link as a parameter, and uses this to talk to the emulator core, instead of the `Func`/`Action` delegates currently used. This means that the compiler knows the exact methods being called at compile time.

By making all implementations of this link interface `struct`s, the compiler can subsequently eliminate pointer chasing (the delegates live on the heap) and virtual calls (`Func.Invoke`/`Action.Invoke`).

For example, on my machine, this leads to a speedup of ~1-2 fps for NESHawk and ~7-9 fps for A2600.

This change needs some more testing across all cores utilising the MOS 6502X implementation (Commodore 64, NES, A2600 and A7800) and by more people, to ensure the JITer consistently applies the optimization across frameworks, or at the very least it's not detrimental to some people's performance.

If effective, a similar construct could be used in the other CPU implementations (or perhaps elsewhere) for similar performance gains.

### DIFF
--- a/BizHawk.Emulation.Cores/BizHawk.Emulation.Cores.csproj
+++ b/BizHawk.Emulation.Cores/BizHawk.Emulation.Cores.csproj
@@ -1299,6 +1299,7 @@
     <Compile Include="CPUs\LR35902\Registers.cs" />
     <Compile Include="CPUs\LR35902\Tables_Direct.cs" />
     <Compile Include="CPUs\LR35902\Tables_Indirect.cs" />
+    <Compile Include="CPUs\MOS 6502X\IMOS6502XLink.cs" />
     <Compile Include="CPUs\W65816\Disassembler.cs" />
     <Compile Include="CPUs\68000\Diassembler.cs" />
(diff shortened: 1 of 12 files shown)


Asnivor:
Appveyor is stumbling over a bunch of things by the looks of it:

https://ci.appveyor.com/project/zeromus/bizhawk-udexo/build/0.0.0.2170#L252

Scepheo:
Seems AppVeyor is using an older version of MSBuild that doesn't support expression-bodied constructors yet. I've replaced them with regular constructors, so the build's fixed again.

Asnivor:
Ahh yes, makes sense. Appveyor uses VS2015 by default (unless you specify otherwise in your appveyor.yml).

nattthebear:
We talked about this one a while back in irc.  I think it's a fine change

Issue #1200, opened 2018-05-25
Analog Y axis is flipped between various cores depending on input method


NarryG:
Summary:
* DInput has positive **right** and **down**
* XInput has positive **right** and **up**
* libinput has positive **?** and **?**
* Genesis has positive **?** and **?** (mouse) - #1602
* N64 has positive **right** and **up** (standard controller)
* Saturn has positive **?** and **?** (3D controller)
* PSX has positive **right** and **down** (analog controller)
* Vectrex has positive **right** and **up** (standard controller)

So we should only be negating DInput Y and virtual PSX Y.

***

The same physical controller will be Y-inverted when switching between DInput and XInput (has this been fixed? --yoshi).
Before #1202, Octoshock was Y-inverted compared to Mupen and Saturnus.

see also #137

NarryG:
Base on Microsoft's documentation
DInput: `The value returned for the x-axis is greater as the stick moves to the right, and the value for the y-axis increases as the stick moves toward the user.`
Xinput:  `Negative values signify down or to the left. Positive values signify up or to the right. `

It appears Microsoft flipped the Y axis between the two.

Even so, the inconsistencies between cores and the defaults should be fixed.

YoshiRulz:
see also #137

YoshiRulz:
Finding docs is hard.

From [cen64](https://github.com/n64dev/cen64/blob/8b89df2fdca20eec2d735e2b3db80b0b8199879f/os/common/input.c#L35-L53), it looks like right and up are positive on the N64. I also found [this](http://n64devkit.square7.ch/pro-man/pro26/26-02.htm#04-01) which includes the specs of the octagonal hardware limit in the controller's case, and [this](https://sites.google.com/site/consoleprotocols/home/nintendo-joy-bus-documentation/n64-specific/standardcontroller) similarly interesting-yet-unhelpful document. I did remember that in my research on Rayman 2 (U), I found that 0x0CF590-0x0CF593 mirror the input. Sure enough, right and up positive.

From [this programmers' guide](http://www.playvectrex.com/designit/chrissalo/joystick.htm), Vectrex also has right and up as positive.

I've edited this into the issue desc.

iamgreaser:
Alright, as requested:

According to psx-spx: http://problemkaputt.de/psx-spx.htm#controllersstandarddigitalanalogcontrollers
On a real PSX analogue controller, +X is right and +Y is down for both sticks.
[...]

PR #1202, opened 2018-05-29
Flip the Y axis for N64 and Saturn 3D-Pad by default. Workaround for #1200

NarryG:
As explained in #1200 , the N64 core and Saturn core take analog positive Y as down, whereas the PSX core takes analog positive Y as up.
In the case of xinput, which is what the default binds are, positive Y is pushing the stick away from the user. This results in a situation where the default binds provide the opposite input compared to what the user would expect.

This pull request flips the Y axis on the default binds for the N64 and Saturn 3D pad to provide input as expected by the user.

This isn't a solution to the core problem, but it results in the default binds at least being correct until the core problems explained in the issue are addressed.


NarryG:
Not sure why github is including those additional commits. One of which was already merged, the other was me syncing me own repo up to master. Only one file was changed.

zeromus:
Nope. Fix "the core issue" by inverting one of them. 

Let's use SDL as a tiebreaker.

This URL says SDL considers up positive:
http://lazyfoo.net/tutorials/SDL/19_gamepads_and_joysticks/index.php
Check a couple of other links to make sure up is really positive in SDL

Then fix whichever of xinput or dinput aren't that way by inverting their Y axis when it's read

NarryG:
Will do

NarryG:
While that does fix xinput and dinput being flipped, that doesn't fix the fact that some of the cores are expecting a different "up".

zeromus:
The cores were presumably configured correctly by whoever set it up, based on whichever orientation was used by the pad they were testing on. If you flip that one on the input side, then youll have to flip the orientations in the core. Maybe it would be better to choose flipping the input method which does not result in having to change the cores. However if the cores are _inconsistent_, then choose it based on SDL and fix all the cores to suit it.

NarryG:
The cores are inconsistent at the moment (N64 and Saturn are one way, PSX is the other). I'll go through and make them all match (+Y as up) and then make the inputs all work the same.

PR #1222, opened 2018-06-21
This solve transparency color problem

Azreally:
It's work fine.

### DIFF
--- a/BizHawk.Client.Common/lua/LuaLibraryBase.cs
+++ b/BizHawk.Client.Common/lua/LuaLibraryBase.cs
@@ -74,10 +74,15 @@ namespace BizHawk.Client.Common
 			var result = double.TryParse(color.ToString(), out tryNum);
 
 			if (result)
-			{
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Does this method properly parse color names, like "red"?

BizHawk.Client.Common/lua/LuaLibraryBase.cs
```diff
@@ -74,10 +74,15 @@ protected static uint LuaUInt(object luaArg)
...
-			}
+			unchecked
+				{
+					var argb = (long)tryNum;
+					return Color.FromArgb(
+						(byte)(argb >> 24),
+						(byte)(argb >> 16),
+						(byte)(argb >> 8),
+						(byte)argb);
+				}
```
vadosnaprimer: This works just fine for me:
```
			int tryNum;
			var result = int.TryParse(color.ToString(), out tryNum);

			if (result)
			{
				return Color.FromArgb(tryNum);
			}
```

Issue #764, opened 2016-12-30
Multi-disk Bundler doesn't handle .zip


mugg91:
When using Multi-disk Bundler, selecting Master Karateka (J).gb works fine.
But Master Karateka (J).zip doesn't.
The problem seems independent of which game I try.

Suggestion; Either don't allow user to select .zip or have the emu extract the .zip

PR #1275, opened 2018-08-15
MultiDiskBundler: Handle archive files properly

Asnivor:
This has been bugging me for a while now.

Open ticket from 2016 - #764

This pull request does the following:

* Allows BizHawk-XMLGame files to include references to ROMs/Tapes/Disks within archives (using the same format that Bizhawk uses elsewhere). Importantly, this is parsed correctly before being sent to RomLoader further down the line (the code was kind of there, but not complete). 

XML example:

image

* MultiDiskFileSelector and MultiDiskBundler have been modified to call ArchiveChooser when an archive is selected through the browse dialog. There is certain conditional logic here. It will basically tell the user to FO if the selected system is PSX, PCFX or SAT (basically systems that will probably die if you try to load files from within an archive).

* The 'SaveAndRun' button on the multidiskbundler form now actually RUNS, rather than just SAVES (the issue was the LoadRomArgs passed in to LoadRom were infact empty before).

I havent managed to break it yet, so looking good so far.

Thanks,

-Asni

### DIFF
--- a/BizHawk.Client.Common/XmlGame.cs
+++ b/BizHawk.Client.Common/XmlGame.cs
@@ -82,11 +82,13 @@ namespace BizHawk.Client.Common
 								using (var hf = new HawkFile(fullpath))
 								{
 									if (hf.IsArchive)
-									{
-										var archiveItem = hf.ArchiveItems.First(ai => ai.Name == filename.Split('|').Skip(1).First());
+									{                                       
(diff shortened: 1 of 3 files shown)


zeromus:
looks fine to me. bundler isn't my department. I like that it supports archive.zip|insidefile.

hegyak:
I am pretty sure I made a comment/issue about this a long while ago.  Glad to see it's getting done.

Issue #1272, opened 2018-08-15
Commodore 64 core borders


QuakerRUS:
1. There is no option for none borders. Can you add this option, please?
2. Full border option crashes program after reboot core.

QuakerRUS:
I need none borders because old-games.ru site requires 320x200 screenshots w/o borders for Commodore 64. Thanks!

QuakerRUS:
Just tested, works perfectly!

PR #1276, opened 2018-08-16
C64Hawk: Fix border issues - #1272

Asnivor:
This is in relation to ticket #1272

This PR:

* Adds a 'None' border option as requested:

image

* Fixes the endless loop that happens in  Vic.TimingBuilder (timingbuilder is expecting -1 values for hblankStart and hblankEnd passed to it when the 'Full Border' option is selected. Without the new changes in this PR Vic.ConfigureBlanking ends up converting hblankStart and vblankStart to the same value (and actually an odd number - which is ultimately what is responsible for the infinite loop).

So basically the 'Full Border' option now works, but it actually is showing the screen including the horizontal and vertical flyback periods. I.E. : it looks weird:

image

Its probably safe to say that the person who raised the issue believed that FULL is what he/she wanted (but didnt know this because it never worked), when actually he/she *actually* just wants to stick with NORMAL.

The question is do we just remove the FULL border as an available option? Its really only a dev/debug thing as far as I can see.

Thanks,

-Asni

### DIFF
--- a/BizHawk.Emulation.Cores/Computers/Commodore64/C64.ISettable.cs
+++ b/BizHawk.Emulation.Cores/Computers/Commodore64/C64.ISettable.cs
@@ -104,7 +104,7 @@ namespace BizHawk.Emulation.Cores.Computers.Commodore64
 
 		public enum BorderType
 		{
-			SmallProportional, SmallFixed, Normal, Full
+			None, SmallProportional, SmallFixed, Normal, Full
 		}
 
 		public enum SidType
--- a/BizHawk.Emulation.Cores/Computers/Commodore64/C64.ISettable.cs
+++ b/BizHawk.Emulation.Cores/Computers/Commodore64/C64.ISettable.cs
@@ -34,7 +34,12 @@ namespace BizHawk.Emulation.Cores.Computers.Commodore64
 		public class C64Settings
 		{
 			[DisplayName("Border type")]
-			[Description("Select how to show the border area")]
+			[Description("Select how to show the border area\n" +
+                "NORMAL:\t Horizontal and Vertical border both set to 32 pixels (although horizontal will appear narrower due to pixel density)\n" +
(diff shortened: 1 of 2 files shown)


Asnivor:
@Scepheo looking over the issue tracker, you may still be doing some work on this core?

Shout out if this is problematic with anything you may have implemented already.

vadosnaprimer:
@SaxxonPike came up with detailed explanations for all of the C64 syncsettings, and I put them to the appropriate dialog. Looks like Border type needs similar notes.

vadosnaprimer:
If it's guaranteed not to affect sync, it's not a syncsetting. @SaxxonPike decided this way, I think it makes sense. I also noticed the core doesn't inform the user that reboot is needed. Usually we don't require reboot when the setting takes effect immediately, and it's not the case here. But I'm unsure of the logistics, just guessing.

Asnivor:
@feos I have added in setting descriptions for the border types as requested.

I've also gone ahead and removed the FULL option as it is fairly meaningless outside of development. This was probably just left in for testing.

Asnivor:
@vadosnaprimer does this meet your requirements to be merged now?

vadosnaprimer:
Can it be easily made to require a core reboot if this is changed?

Asnivor:
Well, the borderType is already passed in via constructor anyway (albeit via Settings), so the core needs a core reboot for the settings to take (I haven't changed this).

If we want to notify a user that a core reboot is needed, then the simplest way is to move it from Settings to SyncSettings. And that is very easy.

vadosnaprimer:
Ugh, C64 completely lacks the logic that demands core reboot, even for syncsettings. It looks like any core whose options are generated on the fly fail to FlagNeedsReboot(), which happens in PutCoreSettings() and PutCoreSyncSettings().

I don't think this should be a part of this PR.

Issue #1290, opened 2018-08-20
Movie recording and IEmulator.DeterministicEmulation


Asnivor:
Currently DeterministicEmulation is ignored when starting to record a movie.

Unless anyone has any objections, we really should be enforcing DE for movie recording.

vadosnaprimer:
Ignored for all the cores? O_O

Asnivor:
I'm assuming so. I dont see anything here: https://github.com/TASVideos/BizHawk/blob/master/BizHawk.Client.EmuHawk/movie/RecordMovie.cs

I have tried recording a movie with a core that exposes DeterministicEmulation as a SyncSetting (ZXHawk in this case) and the movie recording process starts up when DE==false

Unless I am missing a piece of the puzzle, I would expect other cores to behave the same way (the ones that have this setting available to change that is)

vadosnaprimer:
I don't know when it should be false (maybe some cores have casual gaming profiles), but I don't think DE should be a sync setting. SyncSettings is only a thing for movies, and for movies DE must be always true, so it can't change as long as we're (re)recording. And yes, if a core even allows to skip some measures that actually make it deterministic, for such cores we must force DE. If it basically changes nothing, then it doesn't matter, but this also means it should always be set to true for them.

vadosnaprimer:
I don't think the user should be able to decide whether they would like to lose their leg or not. I'll have to look up which cores even depend on this thing, and when they want it to be false.

Asnivor:
Oh, I just re-read what you wrote. 

So DE is pretty much controlled via user profile selection at the moment?

If that is the case, then I should probably just remove this as a syncsetting completed in zxhawk...

Asnivor:
>  I'll have to look up which cores even depend on this thing, and when they want it to be false

Makes sense. Unless there are cores that by default expect a null value to be passed in and a false value generated as a result (which seems highly irregular but I guess might be that way for reasons), the PR will still work.

zeromus:
I would like for DeterministicEmulation to be a sync setting. It _is_. Whether we expose that to the user is a separate question. I don't mind special logic curating what we show the user in the presentation. Buried in the inner logics, it's too complicated. Maybe that means we filter DeterministicEmulation out of a property list for showing to the user. Maybe that means we use reflection to see if a DeterministicEmulation property is available, and set it to true if it is.

Asnivor:
I'm not sure where I would start with that (or even if I understand correctly).

[...]

PR #1291, opened 2018-08-21
EmuHawk: Movies and DeterministicEmulation - ensure this is enforced …

Asnivor:
…whenever a movie is queued - #1290 

See ticket #1290 for background.

Very simple update. Previously when LoadRom() was called, a deterministic bool was passed into RomLoader (then onto a core). This was set to TRUE if a movie session was queued, but ONLY if LoadRomArgs.Deterministic was NOT NULL.

This PR removes any potential edge cases that might come up in the future by:

* If LoadRomArgs.Deterministic is null (as in most cases at the moment) set *deterministic* to TRUE (**previously this was being set as FALSE if no movie was queued - I can't for the life of me figure out why this was the case**)
* Otherwise set *deterministic* based on args.Deterministic's value
* Finally, if a movie is queued, ensure *deterministic* is set to TRUE

### DIFF
--- a/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/BizHawk.Client.EmuHawk/MainForm.cs
@@ -3555,9 +3555,29 @@ namespace BizHawk.Client.EmuHawk
 			}
 
 			try
-			{
-				// If deterministic emulation is passed in, respect that value regardless, else determine a good value (currently that simply means movies require deterministic emulaton)
-				bool deterministic = args.Deterministic ?? Global.MovieSession.QueuedMovie != null;
+			{                
+                bool deterministic;
+
+                if (args.Deterministic == null)
+                {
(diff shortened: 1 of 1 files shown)


Asnivor:
Also, previous ZXHawk commit to master showing how it works for this particular core (no other cores as yet need this): https://github.com/TASVideos/BizHawk/commit/4e470cf17ad30f85c68c9e0c1e2de9457c2705fb

The SyncSettings.json stored in any movie made with this core (or any future core that presents DE as a syncsetting) will actually reflect whatever the core setting for this is, but that shouldn't matter as we are forcing DE for ALL movies.

zeromus:
I discussed this in another thread. I think it should all be done through special logic on sync settings, rather than having lots of other wacky zany wild and crazy things going on

Issue #1241, opened 2018-07-10
Invalid SaveRAM Path disallows closing of game/core/BizHawk


raekuul:
Importing config files with invalid saveram paths (usually by transferring the folder from an old system to a new system with new path topology) will allow games to start, but you get caught in an exception continue loop until you fix the saveram path if you try to close the game, core, or program

Asnivor:
@adelikat @nattthebear @zeromus 

Whats the preferred outcome in this situation? Do we want to be handling the exception before it happens and prompt the user? Set the path to config default? Something else?

zeromus:
Keep in mind I dont think we should be automatically saving saveram in the first place. Then we would not have this problem. But since we do, I think it would be ideal to try saving the saveram as soon as possible when closing and handling a failure with a "Failed flushing the game's Save RAM to your disk. Close without flushing Save RAM? yes/no" kind of box. Flush is the terminology we use in the file menu right now.

Asnivor:
Merged to master

PR #1292, opened 2018-08-21
EmuHawk: Handle SaveRAM IO errors - #1241

Asnivor:
See ticket #1241 

This PR:

* Pushes a client OSD message to the screen when flushing saveRAM fails (also introduces a try/catch here in FlushSaveRAM() to avoid the reported exception loop)
* It also modifies the FlushSaveRAM() method to return a bool. This way CloseGame() is modified so that if FlushSaveRAM() return false, a MessageBox is shown to the user making them aware of the situation, and whether they really want to close.


Asnivor:
Done

Issue #1293, opened 2018-08-22
SNES9x Code Converter is completely broken for GameShark


Kippykip:
Using GameShark codes with the Code Converter throws a **system.nullreferenceexception** if SNES9X is the selected core.

Changing to bSNES solves the problem.
![capture](https://user-images.githubusercontent.com/33023808/44449793-79646b80-a632-11e8-89de-6f2f8bf5e161.png)

NarryG:
Snes9X doesn't expose the System Bus. The code converter is supposed to not even open under Snes9X if I'm not mistaken.
Cheats are already pretty hit or miss in bizhawk due to various domains being write protected.

Asnivor:
The CodeConverter is either greyed out (on the tools menu) or hidden (in the cheats dialog) if _GlobalWin.Tools.IsAvailable`<GameShark`>();_ returns false.

This is based on the SupportedSystems ToolAttribute:

```
[Tool(released: true, supportedSystems: new[] { "GB", "GBA", "GEN", "N64", "NES", "PSX", "SAT", "SMS", "SNES" })]
	public partial class GameShark : Form, IToolForm, IToolFormAutoConfig
```

This check only looks at stuff based on the SystemID rather than the core itself.

I suspect that introducing an **_UnsupportedCores_** ToolAttribute would possibly be a way to solve this without too much adhoc messing around all over the place. I'll have a look and see if it looks doable.

Asnivor:
Fix merged to master a while ago.

PR #1294, opened 2018-08-23
EmuHawk: ToolManager.IsAvailable<T>

Asnivor:
This is because of issue #1293 

Currently the gameshark Tool (called 'Cheat Code Converter' in the menus and dialogs) is made available to the user based on a _GlobalWin.Tools.IsAvailable`<GameShark`>()_ call. As this method only looks at the *supportedSystems* ToolAttribute, you have a situation where SNES IS supported (and the code converter dialog is made available) but the Snes9x core does not expose a 'System Bus' MemoryDomain - so exceptions are shown when the user attempts to convert a code.

This PR:

* Modifies the **ToolAttribute** class to make available an **UnsupportedCores** IEnumerable (which defaults to null)
* Modifies the **.IsAvailable(Type t)** method in **ToolManager.cs**. The logic is pretty much the same as before, except a second check is made to see whether there are any UnsupportedCore overrides for the current core. Similar to before it returns a TRUE if both SupportedSystems & UnsupportedCores are null
* Adds 'Snes9x' to the **GameShark.cs** **UnsupportedCores** ToolAttribute (in order to solve ticket #1293)

As far as I can see, it is only the Snes9x core that needs this within GameShark.cs (for now), but rather than implement an arbitrary kludge, I thought it might be worthwhile doing it in a way that can/will be useful elsewhere in the future.

Thanks,

-Asni

### DIFF
--- a/BizHawk.Client.EmuHawk/ToolAttribute.cs
+++ b/BizHawk.Client.EmuHawk/ToolAttribute.cs
@@ -6,14 +6,17 @@ namespace BizHawk.Client.EmuHawk
 	[AttributeUsage(AttributeTargets.Class)]
 	public class ToolAttribute : Attribute
 	{
-		public ToolAttribute(bool released, string[] supportedSystems)
+		public ToolAttribute(bool released, string[] supportedSystems, string[] unsupportedCores = null)
 		{
 			Released = released;
 			SupportedSystems = supportedSystems;
+            UnsupportedCores = unsupportedCores;
 		}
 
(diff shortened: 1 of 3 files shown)


zeromus:
Sounds sensible to me. The cheat system needs to be reworked entirely. Theres no need to overthink this part for now when it's only used for the cheat system

adelikat:
I agree with zeromus

zeromus:
yes, its fine

PR #1307, opened 2018-09-02
update upstream mgba to latest 0.6.  fixes saveastate issue

nattthebear:
I don't have time to look into this right now, but it compiled without issue


vadosnaprimer:
@spikestuff go check!
https://ci.appveyor.com/project/zeromus/bizhawk-udexo/build/0.0.0.2317/artifacts

vadosnaprimer:
People confirmed that it fixes the thing it attempts to fix.

vadosnaprimer:
People confirmed that it fixes the thing it attempts to fix.

Spikestuff:
Bit late. But on the version you told me to check it desyncs when rewatching the input back:


[Here's the input in the tasproj format](https://github.com/TASVideos/BizHawk/files/2351301/sims.zip)

vadosnaprimer:
Maybe it's a legitimate desync that was hidden by incorrect state loading, and now it should be good after resyncing?

Spikestuff:
Honestly I hope so.
There was another state where after frame 1478 all the following input went to a lag frame.

The game is risky so I'll look into further input and give a shout about it if it goes awry (and one if it's working fine).

Issue #1271, opened 2018-08-14
[Lua request] Get BizHawk version with client.getversion


brunovalads:
In order to make my scripts more accessible, would be interesting a function like `client.getversion` (not sure if `client` is the best class, seems to be) that returns the BizHawk version as a string (this information is stored somewhere), and it can be used to forbid features in the scripts that use lua functions developed later (e.g. a lot of lua functions doesn't exist in 1.X). This is interesting to avoid making multiple versions of the same script.

brunovalads:
Oh nice! 

> Ultimately though, you have no way of differentiating between v1.x.x clients and v2 clients up to 2.3.0 (as none of these will have the new Lua method).

Yea that's exactly what I thought when I was reading your reply, this feature will only be really useful in the future when older BizHawk versions are completely forgotten. So for now, the way I'm used to identify the version is the way Amaraticando did in the SMW script:
```lua
biz.features = {
  backcolor_default_arg = (emu.setislagged == nil), -- < 1.11.5
  gui_text_backcolor = (gui.DrawFinish == nil), -- < 1.11.7
  support_extra_padding = (client.SetGameExtraPadding ~= nil) and (gui.DrawFinish ~= nil), -- 1.11.7
}
```
So in the future, I would need to make the following:
```lua
local bizhawk_version

if client.getversion ~= nil then -- is "2.4.0" or higher
  bizhawk_version = client.getversion()
else -- 2.3.X or lower
  if emu.getregisters().H ~= nil then bizhawk_version = "2.3.0"
  elseif emu.getluacore ~= nil then bizhawk_version = "2.2.2"
  elseif memory.hash_region ~= nil then bizhawk_version = "2.2.1"
  elseif tastudio.getbranches ~= nil then bizhawk_version = "2.2.0"
  --elseif  ~= nil then bizhawk_version = "2.1.1"
  elseif forms.pictureBox ~= nil then bizhawk_version = "2.1.0"
  elseif SQL ~= nil then bizhawk_version = "2.0.1"
  elseif LuaCanvas.defaultForeground ~= nil then bizhawk_version = "2.0.0"
end
```
and make a few more conditionals to detect the 1.X versions as well.

vadosnaprimer:
Yeah even if we add this now, all the previous releases would have to use hacks like this.

Asnivor:
@vadosnaprimer presumably we could hardcode these hacks on the c# side, and return a version specifically for Lua.

The question is...should we..

[...]

PR #1314, opened 2018-09-10
Lua: client.getversion() - #1271

Asnivor:
see ticket #1271 

Adds lua method:

`client.getversion()`

This returns the _Mainversion_ string from _VersionInfo_ regardless of whether this is a DeveloperBuild or not. 

We just need to keep _VersionInfo.Mainversion_ updated at every release.

### DIFF
--- a/BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
+++ b/BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
@@ -391,6 +391,13 @@ namespace BizHawk.Client.EmuHawk
 			return GlobalWin.MainForm.DesktopLocation.Y;
 		}
 
+        [LuaMethodExample("local incbhver = client.getversion( );")]
+        [LuaMethod("getversion", "Returns the current (non-dev) BizHawk version")]
+        public static string GetVersion()
+        {
+            return VersionInfo.GetEmuVersionNonDev();            
+        }
+
 		[LuaMethodExample("local nlcliget = client.getavailabletools( );")]
 		[LuaMethod("getavailabletools", "Returns a list of the tools currently open")]
 		public LuaTable GetAvailableTools()
--- a/Version/VersionInfo.cs
+++ b/Version/VersionInfo.cs
@@ -2,8 +2,9 @@ using System.IO;
 
 internal static class VersionInfo
 {
-	public const string Mainversion = "2.0.0"; // Use numbers only or the new version notification won't work
-	public static readonly string RELEASEDATE = "June 25, 2017";
+    // keep this updated at every major release
+	public const string Mainversion = "2.3.0"; // Use numbers only or the new version notification won't work
+	public static readonly string RELEASEDATE = "June 24, 2018";
 	public static readonly bool DeveloperBuild = true;
 	public static readonly string HomePage = "http://tasvideos.org/BizHawk.html";
 
@@ -13,6 +14,11 @@ internal static class VersionInfo
 	{
 		return DeveloperBuild ? ("GIT " + SubWCRev.GIT_BRANCH + "#" + SubWCRev.GIT_SHORTHASH) : ("Version " + Mainversion);
 	}
+
+    public static string GetEmuVersionNonDev()
+    {
+        return Mainversion;
+    }
 
 	static VersionInfo()
 	{


BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
```diff
@@ -391,6 +391,13 @@ public static int Ypos()
 			return GlobalWin.MainForm.DesktopLocation.Y;
 		}
 
+        [LuaMethodExample("local incbhver = client.getversion( );")]
+        [LuaMethod("getversion", "Returns the current (non-dev) BizHawk version")]
```
vadosnaprimer: "stable"?

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
```diff
@@ -391,6 +391,13 @@ public static int Ypos()
 			return GlobalWin.MainForm.DesktopLocation.Y;
 		}
 
+        [LuaMethodExample("local incbhver = client.getversion( );")]
+        [LuaMethod("getversion", "Returns the current (non-dev) BizHawk version")]
+        public static string GetVersion()
+        {
+            return VersionInfo.GetEmuVersionNonDev();
```
vadosnaprimer: Mainversion is public, why not obtain it directly?

Version/VersionInfo.cs
```diff
@@ -14,6 +15,11 @@ public static string GetEmuVersion()
 		return DeveloperBuild ? ("GIT " + SubWCRev.GIT_BRANCH + "#" + SubWCRev.GIT_SHORTHASH) : ("Version " + Mainversion);
 	}
 
+    public static string GetEmuVersionNonDev()
```
vadosnaprimer: Mainversion is public, why not obtain it directly?

Version/VersionInfo.cs
```diff
@@ -2,8 +2,9 @@
 
 internal static class VersionInfo
 {
-	public const string Mainversion = "2.0.0"; // Use numbers only or the new version notification won't work
-	public static readonly string RELEASEDATE = "June 25, 2017";
+    // keep this updated at every major release
```
vadosnaprimer: It's being updated, just not merged to master. I guess all we need is merging it back after release. @adelikat?

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
Asnivor: Yeah, that might make more sense...

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
Asnivor: I was thinking about appending or prepending something to the string to denote whether it was a release or development build, but then didn't. Can't make up my mind whether this is something that would be useful? If not, we can return directly.

Version/VersionInfo.cs
Asnivor: It makes no difference what the value here is from a release perspective. This appears to be generated automatically in a release build process anyway (if you check the version of 2.3.0 in bizbox it actually returns the correct version).

Therefore I believe we should just set it in master here so that development versions have access to the current stable version number.

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
vadosnaprimer: If we want to provide interim version (somehow) through lua, it could also obtain the interim version directly I think.

Version/VersionInfo.cs
vadosnaprimer: It's set manually in the release branch.

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Client.cs
Asnivor: Done

Version/VersionInfo.cs
Asnivor: Ahh ok. Then it doesn't matter how it gets updated in master, as long as it does :)

Version/VersionInfo.cs
Asnivor: done

Issue #1257, opened 2018-08-01
[Request] Minor QOL changes to RAM Watch


GabrielSorensen:
Two minor changes that would help out in ram watch.

1. Adding a menu option to send the selected watch to the top or bottom of the list. 
2. The ability to add notes to row separators, to act as section headings.

(1.5) Really I would like to be able to send the watch to the next line separator, but I feel this would just end up adding visual clutter to the menus.

Asnivor:
I've created a PR that handles 1. and 2. It may or may not get accepted.

PR #1315, opened 2018-09-10
RAMWatch: Enhancements - ticket #1257

Asnivor:
See ticket #1257 

This PR takes care of 1. and 2.

**1. Adding a menu option to send the selected watch to the top or bottom of the list.**

Implemented in top menu and context menu ('Move Top' and 'Move Bottom') along with keyboard shortcuts and new icons:

image

This works for both single and multiple selections.

**2. The ability to add notes to row separators, to act as section headings.**

I've implemented this by making a few changes to _SeparatorWatch.cs_ and added an additional _SeparatorEditor.cs_ control. The user can now right click on a Separator row and edit it. This brings up the dialog where a name can be set: 

image

This text is technically stored in the watch _Notes_ field (which is not used in a Separator) but is displayed in the Address column.

I've made some small changes to _SeparatorWatch.cs_ and _WatchList.cs_ so that this new Separator data is saved and loaded successfully.

Lastly, a small modification so that the Separator name is displayed correctly within the client when 'Display Watches On Screen' is enabled:

image


vadosnaprimer:
I love the features, but why use a custom dialog when we have InputPrompt class exactly for that?

Asnivor:
Because I know nothing :)

I'll take a look at swapping that out tomorrow.

GabrielSorensen:
Thanks @Asnivor

Asnivor:
@vadosnaprimer requested change has been made

vadosnaprimer:
@GabrielSorensen please check the build
https://ci.appveyor.com/project/zeromus/bizhawk-udexo/build/0.0.0.2350/artifacts

Is it working the way you wanted?

GabrielSorensen:
Looks great, works great. Even displaying on screen is fantastic.
(My c# is garbage or I would be more willing to make my own pull reqs, sorry mates.)
I'm sure other people will also use these changes.
Thanks.

PR #1316, opened 2018-09-10
lua: forms.drawText() horizontal alignment "right" fix

nog5:
The lua function forms.drawText() horizontal alignment "right" did the same as "center"


vadosnaprimer:
Please post comparison pics for easy grasp.

nog5:
Before it was center aligned on both "center" and "right"
befor

After
after

Issue #314, opened 2015-06-22
Import Invalid files in Firmware Dialog


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Open EmuHawk's Firmware Dialog
2. Drag and Drop a large collection of files (doesn't need to have any BIOS 
files)
3. Enjoy waiting for EmuHawk to extract EVERY ROM you just told it to load and 
having invalid files in the Firmware folder.

What is the expected output? What do you see instead?
EmuHawk imports anything that the Import and Drag and Drop Functions tell it 
to.  Even if they are the wrong size or invalid (double check size but I am 
pretty sure).

What version of the product are you using? On what operating system?
r8193

Please provide any additional information below.
Uncompressed Files will also be imported.  What fun!

```

Original issue reported on code.google.com by `hegyak` on 18 Nov 2014 at 4:12

PR #1321, opened 2018-09-13
FirmwaresConfig: Stop indiscriminate importing (and some other things…

Asnivor:
…) - #314 

Another issue from the vaults...

Addresses ticket #314 

This PR:

* Adds a bool return method to _FirmwareManager.cs_ that tests whether a file you are trying to import matches the size and hash of anything in the _FirmwareDatabase_.
* This is then tested in _FirmwaresConfig_ whenever files are imported (either by the 'Import' button or via drag'n'drop) so that files that dont match are not imported
* Added an 'Open Firmware Folder' button in the _FirmwaresConfig_ dialog so users can manually copy in custom firmware if they desire
* Updated the 'Set Customization' context menu click event so that if the file selected does **not** reside in the firmwares folder, the user is given the option to copy it there before the customization is set (so users have a method of getting custom firmware into the firmware folder without manually copying it).

### DIFF
--- a/BizHawk.Client.Common/FirmwareManager.cs
+++ b/BizHawk.Client.Common/FirmwareManager.cs
@@ -128,7 +128,6 @@ namespace BizHawk.Client.Common
 
         /// <summary>
         /// Test to determine whether the supplied firmware file matches something in the firmware database
-        /// For customization overrides then perhaps the user can copy into the firmware directory manually?
         /// </summary>
         /// <param name="f"></param>
         /// <returns></returns>
--- a/BizHawk.Client.Common/FirmwareManager.cs
+++ b/BizHawk.Client.Common/FirmwareManager.cs
@@ -126,7 +126,38 @@ namespace BizHawk.Client.Common
 			public Dictionary<string, RealFirmwareFile> Dict { get; } = new Dictionary<string, RealFirmwareFile>();
 		}
 
-		public void DoScanAndResolve()
+        /// <summary>
+        /// Test to determine whether the supplied firmware file matches something in the firmware database
+        /// For customization overrides then perhaps the user can copy into the firmware directory manually?
+        /// </summary>
+        /// <param name="f"></param>
+        /// <returns></returns>
+        public bool CanFileBeImported(string f)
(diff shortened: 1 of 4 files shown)


hegyak:
This is one of my old tickets.  I found/did weird stupid stuff to BizHawk back then.

Asnivor:
Well tbh, its not that weird/stupid :)

It has always bugged me that if I dragged archives full of firmware into bizhawk there was no checking done.

zeromus:
Uhmmm I'm not sure the concept of "Import" here is to only import recognized files, is it? Using the size as a heuristic for what might be recognized, when recognition isn't even the aim, seems weird. So I think the result is, people now have a seemingly random subset of the files they selected. That's not great.

All the other stuff sounds good though.

IF importing what's recognized is what's desired, then using size to speed up processing of large data sets would be a good idea. But... that isn't what's desired.

Asnivor:
Hmmm, I never thought about the concept of "Import" as being at odds with the originally reported issue. I guess the takeaway here is that I should ask questions of a 3 year old issue before solving it :)

How does this sound:  By default the firmwares dialog will only import things that match known firmwares in the database, BUT we have an additional checkbox on the form that overrides this behavior? Maybe, "Allow unrecognized firmware import"?

vadosnaprimer:
Just a note: I simply have a global firmware folder with all sorts of crap in it, and for every new release of BH, I point to that folder, and it detects everything. The only thing I care about is this approach working.

Asnivor:
@feos 

This PR doesnt touch any of the folder scanning stuff. It is purely modifying the importing functions.

zeromus:
compromise solution sounds good in principle, but as I haven't tested this code, I'm not sure which form would have the checkbox on it.

Asnivor:
Up on the top bar, right of the  "Import" button probably makes sense. Although it could go in the status bar area aligned right I support.

zeromus:
yeah sure checkbox by import works for me.

Asnivor:
@zeromus 

This has been done.

image

PR #1355, opened 2018-10-22
Create JoypadIntersection.lua

Gikkman:
Adds a small script which, when run, only allows input to P1's controller if both P1 and P2 holds down the specific button simultaneously. Also useful since it demonstrates joypad manipulation.


adelikat:
The script looks good, but what is the use case for something like this?

Gikkman:
I guess it is more of a gimmick than anything with an actual use case. If anything, it shows how to properly do the "reset joypad" trick, which isn't entirely straight forward.

I submitted this PR since zeromus asked me to, as compensation for him helping me with researching how to actually unset a button that you programatically pressed.

vadosnaprimer:
This could be mentioned in the docs too.

Issue #786, opened 2017-01-28
Virtual Pad buttons get stuck + crash


staticfox:
### Current setup:
* Windows 10
* Downloaded from AppVeyor Artifact commit dd19a69b6
* Super Mario 64 (USA)

### Steps to reproduce
1) Load EmuHawk.exe
2) File -> Open Rom -> Super Mario 64 USA
3) Tools -> TAStudio
4) Tools -> Virtual Pad
5) Click S (or A, B, Z, doesn't matter)
6) Advance a frame in TAStudio (Or hit play/pause)
7) The selected button from step 5 is now stuck until you start the entire emulator

If you try to close Virtual Pad and TAStudio then load TAStudio it throws this exception:
```
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at BizHawk.Client.EmuHawk.InputRoll.DoBackGroundCallback(PaintEventArgs e, List`1 visibleColumns)
   at BizHawk.Client.EmuHawk.InputRoll.DrawBg(PaintEventArgs e, List`1 visibleColumns)
   at BizHawk.Client.EmuHawk.InputRoll.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.Control.PaintWithErrorHandling(PaintEventArgs e, Int16 layer)
   at System.Windows.Forms.Control.WmPaint(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

Let me know if you need any additional information. Thanks!

vadosnaprimer:
Switch to recording mode in tastudio to use virtualpad.

alyosha-tas:
This also seems fixed I think? It seems to work as intended now without crash.

Might want to double check @vadosnaprimer

vadosnaprimer:
I don't have time atm. If it works just close it.

PR #1368, opened 2018-11-14
Attempt to resolve #832 , needs a lot of testing probably

### LINKED ISSUE #832: Mid values aren't set to dualshock analogs when tastudio autoloads
If you set the controller to dualshock for psx, and tastudio to autoload, the initial values for analog controls are set as all zeros instead of FloatRange Mid.

Turning Recording mode on in tastudio fixes it. And if you restart tastudio, mid ranges are also properly used. But if you restart the whole app, all zeros are again there.

alyosha-tas:
The previous logic was always latching input from the background global controller state, which is not what we want for default state. 

Separately this was also happening when appending new frames by drag and drop, which is obviously not what we want.

I think this PR resolves these issues without any regressions, but of course it needs a lot of verification before merging.

### DIFF
--- a/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
+++ b/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
@@ -69,7 +69,6 @@ namespace BizHawk.Client.Common
 		public void SetFrame(int frame, string source)
 		{
 			ChangeLog.AddGeneralUndo(frame, frame, "Set Frame At: " + frame);
-
 			SetFrameAt(frame, source);
 			InvalidateAfter(frame);
 
@@ -392,7 +391,7 @@ namespace BizHawk.Client.Common
 			int oldLength = InputLogLength;
 			ChangeLog.AddGeneralUndo(oldLength, oldLength + numFrames - 1);
 
-			Global.MovieSession.MovieControllerAdapter.LatchSticky();
+			//Global.MovieSession.MovieControllerAdapter.LatchSticky();
 
 			var lg = LogGeneratorInstance();
 			lg.SetSource(Global.MovieOutputHardpoint); // account for autohold. needs autohold pattern to be already recorded in the current frame
--- a/BizHawk.Client.Common/movie/MovieSession.cs
+++ b/BizHawk.Client.Common/movie/MovieSession.cs
@@ -143,8 +143,11 @@ namespace BizHawk.Client.Common
 			if (Global.Emulator.Frame == 0)
 			{
 				MovieControllerAdapter.LatchFromSource(input);
-				(Movie as TasMovie).GreenzoneCurrentFrame();
-				Movie.RecordFrame(Global.Emulator.Frame, Global.MovieOutputHardpoint);
+				if (Movie.IsRecording)
+				{
+					(Movie as TasMovie).GreenzoneCurrentFrame();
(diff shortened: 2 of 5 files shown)


BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
```diff
@@ -392,7 +391,7 @@ private void ExtendMovieForEdit(int numFrames)
 			int oldLength = InputLogLength;
 			ChangeLog.AddGeneralUndo(oldLength, oldLength + numFrames - 1);
 
-			Global.MovieSession.MovieControllerAdapter.LatchSticky();
+			//Global.MovieSession.MovieControllerAdapter.LatchSticky();
```
adelikat: Maybe just delete thsi?

alyosha-tas: I guess I can delete it permanently once everyone is happy with it going away. Honestly I couldn't grasp what it's intent was in the first place, all I know is that when appending frames to TAStudio without the movie playing that line causes errant float inputs to pop up.

Does it have a different intended use in any other scenario? I don't know.

vadosnaprimer:
You're touching logic here, and this scares the shit out of me after what we've been through. The problem is that dualshock is initialized somehow differently, with regular controller everything works fine. I don't see how anything related to dualshock is being fixed.

zeromus:
Regular controller doesn't have analog axes. I don't understand how this is relevant.

alyosha-tas:
I don't have a dualshock controller. I just tested with NES zapper bound to mouse. This fixed that case. If there are other unrelated cases yeah I can't fix them.

zeromus:
I dont know what default value for the zapper even makes sense. There's at least some justification that a joystick axis should be neutral, even though that's as arbitrary as anything else for an initial value THE USER CAN'T CHOOSE, but whats the neutral zapper coordinate? IMO the whole situation is bolloxed and we need to let users pick input for frame zero at t=0, but nobody likes that idea.

alyosha-tas:
Well, I checked dualshock controller and it seems to be working correctly. TAStudio no longer starts up with zeroes when analog controls are bound to mouse. I tried both drag and drop with autoload and starting TAStudio from the menues. So, it seems fixed to me I think.

@vadosnaprimer I think what was wrong was the logic with latching mouse states , not anything with the controller. I think the reason you were getting zeroes was the actual location of TAStudio relative to the main window. This means when you click play or frame advance in TAStudio, the mouse just happens to be at 0,0 relative to the window because that's physically where the button is located.  Indeed if I move the TAStudio window to the other side before starting, I get 255,0 instead of 0,0 with the old logic.

@zeromus: Well, picking mid values is easy because that is what exists in the controller definition already. Passing user defined defaults to TAStudio seems as simple as putting them there in the SetSticky method I made if such a global option exists. How to get them there in the first place though is a bit beyond me.

vadosnaprimer:
I'll hopefully dig into this within a week.

alyosha-tas:
This might also fix #786 (at least I can't reproduce that crash again)

alyosha-tas:
Fixed a bug with invalidating greenzones when opening a movie, oops.

vadosnaprimer:
Please make a new PR that only fixes the initial problem #832 had. There have been so many bonus commits fixing things that are already fixed in master, this PR is a mess right now.

PR #1371, opened 2018-11-18
This fixes a bug in TAStudio where you cannot use '<'to get to frame 0

alyosha-tas:
This resolves the TODO involving runFrame, which is supposed to be false when we want to be at frame zero, but is currently always true.

I don't believe it has any side effects from other code paths, and I tested it out in my own projects, but it could use a second look just to be sure

### DIFF
--- a/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/BizHawk.Client.EmuHawk/MainForm.cs
@@ -4461,7 +4461,7 @@ namespace BizHawk.Client.EmuHawk
 					if (isRewinding)
 					{
 						runFrame = true; // TODO: the master should be deciding this!
-						Master.Rewind();
+						Master.Rewind(ref runFrame);
 					}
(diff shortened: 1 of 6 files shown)


adelikat:
Ugly bug works. We need to rethink how IControlMainForm interacts with runFrame

vadosnaprimer:
I still need to look at it >_>

BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -4461,7 +4461,7 @@ private bool Rewind(ref bool runFrame, long currentTimestamp, out bool returnToR
 					if (isRewinding)
 					{
 						runFrame = true; // TODO: the master should be deciding this!
```
vadosnaprimer: `runFrame = Emulator.Frame > 1;` here replaces this entire PR.

PR #1376, opened 2018-11-24
Sync TAStudio Recording mode to movie state

alyosha-tas:
Fixes a bug where the current movie can be in a different recording state then the displayed TAStudio state (according to the checkbox.)

This happens when appending frames by clicking and dragging input.

Note: Probably / maybe, this should happen anytime input is changed by manually using the piano roll, but it currently only happens in this instance. Maybe that can be future work?

### DIFF
--- a/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/BizHawk.Client.EmuHawk/MainForm.cs
@@ -480,25 +480,25 @@ namespace BizHawk.Client.EmuHawk
 					GlobalWin.Tools.LuaConsole.ResumeScripts(false);
 				}
 
+				if (GlobalWin.Tools.Has<TAStudio>())
+				{
+					// There are instances where the movie will be switched out of recording mode but TAStudio doesn't know about it
+					// update the state here
+					if (GlobalWin.Tools.TAStudio.TasPlaybackBox.RecordingIsChecked != Global.MovieSession.Movie.IsRecording)
+					{
+						Global.MovieSession.ReadOnly = true;
+						GlobalWin.Tools.TAStudio.WasRecording = false;
+						GlobalWin.Tools.TAStudio.TasPlaybackBox.RecordingIsChecked = Global.MovieSession.Movie.IsRecording;
+					}
+
+				}
+
 				StepRunLoop_Core();
 				StepRunLoop_Throttle();
 
 				Render();
 
 				CheckMessages();
-				
-				if (GlobalWin.Tools.Has<TAStudio>())
-				{
-					// There are instances where the movie will be switched out of recording mode but TAStudio doesn't know about it
-					// update the state here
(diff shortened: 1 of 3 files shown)


BizHawk.Client.EmuHawk/tools/TAStudio/PlaybackBox.Designer.cs
```diff
@@ -192,7 +192,7 @@ private void InitializeComponent()
 		private System.Windows.Forms.CheckBox AutoRestoreCheckbox;
 		private System.Windows.Forms.CheckBox TurboSeekCheckbox;
 		private System.Windows.Forms.CheckBox FollowCursorCheckbox;
-		private System.Windows.Forms.CheckBox RecordingModeCheckbox;
+		public System.Windows.Forms.CheckBox RecordingModeCheckbox;
```
adelikat: You can't do this.  This is auto-generated code.  As soon as a change is made, it will put it back to private again.

Instead, if you need access to a proprety of one of these controls make a public property getter/setter in the non-generated part of this class:
public bool RecordingIsChecked { get { return RecordingModeCheck.Checked; } { set { RecordingModeCheck.Checked = value; } }

vadosnaprimer:
I don't understand what the problem is.

alyosha-tas:
open TAStudio
click on recording mode
click and drag some inputs so that new frames are added

You will find that the recording mode check box is still checked, even though the underlying state changed so that it should be unchecked. (You have to click the box twice before it gets unchecked.)

BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -480,6 +480,19 @@ public int ProgramRunLoop()
 					GlobalWin.Tools.LuaConsole.ResumeScripts(false);
 				}
 
+				if (GlobalWin.Tools.Has<TAStudio>())
+				{
+					// There are instances where the movie will be switched out of recording mode but TAStudio doesn't know about it
```
vadosnaprimer: Reproduced, and it seems it's emuhawk who doesn't know it's disabled read|write, because look at the movie mode icon in the mainform. It seems it doesn't know it's not recording anymore if we're appending frames out of order, as in, at the current frame and after. So instead of fixing just tastudio, we need emuhawk to know to update itself.

alyosha-tas: That can be fixed with a `SetMainformMovieInfo();`
But, does this situation occur anywhere outside of TAStudio? If not I can just throw that in after the checkbox update.

If it does, well I guess I don't even know where to look to reproduce it.

vadosnaprimer:
All in all, please post reproduction steps as issues instead of pull requests. At this point, there's about 1.5 people understanding how tastudio is supposed to work, and it's definitely not working the way that's easy to guess.

PR #1237, opened 2018-07-02
Allow 2.3 to be built and run w/ MSBuild and Wine+Mono on GNU+Linux

YoshiRulz:
Tested on Manjaro, Linux 4.18.0-1-MANJARO, MSBuild 15.6.0.0 (`msbuild-stable` on the AUR), Mono 5.12.0 (`extra` repo), Wine 3.11 (`multilib` repo, needed for its bundled libraries).

Changes:
- Update to latest OpenTK, add input code from `mono-portable` branch (kudos), and fix bugs arisen from version differences
- Move some functionality into Win-specific classes and add Wine+Mono implementation where simple
- Add conditions to main loop so some things only run on Windows

Steps to build and run: 
```
msbuild BizHawk.sln
cd output
ln -s dll/nlua/NLua.dll   # not sure why, but putting this in LD_LIBRARY_PATH wasn't working
LD_LIBRARY_PATH="/usr/lib/wine" mono EmuHawk.exe
```
I promise no functionality beyond Help>About.

***

Cores known to crash:
* Cygne
* Gambatte
* Genesis Plus GX
* Handy
* Mupen64Plus
* Neo-Geo Pocket (dunno what it's called)
* NESHawk
* Octoshock
* Snes9x
* TI83Hawk (or I got a corrupted ROM)
* VBANext

Cores that may work:
* Atari2600Hawk
* A7800Hawk

PR #1380, opened 2018-11-28
Add preliminary Unix compatibility (using Mono; resolves #1384)

### LINKED ISSUE #1384: YoshiRulz portability commit broke opengl display method
@YoshiRulz 

https://github.com/TASVideos/BizHawk/commit/ca021d96cc89235df272c067ce6a2d8abeba22ec

This commit broke opengl display method because you didnt update OpenTK.GLControl.dll at the same time you updated OpenTK.dll. Please update the GLControl dll and fix whatever bugs appear consequently

YoshiRulz:
**tl;dr**:
* Once you've installed the build and runtime dependencies (see third and fourth tables) and built the solution (see end of desc.), EmuHawk will run on Mono using libraries from Wine.
* You can play games with the keyboard (no controllers, sorry) on 9 systems in total (9 cores + NullHawk), plus one more if you build it yourself (see first table).
* You can use the Virtual Pad, savestates, and Hex Editor (and maybe some other stuff, see second table).

Before merging:
* [x] Test everything on all the working cores (on Unix)
* [x] Prevent anything to do with Lua from running on Unix
* [x] See if any of the tools that didn't open have easy fixes
* [x] Fix OpenTK X keyboard interface
* [x] ~~Double-check nothing breaks functionality on Windows (except #1384, which was fixed in 318fd36)~~ ~~Apologize for ruining performance on at least one core~~ Pretend nothing happened

Extra open problems:
* Rewrite `MemoryBlock` in a way that doesn't use `kernel32.dll`, or uses an alternative for Unix platforms
	* see [nattthebear's comment](https://github.com/TASVideos/BizHawk/pull/1380#issuecomment-443875933)
* Figure out if/how to use system `7z.so` instead of bundled `7z.dll`
* See if T.S.T. or PicoDrive work on Unix
* ~~Try to use Mono's `xbuild` instead of `msbuild`~~ Mono devs deprecated `xbuild`
* Try to get TAStudio (+ Macro Tool) to open
	* see [my comment](https://github.com/TASVideos/BizHawk/pull/1380#issuecomment-443399720)
* Try to get Code-Data Logger to open
	* Mono's winforms throws `System.InvalidOperationException` when `ListViewItemCollection.Clear()` is called on [CDL.cs:112 when added](https://github.com/TASVideos/BizHawk/blob/4942ae7905dcf87856871c8a009d579cb3158514/BizHawk.Client.EmuHawk/tools/CDL.cs#L112), [CDL.cs:101 now](https://github.com/TASVideos/BizHawk/blob/master/BizHawk.Client.EmuHawk/tools/CDL.cs#L101)
* Try to get Toolbox working
* Try to get Basic Bot, Cheats, and RAM Search to open
* Try to get the OpenTK gamepad interface working
* Clean up libblip_buf (see [libblip_buf issue tracker](https://gitlab.com/TASVideos/libblip_buf/issues))
* Test libretro
* Create build scripts for your distro
	* libblip_buf: [Arch/Manjaro](https://aur.archlinux.org/packages/libblip_buf)
	* BizHawk

***

Current progress:

Status | Cores
--:|:--
Loads roms! 🎉 (and NullHawk) | A7800Hawk, Atari2600Hawk†, C64Hawk, ColecoHawk, GBHawk, IntelliHawk, NesHawk, NullHawk, PCEHawk, SMSHawk, ZXHawk
Needs the bundled 7zip library | TI83Hawk, Virtu
Needs `MemoryBlock` to be rewritten | BSNES, Genplus-gx, NeoPop, SameBoy, Saturnus, Snes9x, Virtual Boyee†
Tries to load bundled `.dll`‡ | Cygne, Gambatte, Handy, mGBA, Mupen64Plus, Octoshock, QuickNes, VBA-Next

† requires workaround: comment out `DefaultValue` attr for `A2600Settings.BackgroundColor` (Atari2600Hawk) or for all `VirtualBoyee.Settings.*Color` (Virtual Boyee)

‡ The only way I see to get these cores working is to build them from source again but as `.so`. This is probably not feasible given they're at various points behind upstream, but it would in theory move them up with the "working" native cores as their libraries' logics would be the same as on Windows.

Feature | Works? | Cores tested† | N/A to Cores
--:|:--|:--|:--
Batch Runner | opens, untested | - | -
Core Features | works perfectly | - | -
Forums / Online Help | works perfectly | - | -
Multi-disk Bundler | opens, doesn't work | - | -
|||
AutoHawk (experimental) | opens | * | NullHawk
Basic Bot | opens, contents don't draw | * | NullHawk
Cheat Code Converter | opens, untested | GBHawk, NesHawk, SMSHawk | A7800Hawk, C64Hawk, ColecoHawk, IntelliHawk, NullHawk, PCEHawk, ZXHawk
Cheats | doesn't open | * | NullHawk
Code-Data Logger | crashes | GBHawk, NesHawk, PCEHawk, SMSHawk, ZXHawk | A7800Hawk, C64Hawk, ColecoHawk, IntelliHawk, NullHawk
Debugger | opens, untested (crashes on GBHawk) | * | NullHawk
Hex Editor | "works" (on NesHawk) | * | NullHawk
Lua Console | disabled | *, NullHawk | -
Macro Tool | crashes | *, NullHawk | -
New Hex Editor (experimental) | crashes | * | NullHawk
RAM Search | doesn't open | * | NullHawk
RAM Watch | opens, untested | * | NullHawk
Record AVI/WAV | crashes (FFmpeg) | *, NullHawk | -
Savestates | works perfectly | * | NullHawk
Screenshot | works perfectly | *, NullHawk | -
TAStudio | crashes very well | * | NullHawk
Trace Logger | opens, untested | * | NullHawk
Toolbox | crashes ("works" on NullHawk) | *, NullHawk | -
Virtual Pad | "works" | *, NullHawk | -

† Where `*` indicates all cores that load roms: A7800Hawk, C64Hawk, ColecoHawk, GBHawk, IntelliHawk, NesHawk, PCEHawk, SMSHawk, and ZXHawk

I got RAM Search to open once on A7800Hawk, it didn't draw window contents (like the Basic Bot).

***

*Compiling* requires MSBuild and *running* requires Mono and WINE, but **BizHawk does not run under WINE** (nor is that my goal). Only the bundled libraries are required. In the below tables:
* Linux Mint -> Ubuntu
* macOS is assumed to be Homebrew
* Manjaro -> Arch Linux

Distro | Mono | MSBuild | WINE
--:|:--|:--|:--
**Arch Linux** | [`mono-basic`](https://aur.archlinux.org/packages/mono-basic) | [`msbuild-stable`](https://aur.archlinux.org/packages/msbuild-stable) | `wine`
**CentOS** | ? | ? | ?
**Debian** | ? | ? | ?
**elementaryOS** | ? | ? | ?
**Fedora** | `mono-basic` | from source | `wine-core`
**Gentoo** | ? | ? | ?
**Hyperbola** | ? | ? | ?
**macOS** | ? | ? | ?
**MX Linux** | ? | ? | ?
**NixOS** | ? | ? | ?
**openSUSE** | `mono-basic` | from source | `wine-32bit`
**Solus** | ? | ? | ?
**SteamOS** | ? | ? | ?
**Ubuntu** | `mono-devel` | [`msbuild`](https://www.mono-project.com/download/stable/#download-lin-ubuntu) | `libwine`

Additional runtime dependencies:

Distro | libblip_buf
--:|:--
**Arch Linux** | [`libblip_buf`](https://aur.archlinux.org/packages/libblip_buf)
**CentOS** | TODO
**Debian** | TODO
**elementaryOS** | TODO
**Fedora** | TODO
**Gentoo** | TODO
**Hyperbola** | TODO
**macOS** | TODO
**MX Linux** | TODO
**NixOS** | TODO
**openSUSE** | TODO
**Solus** | TODO
**SteamOS** | TODO
**Ubuntu** | TODO

***

Building is as easy as:
```sh
git clone https://github.com/YoshiRulz/BizHawk.git && cd BizHawk   # (or git@github.com:YoshiRulz/BizHawk.git)
msbuild BizHawk.sln
```
* MSBuild doesn't like paths containing "+"! You'll get something like `/usr/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(1348,5): error MSB3073: The command ""..."" exited with code 126. [/path/with/+/Version/Version.csproj]`.

Running is even easier, just execute `EmuHawkMono.sh` in the repo's `output` folder (move/rename this folder wherever you'd like unless you're debugging).
* If you see "Unknown distro" in the script's output, you may need to add your distro to the case statement, setting `libpath` to the location of `d3dx9_43.dll.so` (please do share if you get it working).
* `File > Exit (Alt+F4)` doesn't terminate the process correctly, you'll need to send SIGINT (`^C`).

If libblip_buf isn't yet available for your distro, you can easily build it:
```sh
git clone https://gitlab.com/TASVideos/libblip_buf.git && cd libblip_buf   # (or git@gitlab.com:TASVideos/libblip_buf.git)
make so
libpath="/usr/lib"   # or wherever your distro's library path is
sudo install -Dm755 libblip_buf.so.1.1.1 "$libpath/libblip_buf.so"
# optionally install as libblip_buf.so.1.1.1, then link .so.1 -> .so.1.1.1 and .so -> .so.1
```

See also #1237, in which I make `mono EmuHawk.exe` not crash immediately.
This PR resolves #1384.


YoshiRulz:
Can't get DllMaps to work, I've added this to `output/EmuHawk.exe.config` and nothing changed:
```xml
<dllmap dll="mupen64plus" target="libmupen64plus.so.2"/>
```

YoshiRulz:
Can't get DllMaps to work, I've added this to `output/EmuHawk.exe.config` and nothing changed:
```xml
<dllmap dll="mupen64plus" target="libmupen64plus.so.2"/>
```
I'll continue tomorrow.

vadosnaprimer:
I'm not opposed to trying N64 first, this is your project so you can start with whatever you like. I only suggested a C# core because it's native, so maybe it'd just work as soon as mono is fully set up. I don't know if N64 is currently available even in the Mac release, @Sappharad? But if you somehow manage to make it run, wow that'd be insane!

YoshiRulz:
@vadosnaprimer There are cores written in C# in the codebase? I only changed to N64 because I have the non-libretro libraries installed (the libretro "core" was too confusing).

YoshiRulz:
I can now play Pac-Man! Sorta!

All I did was comment out the `DefaultValue` attribute on `A2600Settings.BackgroundColor`, as you can see behind the game. What I haven't figured out is why the reflection fails in `SettingsUtil.CreateSetter`. With the attribute uncommented, `object value = attr.Value` (line 69) makes `value` null, causing `value.GetType()` (line 71) to throw a NullReferenceException. I assume these attributes are actually being used as defaults somewhere and commenting isn't an acceptable solution?

vadosnaprimer:
Hahaha nice! We have 6502 and z80 implemented in C#, so cores that use them are also in C#: NESHawk, A2600Hawk, GBHawk, ZXHawk, some others maybe.

NarryG:
Useless knowledge to the rescue:

C# Cores:  
NESHawk
A2600Hawk
A7800Hawk
GBHawk
ZXHawk
SMSHawk
IntelliHawk
ColecoHawk
C64Hawk
Virtu 
TI83Hawk


Unmanaged cores:
mGBA
VBA-Next
Mupen64Plus
Octoshock
Saturnus
PCEHawk
Bizswan
Bizlynx


Waterbox (magic) cores:
bsnes
snes9x
GPGX
PCFX
Sameboy
Picodrive
Virtualboy

NarryG:
Also, those defaults are being used in the settings config, so commenting them out is going to break things as it stands.

image
image

YoshiRulz:
Thank you for the useless knowledge! Going back to the original NES plan (but with NesHawk, which I somehow overlooked), I've immediately run into the same .dll loading issue as the unmanaged cores. That being that mono is unable to find them even when they're in the executable's directory and said directory is in `LD_LIBRARY_PATH`. In this case it's `blip_buf.dll`, required for an instance variable in `NES.MagicSoundProvider`.

I'll leave it there for today but I'll be back! I'm thinking sound isn't as important as playing the game, maybe there's a way to discard it (I've not yet looked).

alyosha-tas:
blip_buf is used in most of the C# cores. It's only due to A2600's age that it isn't used there yet (though it probably should be at some point.)

It would be relatively easy to comment out all the blip_buf sound stuff though, if there is no other option.

Sappharad:
Several years ago before I started doing native builds of cores for macOS, I had thrown together some shitty C# replacement code for blip_buf calls. At some point the portable branch was re-forked from scratch and I just compiled blip_buf as a native library instead so I could replace less code. If it's at all useful, maybe you could try and dig up the crappy C# implementation and use that. I don't know how far back you'd have to go to get it, and actually wouldn't recommend it.

Why don't you just build the library and get it over with? mono will automatically look for lib<dllname>.so and load it instead of the windows .dll for you. No mappings required.

YoshiRulz:
IT'S-A ME
`if (unix) return;` works!

@Sappharad by "just build the library" do you mean compile [this](https://code.google.com/archive/p/blip-buf/)? I'm afraid I know less about C than I do about C#.

YoshiRulz:
zxhawk
*Ghostbusters (1984)* running in ZXHawk... I think...

Asnivor:
@YoshiRulz switch the core to 128k mode and you can just press enter to start loading the game.
Be nice to see whether the core loading/playing processes are working correctly.

BizHawk.Emulation.Common/Sound/Utilities/BlipBuffer.cs
```diff
@@ -147,7 +147,7 @@ public void EndFrame(uint clockDuration)
 
 		public int SamplesAvailable()
 		{
-			if (PlatformLinkedLibSingleton.RunningOnUnix) return 0; // temporary workaround
+			if (PlatformLinkedLibSingleton.RunningOnUnix) return 882; // temporary workaround
```
Asnivor: So this is *i think* consistently true for ZXHawk. Probably not for other cores that utilise blipbuffer

YoshiRulz: I was hoping Sappharad or someone else would point me to either a source for blip_buf.so or build instructions because I have no idea how to make such a file. If that can't be done and this value does change with the core, maybe I could add that as a parameter, as `SamplesAvailable` here is actually wrapping the library.

Sappharad: I’ll look tonight.

Rough steps:
Find C++ IDE for Linux
Create library project
Point project to source files
Compile

Normal people would probably use command line tools to do it instead of an IDE. But many Linux IDEs probably generate command line makefiles anyway.

vadosnaprimer:
Man this is incredible!

BizHawk.Emulation.Common/Sound/Utilities/BlipBuffer.cs
NarryG: The blip_buf code is actually in the root of the repo. Modifying the make to output an so instead of a dll and running the makefile built fine. Completely untested as I only have the Linux Subsystem for Windows set up at the moment though.

vadosnaprimer: > Modifying the makefile to output an so instead of a dll and running make built fine.

I really don't think that's how it works. One has to use mingw to build a .dll out of code that builds .so if you just compile on linux.

Sappharad: Windows subsystem for Linux uses native Ubuntu binaries. The .so built by it should be fine for Linux. 

As long as it’s called libblip_buf.so and located in the same place as the windows DLL mono will automatically use it. No Cygwin, that would make a windows DLL and he won’t want that.

vadosnaprimer: I mean it's not achieved by hacking the makefile, is it?

NarryG: When I said modified the makefile I should have been more clear. I literally just modified it to name the output .so rather than .dll. The makefile within the directory literally just calls gcc with a few specific options

NarryG: I went and set up a VM to test it. Got it all built based on the branch but it's yelling at me about not being able to find DX9, VC++ 2010, 2012, and 2015. I'm clearly missing something. @YoshiRulz

vadosnaprimer: It's looking for prereqs. Are these even available on linux?

NarryG: I assumed the Mono branch would deal with that but it appears it doesn't. There may be some dependency besides just mono required? @YoshiRulz Could probably elaborate

Sappharad: Yeah, I’m curious how that’s supposed to work too. I ran into that a few weeks back when I pulled the latest code into my fork and had to replace a bunch of if(runningOnUnix) checks with #if WINDOWS again. Someone mentioned in IRC that it was for WINE so I assumed that was valid.

vadosnaprimer: Wine doesn't run 2.x bizhawk at all. There are some insane problems with dependencies. It runs 1.x though, and I dunno if that was intended wither when those edits were made. 1.x running in wine was found out like this year, but never seriously verified or examined.

YoshiRulz: @NarryG @Sappharad It doesn't need WINE, just mono! You'll need WINE *installed* to get those missing libraries, but it's simply `LD_LIBRARY_PATH="/usr/lib/wine" mono EmuHawk.exe` to run it. Also I've forked `master` not `mono-portable` in case that was unclear.

Sappharad: I took a look at it, and the changes @NarryG made seem to be all that is needed. Just update makefile to change the output name from blip_buf.dll to libblip_buf.so and run make at the terminal. It looks like the -shared flag is the key to making an .so and it's already set.

For some reason, GitHub isn't letting me attach it directly to this comment, so here's a compiled build that I made with Ubuntu:
https://www.sappharad.com/junk/libblip_buf.so.zip

Regarding WINE, I think the problem there is that .NET stuff doesn't really work correctly at all in 64-bit WINE. I just tried it myself and the dependencies didn't give me trouble, I created a 64-bit bottle and those installed fine. But the BizHawk blows up trying to use classes that are supposed to be part of the .NET Framework. I guess it'll probably be another year or so before .NET is working properly.

NarryG: I attempted running it that way but it didn't work. Might have been a mistake on my side.

YoshiRulz:
I've copied the build instructions from the other PR and added a table of package names.

BizHawk.Emulation.Common/Sound/Utilities/BlipBuffer.cs
YoshiRulz: Worked for me, though the copy I built was different from @Sappharad's.

YoshiRulz:
If someone has the C64 "Drive1541" firmware, could they link it or try a rom with Mono?

Also, the file selector is hiding .d64 files.

YoshiRulz:
[SEEEEGAAAAAA](youtu.be/GguzZTDGNnQ)
gotta go fast

YoshiRulz:
Opening TAStudio causes `GDIRenderer.CreateNormalHFont` to throw an `EntryPointNotFoundException` on the external function `CreateFontIndirect` from `gdi32.dll`. But the function exists:
```
> objdump -T /usr/lib/wine/gdi32.dll.so | grep CreateFontIndirect
000000000007e810 g    DF .text  0000000000000129  Base        CreateFontIndirectA
000000000007e690 g    DF .text  0000000000000171  Base        CreateFontIndirectW
000000000007e470 g    DF .text  000000000000021c  Base        CreateFontIndirectExA
000000000007e160 g    DF .text  000000000000030c  Base        CreateFontIndirectExW
```
From [what I can tell](https://docs.microsoft.com/en-us/dotnet/framework/interop/specifying-a-character-set), there are two copies for "ANSI" and "wide" strings. Would changing lines 346-347 to read:
```csharp
[DllImport("gdi32.dll")]
private static extern IntPtr CreateFontIndirectW(
```
be a good idea? The linked docs say "wide" has been the default for `CharSet.Auto` since Windows XP.

vadosnaprimer:
The firmware file is called `dos1541` with no extension, and I don't know if it's safe to distribute it.

alyosha-tas:
What part of GBHawk needs 7zip lib? If there is some unintended dependency there I'll gladly remove it.

alyosha-tas:
Looks like it's mis-detecting a compressed format, I think.

Can you try with this file: https://github.com/mills32/CUTE_DEMO/tree/master/0_rom

YoshiRulz:
@alyosha-tas is it supposed to do this? More importantly, how did I grab the one rom that tripped it up...
uhhh

alyosha-tas:
Yeah that's jut a tech demo rom that I test the core with.

Whatever Shantae rom you have is making it by this check (Hawkfile.cs line 466):

`if (ArchiveHandlerFactory.CheckSignature(path, out offset, out isExecutable)'

Then it tries to do 7ip stuff.

YoshiRulz:
@alyosha-tas The call to `FileChecker.CheckSignature` returns `InArchiveFormat.None` for Pokemon Gold and the test rom, and `InArchiveFormat.Tar` for Shantae. It is not a tape archive.

alyosha-tas:
```
            #region Last resort for tar - can mistake
            if (stream.Length >= 1024)
            {
                stream.Seek(-1024, SeekOrigin.End);
                byte[] buf = new byte[1024];
                stream.Read(buf, 0, 1024);
                bool istar = true;
                for (int i = 0; i < 1024; i++)
                {
                    istar = istar && buf[i] == 0;
                }
                if (istar)
                {
                    return InArchiveFormat.Tar;
                }
            }
```

So any file padded with zeroes will be mistakenly detected as tar. That's just silly.

YoshiRulz:
Just realized I didn't show any "proof" for the A7800Hawk core, so here you go:
pew pew
That's the 8 "working" cores, only 17 more to cover every system!

alyosha-tas:
> More? There are 29 cores above, but the `Core` attribute is on 36 classes...

The other cores are:

cpchawk (in development)
dualgambatte (for linked play)
dualneopop (for linked play)
ppsspp (dormant)
libretro (for libretro cores)
02em (not sure if it runs)
uzem (homebrew console)

alyosha-tas:
I tried playing around with the background color problem a little bit. I can get the same null reference exception error in windows if I give it a default value that is not a recognized color (ex. "q"). What seems to happen is that the converter can't convert it from the string to the Type. The known colors are defined in System.Drawing.KnownColor

So, maybe this enumeration isn't present in the package?

YoshiRulz:
> So, maybe this enumeration isn't present in the package?

100 points.
`error CS0234: The type or namespace name 'Drawing' does not exist in the namespace 'System' (are you missing an assembly reference?)`

YoshiRulz:
Question, where do I put the run script so that it gets copied to the output? The script uses `$(dirname $0)` to find itself so it just needs to be in there.
```sh
#!/bin/sh
cd "$(dirname "$0")" && if [ -z "$(ps -C "mono" -o "cmd" --no-headers | grep "EmuHawk.exe")" ]; then LD_LIBRARY_PATH="/usr/lib/wine" mono ./EmuHawk.exe; fi
```

vadosnaprimer:
If you mean you want it to be copied to the output folder, put it into Assets.

YoshiRulz:
I've got blip_buf in [a separate repo](https://gitlab.com/TASVideos/libblip_buf) and I'm going to submit [this PKGBUILD](https://gitlab.com/TASVideos/libblip_buf/snippets/1786798) to the Arch User Repository. If anyone is familiar with the Fedora or Debian/Ubuntu packaging systems, please make a package for it so we don't have to distribute it with BizHawk on those OSes (and can revert d3e978c). I should have a working NixOS machine soon so if nobody beats me to it I'll do the Nix expression.

Warepire:
> I've got blip_buf in [a separate repo](https://gitlab.com/TASVideos/libblip_buf) and I'm going to submit [this PKGBUILD](https://gitlab.com/TASVideos/libblip_buf/snippets/1786798) to the Arch User Repository. If anyone is familiar with the RPM, dpkg, or brew packaging systems, please make a package for it so we don't have to distribute it with BizHawk on those OSes (and can revert [d3e978c](https://github.com/TASVideos/BizHawk/commit/d3e978c20ca74b8916c0880680d2737337a7c0e6)). I should have a working NixOS machine soon so if nobody beats me to it I'll do the Nix expression.

I was asked by @vadosnaprimer to review your PKGBUILD:
Since you're in control of the repository, please implement a proper install rule instead and utilize that.

YoshiRulz:
@Warepire I think [this second draft](https://gitlab.com/TASVideos/libblip_buf/snippets/1787175) does that? Unless you're talking about changing the Makefile, in which case I can only add it to the issue tracker.

Warepire:
I meant to modify the makefile of blip_buf to provide an install target.

YoshiRulz:
For distribution, I was thinking that the BizHawk package in a repo could depend on a virtual package for libraries. This would be provided by both a prebuilt package which would inevitably lag behind, and a distro-native package which in turn would depend on whatever names the libraries go by in that distro. The former putting libraries in `/usr/lib/bizhawk` or equivalent and the latter putting symlinks in the same place. If the distro doesn't have the libraries we need, only the prebuilt ones would be available.

Again, I'm not familiar with every package manager, some might not have virtual packages, this is just an idea.

YoshiRulz:
No clue know how to use it but it didn't crash:
?SYNTAX  ERROR

Asnivor:
@YoshiRulz I believe you just need to create a bundler xml with two ROMs (using the multi-disk bundler tool for GB) then open that.

At least that's worked for me in the past (although not sure whether you need to have gambatte set as the GB core).

vadosnaprimer:
Guide: http://tasvideos.org/forum/viewtopic.php?t=17307

vadosnaprimer:
Re: Lua. There's this:
https://github.com/TASVideos/BizHawk/issues/1093
I dunno why the author never made a pull request.

It has its own issue related to Linux btw:
https://github.com/lsthiros/BizHawk/issues/1

It'd really make sense to make use of this sooner or later.

YoshiRulz:
@vadosnaprimer I'll be honest, Lua is pretty far down my list of things to port. @lsthiros is welcome to it.

YoshiRulz:
And we have keyboard input! I think that's enough to ship, is there some kind of test suite for Windows or is it all manual?

vadosnaprimer:
Everything is manual. If it's ready, we can merge this as a branch and ask people to actually try it out, fixing stuff in the main repo as needed, or with PRs if the fixes or improvements need to be reviewed.

YoshiRulz:
@vadosnaprimer Got it. There certainly is a performance hit by the way: a quick test on Windows has SMB1 running at 60 fps on the 2.3 release and 19 fps on my HEAD (though I get about 30 fps on Linux, gonna credit a slightly better graphics card for that). SM64 ran at 60 fps on both, so it's not every core. I didn't check if there was a NesHawk issue in whatever commit I forked from, so it might not have been me. Input and sound seem unaffected on Windows, and Lua still works.

vadosnaprimer:
Yes it was a performance hit caused by emulation improvements in some in-house cores. Don't worry. Do you have anything else to clean up/finish?

alyosha-tas:
19 fps? That sounds like some logging is happening, since NESHawk should have actually gotten a slight boost in performance.

Either that or it was built in debug mode.

YoshiRulz:
@vadosnaprimer Oh thank goodness. Yes, I'm done (for now), I've got a list of tiny things to do that aren't Mono-related which I'll be working through once this is merged.

Issue #1363, opened 2018-11-04
PCEHawk can crash due to invalid DMA resulting in an out of range read


NarryG:
![image](https://narry.land/zwKLaA.png)

Call stack (old version of Bizhawk but the code in question hasn't changed)
``` 	
BizHawk.Emulation.Cores.dll!BizHawk.Emulation.Cores.PCEngine.VDC.UpdatePatternData(ushort addr) Line 271	C#
 	BizHawk.Emulation.Cores.dll!BizHawk.Emulation.Cores.PCEngine.VDC.RunDmaForScanline() Line 231	C#
 	BizHawk.Emulation.Cores.dll!BizHawk.Emulation.Cores.PCEngine.VDC.ExecFrame(bool render) Line 95	C#
	BizHawk.Emulation.Cores.dll!BizHawk.Emulation.Cores.PCEngine.PCEngine.FrameAdvance(BizHawk.Emulation.Common.IController controller, bool render, bool rendersound) Line 28	C#
```  

Note: To cause this, I had to override read-only on the 21-bit bus and then I fuzzed it. That doesn't mean a buggy game couldn't cause the behavior though.

zeromus:
one of you guys ought to be able to fix this pretty easy...

zeromus:
More likely it needs to be masked down by 1 bit

Asnivor:
Ahh ok. Thats a much better idea as its almost certainly what would happen on real hardware.

Asnivor:
This should now be fixed in the latest dev build.

@NarryG If you do happen to be able to break this in the same way as before, please reopen this issue.

PR #1387, opened 2018-12-03
PCEHawk - DESR enforce 16bit register size - #1363

Asnivor:
So I think this should do for #1363

According to here:  http://www.emulatronia.com/doctec/consolas/pce/vdcdox.txt
DESR is only 16bit.

```
17($11) R?/W    DESR - '(DMA) Destination Address Register'

                     b 15-0  This register sets the destination 
                             address for DMA transfers.
                             All bits used 
                             (although no VRAM above $7FFF).
```

@NarryG can you test?

@zeromus can you make sure I have not been stupid?


zeromus:
That's not the way I would do it. The docs say that all the bits are there, and there simply isn't that much VRAM available. That suggests the logic should be straightforward and unmasked up until the last minute where the address is attached to the memory device. at that last moment, it should be masked to the memory device size

Asnivor:
@zeromus like so?

zeromus:
yep. i dont even need to think about that to know it's an improvement. any time a 16bit value is used as an index to a 15bit array, it should be masked like that. even if you could do the analysis properly and ensure it was a small value earlier, that probably isn't the way the hardware worked--frequently it's possible to address stuff at crazy locations and count on mirroring behaviour
you can merge this

Issue #1186, opened 2018-04-27
[Request] Apple II joystick implementation


phoenix1291:
Some games can only be played with the apple II joystick. (Can't found a list, sadly)
So is it possible to implement support for it?

I take this opportunity to ask another question, is support for.nib file format planned? Many games are only in this format and conversion software like dsk2nib doesn't really seem to work in the nib->dsk direction.

alyosha-tas:
I'll add the .nib file extension to the existing issue #1121 

The joystick is already there.

Issue #1169, opened 2018-03-31
Scrolling down in the hex editor while viewing the System Bus in the Apple II core results in a crash


NarryG:
```System.NullReferenceException: Object reference not set to an instance of an object.
   at Jellyfish.Virtu.DiskIIDrive.get_IsWriteProtected()
   at Jellyfish.Virtu.DiskIIController.ReadIoRegionC0C0(Int32 address)
   at Jellyfish.Virtu.Memory.ReadIoRegionC0C0(Int32 address)
   at Jellyfish.Virtu.Memory.ReadIoRegionC0CF(Int32 address)
   at Jellyfish.Virtu.Memory.Peek(Int32 address)
   at BizHawk.Emulation.Cores.Computers.AppleII.AppleII.<SetupMemoryDomains>b__85_2(Int64 addr) in C:\Users\Daniel\Documents\BizHawk\BizHawk.Emulation.Cores\Computers\AppleII\AppleII.IMemoryDomains.cs:line 44
   at BizHawk.Client.EmuHawk.HexEditor.MakeValue(Int32 dataSize, Int64 address, Boolean& is_cheat) in C:\Users\Daniel\Documents\BizHawk\BizHawk.Client.EmuHawk\tools\HexEditor\HexEditor.cs:line 603
   at BizHawk.Client.EmuHawk.HexEditor.GenerateMemoryViewString(Boolean forWindow) in C:\Users\Daniel\Documents\BizHawk\BizHawk.Client.EmuHawk\tools\HexEditor\HexEditor.cs:line 539
   at BizHawk.Client.EmuHawk.HexEditor.UpdateValues() in C:\Users\Daniel\Documents\BizHawk\BizHawk.Client.EmuHawk\tools\HexEditor\HexEditor.cs:line 156
   at BizHawk.Client.EmuHawk.ToolManager.UpdateAfter() in C:\Users\Daniel\Documents\BizHawk\BizHawk.Client.EmuHawk\tools\ToolManager.cs:line 445
   at BizHawk.Client.EmuHawk.ToolManager.UpdateToolsAfter(Boolean fromLua) in C:\Users\Daniel\Documents\BizHawk\BizHawk.Client.EmuHawk\tools\ToolManager.cs:line 673
   at BizHawk.Client.EmuHawk.MainForm.StepRunLoop_Core(Boolean force) in C:\Users\Daniel\Documents\BizHawk\BizHawk.Client.EmuHawk\MainForm.cs:line 2988
[...]

Issue #1121, opened 2018-02-24
Apple II: Virtu Bugs


alyosha-tas:
- Somce games need Black or White apple buttons
- need Joystick Support
- Donkey Kong is unplayable due to buggy controls.

nattthebear:
@alyosha-tas There's an Apple II core in Mednafen.  Do you have any idea how good or bad that is?

alyosha-tas:
I'm no expert, but looking over the code, it looks ok, but not close to the level of completeness you'd get from a mature apple II focused emulator. It looks like it doesn't do IIe yet either, also missing some peripherals.

But it does have better controller support which is a big plus.

PR #1394, opened 2018-12-07
[WIP] Trying to get Virtu fixed up a bit

Asnivor:
**WIP - DO NOT MERGE YET**

So this might not go anywhere, but there is so much wrong with the Apple II core right now, I thought i'd take a look over the next week or two.

If no-one minds I'll keep this PR open until it's done (or we give up). Its just easier for me to track stuff and ask questions here.

So, problems faced:

- [x] Apple 'Open' and Apple 'Close' buttons are not currently implemented
- [ ] Reset button not currently implemented
- [ ] Joysticks are not implemented
- [ ] Paddles are not implemented
- [ ] Controls are buggy generally (see prince of persia - character auto-walks left at the start until you press up-arrow and Donkey Kong)
- [ ] It looks like *.nib support is in the original Virtu source, but the core will not load these files correctly
- [ ] Currently, all client methods that read the system bus (RAMWatch, RAMSearch, HexEditor etc.) have the potential to trigger the peripheral card soft-switches in the core. Altering core state just by observing it is some next-level quantum terribleness..

Can't promise anything, but the core has been annoying me for a while now. Speak up here if anything else is wrong that I have missed above.

Current reference issues: 
#1121 
#1186 
#1169


alyosha-tas:
Just throwing this out there as an idea, but would it be worth it to just make an in-house core for Apple II instead? We already have a 6502 and can borrow from the existing core infrastructure. Virtu seems like kind of a dead end to me.

Asnivor:
@alyosha-tas You know, that might actually be doable. I'll think about it a bit more next week.

Asnivor:
So mednafen has Apple II/II+ now (with potentially IIe coming later). Its probably a waste of my time to start a new apple II core if there is any chance of waterboxing this....

adelikat:
So, this has been sitting for awhile now.  Are the other 6 things on the checklist going to happen?  Can we just merge this one since it is a solid improvement?

Asnivor:
Yeah, I'm not sure I will have time for this anytime soon. Also, github is showing a merge conflict on bizhawk.sln and I don't have time to fix that and merge right now. If someone else wants to fix and merge that would be cool....

PR #1422, opened 2019-01-06
[WIP] .NET Core 3.0 support (resolves #1415)

### LINKED ISSUE #1415: Consider moving to .NET Core
With .NET Core version 3, WinForms is fully supported in .NET Core.  I think we should move Bizhawk over.

Benefits:

1. Can deploy exactly the runtimes we want with our application, no more mucking around with prereq installers.
2. Can update to new versions of .NET Core without needing to worry about our users updating.
3. Can get some minor performance improvements right now - much has been improved since .NET 4.6.1.

Drawbacks:

1. Fully self-contained builds will be much larger than current builds.  We could also create a release that requires .NET Core to be installed on the target machine, which would lose most of the benefits from above but bring the size back down.

nattthebear:
DO NOT MERGE - NOT READY.

Right now, it compiles, but that's about it.  Major problems/blockers:

1. OpenTK support.  Seems promising, but still a WIP: https://github.com/opentk/opentk/issues/823
2. Lua support.  MoonSharp seems promising, but would require a decent number of things to be rewritten
3. ~~NET Core 3 needs to release.  probably 4+ months away?~~ Core 3.1 is out
4. ~~SlimDX support?  The project seems absolutely dead.  If the things it did are important, probably better to try replacing with an alternative.~~ DirectX now strictly optional, and only loads on Windows anyway
5. ~~7zip support.  I assume it's not working.  There don't seem to be any very good alternatives.~~ `SharpCompress` replaced `SevenZipSharp` a while ago


Asnivor:
A couple of possibly-related things that have come up in our work on mono compatibility.

`BizHawk.Client.EmuHawk.CustomControls.GDIRenderer.cs`

This is going to be problematic in .netcore as it is in mono. You will probably have to wait and see how MS implements System.Drawing. But it may be GDI+ (or libgdiplus on linux) or maybe even another assembly like ImageSharp.Drawing. I guess we will have to wait and see (unless they have made a decision already that I have missed). Regardless, using `DllImport("gdi32.dll")` is not going to cut it. The safe bet would maybe be to rework GDIRenderer to use GDI+ along with the controls that rely on this. There are only really two custom controls that make use of this at the moment as far as I can see, HexView.cs (the experimental 'new' hexviewer) and InputRoll.cs (which I'm assuming is just TAStudio).

I've had a look at doing this myself today for the mono compatibility piece, but I really do not know what I am doing (and hence, am not getting anywhere).

`SevenZipSharp`

This is another thing that will not work. As discussed in IRC, SharpCompress looks like a suitable replacement. In fact, we now have this in place for mono (albeit in a fork).

You can see the implementation here:  https://github.com/Asnivor/BizHawk/pull/2
At the moment it kinda copies the SevenZipSharp way of implementation by just dumping the source in its own folder within `BizHawk.Client.Common.SharpCompress`, but obviously this could just be a seperate project that builds a DLL that we reference (at the moment its easy as it is as it builds on both windows and mono/linux without too much messing around).

At the moment it is using SharpCompress for mono and SevenZipSharp for Windows (as can be seen here: https://github.com/Asnivor/BizHawk/pull/2/files#diff-519dc7f6539429d310f51319c3e33ec4R115), but I have tested both mono and windows with the SharpCompress IHawkFileArchiveHandler and it is working perfectly (it may be slightly less performant than SevenZipSharp, but its really not noticeable). But as you can see it's just one line to change if:

1. You actually want to use my implementation
2. We actually want to use SharpCompress going forward

It does have other benefits too. eg, support for RAR, LZIP and GZIP archives just by changing the frontend file extension filters and a couple of lines in the `SharpCompressArchiveHandler.cs` class. And possibly TAR as well, although SharpCompress does appear to detect *.gbc roms as tar archives, so a little more work would be needed on that. I mean, gzip, tar and lzip are not particularly important archive formats for us, but potentially with .netcore and *nix users supporting the opening of them may become more important I guess...

Sappharad:
> `BizHawk.Client.EmuHawk.CustomControls.GDIRenderer.cs`
> 
> This is going to be problematic in .netcore as it is in mono. You will probably have to wait and see how MS implements System.Drawing. But it may be GDI+ (or libgdiplus on linux) or maybe even another assembly like ImageSharp.Drawing. I guess we will have to wait and see (unless they have made a decision already that I have missed). 

You did. The WinForms implementation for .NET Core 3 is available on GitHub:
https://github.com/dotnet/winforms
Note that it only supports Windows, as they originally announced. Mono still requires their separate implementation.
System.Drawing is here:
https://github.com/dotnet/corefx/tree/master/src/System.Drawing.Common/src/System/Drawing

Asnivor:
Ok, so it will be gdiplus (and libgidplus on linux) looking at the corefx System.Drawing.Common. My point still stands though. Someone with more knowledge than me will have to do away with our existing gdi32 calls.

YoshiRulz:
The Mono project has something called *Cairo*, haven't looked into it but it's here: https://www.mono-project.com/docs/gui/drawing

Asnivor:
Yup, iirc mono uses Cairo to do some of the heavy lifting behind libgidplus. Mono's gdiplus implementation seems to be working quite well so far in our mono testing (I mean, we have core video playback - slow as it may be). Replacing our managed gdi32 calls (especially with relation to TAStudio) is going to be the hurdle for both this .netcore impl. and the mono compatibility work. I'm personally hoping that the mono work we have been doing becomes irrelevant eventually with .netcore, but much of the mono work we are doing will aid the switch to .netcore anyway (as many of the goals are the same).

YoshiRulz:
Will targetting .NET Core be like an extension of our Mono work, or will it also require changes to the C# cores?

Asnivor:
The mono work we have done so far should be applicable when we move to .netcore. Mono will still be required on *nix (because Winforms will still be windows only).

The work @nattthebear as done here looks to be addressing all the non-.netcore things that we (mostly) havent looked at. I imagine the easiest thing would be for the changes here to be re-done against master once we have merged the MoreUnixMonoCompat branch (as it does look like we have touched a few of the same things). But hopefully by the time we merge said branch, points 2, 3, 4 and 5 in the original description here will already have been put to bed.

YoshiRulz:
[A bunch of stuff has been announced](https://devblogs.microsoft.com/dotnet/introducing-net-5/), namely:
* .NET Core 3.0 releases September 2019, .NET Core 3.1 releases November 2019
* .NET Core 5.0 releases November 2020 ("We’re skipping the version 4 because it would confuse users")
* .NET Framework 4.8 will be the last version of .NET Framework, all products but .NET Core will only recieve security updates after 2020
* .NET Core 5.0 and later will have more APIs open-sourced and/or ported to Android, iOS, Linux, and macOS, meaning no more Mono or Xamarin.*

YoshiRulz:
If any NuGet packages are required by master when we get around to this, it'll be with `packages.config`. Apparently this is different in *Core*: [MSDN](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files) (third paragraph at time of writing).

adelikat:
The reason that GDIRenderer exists is because GDI+ isn't going to cut it in terms of performance for TAStuido.  Rather than looking into a GDI+ implementation we need to be wiring up the native rendering (opengl, directx) that the mainform is using to do the drawing.

YoshiRulz:
Rebased. Old branch is backed up as `netcore-201901`.

nattthebear:
We're moving in this direction, but this exact pull request is no longer relevant

PR #1425, opened 2019-01-06
Fix quicknes (Fixes #1423)

### LINKED ISSUE #1423: QuickNES Crashes in Latest Dev Build
Load a game (I tried Streemerz) and the core will silently crash and the emulator close.

NarryG:
Fixes issue #1423  (reversion from #1380)

Why this fixes it? No idea. I'm assuming it has to do with the static but there's no real documentation on DynamicLibraryImportResolver


nattthebear:
It's 40 lines of code, how much documentation would you expect it to have?

`DynamicLibraryImportResolver` owns the library it loads, so on dispose or finalize, it frees that.  Since QuickNES is loaded exactly once and kept there forever, the resolver needs to stay around.

NarryG:
I didn't really expect it to have anything. I just didn't actually look at what it was doing either.

PR #1426, opened 2019-01-06
Sync up subneshawk

alyosha-tas:


nattthebear:
It's impossible to see what this is doing since it includes some other PRs (like the whitespace PR) in it.  I don't want it merged until that's resolved and then I can get an actual look at it.

nattthebear:
Disregard my previous comment, didn't see that this was a request for master into another branch, not the other way around

alyosha-tas:
wow this is a mess.

PR #1427, opened 2019-01-06
SubNESHawk

alyosha-tas:
Adds Sub Frame input Capability for use with NESHawk.
- Can use TAStudio (each frame is now an input frame instead of a video frame)
- Supports 2 default NES Controllers
- Record only complete video frames / audio
- Can still use NESHawk as normal core

TODO:
- Testing
- Add other NES tools (i.e. PPU viewer)
- midframe resets (how to implement them?)

Questions, thoughts, comments?


vadosnaprimer:
Repeating my comment. Check out this implementation:
https://github.com/TASVideos/BizHawk/issues/676#issuecomment-255526749
https://github.com/TASVideos/BizHawk/issues/676#issuecomment-407103055

Tweaks to client could be relevant.

alyosha-tas:
I think what wyrster was trying to do is much different then this. SubNESHawk is actually just a normal core instance. The only thing I had to do was give it the ability to return whether or not it hit a frame ending so encoding works correctly. The rest is just running the underlying NESHawk core one step at a time and passing it inputs when it asks for it. It's really just a trick.

That being said, you could build a full featured debugger on top of this, since you have access to the entire core state at every step.

Asnivor:
So I don't really understand (yet) what this implementation is doing, but for some reason SubNESHawk performance is much better then NESHawk?

**NESHawk**
image


**SubNESHawk**
image

What is this black magic?

Asnivor:
Also, is this expected TAStudio stuff (with no inputs) for this new implementation?

**SubNESHawk**
image

compared with....

**NESHawk**
image

alyosha-tas:
It’s not faster, each frame is just only part of a full video frame , but the FPS calculation treats it like a full frame . Maybe this is something I can fix in the FPS calculation .

Yes that’s expected . The lag frames are video frame edges, the green frames are the input polls. So that pattern means you are getting one input poll per frame (so subframe input won’t do anything for you unless it’s a reset which is todo.)

nattthebear:
I really, really, really don't want this code.

Yes, the concept is simple enough, but the idea of a "frame" meaning one particular thing is pretty pervasive throughout Bizhawk, and there are a lot of things that need to be changed in a lot of messy files (like MainForm) for this to work at all.  And even then, it's another half-workable hack because it's not the implementation anyone would really want and everything is still hidden behind fake "frames".

Compare with frame length options for the Gambatte core; it's a simpler way to accomplish some of the same results, and people have been arguing about it from the beginning of time and will continue arguing about it until the end of time, because it's all still a hack and everything sucks.

I'd challenge the TAS community at large to figure out what they want in an arbitrary input emulator; one that can handle both video frame end events and input poll events occurring at any time with no particular relationship to one another.  What would the UI look like?  How would controls work, especially time control (replacement for "frame advance")?  Things like TASStudio?  Come up with answers to all of that, and maybe some day 10 years from now someone will have built it.

nattthebear:
I'd like to actually review the content of this, but there are over 1000 changed files.  Could we not do project-wide whitespace changes at the same time as significant functionality changes?

alyosha-tas:
Huh, I was certain I made the branch after the whitespace PR was merged. I don't know why it's saying those changes are included now.

nattthebear:
I'll see if I can git wizard it tomorrow

alyosha-tas:
Oh that’s what happened .

Well if it’s simpler I can just delete the branch and remake it again ( third times the charm)

nattthebear:
Thanks for looking into that.  @alyosha-tas , could you look at that branch and if it's too your liking, update the PR to point from it?  (Or create a new PR if you can't change refs in an existing PR, not sure)

vadosnaprimer:
Note that all conflicts were false positives, all have been resolved before committing.

alyosha-tas:
Ok cool thank you @vadosnaprimer 

I'll close close this and pull from there. Fresh start.

PR #1429, opened 2019-01-07
Inconsequential changes (no whitespace)

YoshiRulz:
I'll do something more helpful next, I promise.

Edit: to save you the trouble, the commit desc is "Remove useless semicolon, remove useless `? true : false`, move period to next line". Pretty self-explanatory stuff.

***

I've got a list of things like this that aren't removed pre-compile, the last few might be controversial. Input on what is and isn't project convention would be appreciated so I have something to do while Mono work is being tested.
* `"a" + b + "c"` -> `$"a{b}c"`
* `String.Format("a {0} c", b)` -> `$"a{b}c"` (I think this can't be done in some places)
* `Type Property { get; }` -> `readonly Type Property` where possible
* renaming files where it doesn't match the contained class name, excluding partial classes ofc
* LF -> CRLF, not sure how LFs happened in a Windows codebase
* `@"C:\a\path\to\something"` or `"C:\\a\\path"` -> Path.Combine
* CS0162 unreachable code (most likely a fallthrough return where the code above has since been fixed)
* CS0168/CS0169 var declared but never used
* CS0067/CS0219/CS0414 var declared and assigned to but never used
* CS0612 deprecated API used (where API change is easy)
* CS0649 var declared and used but never assigned to (should be readonly?)
* `for (var item in arr) if (item.something == something) return item.something; return default;` -> `return Array.Find(arr, item => item.something == something)?.something ?? default`
* `if (something != null) return something.prop; else return default;` -> `something?.prop ?? default`
* `if (disposable != null) disposable.Dispose();` -> `disposable.?Dispose();`
* typos in docs
* remove \\-escaped newline (C# doesn't do this, thankfully, and it's only in two comment lines. Still, better safe than sorry)
* I think there's three files indented with two spaces, that and #1437 should leave only 4-space indents which are easy to convert should you decide to try again
* `if(`, `for(` -> `if (`, `for (` for all keywords
* `{ a , b , c }` -> `{ a, b, c }`
* `var.Method (arg)` -> `var.Method(arg)`
* `if (short expression) varName = MethodA(); else varName = MethodB();` -> `varName = short expression ? MethodA() : MethodB();` (or the reverse for long expressions?)
* enforce Allman braces

### DIFF
--- a/BizHawk.Client.DiscoHawk/DiscoHawkDialog.cs
+++ b/BizHawk.Client.DiscoHawk/DiscoHawkDialog.cs
@@ -95,7 +95,7 @@ namespace BizHawk.Client.DiscoHawk
 			}
 
 			var cueBin = boundDisc.DumpCueBin(boundDiscRecord.BaseName, GetCuePrefs());
-			txtCuePreview.Text = cueBin.cue.Replace("\n", "\r\n"); ;
+			txtCuePreview.Text = cueBin.cue.Replace("\n", "\r\n");
 		}
 
 		private void btnPresetCanonical_Click(object sender, EventArgs e)
--- a/BizHawk.Client.EmuHawk/tools/BasicBot/BasicBot.cs
+++ b/BizHawk.Client.EmuHawk/tools/BasicBot/BasicBot.cs
@@ -350,8 +350,8 @@ namespace BizHawk.Client.EmuHawk
 
 			set
 			{
-				var item = StartFromSlotBox.Items.
-					OfType<object>()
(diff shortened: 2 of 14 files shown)


zeromus:
do not modify lzma code from a library we've embedded. that's their problem, not ours.
do not replace for(;;) with while(true) ; this is purely a matter of preference

YoshiRulz:
1. oops
2. reverted

vadosnaprimer:
Other suggestions seem to be fine, tho their priority is very low. And please no all-in-one-commit PR.

PR #1434, opened 2019-01-08
[Untested] Banish most gotos from client

YoshiRulz:
a.k.a. how many PRs can I have open at once

Skipping all `goto case`s until I look up what that is. Also skipping `StreamBlobDatabase.Enqueue` because it has TWO labels. Also skipping `ArchiveExtractCallback.GetStream` because it mixes `goto` and `return`.

TexAtlas?:
* [ ] no idea

Firmwares:
* [x] Really doesn't need testing, literally replaced goto w/ break.

TAStudio:
* [ ] whatever StateManagerDecay does at ([callsite](https://github.com/TASVideos/BizHawk/blob/938154f72eda86d380e58c25c5b0e3dc01529391/BizHawk.Client.Common/movie/tasproj/TasStateManager.cs#L298))

Rewind:
* [ ] works?

DiscoHawk
* [ ] file comparison?

MultiHawk keybinds:
* [ ] keydown ignored except escape

EmuHawk keybinds:
* [ ] keydown ignored except escape

DiscSystem:
* [ ] Timestamp from string in [DiscTypes.cs](https://github.com/TASVideos/BizHawk/blob/938154f72eda86d380e58c25c5b0e3dc01529391/BizHawk.Emulation.DiscSystem/DiscTypes.cs)
* [ ] iso to cue in [DiscMountJob.cs](https://github.com/TASVideos/BizHawk/blob/938154f72eda86d380e58c25c5b0e3dc01529391/BizHawk.Emulation.DiscSystem/DiscMountJob.cs)

FilterManager:
* [ ] shader compilation?

MainForm/Program:
* [ ] works
* [ ] frame advance while (video) recording?

PCESoundDebugger:
* [ ] works?

### DIFF
--- a/BizHawk.Client.Common/movie/tasproj/StateManagerDecay.cs
+++ b/BizHawk.Client.Common/movie/tasproj/StateManagerDecay.cs
@@ -59,124 +59,108 @@ namespace BizHawk.Client.Common
 			_align = false;
 		}
 
-		// todo: go through all states once, remove as many as we need. refactor to not need goto
+		// todo: go through all states once, remove as many as we need. ~~~refactor to not need goto~~~ yw -Yoshi
 		public void Trigger(int decayStates)
 		{
-			for (; decayStates > 0 && _tsm.StateCount > 1;)
+			while (decayStates > 0 && _tsm.StateCount > 1) Trigger_Sub(decayStates);
+		}
+
+		private void Trigger_Sub(int decayStates)
+		{
+			int baseStateIndex = _tsm.GetStateIndexByFrame(Global.Emulator.Frame);
+			int baseStateFrame = _tsm.GetStateFrameByIndex(baseStateIndex) / _step;	// reduce right away
+			int forwardPriority = -1000000;
+			int backwardPriority = -1000000;
+			int forwardFrame = -1;
+			int backwardFrame = -1;
+
+			for (int currentStateIndex = 1; currentStateIndex < baseStateIndex; currentStateIndex++)
 			{
-				int baseStateIndex = _tsm.GetStateIndexByFrame(Global.Emulator.Frame);
-				int baseStateFrame = _tsm.GetStateFrameByIndex(baseStateIndex) / _step;	// reduce right away
-				int forwardPriority = -1000000;
-				int backwardPriority = -1000000;
-				int forwardFrame = -1;
-				int backwardFrame = -1;
+				int currentFrame = _tsm.GetStateFrameByIndex(currentStateIndex);
 
-				for (int currentStateIndex = 1; currentStateIndex < baseStateIndex; currentStateIndex++)
+				if (_tsm.StateIsMarker(currentFrame))
 				{
-					int currentFrame = _tsm.GetStateFrameByIndex(currentStateIndex);
+					continue;
+				}
+				else if (currentFrame + 1 == _tsm.LastEditedFrame)
+				{
+					continue;
+				}
+				else if (currentFrame % _step > 0)
+				{
+					// ignore the pattern if the state doesn't belong already, drop it blindly and skip everything
+					if (_tsm.RemoveState(currentFrame))
+					{
+						// decrementing this if no state was removed is BAD
+						decayStates--;
 
-					if (_tsm.StateIsMarker(currentFrame))
-					{
-						continue;
-					}
-					else if (currentFrame + 1 == _tsm.LastEditedFrame)
-					{
-						continue;
-					}
-					else if (currentFrame % _step > 0)
-					{
-						// ignore the pattern if the state doesn't belong already, drop it blindly and skip everything
-						if (_tsm.RemoveState(currentFrame))
-						{
-							// decrementing this if no state was removed is BAD
-							decayStates--;
-
-							// this is the kind of highly complex loops that might justify goto
-							goto next_state;
-						}
-					}
-					else
-					{
-						// reduce to imaginary integral greenzone for all the decay logic
-						currentFrame /= _step;
-					}
-
-					int zeroCount = _zeros[currentFrame & _mask];
-					int priority = ((baseStateFrame - currentFrame) >> zeroCount);
-
-					if (_align)
-					{
-						priority -= ((_base * ((1 << zeroCount) * 2 - 1)) >> zeroCount);
-					}
-
-					if (priority > forwardPriority)
-					{
-						forwardPriority = priority;
-						forwardFrame = currentFrame;
+						return; // continue loop in Trigger
 					}
 				}
-
-				for (int currentStateIndex = _tsm.StateCount - 1; currentStateIndex > baseStateIndex; currentStateIndex--)
+				else
 				{
-					int currentFrame = _tsm.GetStateFrameByIndex(currentStateIndex);
-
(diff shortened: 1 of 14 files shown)


BizHawk.Emulation.DiscSystem/DiscTypes.cs
```diff
@@ -105,29 +105,26 @@ public static bool IsMatch(string str)
...
-			FRAC = (byte)((str[6] - '0') * 10 + (str[7] - '0'));
-			Valid = true;
-			Negative = false;
-			return;
-		BOGUS:
-			MIN = SEC = FRAC = 0;
-			Valid = false;
+			Valid = str.Length == 8 && str[0] >= '0' && str[0] <= '9' && str[1] >= '0' && str[1] <= '9' &&
+				str[2] == ':' && str[3] >= '0' && str[3] <= '9' && str[4] >= '0' && str[4] <= '9' &&
+				str[5] == ':' && str[6] >= '0' && str[6] <= '9' && str[7] >= '0' && str[7] <= '9';
```
vadosnaprimer: If you move `str.Length == 8`, `str[2] == ':'`, and `str[5] == ':'` to the end of the corresponding lines, it'd allow to align the lines better, which helps to understand what's going on. Removal of goto should improve readability, so let's actually improve it.

YoshiRulz: That's only because GitHub displays tabs as 8 x's wide though, the "s" starting line 113 is aligned with the "d" on line 112 in my IDE.

BizHawk.Emulation.DiscSystem/DiscMountJob.cs
```diff
@@ -147,8 +146,7 @@ void RunBizHawk()
 				catch (DiscJobAbortException) { okParse = false; parseJob.FinishLog(); }
 				if (!string.IsNullOrEmpty(parseJob.OUT_Log)) Console.WriteLine(parseJob.OUT_Log);
 				ConcatenateJobLog(parseJob);
-				if (!okParse)
-					goto DONE;
+				if (!okParse) return RunBizHawk_Done();
```
vadosnaprimer: I think we always try to break these into 2 lines at least. But usually this:
```
if (prettyPlease)
{
    OneLinerExpression();
}
```

YoshiRulz: The return has to be two lines anyway so it has to use braces.

BizHawk.Emulation.DiscSystem/DiscTypes.cs
vadosnaprimer: This would look nicer IMO:
```
Valid =
	str[0] >= '0' && str[0] <= '9' && str[1] >= '0' && str[1] <= '9' && str.Length == 8 &&
	str[3] >= '0' && str[3] <= '9' && str[4] >= '0' && str[4] <= '9' && str[2] == ':' &&
	str[6] >= '0' && str[6] <= '9' && str[7] >= '0' && str[7] <= '9' str[5] == ':';
```

YoshiRulz: Gonna call that out of scope :P

Bizware/BizHawk.Bizware.BizwareGL/TexAtlas.cs
```diff
@@ -59,106 +59,105 @@ public static PackedAtlasResults PackAtlas(IEnumerable<RectItem> items)
...
-					int w = 1 << i;
-					int h = 1 << j;
-					TryFitParam tfp = new TryFitParam(w, h);
-					todoSizes.Add(tfp);
-				}
-			}
-
-			//run the packing algorithm on each potential size
-			Parallel.ForEach(todoSizes, (param) =>
+			while (true)
```
vadosnaprimer: Can't comment how well this one handles the job (@zeromus required), but looks reasonable at least.

BizHawk.Emulation.DiscSystem/DiscTypes.cs
vadosnaprimer: What do you mean?

Asnivor: This may be one of those situations where use of `goto` is preferable purely from a readability perspective...

YoshiRulz: @vadosnaprimer I mean I couldn't be bothered. Changes made.

BizHawk.Client.Common/movie/tasproj/StateManagerDecay.cs
```diff
@@ -59,152 +59,149 @@ public StateManagerDecay(TasStateManager tsm)
 			_align = false;
 		}
 
-		// todo: go through all states once, remove as many as we need. refactor to not need goto
+		// todo: go through all states once, remove as many as we need. ~~~refactor to not need goto~~~ yw -Yoshi
 		public void Trigger(int decayStates)
 		{
-			for (; decayStates > 0 && _tsm.StateCount > 1;)
+			while (decayStates > 0 && _tsm.StateCount > 1) Trigger_Sub(decayStates);
```
YoshiRulz: Does this need to be `while (...) decayStates = Trigger_Sub(decayStates);` (with the returns and fallthrough in `Trigger_Sub` returning the modified value of `decayStates`)?

vadosnaprimer: Can't tell without testing. This logic is quite complex and it can break in ways that, when fixed, lead to new bugs recursively. This function is so influential and fragile that I'd even move any refactors of it to yet another PR. Or just give up until we hit new bugs in it, then we'd just **have** to refactor it. Let's not touch it until I have time to meditate on this once again.

YoshiRulz: What I meant was, in C#, method int params are mutable, but do changes affect the copy passed to the function?

vadosnaprimer: Which function? I may only know what was happening in the old code.

YoshiRulz: HEAD / sans goto, I've added the _Sub function and passed decayStates to it (I've done this elsewhere too), but the sub function calls decayStates--.

vadosnaprimer: Not resolved until I can take a proper closer look.

adelikat:
Why is this say "banish gotos" and then have a thousand changes that aren't about GOTOs? I can't make sense of this.  And it looks like it touches some very brittle things and I have no idea what it might break.

This needs to be broken up into smaller changesets and they need to have a good reason and be easier to review for potential errors

BizHawk.Client.Common/movie/tasproj/StateManagerDecay.cs
YoshiRulz: I have since learned what value types are and int is one of them, so yes.

YoshiRulz:
Ok, I might do that later so I've kept a copy of this branch.

PR #1435, opened 2019-01-08
EmuHawk: _LoadRom() - Change deterministic back to how it was before. 

### LINKED ISSUE #1432: Gameboy Advance - Skip Bios not working
Load any game, then go to "GBA - Settings... - Sync Settings - Skip Bios: True"

Bios shouldn't loading, but looks like the Bizhawk (* - not sure what is) ignoring that setting

In zip-archive video with issue. Tested on Bizhawk Developer build, commit b9fb6d3f7 and on Bizhawk 2.3.1

[bandicam 2019-01-07 22-25-41-793.zip](https://github.com/TASVideos/BizHawk/files/2734197/bandicam.2019-01-07.22-25-41-793.zip)

Asnivor:
This fixes #1432

A while ago in a PR, I rather naively forced `deterministic` to True when `args.Deterministic == null` in `MainForm._LoadRom()`. This broke a bunch of stuff related to how the mGBA was initialised (specifically `SkipBios` and `RTCTime()` were being handled). It also meant that all cores that don't specifically handle Deterministic Emulation within their SyncSettings (which is many/most of them looking at it) always had DE set to TRUE with no possibility of being able to use frame or audio skip.

This PR reverts determinism handling in MainForm back to how it was:

* If `Global.MovieSession.QueuedMovie != null` then `deterministic = true`
* If `args.Deterministic != null` then `deterministic = args.Deterministic`
* Otherwise, `deterministic = false`

Basically, if there is a threat of a movie, determinism must be enforced, otherwise turn it off and let individual cores handle this themselves (if they want to).

This makes far more sense as a baseline (and clearly why it was this way originally).

### DIFF
--- a/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/BizHawk.Client.EmuHawk/MainForm.cs
@@ -3589,23 +3589,12 @@ namespace BizHawk.Client.EmuHawk
 			}
 
 			try
-			{                
-                bool deterministic;
-
-                if (args.Deterministic == null)
-                {
-					// movies should require deterministic emulation in ALL cases
-					// if the core is managing its own DE through SyncSettings a 'deterministic' bool can be passed into the core's constructor
-					// it is then up to the core itself to override its own local DeterministicEmulation setting
-					// if a movie is *not* queued then make this false (this is needed for mGBA logic and is how things were working before I changed them -Asni)
-					deterministic = Global.MovieSession.QueuedMovie != null;
-				}
-                else
-                {
-                    // a value was actually supplied somehow - use this
-                    deterministic = args.Deterministic.Value;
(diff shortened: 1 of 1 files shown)


nattthebear:
LGTM

YoshiRulz:
Pretty sure this is just "revert c1c34d3"

Asnivor:
@YoshiRulz True, but the somewhat more verbose comment in this commit may help folks coming to this in the future (including possibly me in 2 months time when I have forgotten everything). The way DE is handled in the BizHawk client vs the way different cores manage this (via SyncSettings, via slightly weird logic within the core's constructor, or *not at all*) is fraught with danger at the moment. So every little helps :)

PR #1436, opened 2019-01-08
[Untested] Banish all gotos from cores

YoshiRulz:
Is 4 simultaneous PRs a record? I feel like no-one's been stupid enough to do it before.

Excluding `goto case` because I don't know how it works.

NES:
* [ ] Light Gun - either I broke it or I'm not good at aiming with the Virtual Pad
* [ ] ExROM cart (do they exist?)

TG-16/PCE:
* [ ] Op-codes 0x28, 0x40, and 0xF4?

### DIFF
--- a/BizHawk.Emulation.Cores/CPUs/HuC6280/Execute.cs
+++ b/BizHawk.Emulation.Cores/CPUs/HuC6280/Execute.cs
@@ -64,6 +64,7 @@ namespace BizHawk.Emulation.Cores.Components.H6280
 
                 if (CDL != null && CDL.Active) CDLOpcode();
 
+                var skipClearTFlag = false;
                 byte opcode = ReadMemory(PC++);
                 switch (opcode)
                 {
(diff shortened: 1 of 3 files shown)


nattthebear:
```c#
// Do not modify this file directly! This is GENERATED code.
// Please open the CpuCoreGenerator solution and make your modifications there.
```

alyosha-tas:
Do we need CpuCoreGenerator anymore? it seems like it's an idea that never really got anywhere and kind of died out. Just keeping the CPU core code seems simpler / cleaner.

nattthebear:
I'm not sure how it needs to catch on?  All a particular CpuGenerator needs to do is generate that one Cpu.   If it can do that easily and correctly, I'd say it's complete.

The value to keeping the generator around is if you ever need to make significant changes that sweep across multiple parts of the core, it's easier to do it in the generator then potentially in many places.

YoshiRulz:
[Looks like it's been manually edited](https://github.com/TASVideos/BizHawk/commits/master/BizHawk.Emulation.Cores/CPUs/HuC6280/Execute.cs), and CpuCoreGenerator hasn't been touched in 5 years so I figured it was ok. The gotos have been there [since the repo was created](https://github.com/TASVideos/BizHawk/commit/89e4c5a67444623a9c80b14c1cfa8f5e05fd59f7#diff-2726e9eb9fcd11e87238d48074cb821d).

@alyosha-tas goto can lead to code that doesn't behave intuitively (though in this case they're pretty simple), but more importantly, removing them makes it easier to change language in case C# isn't around in another 5 years.

alyosha-tas:
I removed GOTOs from NESHawk seperately.

Personally I don't really care about the cpugenerator, but I think it would be better to sort the gotos there out on their own.

PR #1439, opened 2019-01-10
SubNESHawk

alyosha-tas:
SubNESHawk:

- Subframe input using the underlying NESHawk core.
- Selected by having SubNESHawk checkbox checked and using NESHawk core.
- Can still use NESHawk as normal (by default)
- Currently supports all NESHawk controllers
- TAStudio, trace logger, hex editor, and RAM search supported
- Encoding is supported with no extra work
- FrameAdvance now returns bool (all cores except SubNESHawk always return true, so they behave as before.)
- Existing NES menues used for controller selection and other graphics settings / tools
- bk2 header (implemented but only reports correct value if saving at end of movie)

TODO / Future work: 
- maybe a toggle for subframe mode
- subframe poll indicator


alyosha-tas:
SMB3 DMC glitch to credits verified to work as a test case, see:

http://tasvideos.org/forum/viewtopic.php?p=479842#479842

BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -2997,7 +2997,7 @@ private void StepRunLoop_Core(bool force = false)
 				}
 
 				bool render = !_throttle.skipNextFrame || (_currAviWriter?.UsesVideo ?? false);
-				Emulator.FrameAdvance(Global.ControllerOutput, render, renderSound);
+				bool new_frame = Emulator.FrameAdvance(Global.ControllerOutput, render, renderSound);
```
nattthebear: How hard would it be to compensate for this in the FPS calculation on screen as well?

BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -344,14 +344,16 @@ internal void FrameAdvancePrep(IController controller)
 			LibGambatte.gambatte_setlayers(GambatteState, (_settings.DisplayBG ? 1 : 0) | (_settings.DisplayOBJ ? 2 : 0) | (_settings.DisplayWindow ? 4 : 0));
 		}
 
-		internal void FrameAdvancePost()
+		internal bool FrameAdvancePost()
```
nattthebear: Since the gambatte internal code doesn't actually do any of this stuff (although it could easily be made to), would it be more appropriate to leave this method alone and only make the `return true;` on the `FrameAdvance` method?

BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.Core.cs
```diff
@@ -736,21 +774,63 @@ void write_joyport(byte value)
 			var si = new StrobeInfo(latched4016, value);
 			ControllerDeck.Strobe(si, _controller);
 			latched4016 = value;
+			new_strobe = (value & 1) > 0;
+			if (current_strobe && !new_strobe)
```
nattthebear: From what I remember, the classes in the ControllerDeck already handle all aspects of strobing and bitshifting, and they can be savestated at any point in the process without issue.  Why is all of that logic being duplicated here?

BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
alyosha-tas: Oops, that's my bad. Yeah it should just be void.

BizHawk.Client.EmuHawk/MainForm.cs
alyosha-tas: I was thinking about this, and I'm still playing with the idea in my head. With subframe inputs it's not really meant for casual play, so the FPS doens't matter too much, but maybe there is something that can be done to make it have some utility.

BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.Core.cs
alyosha-tas: It does, but it wasn't obvious to me how to get all the menus working to pick controllers and I wanted to have something put together for testing. I'm hoping over the next few days I'll have some time to get the NES menu and ppu viewer up and running. Then that redundant stuff will go away.

nattthebear:
Subframe movies won't have useful lengths as reported on the movie screen, and things like the PlatformFrameRates will have no value there.  @adelikat is this important?

vadosnaprimer:
VBlank count can still be put into the header, along with calculated real-time length if needed. That's what VFR emulators have to do.

alyosha-tas:
Alright I enabled menues, use existing NES controllers, and cleaned up the code base

YoshiRulz:
Can't say it works as intended (because TAStudio), but SMB1 loads and is playable on Linux.

BizHawk.Client.EmuHawk/MainForm.cs
alyosha-tas: I just moved the fps calculation to after the frameadvance. Seems to work fine.

alyosha-tas:
@YoshiRulz cool thanks. I think this is a TAS only core so it kind of needs TAStudio, but at least none of the funny logic with sync settings and such broke it.

I fixed up all the relatively easy stuff. 

Probably the only really necessary thing left to do is put VBlanks in the BK2, but I might need some help with that as I'm not following how everything moves through the movie code.

alyosha-tas:
Ok, so I got VBlank count into the Bk2 movie header easy enough. But there is a catch. Unlike frame count, you don't know ahead of time how many VBlanks are in your movie until you run it. So, it will only report the correct value if you export the movie from the end.

I'm thinking of ways around this, but haven't gotten it worked out yet

alyosha-tas:
Yes it does .

vadosnaprimer:
Sensible people have final version of their movies stored as savestates, so this info shouldn't get lost forever.

Another trick would be to extend the movie format by an additional bit of the input string per frame, marking either VBlank or poll, allowing to count them separately.

alyosha-tas:
Unfortunately the count will be invalidated if returning to a previous state ad modifying it. In this case lag is the exact same as VBlank, so it shares the same issues. Probably there isn't much that can be done about it. you'll just have to verify that proper number of VBlanks is in the file.

alyosha-tas:
Ah ok I get what you are saying . I guess that would be less likely to be wrong, sounds complicated though.

Maybe that can be for a future subneshawk 2.0 that includes a poll counter and other high end features

nattthebear:
There really shouldn't be lag frames in this for any reason other than trying to sledgehammer it into old bizhawk code.  Once you're deciding to prompt for input every time it'd be needed, the obvious next step is to not prompt for input when it's not needed.

nattthebear:
mupen's handling of vblanks is probably the best idea (the idea, not its buggy execution)

The vblank count is stored in the movie, but since it could be wrong, the movie stops when either the input frame count end is reached or the vblank count end is reach.  So falsifying your vblank count doesn't help you since mupen will just stop playing your movie.

alyosha-tas:
Lag Frames: Not really in the scope of what I wanted to do. There's a lot of extra stuff that could be done, but I don't want to go down the (probably very deep) rabbit hole of changing underlying movie functions.

VBlanks: That's a good idea too, I might look into it, but I'm honestly not that concerned since it's something easy to verify anyway.

nattthebear:
> Lag Frames: Not really in the scope of what I wanted to do. There's a lot of extra stuff that could be done, but I don't want to go down the (probably very deep) rabbit hole of changing underlying movie functions.

I don't blame you, but the road to Bizhawk is paved with half baked ideas where we "just needed to get something in" and "it'll get cleaned up later."

We'll be gaining a very specialty core for one system with a bunch of bad UX around its use and a bunch of bad code around its use.  Why?

alyosha-tas:
Why? There's not really a good reason I guess. Over the years I've seen various people ask for this capability at various times, and I just happen to realize a way to do it. When I first started working on BizHawk I kind of expected it would be obsolete by now with a bunch of Mesen-like equivalents for various systems taking over. Needless to say that hasn't happened. So what we have here is better then nothing and at least a step in the right direction. I think it leaves a light enough footprint to be worth it.

Maybe it can inspire someone to go even further or we'll get some cool runs out of it. That's all I'm hoping for.

vadosnaprimer:
I forgot, what is there yet to be done properly right now?

alyosha-tas:
I still want to add in a subframe reset option.  But I'm still rolling over in my head exactly how to do it. And I'm not sure there is a test case for it.

Other then that, this basically works (as is evidenced in the thread with the successful SMB3 ACE test.)

vadosnaprimer:
Reset is more complicated since it can happen at any cycle (or only any instruction?). So you can't just record input for every poll and automatically be ready for a reset somewhere there. And you can't record every instruction either, so reset coming from a movie should be scheduled in advance somehow.

alyosha-tas:
Yeah I can schedule for a reset at a certain cycle past the previous video frame edge pretty easily. It's not really a hurdle in itself. Maybe this weekend I will do it.

vadosnaprimer:
So you're not moving to _just poll_ frames?

alyosha-tas:
What you see now is basically the final version of what I have planned. Green frames = poll, red frames = video edge.

PR #1438, opened 2019-01-10
[WIP][🐧] More Linux/Mono Compatibility Work

Asnivor:
TODO rewrite this checklist, removing things that were committed to master in the interim

### Progress
- [x] Resolving Assemblies
  - [x] AssemblyResolve - find .NET DLLs not in the executing directory (https://github.com/TASVideos/BizHawk/commit/39afe3eb732bdb0b3387505148311612ecc3ad24)
  - [x] Microsoft.VisualBasic namespace should be usable in Mono (https://github.com/TASVideos/BizHawk/commit/1be2738a23209e6307b87023a96f2b7a08f434ff)
  - [x] SevenZipSharp assembly exception (need to remove SevenZipSharp) (https://github.com/Asnivor/BizHawk/commit/e76df2732a302a3d4435740d7035544deaca846d)
- [x] Misc UI related stuff
  - [x] Menubar mnemonics (disabled on unix to prevent a crash caused by a WinForms reflection hack) (https://github.com/Asnivor/BizHawk/commit/632768d5f8b325441ebbb103b0bb24aeaff7a2d6)
  - [x] Path settings throwing exceptions and getting non-relative paths when setting through the folder browser dialog (https://github.com/Asnivor/BizHawk/commit/dccaa2b87807d261adc2611ccec14782dbcc7e5b & https://github.com/Asnivor/BizHawk/commit/519e114721104b9cb69385ef798f820547b725d9)
- [ ] ToolBox
  - [x] Opens with NullHawk (https://github.com/TASVideos/BizHawk/commit/39afe3eb732bdb0b3387505148311612ecc3ad24 and I think https://github.com/TASVideos/BizHawk/commit/42b5931c1741be1694ede9904e54ce5e8d86b906)
  - [ ] Opens with core loaded
- [x]  Archive Handling
  - [x] Replace SevenZipSharp with .NET native SharpCompress for mono (https://github.com/Asnivor/BizHawk/commit/e76df2732a302a3d4435740d7035544deaca846d)
  - [x] Check implementation also works on Windows (yes, but currently not wired up) 
- [x] Sound
  - [x] OpenAL using latest OpenTK (https://github.com/Asnivor/BizHawk/commit/3aefa3c75f2f5bb39559701dbb6d46b4b8b8b3b3)
- [x] Video (https://github.com/Asnivor/BizHawk/commit/196d51da94851b118dbaa917cc9dc72fa3fc89c2)
  - [x] OpenGL support (using latest OpenTK)
[...]

PR #1448, opened 2019-01-15
Add a WIP implementation of memoryblock in loonix.

nattthebear:
I don't have time to test this right now, but it should be a good start for whenever we want to go farther with this.

@YoshiRulz @Asnivor


YoshiRulz:
draft is in #1438 and commented out, confirmed no regression on Windows

PR #1451, opened 2019-01-16
NES Mapper034.s is missing SyncState

bbbradsmith:
Someone was doing a TAS of Lizard and frequently causing the game to trigger its "crash" handler:
https://twitter.com/Choaslegion00/status/1084376804171939841

Appears to be due to the PRG bank not being Synced whenever a savestate is used. This patch should clear it up, I think?


nattthebear:
Woops, guess I forgot to finish up this one.  The idea looks fine to me, just have a few style questions

bbbradsmith:
I have not done much work with BizHawk's code, so feel free to adapt the style if it's out of place.

The Serializer didn't work with int[], which is why I changed it to ByteBuffer. Some other mapper code I looked at was using it the same way.

The change to ByteBuffer created the need for a cast. I wasn't sure if (byte) or Convert.ToByte was better, the latter seemed more explicit to me. Feel free to change if it's out of style here.

bbbradsmith:
Switched it to (byte). If there's a better alternative to ByteBuffer feel free to correct it.

PR #1468, opened 2019-01-28
[Waiting on upstream] Vulkan renderer

### LINKED ISSUE #575: [Feature request] Add Vulkan support
https://www.khronos.org/vulkan/

Since Vulkan  is out, I was wondering if support can be added to Bizhawk?

YoshiRulz:
TODO:
* [x] (Vulkan as a display option already in Mono branch)
* [x] import VulkanSharp
	* [x] without committing files the builder will redownload anyway
* [x] make Vulkan API calls
* [ ] draw stuff
* [ ] replace OpenTK entirely in IGL_Vulkan
* [ ] Vulkan is faster than OpenGL on Linux

resolves #575


Sappharad:
I don’t know what the standard is here, but it’s generally followed that you don’t commit NuGet packages to source control because they’ll be restored by msbuild at compile time. With .Net core, Microsoft even tries to force you not to commit them by having a system level package cache and none relative to the project.

YoshiRulz:
VulkanSharp doesn't actually have the correct binds, and WINE's copy doesn't have the Win32 symbols. mono/VulkanSharp#52

YoshiRulz:
VulkanSharp still dead, still no Linux and/or .NET Core. By the time this becomes relevant again, hopefully we'll be using Veldrid or similar instead of rolling our own graphics abstraction.

zeromus:
veldrid looks cool. bug me repeatedly and I might do some experiments

YoshiRulz:
I pushed a proof-of-concept for a portable (Vulkan/OpenGL/whatever) GLControl using [Veldrid](https://veldrid.dev) to branch [`veldrid-poc`](https://github.com/TASVideos/BizHawk/tree/veldrid-poc). Open a new issue if you end up do anything with it.

While working on Veldrid, I found what seems to be a similar project under the .NET Foundation: [Silk.NET](https://github.com/dotnet/Silk.NET).

PR #1480, opened 2019-02-05
Mame core for TAS on BizHawk (Not the libretro beta), please!

hybrapavilion:
This is the only good emulator capable to make a good tas for mame, Mame-rr is a pain in the a**...


vadosnaprimer:
WHAAAAAAAAAAAAAAAAAA~

PR #1485, opened 2019-02-14
[Untested] Change solution to C# 7.3

YoshiRulz:
Changes from 6.0:
* inline out vars
* tuples
* write-only vars (for ignoring returned values)
* strong typing in switch where cases are `is Type`?
* safe pointers with `ref`
* `throw` is an expression (yay `val ?? throw new Ex()`)
* binary literals
* smarter `= default();`
* various minor syntax changes

### DIFF
--- a/BizHawk.Client.ApiHawk/BizHawk.Client.ApiHawk.csproj.DotSettings
+++ b/BizHawk.Client.ApiHawk/BizHawk.Client.ApiHawk.csproj.DotSettings
@@ -1,2 +1,0 @@
-﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
-	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String></wpf:ResourceDictionary>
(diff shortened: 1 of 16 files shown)


vadosnaprimer:
I'd rather have each of these separated.

YoshiRulz:
Each of what?

Asnivor:
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7

> The new tuples features require the ValueTuple types. You must add the NuGet package System.ValueTuple in order to use it on platforms that do not include the types.

> You need to add the NuGet package System.Threading.Tasks.Extensions in order to use the ValueTask<TResult> type

I'm assuming in order to not have to do the above requires .net 4.7 nad VS2017

YoshiRulz:
Yeah, [.NET framework 4.7](https://docs.microsoft.com/en-us/dotnet/api/system.valuetuple?view=netframework-4.7.2#applies-to). The package can't be very big though, we don't have to switch immediately.

adelikat:
Upgrading .net should be avoided or at least carefully considered.  Because the user must have this installed.  And not all OSes even support it.  We picked .net 4.6.1 for good reason.  There are versions of windows 10 that can't go past this (even if you have auto-update on).

zeromus:
Weird, up to 4.7.2 downloads pages are saying it's even supported on windows 7.

NarryG:
It's a bit odd. Microsoft doesn't support 4.7+ on 1507 and 1511, but AdDuplex's statistics show that Windows 10 1507 and Windows 10 1511 only make up for 0.6% of all Win10 installs as of this point in time  (https://reports.adduplex.com/#/r/2019-01)

YoshiRulz:
Tuples are backported in the [`System.ValueTuple` package](https://www.nuget.org/packages/System.ValueTuple).

adelikat:
This PR requires an upgrade to .net 4.7.  And that's a decision we do not want to make right now.  That's too much of an impact on the user base.  I do not approve of this PR.

I don't what the high C# version is that doesn't require a .net change for the user.  But if we want to set it to that verison and use some of those language features, that would be fine, and I'd like to see it.

YoshiRulz:
[4.6.1 supports 6.0](https://en.wikipedia.org/wiki/C_Sharp_%28programming_language%29#Versions)

PR #1498, opened 2019-02-26
Rewrite AnalogStickPanel

YoshiRulz:
This caught my eye:
https://github.com/TASVideos/BizHawk/blob/ebe789eed22954d1eb40a0ef0a5937181739e1c3/BizHawk.Client.EmuHawk/tools/VirtualPads/controls/components/AnalogSticklPanel.cs#L8

Also includes `MutableIntRange` class and `float.RoundToInt` extension method which I felt were missing in the codebase (well, a range class at least, I didn't commit an immutable version to this branch though).

The copy in master isn't just messy, it has bugs too. The two I caught were using X in `GfxToRealY`, and never writing +1 to `flipx`/`flipy` so they'd get stuck on -1, I may have inadvertently squished others.


BizHawk.Common/MutableIntRange.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+
+namespace BizHawk.Common
+{
+	public class MutableIntRange
+	{
+		public int Min;
```
adelikat: My one nag is that these should be properties instead of fields

BizHawk.Client.EmuHawk/tools/VirtualPads/controls/components/AnalogStickPanel.cs
```diff
@@ -36,27 +33,25 @@ public int Y
...
 
-		public bool HasValue = false;
-		public bool ReadOnly { get; set; }
+		public bool HasValue;
+		public bool ReadOnly { private get; set; }
 
-		public string XName = "";
-		public string YName = "";
+		public string XName = string.Empty;
+		public string YName = string.Empty;
```
zeromus: string.empty is no improvement over ""

BizHawk.Common/MutableIntRange.cs
YoshiRulz: Fixed

Issue #1177, opened 2018-04-14
[Request]  Add a sleep LUA command


Netbrian:
The Bizhawk implantation of Lua does not seem to include a "sleep" command, and there isn't one built-in to the language.  I'd like to request one be added.

The most common use case would be for use with the comm functions, such as sockets or HTTP requests.  Being able to sleep while waiting for socket communications without changing emulator state would be useful.

The LUA sockets library should already have a sleep command built in if Bizhawk uses that.

Ashafix:
When you call `comm.socketServerSend(message)` (or a related function) the emulator is frozen until one of the following conditions occurs
* the function returns something as expected
* the function times out
* the function crashes

You can set the time out via `comm.socketServerSetTimeout()` and `comm.httpSetTimeout()`.

@Netbrian : Is that what you were looking for?

Netbrian:
I was looking for a generic sleep function, for cases when I'm using a socket library other than what Bizhawk has built in (for instance, ZMQ).

PR #1080, opened 2017-12-23
new library for communicating with other programs via Lua

Ashafix:
A library which allows easier communication between Emuhawk and other programs via Lua
* Socket communication
* HTTP communication (GET/POST)
* Memory mapped files

All new functions are written in C# and exposed in Lua.

All Lua functions have "unit tests".
A simple Python package which tests the communications features is available here:
https://github.com/Ashafix/PyEmuHawk

**New Lua functions**
_SocketServerScreenShot()_
sends a screenshot to the Socket server

_SocketServerScreenShotResponse()_
sends a screenshot to the Socket server and retrieves the response

_SocketServerSend(message)_
sends a string [message] to the Socket server

_SocketServerResponse()_
receives a message from the Socket server

_SocketServerSuccessful()_
returns the status of the last Socket server action
boolean response: true or false

_SocketServerSetTimeout(timeout)_
sets the timeout in milliseconds for receiving messages

_mmfSetFilename(filename)_
Sets the filename for the screenshots

_mmfGetFilename()_
Gets the filename for the screenshots

_mmfScreenshot()_
Saves screenshot to memory mapped file

_mmfWrite(filename, message)_
Writes a string [message] to a memory mapped file [filename]

_mmfRead(filename, expectedLength)_
Reads a string from a memory mapped file [filename]. The expected length needs to be provided to deliver accurate reads.

_httpTest()_
tests HTTP connections

_httpTestGet()_
tests the default HTTP GET connection

_httpGet(url)_
makes a HTTP GET request, i.e. retrieves information from a URL

_httpPost(url, message)_
makes a HTTP POST request to the url and sends the string as {payload: message}

_httpPostScreenshot()_
HTTP POST requests to the default POST URL and sends the screenshot as base64 encoded string in {screenshot: base64_encoded_string}

_httpSetTimeout(timeout)_
Sets HTTP timeout in milliseconds

_httpSetPostUrl(url)_
Sets HTTP POST URL

_httpSetGetUrl(url)_
Sets HTTP GET URL

_httpGetPostUrl()_
Gets HTTP POST URL

_httpGetGetUrl()_
Gets HTTP GET URL

**Changes**
* ArgParser.cs
new options introduced
--socket_port
Sets the default socket port
--socket_ip
Sets the default socket IP
--mmf
Sets the default memory mapped filename
--url_get
Sets the default URL for GET requests
--url_post
Sets the default URL for POST requests

* New dependencies
System.Net.Http

**New modules**
* Communication.cs
[...]

PR #1505, opened 2019-03-03
set socketServer IP/port via Lua (resolves #1495)

### LINKED ISSUE #1495: [Request] Add ability to set socket host and port via lua api
Currently you are only able to set socket ip and port by executing emuhawk.exe with arguments behind it. (--socket_port and --socket_ip)
It took me a long time to figure out how to change these values since they seem to be defaulted at 192.168.178.21 for ip and 9999 for port. Maybe atleast change these to 127.0.0.1 and 8080?

Or Is it possible to simplify setting these with lua commands like:
comm.setSocketServerIP
comm.setSocketServerPort

and adding
comm.getSocketServerIP
comm.getSocketServerPort

Ashafix:
Addressing issue #1495

* All defaults were removed
* An error message is shown when only IP or port is set for the socketserver in the command line parameters (`ArgParser` will throw an exception which is caught and the error message is shown).
* `socketServerSetIp`/`setSocketServerPort`/`socketServerGetIp`/`socketServerGetPort`/`socketServerGetInfo` Lua functions were added
* When a socket server is initialized from the command line, its IP and port can be changed later via Lua functions.
* All communication (socket, HTTP, mmf) global classes are now set to null by default and only initialized if requested by command line parameters
* Some Lua log messages were added
* Minor code, whitespace, etc. cleanup


Gikkman:
I was just thinking about trying to implement this feature, but I thought I'd go over here and see if there were any mentions of it. Good going!

BizHawk.Client.EmuHawk/ArgParser.cs
```diff
@@ -135,39 +135,63 @@ public void ParseArguments(string[] args)
...
+			//initialize mapped memory files
 			if (mmf_filename != null)
 			{
-				GlobalWin.memoryMappedFiles.initialized = true;
+				GlobalWin.memoryMappedFiles = new Communication.MemoryMappedFiles();
 				GlobalWin.memoryMappedFiles.SetFilename(mmf_filename);
 			}
 		}
+
+		public static string GetCmdConfigFile(string[] args)
```
YoshiRulz: method can be
```csharp
return args.FirstOrDefault(arg => arg.StartsWith("--config=", StringComparison.InvariantCultureIgnoreCase));
    ?.Substring(9);
```

BizHawk.Client.EmuHawk/Communication.cs
```diff
@@ -194,42 +193,63 @@ public void Connect()
...
+				port = port_;
+				Connect();
+			}
+
+			public int GetPort()
+			{
+				return port;
+			}
+
+			public string GetIp()
```
YoshiRulz: There are getters and setters in C#, see [MSDN](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties). You make a "backing field" (private) and use the get/set keywords to control access or call e.g. `Connect` after assignment.

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Communication.cs
```diff
@@ -37,70 +33,129 @@ public static string GetLuaFunctionsList()
...
 		{
-			return GlobalWin.socketServer.SendScreenshot(1000).ToString();
+			return GlobalWin.socketServer?.SendScreenshot(1000).ToString();
 		}
 
 		[LuaMethod("socketServerSend", "sends a string to the Socket server")]
 		public string SocketServerSend(string SendString)
 		{
-			return "Sent : " + GlobalWin.socketServer.SendString(SendString).ToString() + " bytes";
+			return "Sent : " + GlobalWin.socketServer?.SendString(SendString).ToString() + " bytes";
```
YoshiRulz: Why not interpolate here as you did in ApiHawk? Also, don't you need to use conditional access in ApiHawk as well?

YoshiRulz:
@Ashafix Off-topic, but does #1080 (adding socket) sufficiently address #1177?

BizHawk.Client.EmuHawk/tools/Lua/Libraries/EmuLuaLibrary.Communication.cs
zeromus: For the record, "dunno" and "probably, maybe later" are acceptable answers

Ashafix: Since I didn't implement ApiHawk _dunno_ is appropriate answer.

BizHawk.Client.EmuHawk/Communication.cs
Ashafix: Makes sense, will be changed accordingly.

BizHawk.Client.EmuHawk/ArgParser.cs
Ashafix: Thanks for the suggestion, will be changed accordingly.

Ashafix:
> @Ashafix Off-topic, but does #1080 (adding socket) sufficiently address #1177?

From #1177 
> The most common use case would be for use with the comm functions, such as sockets or HTTP 
> requests. Being able to sleep while waiting for socket communications without changing emulator 
> state would be useful.

Yes, the emulator is frozen until the Lua function either returns something, times out or crashes.

PR #1508, opened 2019-03-13
Use SystemColors, not Color.FromName() to avoid the extra hop

Sappharad:
And work around unimplemented stuff in the System.Drawing implementation that I'm using

### DIFF
--- a/BizHawk.Client.Common/config/Config.cs
+++ b/BizHawk.Client.Common/config/Config.cs
@@ -406,9 +406,9 @@ namespace BizHawk.Client.Common
 		public bool DisplayRamWatch = false;
 
 		// Hex Editor Colors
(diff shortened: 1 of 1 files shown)


Sappharad:
I started 2 branches on my end. One is a clone of your UnixCompat branch, the other is a branch of that with macOS specific changes that I need that you can’t/shouldn’t have that only need to exist because I’m using a non-standard implementation of WinForms and have to replace all of the references. (And comment some things out from third party libraries like SlimDX that reference the real one) I’m going to put changes that you can probably use in pull requests for the vanilla fork.

This first one is a simple one. Color.FromName("Control") isn't working in the System.Drawing implementation that I'm using, and SystemColors.Control is faster / more direct anyway.

YoshiRulz:
Recommitted as 72c052622.

PR #1511, opened 2019-03-15
Use nameof

YoshiRulz:
Prevents type/method/etc. identifiers and their string representations in literals becoming mismatched, which already happened and I had to dig through blames to check what the correct field was.

see also #1514


YoshiRulz:
no clue why build failed, at least the other one was an actual bug

adelikat:
I think maybe this is conflicted because I approved future PRs before this one.  Either way it's approved.  Fix the conflicts and go ahead and merge it.  In general I'd rather the author merge the PR's and requestors just approve them.

PR #1514, opened 2019-03-18
Use nameof in BizHawk.Emulation.Cores

YoshiRulz:
Prevents type/method/etc. identifiers and their string representations in literals becoming mismatched.

edit: zeromus tells me this breaks savestates and that they were assumed broken already

see also #1511


adelikat:
If you can fix the conflicts, go ahead and merge this.  Might be tricky though, breaking up the interpolation and name of in places that had both

PR #1525, opened 2019-03-24
Better use of LINQ / collections in general

YoshiRulz:
Avoids superfluous LINQ:
* e790d904c
* 604599246
* 5f80e9d8e
* 718e60cbc
* c82ff6077
* f060bc121

Utilises different/lesser-known method(s) to simplify code:
* 4f5beb09a
* 2900484ad
* 90623d36e
* ac1b24ae6
* 6abcf7953
* c847b8cdb
* 03d4c51c4
* 4d465a4a5
* 56c33ac99
* f256c164d

Just for readability:
* b72f40808
* f45201ce7
* 2b7443bd8
* d09003c11 (also fixes typo)

### DIFF
--- a/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
+++ b/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
@@ -51,88 +51,76 @@ namespace BizHawk.Client.Common
 		public void CallSaveStateEvent(string name)
 		{
 			var lfs = _luaFunctions.Where(l => l.Event == "OnSavestateSave").ToList();
-			if (lfs.Any())
+			try
 			{
-				try
+				foreach (var lf in lfs)
 				{
-					foreach (var lf in lfs)
-					{
-						lf.Call(name);
-					}
+					lf.Call(name);
 				}
-				catch (Exception e)
-				{
-					Log(
-						"error running function attached by lua function event.onsavestate" +
-						"\nError message: " +
-						e.Message);
-				}
+			}
+			catch (Exception e)
+			{
+				Log(
+					"error running function attached by lua function event.onsavestate" +
+					"\nError message: " +
+					e.Message);
 			}
 		}
 
 		public void CallLoadStateEvent(string name)
 		{
 			var lfs = _luaFunctions.Where(l => l.Event == "OnSavestateLoad").ToList();
-			if (lfs.Any())
+			try
 			{
-				try
+				foreach (var lf in lfs)
 				{
-					foreach (var lf in lfs)
-					{
-						lf.Call(name);
-					}
+					lf.Call(name);
 				}
-				catch (Exception e)
-				{
-					Log(
-						"error running function attached by lua function event.onloadstate" +
-						"\nError message: " +
-						e.Message);
-				}
+			}
+			catch (Exception e)
+			{
+				Log(
+					"error running function attached by lua function event.onloadstate" +
+					"\nError message: " +
+					e.Message);
 			}
 		}
 
 		public void CallFrameBeforeEvent()
 		{
 			var lfs = _luaFunctions.Where(l => l.Event == "OnFrameStart").ToList();
-			if (lfs.Any())
+			try
 			{
-				try
+				foreach (var lf in lfs)
 				{
-					foreach (var lf in lfs)
-					{
-						lf.Call();
-					}
+					lf.Call();
 				}
-				catch (Exception e)
-				{
-					Log(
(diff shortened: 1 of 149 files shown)


vadosnaprimer:
Overall please add notes why the new approach is known to be better.

BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
```diff
@@ -51,89 +51,77 @@ public void CallExitEvent(Lua thread)
 		public void CallSaveStateEvent(string name)
 		{
 			var lfs = _luaFunctions.Where(l => l.Event == "OnSavestateSave").ToList();
-			if (lfs.Any())
+			try
```
vadosnaprimer: What if there's nothing attached?

YoshiRulz: If `Where` doesn't match anything, `lfs` will be an empty list. The `foreach` has nothing to iterate over so it can't throw, and the `catch` block won't be entered.

zeromus:
"Utilises different/lesser-known method(s) to simplify code:"
More obscure code isn't _simpler_. It's just smaller and more obscure.
`.Where(pred).Count()` -> `.Count(pred) `
`.Where(pred).FirstOrDefault()` -> `.FirstOrDefault(pred)`
Replace one commonly reused thing and one obvious thing with one lesser-known thing.
Not a net plus, in my book. I don't think it reads as well, either. I see where and know there's a predicate there because that's all it does. I don't know what weird semantics theyve jammed together with predicates in the other methods.
I don't necessarily object to any of these instances, but I can't accept the premise that they're improvements without objection.

You changed a Parallel.ForEach to a AddRange that isn't parallel. Whoever used Parallel.ForEach in such a strange place must have known it would help some odd cases of loading large watch lists.
What's more, I think you changed the algorithm. It's actually baffling now: the algorithm uses _watchList on both sides. I'm concerned that it might add the same thing twice if it was listed multiply in `watches`
Previously it would not do that, but it would probably also crash due to _watchList not being protected, so....

YoshiRulz:
> `.Where(pred).FirstOrDefault()` -> `.FirstOrDefault(pred)`
> Replace one commonly reused thing and one obvious thing with one lesser-known thing.

We'll have to agree to disagree on which methods are obscure.

> You changed a Parallel.ForEach to a AddRange that isn't parallel.

A quick search tells me that [`List` doesn't like concurrency](https://stackoverflow.com/questions/8001748/parallel-foreach-with-adding-to-list) and a multi-thread version would need to `_watchList` to be a `ConcurrentBag<Watch>` or similar.

> It's actually baffling now: the algorithm uses _watchList on both sides. I'm concerned that it might add the same thing twice if it was listed multiple times in `watches`.

`Except` is [set difference](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement). I've written `A ∪ (A ∖ B)`, which equals the simpler `A ∪ B`, but I can't do `_watchList = _watchList.Union(watches).ToList()` because it's read-only.

zeromus:
Except doesn't return duplicates!? What a shitty name for that function. when I see A.Except(B) I think it's going to return "everything in A except what's in B", not "everything in A except what's in B, but only once" because nowhere in the name is "only once" implied.
I'm putting this function on my shitlist. You win this time, but if I ever have to debug code with Except in it, I'm going to begin by removing the Except.

YoshiRulz:
You've just now realised that all the LINQ methods are named terribly?

zeromus:
Do I look like a guy who likes linq enough to learn what what all the stupid functions do so I can compare it to the names and decide they're stupid names? No, I look like the guy that's going to undo all your linqing if I ever have to debug anything you ever over-linq'd.

adelikat:
There's some good things in here and some objectionable things.  I wish it wasn't such a massive PR so we could keep some of these things.  As such, I can't approve this.

YoshiRulz:
@adelikat If you look at [individual commits](https://github.com/TASVideos/BizHawk/pull/1525/commits/4f5beb09a8efe9ff7e7121e290bc302d59e29371) it should be easier to read.

adelikat:
@YoshiRulz : my comment wasn't about readability, it was about the fact that I want to reject some of these, and accept others.  Simple things like .Where().FirstOrDefault() => .FirstOrDefault() are good.

Removing .Any() for .Count > 0 in places I specifically chose .Any(), not so much.

I'm going to reject this PR.  If you want, you can open some smaller PRs that just do one thing, so we can more easily review and accept them.

PR #1532, opened 2019-04-03
Remaining string interpolation (excluding BizHawk.Emulation.Cores)

YoshiRulz:
Everything that didn't make the other PRs.

~~[This needs review](https://github.com/TASVideos/BizHawk/blob/e6374ef4771ff0f0b973bd5f671675269e3f7a96/BizHawk.Client.EmuHawk/tools/SNES/SNESGraphicsDebugger.cs#L728)~~ clearly [copied from the line above](https://github.com/TASVideos/BizHawk/commit/3dd1c5c493d16a90264ab3d0a884c3c6f59d92fc#diff-a1d4c970bc553c6c0a85f7c2cb2ac433R695) and the unnecessary `string.Format` call was never removed. edit: after fixing that, appveyor stopped building this branch ¯\\\_(ツ)\_/¯

### DIFF
--- a/BizHawk.Client.EmuHawk/tools/Debugger/GenericDebugger.Disassembler.cs
+++ b/BizHawk.Client.EmuHawk/tools/Debugger/GenericDebugger.Disassembler.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using System.Text;
 using System.Windows.Forms;
 
+using BizHawk.Common.NumberExtensions;
+
 namespace BizHawk.Client.EmuHawk
 {
 	public partial class GenericDebugger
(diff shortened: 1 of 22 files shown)


BizHawk.Client.EmuHawk/tools/Debugger/GenericDebugger.Disassembler.cs
```diff
@@ -75,7 +77,7 @@ private void DisassemblerView_QueryItemText(int index, int column, out string te
 			{
 				if (column == 0)
 				{
-					text = string.Format($"{{0:X{_pcRegisterSize}}}", _disassemblyLines[index].Address);
+					text = _disassemblyLines[index].Address.ToHexString(_pcRegisterSize);
```
adelikat: Why are you removing ToHexString() in some places, and adding it in others?

YoshiRulz: I'm removing it "where param numdigits is constant" (in the commit message). In the two places I switch to the helper method, the width is only known at runtime (as far as I could tell).

PR #1534, opened 2019-04-08
Serializer: Add support for float[], double and double[]

Asnivor:
As the title says.

Implements required logic to serialize and deserialize double values and double arrays, along with float arrays. Works with both binary and text sync modes.

It uses Buffer.BlockCopy for the array stuff (hopefully this is quick enough).


zeromus:
looks like we already had precedent from int[] and similar so I don't see why there's anything risky about this. looked good to me

PR #1536, opened 2019-04-12
Added sleep functions for Lua to client module (resolves #1177)

### LINKED ISSUE #1177: [Request]  Add a sleep LUA command

Ashafix:
Two new functions for the Lua `client` module, addresses #1177 

* sleep
* exactsleep

Both functions freeze the whole emulator for n milliseconds. `sleep` is a bit more imprecise but uses less resources, `exactsleep` is precises but eats CPU resources.


zeromus:
does anyone need exactsleep? i suppose in some rare cases it could be used for video-gamey timing hacks. but it's a rare requirement. anyway, looks good to me

Issue #1398, opened 2018-12-15
Alternative sync description


vadosnaprimer:
The checkbox should inform what it does and what it affects. In some cases it must be turned off to dump frames accurately. For example, Gambatte drops frames with altsync, GBHawk duplicates them. Some info is [here](https://github.com/TASVideos/BizHawk/issues/1176#issuecomment-381446820) and I totally disagree that it's not a huge deal overall. We just need to know which cores need it on and off.

fsvgm777:
I, myself, would prefer if alternate sync dumped *all* of the frames and not dropping or duplicating them. For instance, the PS1 boot sequence drops a LOT of frames with alternate sync enabled when dumping, while they're all there with it disabled. I once checked at which framerate the PS1 boot sequence was meant to dump, and it turned out to be 60000/1001 FPS and not ~59.292 FPS.

Spikestuff:
_sitting here always disabling it_ (unless it's PAL PS1 games)
Heh.

Remember when this was automatically disabled in earlier versions? I do.
That was a better time.

synnae:
Now whenever I dump a movie I have to go to "Config and Record AVI/WAV" instead of straight "Record AVI/WAV" in order to uncheck it, which is kinda bothersome.

If at least it was possible to save the unchecked box as default, it'd not be necessary to uncheck it every time.

nattthebear:
For cores with weird frame timing (GB in adaptive mode), it's easier for most people to spot an audio hiccup than a video hiccup, so alternate should be the default.  For cores with strict sound timing (NES, SNES, etc), alternate is always better, so alternate should be the default.  For cores with completely broken sound timing (N64), alternate is always better, so alternate should be the default.

Would you all stop crying if you could turn on the "fuck up my audio" option permanently in the config?

vadosnaprimer:
I think the best solution would be to give meaningful hints right in the dialog near the checkbox instead. Upon pondering I saw that audio is supposed to be reliable source of global sync when video and input have weird timing, but in some cases it's just not the case, for example, in PS1 NTSC mode alt sync breaks the interlaced field order. Why would it have to break it there? Seems strange.

nattthebear:
[...]

PR #1542, opened 2019-04-15
added new command line parameter audiosync

Ashafix:
Based on the discussion here #1398 it seemed like a good idea to be able to set 'alternate sync' in movie settings via the command line. Use case for me would be to create PNG imageseries from multiple movie files. Currently I need to do it manually since `alternate sync` is enabled by default.

Not sure about the best name for this parameter, internally it is called `VideoWriterAudioSync ` but in the user interface it is called `alternate sync`.


vadosnaprimer:
Please post the commands that demonstrate this tweak.

Ashafix:
**Usage**

`--audiosync=true`

has the same effect as ticking the `Alternate Sync` box in the 'Choose A\V writer' dialog.

image

Any other value, e.g. `--audiosync=false` will 'untick' the box.

The default values are left untouched.

PR #1544, opened 2019-04-19
Add JetBrains Annotations (useless from C# 8)

YoshiRulz:
Includes `NotNullAttribute`/`CanBeNullAttribute` for design-time null checks by Rider, plus [some other stuff](https://www.jetbrains.com/help/resharper/Code_Analysis__Code_Annotations.html). This is effectively a backport of nullable reference types from C# 8.

The package description claims "All usages of ReSharper Annotations attributes are erased from metadata by default, which means no actual binary reference to 'JetBrains.Annotations.dll' assembly is produced."

edit: consensus is that the benefits are too few, but I'm leaving this open as an extra nag to move to Framework 4.8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,5 @@ waterbox/**/*.wbx.in
 mono_crash*
 
 .idea
+
+packages
--- a/BizHawk.Client.ApiHawk/BizHawk.Client.ApiHawk.csproj
+++ b/BizHawk.Client.ApiHawk/BizHawk.Client.ApiHawk.csproj
@@ -35,6 +35,10 @@
     <TargetFrameworkProfile />
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="JetBrains.Annotations, Version=2019.1.1.0, Culture=neutral, PublicKeyToken=1010a0d8d6380325">
+      <HintPath>..\packages\JetBrains.Annotations.2019.1.1\lib\net20\JetBrains.Annotations.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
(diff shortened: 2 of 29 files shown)


vadosnaprimer:
Does [this comment](https://github.com/TASVideos/BizHawk/pull/1468#issuecomment-458122396) relate here?

zeromus:
HOW would those annotations get erased from metadata? Does the nuget package add msbuild logic? Sounds rickety...

YoshiRulz:
`packages` dir can be left out of version control / added to `.gitignore`.

vadosnaprimer:
I'm only fine with this if it can be disabled by default (or at least just locally). This link looks relevant?
https://stackoverflow.com/a/45445515/2792852

YoshiRulz:
Checkout branch [`jb_annot_emuhawk`](https://github.com/TASVideos/BizHawk/tree/jb_annot_emuhawk) to see what kind of effect this has on the build duration (it's based on e60896c1b).

vadosnaprimer:
It took a few seconds to update the packages, but then everything kept building at normal speed, without nuget notifications. And it explicitly said fetching them upon build can be disabled. So until newer packages slow us down too much, it's fine.

adelikat:
Not a fan of this. It is SO intrusive, for a not-free IDE that few people are using

Scepheo:
> HOW would those annotations get erased from metadata? Does the nuget package add msbuild logic? Sounds rickety...

To answer this: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=netframework-4.8

YoshiRulz:
First-party nullability is opt-in enabled in `BizHawk.Common` from 8de035522, and after #1801 will extend to more projects.
https://github.com/TASVideos/BizHawk/blob/4ed402fc59c5ebac072ce8888f62b09b50e0f085/BizHawk.Common/IImportResolver.cs#L11-L17

Issue #1085, opened 2018-01-01
[N64] GLide N64 crashes bizhawk on Intel Sandybridge GPU


Yadokingu:
Once you load an N64 ROM on an Intel Sandybridge GPU, with GLideN64 set as your graphics plugin, bizhawk crashes. Afaik there is nothing that can be done to fix this, since the OpenGL implementation on Sandybridge is shoddy, but maybe it could fail a bit more gracefully than just crashing. Maybe even prevent the user from loading an N64 ROM with this plugin when they are on Sandybridge, display an error message and fall back to a different plugin.

Bizhawk 2.2.1

Error messages here: http://tasvideos.org/forum/viewtopic.php?p=463237#463237

YoshiRulz:
If this is a problem with the iGPU drivers, it doesn't belong here.

YoshiRulz:
Warning users about HW support is not our problem. If you *really* want OpenGL with that iGPU, try Linux ([Intel GPU feature matrix](https://wiki.gentoo.org/wiki/Intel#Feature_support)).

PR #1545, opened 2019-04-19
fix prevent nullable DX objects to dispose

CodeClericChar:
Fixed problem where was mainly here:
slimdx_dispose_ex

But still some sort of code should be disposed in this similar way (this prevents **nullable disposing**).

### DIFF
--- a/Bizware/BizHawk.Bizware.BizwareGL.SlimDX/IGL_SlimDX9.cs
+++ b/Bizware/BizHawk.Bizware.BizwareGL.SlimDX/IGL_SlimDX9.cs
@@ -69,11 +69,8 @@ namespace BizHawk.Bizware.BizwareGL.Drivers.SlimDX
 
 		private void DestroyDevice()
 		{
-			if (dev != null)
-			{
-				dev.Dispose();
+			using (dev)
 				dev = null;
-			}
 		}
 
 		PresentParameters MakePresentParameters()
@@ -132,8 +129,8 @@ namespace BizHawk.Bizware.BizwareGL.Drivers.SlimDX
 
 		void IDisposable.Dispose()
 		{
-			DestroyDevice();
-			d3d.Dispose();
+			using (d3d)
+				DestroyDevice();
 		}
 
 		public void Clear(OpenTK.Graphics.OpenGL.ClearBufferMask mask)
@@ -160,7 +157,7 @@ namespace BizHawk.Bizware.BizwareGL.Drivers.SlimDX
 
 		public void FreeTexture(Texture2d tex) {
 			var tw = tex.Opaque as TextureWrapper;
-			tw.Texture.Dispose();
+			using (tw.Texture) { }
 		}
 
 		class ShaderWrapper // Disposable fields cleaned up by Internal_FreeShader
@@ -518,19 +515,15 @@ namespace BizHawk.Bizware.BizwareGL.Drivers.SlimDX
 
 		public void FreePipeline(Pipeline pipeline)
 		{
-			var pw = pipeline.Opaque as PipelineWrapper;
-
-			pw.VertexDeclaration.Dispose();
-			pw.FragmentShader.IGLShader.Release();
-			pw.VertexShader.IGLShader.Release();
+			using (pipeline.Opaque as PipelineWrapper) { }
 		}
 
 		public void Internal_FreeShader(Shader shader)
 		{
 			var sw = shader.Opaque as ShaderWrapper;
-			sw.bytecode.Dispose();
-			if (sw.ps != null) sw.ps.Dispose();
-			if (sw.vs != null) sw.vs.Dispose();
+			using (sw.bytecode) {
+			using (sw.ps) { }
+			using (sw.vs) { }}
 		}
 
 		class UniformWrapper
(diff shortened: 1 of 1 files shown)


YoshiRulz:
Can you do it without replacing tabs w/ spaces? Also I'm concerned with some of your `using` structures.

CodeClericChar:
No problemo.

CodeClericChar:
There You go! :-)

zeromus:
No. Whatever you were trying to fix, fix it some other way.
1. this syntax of using(disposable) {} to dispose -- is crazy. I've never seen it before in my life. And anyway, these changes just hide bugs. if it crashes because it's null when it's not supposed to be, then find out why it's null
2. `using(tw.Texture) tw.Texture=null` .. no. untangling that is like a brain teaser for students.
3. "this dispose is needed for XYZ reasons" -- and then you remove the dispose, hiding it inside the using. no.
4. `DestroyDevice(); d3d.Dispose();` -> `using (d3d) DestroyDevice();`
Great, you've taken the code and made it run in the opposite order that it's written. This won't confuse anyone at all.
5. the more normal disposable syntax you've applied elsewhere doesn't really make sense because the code is currently reflecting d3d COM acquire/release semantics. the existing dispose calls correspond to release calls you'd see in d3d c++ samples. It isn't always clear when those are required. By moving that logic into dispose, you risk disposing them at inappropriate times and ensure the code is further diverged from the guidance of the reference materials. However, slimdx did make the things IDisposable because it made sense to them to view it that way at least, so I won't argue with this one much. 

However, I realize now, you're just applying a principle of "never call Dispose explicitly" which is a shitty principle if it's lead to this kind of brain-bending contortions.

Bizware/BizHawk.Bizware.BizwareGL.SlimDX/IGL_SlimDX9.cs
```diff
@@ -518,19 +515,15 @@ public Pipeline CreatePipeline(VertexLayout vertexLayout, Shader vertexShader, S
 
 		public void FreePipeline(Pipeline pipeline)
 		{
-			var pw = pipeline.Opaque as PipelineWrapper;
-
-			pw.VertexDeclaration.Dispose();
-			pw.FragmentShader.IGLShader.Release();
-			pw.VertexShader.IGLShader.Release();
+			using (pipeline.Opaque as PipelineWrapper) { }
```
YoshiRulz: You didn't even fix the bug you claim to be fixing, you just removed the failing calls entirely.

CodeClericChar:
1) For example: Don't ever seen before code like this?

> using(var ofd = new OpenFileDialog) { ... }  // I know this not fixes base problem but looks better to me

Instead of

> var ofd = new OpenFileDialog;
...
ofd.Dispose();

2) I know that after Dispose it'll be null (I just leave this null - nothing's wrong with that)
3) keyword "using" call dispose after code ends and 4) It isn't in the opposite order (Dispose run after DestroyDevice function).
5) I know that. I just fix some Disposing rules eg. (for PipelineWrapper class).

This is temporary proposition not last bug fix, anyway (last fix required more time to analyze as you say). Nevermind...

Bizware/BizHawk.Bizware.BizwareGL.SlimDX/IGL_SlimDX9.cs
CodeClericChar: > you just removed the failing calls entirely.

I'm not! Code was moved to PipelineWrapper class look:
`		class PipelineWrapper : IDisposable // Disposable fields cleaned up by FreePipeline
		{
			public d3d9.VertexDeclaration VertexDeclaration;
			public ShaderWrapper VertexShader, FragmentShader;
			public int VertexStride;
			public d3d9.ConstantTable fsct, vsct;

			public void Dispose()
			{
				using (VertexDeclaration)
				{
					using (fsct)
						FragmentShader.IGLShader.Release();
					using (vsct)
						VertexShader.IGLShader.Release();
				}
			}
		}`

Btw how using works (I know that student could be confused reading code like this, but he/she must know how it works!). Just look:
https://ufile.io/1unmuefh

YoshiRulz: I'll give you fixing it, but I still think it should be `?.`.

YoshiRulz:
From the stacktrace, it looks like it crashed upon closing the main window while a retro shader is active and while using DirectX video. Does it only crash under those conditions?

Also please tell me that's Windows 7 and not XP.

CodeClericChar:
It's Windows 7 SP1 in VirtualBox limited card drivers (I make new pull request only for this problem here: https://github.com/TASVideos/BizHawk/pull/1546)

YoshiRulz:
My concern was with WinXP, which is no longer getting security updates.

Given it's in a VM, I'm worried the guest drivers might not work with DirectX (see #1085 re: certain Intel iGPUs and OpenGL). Is the host also Win7 and, if so, does it crash in the same way?

> Btw, emulator shouldn't be written for everyone? (not only Windows 7+ users).

We're working on Linux support (#1438) so those with older hardware and those who don't want to buy Windows can still run it.

zeromus:
bizhawk should do one of two things in a vm when using the d3d driver
1. crash in 1000000 ways and we will never care or fix it
2. work

if it does
3. crash in one  way

then we should fix it

YoshiRulz:
Resolved by ab020cfad.

PR #1546, opened 2019-04-19
fix prevent nullable PipelineWrapper object to dispose

CodeClericChar:
This enough to solve problem showing here:
https://user-images.githubusercontent.com/49760402/56416450-6ae15900-6280-11e9-90c4-a4252e90e3f4.PNG

### DIFF
--- a/Bizware/BizHawk.Bizware.BizwareGL.SlimDX/IGL_SlimDX9.cs
+++ b/Bizware/BizHawk.Bizware.BizwareGL.SlimDX/IGL_SlimDX9.cs
@@ -516,17 +516,17 @@ namespace BizHawk.Bizware.BizwareGL.Drivers.SlimDX
 			return new Pipeline(this, pw, true, vertexLayout, uniforms,memo);
 		}
 
-        public void FreePipeline(Pipeline pipeline)
-        {
-            var pw = pipeline.Opaque as PipelineWrapper;
-            if (pw != null)
-            {
-                if (pw.VertexDeclaration != null)
-                    pw.VertexDeclaration.Dispose();
-                pw.FragmentShader.IGLShader.Release();
-                pw.VertexShader.IGLShader.Release();
-            }
-        }
+		public void FreePipeline(Pipeline pipeline)
+		{
+			var pw = pipeline.Opaque as PipelineWrapper;
+			if (pw != null)
+			{
+				if (pw.VertexDeclaration != null)
+					pw.VertexDeclaration.Dispose();
+				pw.FragmentShader.IGLShader.Release();
+				pw.VertexShader.IGLShader.Release();
+			}
+		}
 
 		public void Internal_FreeShader(Shader shader)
 		{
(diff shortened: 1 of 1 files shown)


zeromus:
Here's how to fix it properly
1. Proceed with making `Pipeline` disposable
2. Don't use the absurd using syntax to dispose the pipeline; use `pipeline.Dispose`
3. If `pipeline` is null in `FreePipeline`, find out why
4. Otherwise, if the problem is with one of the managed resources `VertexDeclaration` or `FragmentShader` or `VertexShader` being disposed, please identify that; or if the problem is with `fsct` or `vsct` being disposed, please identify _that_
5. Fix the identified problem, without randomly scrambling the rest of the code

CodeClericChar:
Your description about "using" may be wrong received (eg. looking at the code from the project like this Program.cs:305, I guess), but I got your point.

Alright back to the problem, all starts in function "CreatePipeline" in the same file "IGL_SlimDX9.cs".
Function at the moment looks like this:

```
		public Pipeline CreatePipeline(VertexLayout vertexLayout, Shader vertexShader, Shader fragmentShader, bool required, string memo)
		{
			if (!vertexShader.Available || !fragmentShader.Available)
			{
				string errors = $"Vertex Shader:\r\n {vertexShader.Errors} \r\n-------\r\nFragment Shader:\r\n{fragmentShader.Errors}";
				if (required)
					throw new InvalidOperationException($"Couldn't build required GL pipeline:\r\n{errors}");
				var pipeline = new Pipeline(this, null, false, null, null, null);
				pipeline.Errors = errors;
				return pipeline;
			}
...
```

Us interest only this line (because is a source of problem):
`var pipeline = new Pipeline(this, null, false, null, null, null);`

As we see second argument is a null.

Going to the constructor of Pipeline we have this code:

```
		public Pipeline(IGL owner, object opaque, bool available, VertexLayout vertexLayout, IEnumerable<UniformInfo> uniforms, string memo)
		{
			Memo = memo;
			Owner = owner;
			Opaque = opaque;
			VertexLayout = vertexLayout;
			Available = available;
...
```

So we know now why "Opaque" as argument was passed as null.

Change this line:

`var pipeline = new Pipeline(this, null, false, null, null, null);`

to this line:

`var pipeline = new Pipeline(this, new PipelineWrapper(), false, null, null, null);`

Not fixing problem!, because in the "FreePipeline" method are disposing "PipelineWrapper" components (eg. like "VertexDeclaration").
So just I decided to fix it in easiest way using "using" and "IDisposable" interface (but this solution was refused by you).
Now code looks as you wanted.

Btw, checking if nullabe object is, before we want to dispose object it's absolutely brilliant idea (I no denying it) but what for? if "using" do it for me.

zeromus:
FINALLY, some details
1. This bug happens because you selected a retroshader (probably bicubic)
2. It doesnt work in your VM
3. This bug only happens when exiting the emulator, after the (non-functional) retroshader is selected

Can you please write some details about what youre doing specially next time?

CodeClericChar:
Nothing, just I launched compiled BH and exit it.

PR #1565, opened 2019-05-13
Fix GB bios buzz (Fixes #1539)

### LINKED ISSUE #1539: [Gambatte] Weird buzzing noise at Game Boy Boot Rom
A weird buzzing noise is heard during the boot until you heard the Nintendo logo's chimes. Bizhawk 2.2 doesn't have that weird buzz and it seems something broke about that on Bizhawk 2.3 and later.

sornerol:
This fixes #1539 . Prior to 2.2.2, `state.spu.ch1.master` was set to false when using a BIOS. Starting with 2.2.2, the Game Boy BIOS is required. Setting this to true causes a buzzing sound while the Nintendo logo drops during BIOS.


alyosha-tas:
Somehow I never noticed that this wasn't supposed to be present. Thanks for cleaning this up.

PR #1569, opened 2019-05-20
Rewrite ToolManager.IsAvailable

YoshiRulz:


vadosnaprimer:
Motivation behind this PR?

YoshiRulz:
In IRC, @zeromus mentioned that we should be caching the result of `Assembly.GetTypes`, and also that the assembly should be referenced in a reflection-only way.

zeromus:
The code was adding some inefficient bloat when using the menus, and I thought changing the reflection method might fix the bug some other guy was having which was due to code executing (or getting assessed by overzealous security systems) when we were simply trying to reflect over it.

As long as it does this, the PR is fine, but I can't be arsed to solve the puzzle box of multi-line parenthetical conditionals with linqs in order to figure out if it actually works. It's easy to test and easy to fix if it doesnt work so I don't care.

vadosnaprimer:
Yeah looks small and easy enough to test by usage. Yet I'd like @YoshiRulz to check if it even runs well.

YoshiRulz:
@adelikat The parameter is supposed to be a class annotated with a `ToolAttribute`, so it would be unexpected for `Single` to fail. I got the return value of `true` in the catch block from the existing code, thinking through it now, I would think `IsAvailable` should be returning `false` if the parameter is missing the attribute.

PR #1578, opened 2019-05-29
Fix broken/empty doc comments

YoshiRulz:


vadosnaprimer:
- What's wrong with exceptions?
- Are we sure we want to patch 3rd party code?

YoshiRulz:
* Build errors make [build otter](https://i.imgur.com/44XszNN.png) sad :(
* I figured changing comments was OK, if not, that's an easy fix. I'm of the opinion that untouchable code shouldn't be in the repo (in a NuGet package, for instance, like we're planning to do with SharpCompress in the mono branch).

BizHawk.Client.Common/tools/Watch/WatchList/WatchList.cs
```diff
@@ -102,9 +102,6 @@ public bool Contains(Watch watch)
 		/// </summary>
 		/// <param name="array">The one-dimension <see cref="Array"/> that will serve as destination to copy</param>
 		/// <param name="arrayIndex">Zero-based index where the copy should starts</param>
-		/// <exception cref="ArgumentNullException"></exception>
-		/// <exception cref="ArgumentOutOfRangeException"></exception>
-		/// <exception cref="ArgumentException"></exception>
```
vadosnaprimer: I mean these.

YoshiRulz: those are now untouched

zeromus:
I'm not of the opinion that untouchable code shouldn't be in the repo.
Don't change comments in embedded library sources.

PR #1582, opened 2019-06-02
Update Gambatte core

MrWint:
This updates the Gambatte core with changes from https://github.com/sinamas/gambatte and https://github.com/Dabomstew/gambatte-speedrun, bringing it more in line with those versions and improving the accuracy in order to be able to console-verify http://tasvideos.org/6395S.html.


vadosnaprimer:
Awesome job! While we're here, can this core be turned into a submodule with proper fork/patch/upstream setup? Since there's demand for updates, it'd be easier to do the way mGBA is set up, allowing to pull changes from wherever, whenever.

MrWint:
I think that's a good idea in general, but there's no real fitting upstream project to pull from right now. Both repos mentioned above don't only contain libgambatte but also executable Qt and SDL wrappers to run it, which BizHawk won't need. Maybe if gambatte-speedrun's libgambatte was externalized into a separate repo, we could create a TASVideos fork of it and use that fork as a submodule here.

vadosnaprimer:
So right now it boils down to reapplying semi-patches, and even cherry-picking is unfeasible?

adelikat:
I think making it a proper submodule is out of scope for this PR and it should be merged as is.

In the future though, for upstream, I think we should fork one of these repos (in the TASVideos organization), and strip it down to what we need, then submodule that

MrWint:
The problem this time specifically was that BizHawk's version of Gambatte was forked in 2012, back when both were svn repos, and both it and upstream repos had seven years of divergence I needed to reconcile, which would have been messy either way. Part of the goal of the PR is also to make future patches easier by reducing the diff between the versions, so that it will be less painful and the diff will be much clearer and possibly automatically patchable some of the time.

vadosnaprimer:
For future reference, this seems to be a way to split the gambatte folder into a repo without losing the history, that could then be turned into a submodule. Though if it's already easy to update it directly (after this PR), this plan can wait.

alyosha-tas:
Looks good to me, thanks for the hard work, that must have been a grind.

PR #1588, opened 2019-06-14
Polar cleanup

### LINKED ISSUE #581: TAStudio: [Enhancement]: Polar coordinates
Be able to use polar coordinates in TASstudio. I love that idea from the virtual pad and I could see a lot of use in TAStudio. Could be an option to replace the aX and aY with a column for the rayon and a column for the angle be added to ?

Something like:

Settings 
-> Coordinates 
----> Polar
---->Cartesian

YoshiRulz:


YoshiRulz:
Polar in TAStudio isn't actually possible as the analog X and Y columns in TAStudio aren't linked together. I'll just merge what I have.

PR #1589, opened 2019-06-15
Inline a static multidimensional array

YoshiRulz:
edit: PR would previously replace jagged arrays with multidimensional ones. It theoretically caused a tiny performance hit but saved a tiny bit of memory, however it was deemed no easier to read so I did this inlining instead.

### DIFF
--- a/BizHawk.Emulation.Cores/Sound/YM2413.cs
+++ b/BizHawk.Emulation.Cores/Sound/YM2413.cs
@@ -280,8 +280,6 @@ namespace BizHawk.Emulation.Common.Components
 		static short[] fullsintable = new short[PG_WIDTH];
 		static short[] halfsintable = new short[PG_WIDTH];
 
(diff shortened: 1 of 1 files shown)


nattthebear:
Perhaps these things are all on the cold path so perf doesn't matter too much, but multidim arrays are quite slow.  The article you linked even admits as much.  For the basic data access test:

>The timings are pretty much what you expect: 40 ms for the rectangular array and 24 ms for the jagged array.

That's almost twice as slow :(

adelikat:
Performance matters more than memory for C# cores.

Also, I don't find this to be an objective readability improvement, so I'm not sure the motivation

YoshiRulz:
TBH I'm just going through my todo list. If it's no easier to read, should I just revert it (and do the actually useful changes from the desc instead)?

adelikat:
That's what I would prefer

PR #1590, opened 2019-06-15
Allow gaps in laglog

### LINKED ISSUE #1461: Allow gaps in laglog
The most obvious scenario is when you load a branch that's inside invalidated area. You now have known state for that frame, so you can capture it to greenzone, and you don't have to know all states that are immediately above, yet you should be able to navigate below. Currently this scenario results in some logic flaws that aren't pleasant to work around.

vadosnaprimer:
fixes #1461

I tried `Dictionary<int, bool>` and the problem was that it had to be consecutive, because of all the stuff that laglog has to do aside from "is this a lag frame?". While moving things to this approach I kept drowning deeper and deeper in unpleasant code and it felt more and more like a hack, intrusive and ugly.

Then I tried a ternary enum exactly like taseditor does it, and once again drowned in hacks, partly caused by the way laglog logic was written, partly by having to redo everything where it's used.

Then I tried nullable bool and it was the least intrusive solution, because `this[]` is already a nullable bool! Only one file had to be changed, and very slightly. I think it's the cleanest way, aside from total overhaul of the whole thing. Also laglog from old projects imports just fine.

### DIFF
--- a/BizHawk.Client.Common/movie/tasproj/TasLagLog.cs
+++ b/BizHawk.Client.Common/movie/tasproj/TasLagLog.cs
@@ -9,8 +9,8 @@ namespace BizHawk.Client.Common
 	public class TasLagLog
 	{
 		// TODO: Change this into a regular list.
-		private List<bool> _lagLog = new List<bool>();
-		private List<bool> _wasLag = new List<bool>();
+		private List<bool?> _lagLog = new List<bool?>();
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Won't fly in this form since it needs to be saving and loading nulls from tasproj, which makes no sense.

PR #1592, opened 2019-06-15
Tsm interface

adelikat:
Extracts out an Interface IStateManager from TasStateManager and formalizes the API between it and the calling code.  Simplifies that interaction greatly, and removes some no longer used code such as the ndbDatabase logic.


BizHawk.Client.Common/movie/tasproj/IStateManager.cs
```diff
@@ -36,8 +35,7 @@ public interface IStateManager : IDisposable
 
 		int Count { get; }
 
-		// TODO: rename to Last
-		int LastStatedFrame { get; }
+		int Last { get; }
```
vadosnaprimer: The only unpleasant thing here is that it's now way harder to search where these are called.

BizHawk.Client.Common/movie/tasproj/TasStateManager.cs
```diff
@@ -404,26 +404,6 @@ public int GetStateFrameByIndex(int index)
...
-		{
-			get
-			{
-				return _used;
-			}
-
-			set
-			{
-				// TODO: Shouldn't we throw an exception? Debug.Fail only runs in debug mode?
-				if (value > 0xf000000000000000)
```
vadosnaprimer: It was a problem when branch states were changing `Used` all the time while being terrible at it. Maybe not needed anymore.

BizHawk.Client.Common/movie/tasproj/IStateManager.cs
```diff
@@ -7,16 +7,29 @@ namespace BizHawk.Client.Common
 	public interface IStateManager
 	{
 		// byte[] this[int frame] { get; } // TODO: I had it refactored to this back in the day
+
+		/// <summary>
+		/// Retrieves the savestate for the given frame,
+		/// If this frame does not have a state currently, will return an empty array
+		/// </summary>
+		/// <returns>A savestate for the given frame or an empty array if there isn't one</returns>
```
vadosnaprimer: `returns` just repeats `summary`, looks unnecessary.

BizHawk.Client.Common/movie/tasproj/TasStateManager.cs
```diff
@@ -453,110 +273,114 @@ public void Load(BinaryReader br)
...
 		public int GetStateFrameByIndex(int index)
 		{
-			// feos: this is called super often by decay
-			// this method is hundred times faster than _states.ElementAt(index).Key
 			return _states.Keys[index];
 		}
 
-		private ulong _used;
-		private ulong Used
+		private bool IsMarkerState(int frame)
```
vadosnaprimer: So it survived in the end?

adelikat: As a private method within the specific IStateManager implementation, yes.  Things like this are perfectively fine.  It did not make it into the interface however.

BizHawk.Client.Common/movie/tasproj/IStateManager.cs
adelikat: find all references?

BizHawk.Client.Common/movie/tasproj/TasStateManager.cs
adelikat: Yes, pretty sure this is no longer needed

BizHawk.Client.Common/movie/tasproj/IStateManager.cs
adelikat: "correct" documentation is supposed to be like that.
The returns shows in a different situation than the summary sometimes and stuff

PR #1607, opened 2019-07-04
C64: General improvements (disk writing, CIA/VIA timers, 6502X decimal mode fixes)

SaxxonPike:
There were some changes I did in a branch in 2016 that was left unmerged. It can't reasonably be merged now. So, this is meant to bring those changes forward into 2019.

- Disk write support has been added. Still has trouble formatting disks, but writing new data should work (try writing some BASIC programs and test this out!)
- Some missing functionality in the CIA and VIA timer chips has been added. These are very seldom used features of the chip, and I don't recall any games using them, but they can be used so they should be emulated.
- ADC in the 6502X core had a bug in decimal mode where the addition+flags weren't being set correctly in all cases. This is resolved.


alyosha-tas:
Cool! Does this fix that new Mario game with the scrolling?

SaxxonPike:
Currently working on that! These are old fixes, VSP is still not yet addressed. I'm working on that currently.

SaxxonPike:
69f8b14 should fix I-Xera's corrupted top and bottom borders.

SaxxonPike:
There was a change that potentially affects the A7800 core (because it also uses 6502X) related to RDY. I think before merging this, we’ll want to be sure what the effects of that will be. I’m not familiar with that system and how RDY is used. I’d like another pair of eyes on that if possible!

alyosha-tas:
I ran it through the available tests I have and there are no changes I can see, looks good. I also tested nes and don’t see any regressions .

SaxxonPike:
I see that there are some more bugs in how RDY and IRQs interact. Changes there are not likely to cause issues.

However..

Dummy fetches were also not happening in the 6502X core during PHA/PLA operations, and this will absolutely impact other cores that use this CPU.

alyosha-tas:
Are there any test ROMs available to check the things you are looking at? I'm not too worried about dummy stack reads, I can't think of any cases where addresses around the stack are sensitive to address bus activity. RDY and IRQ / NMI interactions could theoretically impact DMC IRQs on NES, but there aren't any tests that would point out deficiencies.

SaxxonPike:
There are some C64 tests around accesses where RDY comes into play. The situations are:
1) rasterlines where the VIC needs to access memory to fill the 40x12-bit video matrix
2) rasterlines where the VIC needs to fill sprite buffers

I'll see if I can locate and upload them here.

SaxxonPike:
There's a 1 cycle jitter when running this test that I cannot explain.
[nmirecord6.zip](https://github.com/TASVideos/BizHawk/files/3388541/nmirecord6.zip)

SaxxonPike:
It looks like there's a 1 cycle jitter for all interrupt requests when interacting with the RDY signal. That's what's ultimately causing the jerky visuals for VSP related effects. I wrote some automated tests that verified that the IRQ and RDY signal generation from the VIC are 100% as expected. So... not sure where to go from here besides look at the interrupt timing and accesses in 6502X.

SaxxonPike:
This is interesting, regarding the interplay between branches and RDY: http://breaknes.com/info/16

SaxxonPike:
These commits make the Nostalgia intro (39) work, which uses VSP for the background graphics
[no-39.zip](https://github.com/TASVideos/BizHawk/files/3389260/no-39.zip)

SaxxonPike:
Going to patch up the savestate stuff today.

alyosha-tas:
That comment about jitter reminded me of something saxxon. In the nes core it is required to delay sending the Irq to the cpu by one cpu cycle. This is because the cpu core checks for interrupts at the wrong time, one cycle after it should. Making it check on the correct cycle would require a rewrite of the entire instruction handling scheme, so I put it out of my mind. This might be the reason for some anomalies you are seeing.

Seems like you are making good progress anyway though !

SaxxonPike:
@alyosha-tas Thanks! I needed to delay the IRQ handling by _one and a half_ cycles from the VIC in order to make things work. Everything just fell into place after that. I think in addition, some interrupt logic still happens even if the system is halted via RDY (which we account for now)

There's still a VIC related slowdown somewhere, as evidenced by the Powerchip 01 intro:
https://intros.c64.org/intro/powerchip-01

But we're a whole lot closer to solid emulation.

SaxxonPike:
Some testing shows that scrolling in ZeroPaige's Super Mario game has working scrolling, _however_ everything on the screen is a whole character to the left (hence the sprites appearing too far to the right).

SaxxonPike:
It just occurred to me that writes to the VIC during phase 2 shouldn't change any of the rendering that the VIC itself does in phase 2 (such as setting YSCROLL, DEN), not until the following phase 1, and since we can't really parallelize these tasks we had to stick them in _some_ order - as it stands the VIC phase 2 is computed _after_ the CPU, which means the CPU could theoretically write something during phase 2. This might be the cause of what we're seeing with respect to the shifted characters. I'm experimenting with this now.

SaxxonPike:
mario
Think we're going to be in good shape.

alyosha-tas:
Looks good!

I think after the next stable release it will be time to modernize the 6502 core so we can do away with some of this messy stuff.

SaxxonPike:
The games I have been testing with all look good. There are a couple known issues with some fastloaders and copy protections still, and in a couple demos, subtle visual artifacts. But the fixes here are so broad and significant that we really want to get them into the mainline so that we can hit the other little bugs in a more narrow and controlled manner.

Issue #1613, opened 2019-07-10
Basic Bot: Best Attempt Playback Does Not Give Expected Value


ShinobiWannabe:
I have posted something on using the Basic Bot here, but I have been encountering issues.  The main issue is that after getting a great Best Attempt, the playback does not give the same result.  I have attached a picture showing the difference.  

![Basic Bot Issue - Playback Does Not Give Expected Value Sometimes](https://user-images.githubusercontent.com/34147391/61011630-655c5200-a349-11e9-997e-5ca79b4528ff.PNG)

Video on Basic Bot usage from what I could figure out.  If someone who is not the originalauthor wants to take a look.
http://tasvideos.org/forum/viewtopic.php?p=485873#485873

ShinobiWannabe:
Ok noticed one thing to help reproduction.  When you click Stop, then press Play Best, it will sometimes play the CURRENT attempt. Then if you press Play AGAIN to play the best it will play the best.  Not sure if this is consistent to always get the best.

ShinobiWannabe:
Well, I think I may have fixed the issue.  It was irritating me so I pulled down the project and started screwing with it and figured out what was going on.  Seems to be very consistent so far.  May have been an issue with TAS Studio only being up, but I don't know the flow of this program well enough.  I am not familiar with Git so tomorrow I will figure out how to commit this after doing some more testing.  But for now explain here what I have been doing.

	//Upon Load State, TASStudio uses GlobalWin.Tools.UpdateBefore(); as well as GlobalWin.Tools.UpdateAfter(); 
		//Both of which will Call UpdateValues() and Update() which both end up in the Update() function.  Calling Update() will cause the Log to add an additional log.  
		//By not handling both of those calls the _currentBotAttempt.Log.Count will be 2 more than expected.
		//However this also causes a problem with ramwatch not being up to date since that TOO gets called.
		//Need to find out if having RamWatch open while TasStudio is open causes issues.
		//there appears to be  "hack"(?) line in ToolManager.UpdateBefore() that seems to refresh the RamWatch.  Not sure that is causing any issue since it does look like the ramwatch is ahead too much..
	

To resolve this issue I created 
	public int LastFrameAdvanced { get; set; }
		#endregion

		public bool HasFrameAdvanced()
		{
			//If the emulator frame is different from the last time it tried calling
			//the function then we can continue, otherwise we need to stop.
			return LastFrameAdvanced != Emulator.Frame;
		}


[...]

PR #1615, opened 2019-07-15
Basicbot tasstudio issue

ShinobiWannabe:
Unable to resolve the issue completely.  Still When TasStudio is open sometimes the value it reads from ram is not the same that RamWatch displays.  I believe at least part of the problem has to do with TasStudio calling Updating multiple times.  I had tried various checks, but was unable to fully resolve the issue when TasStudio is open.

1.  Placed a check in the BasicBot Update() function so that it will only continue when the Emulator.Frame has actually advanced.  This prevents more inputs being generated for the bot over the expected number.  Ex: You select 30 frames for the bot to attempt, but it would generate 32.

Two other minor updates.
2. Added a Help Menu Item in the BasicBot form that will open the Basic Bot help page on TasVideos upon click.
3. Added a Copy button with the Tooltip "Copy to Clipboard.  Then possible to paste to text file or directly into TasStudio.".  It directly copies the displayed best bot input and can be used to copy paste directly into TasStudio or a text file.

----
Not really familiar with github so if there is a problem with this tell me.   
Issue #1613


ShinobiWannabe:
Not really seeing what the issue is here exactly.  Just looks like it is complaining about different form designer code.

YoshiRulz:
It says on the CI card: "AppVeyor was unable to build non-mergeable pull request " and that would be because you are "3 commits ahead, 1459 commits behind TASVideos:master". Once you've fixed that, force push branch `basicbot_tasstudio_issue` to your fork and the build will run again.

FWIW I checked and there *are* build errors, even after I fixed filename cases. I'm assuming they're old bugs and not your fault.

vadosnaprimer:
Please don't merge yet, I'll check it this weekend.

vadosnaprimer:
Okay I haven't found anything alarming.

PR #1616, opened 2019-07-15
Emergency patch to fix and/or workaround changes in Mono

YoshiRulz:
Assembly version check shouldn't be necessary if installed with a package manager.

I have no idea how `GetManifestResourceStream` works so I can't comment on how good the new code is.


YoshiRulz:
Second commit unnecessary (and caused heisenbug, see #1633). Recommitted the first as 080343f90, and closing.

PR #1698, opened 2019-10-21
Use WebUtility.HtmlDecode() to decode game names

### LINKED ISSUE #1667: Incorrect symbol "&" on Bizhawk title bar
If romname contains `&` symbol and not recognized by internal database, like this:
"Battletoads & Double Dragon - The Ultimate Team (U) [t1].nes"
then `&` turns to `&amp;` on the title bar

name_issue

sornerol:
Some game names in gamedb files are HTML-encoded. This change decodes those game names so they display properly in BizHawk's title bar  (resolves #1667).


zeromus:
please unescape it in the goodnes textfile in the assets directory. we generated that file ourselves; we can fix it. (I guess @hegyak generated it?). This will be faster than processing a string in every line during loading as youve done

sornerol:
Yeah, I guess I was way overthinking it. I reverted my changes and fixed gamedb_goodnes.txt.

PR #1715, opened 2019-11-03
Virtu: Make savestates smaller

jdpurcell:
After these changes, savestates were about 76% of their previous size. Should be a bit faster too.


ExternalCoreProjects/Virtu/Cpu.cs
```diff
@@ -3434,7 +3443,6 @@ private void Execute65C02Sta92() // sta (zpg)
 		}
 		#endregion
 
-		[JsonIgnore]
 		public bool Is65C02 { get { return _is65C02; } set { _is65C02 = value; _executeOpCode = _is65C02 ? ExecuteOpCode65C02 : ExecuteOpCode65N02; } }
```
jdpurcell: Removed the [JsonIgnore] from this attribute and added it to _is65C02 and _executeOpCode instead. When loading, this property will get set and initialize both backing variables accordingly.

PR #1731, opened 2019-11-14
Allow getting the version from the command line

TheGreatRambler:
Adding the ability to just type `bizhawk.exe --version` to get the the version and exit. That's it!


vadosnaprimer:
Sounds like the standard way to do this. @zeromus?

PR #1744, opened 2019-11-28
Watch sorting - new algorithm that handles separators. Fixes #1713

### LINKED ISSUE #1713: Watch sort does not account for Separators
Open Ram Watch, add some watches, add a separator, sort by address.  It doesn't sort because it throws an invalid operation exception.  The sort logic doesn't consider separators

Zinfidel:
New algorithm sorts Watches by dividing them up into "blocks" based on separators. Each block of Watches will get sorted with respect to other Watches in the block. This seemed like the most logical implementation, versus just sorting the entire watch list without respect to the separators.

The new algorithm involves a linear scan of the current watch list, but still uses the in-place sorting function of List. Thus time complexity is still n lg(n). There really shouldn't be a noticeable impact unless you are watching millions of addresses.

Also moved IComparers out of fields and into a map to make selecting them more concise when sorting.


Zinfidel:
This bug was annoying me so I decided to do something about it. I was fairly liberal with the changes I made here with the expectation that they may need to be reverted to make this PR acceptable (specifically the moving of IComparers into a map). It was done because it made using them in the sorting routine later more concise.

I'm not married to the implementation. There could just be a case statement that selects the correct IComparer. Either way, many lines of duplicate code can be removed.

Here is a gif of the new sorting algorithm:
new sorting

adelikat:
Nice! Thanks for this PR.  I was going to do this eventually but can't find the time.  This reduce a LOT of code too :)

Zinfidel:
Glad to help. I always try to reduce net LOC if I see an opportunity to do it!

PR #1754, opened 2019-12-11
All Memory Execute

warmCabin:
I'm trying to get this [arcane cycle-counting Lua script](https://cdn.discordapp.com/attachments/606046567232831500/653498902967746570/rm1delay_bizhawk.lua) working on BizHawk so we can get guaranteed console-valid Mega Man TASes. There's a race condition that lets you execute arbitrary code, and this script needs a callback on every cycle to determine when it will happen.

I find it suspicious that this behavior was specifically checked for and blocked. But whatever problems it has, it can't be worse than when I tried to individually register 65,536 callbacks! Right...?

For reference, here's the [original script](https://cdn.discordapp.com/attachments/516260276425981952/653457271006691334/rm1delay_rev6.lua), as seen on FCEUX, and my [refactor](https://cdn.discordapp.com/attachments/516260276425981952/653479310039973898/rm1delay_cabin.lua) of it.

### DIFF
--- a/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
+++ b/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
@@ -185,7 +185,7 @@ namespace BizHawk.Client.Common
 
 		[LuaMethodExample("local steveonm = event.onmemoryexecute(\r\n\tfunction()\r\n\t\tconsole.log( \"Fires after the given address is executed by the core. If no address is given, it will attach to every memory read\" );\r\n\tend\r\n\t, 0x200, \"Frame name\", \"System Bus\" );")]
 		[LuaMethod("onmemoryexecute", "Fires after the given address is executed by the core. If no address is given, it will attach to every memory read")]
-		public string OnMemoryExecute(LuaFunction luaf, uint address = null, string name = null, string domain = null)
+		public string OnMemoryExecute(LuaFunction luaf, uint? address = null, string name = null, string domain = null)
 		{
 			try
 			{
--- a/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
+++ b/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
@@ -183,9 +183,9 @@ namespace BizHawk.Client.Common
 			return nlf.Guid.ToString();
 		}
 
-		[LuaMethodExample("local steveonm = event.onmemoryexecute(\r\n\tfunction()\r\n\t\tconsole.log( \"Fires after the given address is executed by the core. If no address is given, it will attach to every memory read\" );\r\n\tend\r\n\t, 0x200, \"Frame name\", \"System Bus\" );")]
-		[LuaMethod("onmemoryexecute", "Fires after the given address is executed by the core. If no address is given, it will attach to every memory read")]
-		public string OnMemoryExecute(LuaFunction luaf, uint? address = null, string name = null, string domain = null)
+		[LuaMethodExample("local steveonm = event.onmemoryexecute(\r\n\tfunction()\r\n\t\tconsole.log( \"Fires after the given address is executed by the core. If is explicitly nil, it will attach to every memory read\" );\r\n\tend\r\n\t, 0x200, \"Frame name\", \"System Bus\" );")]
+		[LuaMethod("onmemoryexecute", "Fires after the given address is executed by the core. If the address is explicitly nil, it will attach to every memory read")]
+		public string OnMemoryExecute(LuaFunction luaf, uint? address, string name = null, string domain = null)
 		{
 			try
 			{
--- a/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
+++ b/BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
@@ -183,9 +183,9 @@ namespace BizHawk.Client.Common
 			return nlf.Guid.ToString();
 		}
 
-		[LuaMethodExample("local steveonm = event.onmemoryexecute(\r\n\tfunction()\r\n\t\tconsole.log( \"Fires after the given address is executed by the core\" );\r\n\tend\r\n\t, 0x200, \"Frame name\", \"System Bus\" );")]
-		[LuaMethod("onmemoryexecute", "Fires after the given address is executed by the core")]
-		public string OnMemoryExecute(LuaFunction luaf, uint address, string name = null, string domain = null)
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
Dunno why but here's when: https://github.com/TASVideos/BizHawk/commit/be8563469c48c2c28598f98d45ea0876c1e39e33#diff-29e9327e24d1195e8033095d303ea55fR120

vadosnaprimer:
@warmCabin if there's a way to check that the user is explicitly passing lua nil to the hook, as opposed to just skipping the argument, then this thing would be perfect. The reason hooking to every instruction was disabled is to not punish for mistakes like forgetting to pass the address. So it has to be something explicit.

BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
```diff
@@ -183,9 +183,9 @@ public string OnLoadState(LuaFunction luaf, string name = null)
 			return nlf.Guid.ToString();
 		}
 
-		[LuaMethodExample("local steveonm = event.onmemoryexecute(\r\n\tfunction()\r\n\t\tconsole.log( \"Fires after the given address is executed by the core\" );\r\n\tend\r\n\t, 0x200, \"Frame name\", \"System Bus\" );")]
-		[LuaMethod("onmemoryexecute", "Fires after the given address is executed by the core")]
-		public string OnMemoryExecute(LuaFunction luaf, uint address, string name = null, string domain = null)
+		[LuaMethodExample("local steveonm = event.onmemoryexecute(\r\n\tfunction()\r\n\t\tconsole.log( \"Fires after the given address is executed by the core. If no address is given, it will attach to every memory read\" );\r\n\tend\r\n\t, 0x200, \"Frame name\", \"System Bus\" );")]
+		[LuaMethod("onmemoryexecute", "Fires after the given address is executed by the core. If no address is given, it will attach to every memory read")]
+		public string OnMemoryExecute(LuaFunction luaf, uint? address = null, string name = null, string domain = null)
```
adelikat: The reason we throw an error originally is my concern is that it would be easy to leave out an address in a script, and that it would have huge concequences (a callback on all executes is insanely slow, depending on core).

I think address should not have a default value here and make the user explicitly add nil if they want this functionality.

Other than this small change, I think this is a great PR :)

warmCabin:
> Dunno why but here's when: [be85634#diff-29e9327e24d1195e8033095d303ea55fR120](https://github.com/TASVideos/BizHawk/commit/be8563469c48c2c28598f98d45ea0876c1e39e33#diff-29e9327e24d1195e8033095d303ea55fR120)

Wow, that's been here a while! It was probably brand new when I was using BizHawk in high school :D

BizHawk.Client.Common/lua/EmuLuaLibrary.Events.cs
warmCabin: Thanks!

Would it be sufficient to just remove the default value from the header? Because in Lua, explicit nil is indistinguishable from a missing parameter. I'm not super familiar with how the bridging works.

Maybe in the future we could have explicit registerall and even registerrange type callbacks to prevent this 100%.

vadosnaprimer: We're not sure either. Please test both cases.

YoshiRulz: Can you confirm that omitting the second parameter terminates the script rather than filling it with `null`? Might want to try both Lua engines.

warmCabin: Haven't had the chance to test this, sorry. Hopefully I can get to it tonight

warmCabin:
retest this please

warmCabin:
(I hope that works with appveyor. I'm too lazy to build it)

YoshiRulz:
only new commits trigger it, the last one is still up tho https://ci.appveyor.com/project/zeromus/bizhawk-udexo/builds/29474432/artifacts

warmCabin:
😲
My bad. Never used appveyor before

Seems like the magic Jenkins phrase summons _you!_

warmCabin:
I didn't actually get the chance to test this. Did you, @adelikat ?

warmCabin:
Still getting NLua.Exceptions.LuaScriptException: [string "main"]:8: invalid arguments to method call. When skipped or explicit nil.

warmCabin:
OK, confirmed. The commit with the default null value works as expected. **Currently, there is no behavioral change in master.**

AFAICT, the default values in C# are used when nil is passed from Lua (which is equivalent to leaving the parameter off). **There is no way to differentiate between explicit and implicit nil.** @adelikat  If you're worried about people accidentally leaving off the address, we may need to create an extra callback for this. `event.allmemoryexecute()` or similar.

We want to merge this commit: [715761f5](https://github.com/TASVideos/BizHawk/pull/1754/commits/715761f563e7319f16f9fae14479f82a3b756fb2)

Probably easiest to just revert this one: [201a4b85](https://github.com/TASVideos/BizHawk/commit/201a4b85a4b202f0c2f7f650343aaac63db872ea)
Although the documentation would be slightly wrong.

This is a good case against Squash and Merge!

warmCabin:
You get _really_ weird errors if you try to add the name parameter! Which is especially weird because this does basically the same thing as `onmemoryread`.

Can anyone confirm? Here's a [test script](https://cdn.discordapp.com/attachments/606046567232831500/655303628847972353/all-memory-execute-test.lua) that should save you a bit of trouble.

```lua
local function test()

  print(string.format("$%04X", emu.getregister("PC")))

end
-- remove "name" and it will work just fine.
-- Add the domain param and you'll get a totally different error.
event.onmemoryexecute(test, nil, "name" --[[, "domain"]])


print("registers:")
for k,v in pairs(emu.getregisters()) do
  print(tostring(k))
end
```

The exception:
```
NLua.Exceptions.LuaException: unprotected error in call to Lua API (0)
   at NLua.Lua.PanicCallback(IntPtr luaState)
   at lua_error(lua_State* )
   at NLua.ObjectTranslator.throwError(IntPtr luaState, Object e)
   at NLua.Lua.SetPendingException(Exception e)
   at NLua.LuaMethodWrapper.call(IntPtr luaState)
   at NLua.MetaFunctions.runFunctionDelegate(IntPtr luaState)
   at lua_resume(lua_State* , Int32 )
   at BizHawk.Client.EmuHawk.EmuLuaLibrary.ResumeScript(LuaFile lf)
   at BizHawk.Client.EmuHawk.LuaConsole.<>c__DisplayClass56_1.<ResumeScripts>b__1()
   at BizHawk.Client.Common.LuaSandbox.Sandbox(Action callback, Action exceptionCallback)
```

vadosnaprimer:
Can't we just pass -1 as address if we want all?

YoshiRulz:
The address is a `uint`.

vadosnaprimer:
We can make it a 64-bit signed integer.

vadosnaprimer:
Okay, NLua actually works like a charm with explicit nil in the script.

PR #1765, opened 2019-12-19
Cleaning up and hopefully removing some Win32 hacks

YoshiRulz:
I've done a lot of method renaming and refactors to new C# features, hopefully the diff isn't too hard to understand. The main two goals I had were firstly to limit `DllImportAttribute` usage to `BizHawk.Common`, and secondly to prevent any external functions from being called at runtime on Unix and maybe causing a crash.

There were a lot of duplicate imports of `LoadLibrary`/`GetProcAddress`/`FreeLibrary`, so I spent some time cleaning up the call hierarchy for those in 001822413, making sure that the call-site choice between null/optional and throwing was made clear.

~~I'd still like to try backporting the .NET Core `FolderBrowserDialog`.~~ Nevermind, their code calls `shell32.dll` and `user32.dll`.


InfamousKnight:
Nice! I've looked at some of the code in the commits, does this mean PSX and n64 tasing will work on linux? That'll be awesome.

YoshiRulz:
I think I can get at least Octoshock working. Haven't actually tested TASing yet but TAStudio opens on other cores.

InfamousKnight:
I tried loading a psx game, and it still said the octoshock dll was not found.

PR #1770, opened 2019-12-26
Allow informing the core that we're bulk-accessing the memory domain

vadosnaprimer:
If the core runs a bunch of its own threads, we have to wait for them to be ready to access the memory. Hex editor and ram search read a ton of addresses at once, and we have to wait for the core for every single read.

With this new callback we set the flag once before bulk-accessing, hawk waits once for the core threads to get ready, then reads all the addresses in one go, and lets the core threads go.

Speeds up hex editor with mame, even tho its mem reading is slow in general. RAM search is more complex, not touching for now.

The callback is optional.

Aggressive review is required, since this looks like a huge thing, even tho it's surprisingly little new code.

`<natt> feos:  I haven't looked at your commit at all, but just from reading the commit message, this may be something that helps waterbox as well`

### DIFF
--- a/BizHawk.Client.EmuHawk/tools/HexEditor/HexEditor.cs
+++ b/BizHawk.Client.EmuHawk/tools/HexEditor/HexEditor.cs
@@ -512,6 +512,8 @@ namespace BizHawk.Client.EmuHawk
 		{
 			var rowStr = new StringBuilder();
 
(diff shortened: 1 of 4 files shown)


adelikat:
Closing this in favor of https://github.com/TASVideos/BizHawk/pull/1834

PR #1775, opened 2019-12-29
Remove Log windows as console option, and make LogWindow a Tool

adelikat:
This PR removes and consolidates a lot of code, and makes the Log window not a console window.  In addition, it makes it a tool which gives it auto-loading capabilities.

In addition this remove more windows specific code from the codebase.


YoshiRulz:
Log messages aren't printed to stdout after closing the log window. [demo](https://youtu.be/q05LaClQg_Q)

adelikat:
Nice catch, I'll look into it

adelikat:
Fixed.  Note that this behavior was there before this PR as well, seems it was always broken

Issue #1069, opened 2017-11-26
Multiple issues exist with the prerequisite checker leading to crashes despite the prereq checker passing


NarryG:
There are currently multiple issues with the prerequisite checker. I went through manually to check the various configurations and these are the results. I intentionally avoided the prereq installer as the goal was detecting issues with the checker built into Bizhawk.
All of this was tested on 2.2.1 release & Windows 10 Creator's Update.

Issue 1: 
Mis-detects the presence of VC2015. If all other dependencies are met, this results in the emulator seeming to work until you try launching certain cores (for example N64) where it'll proceed to crash on you.
![fkgh1gn](https://user-images.githubusercontent.com/3047793/33236176-e5ec757a-d218-11e7-9cf3-f67a49e03c2c.png)  

Issue 2: 
Even with all of the dependencies seemingly installed (according to the prereq checker), Bizhawk still failed to launch (crashed while launching). 
As it turns out, the culprit was that DirectX wasn't fully installed. The prereq checker thought that Direct3D9 was in place, but in reality it was most likely detecting the bundled Windows 10 DLLs and not the full runtime. This resulted with a crash at launch (no error message or anything, just a silent crash).
You can see in the image posted before that it thought that Direct3D9 was installed despite the full DirectX runtime not being installed. Manually installing the full runtime fixed it.  

Issue 3: 
It doesn't check for VC2013. According to the prereq installer, this is required for Bizhawk to run properly. I didn't run into any issues without 2013 in place, but I did a very quick check. I didn't do a full runthrough of all cores and features.

zeromus:
1. NOPE: I doubt your method for confirming the vc2015 runtime is installed is correct; in a newer OS it's likely to be a core part of the OS not listed anywhere. I also doubt this is what makes "certain" cores crash. I can't diagnose "certain". Provide more details instead of jumping to conclusions. Your vc2015 runtime may have been broken. About the only version mupen doesn't require right now is 2015. 

2. NOPE: Bizhawk will fall back to gdi+ if direct3d9 can't startup. Since that didn't happen in your case, it's more likely your direct3d9 was broken, and manually installing the full runtime repaired the brokenness. We can't detect a 100% fully operational direct3d runtime, otherwise our detector would crash the same way bizhawk did.

[...]

PR #1786, opened 2020-01-11
Simplify Windows prereqs

YoshiRulz:
Copied from commit desc:
* Changed default sound/display methods to OpenAL/OpenGL on Windows to match non-Windows, and replaced DirectInput with OpenTK input (needs regression testing on Windows).
* (meta) As a result of the above, DirectX is no longer a prerequisite, leaving only: .NET Framework 4.6.1 (comes with Win10, as does 4.8), MSVC 14 (also comes with Win10), MSVC 12 for Mupen, and MSVC 10 for BSNES/Mupen.
* Refactored static Program ctor, using ExceptionBox for missing prereqs, and removed PrereqsAlert.
* Added OS version check for Windows, with a flag in the config to skip it.
* Updated readme for Win7 EOL (and generally cleanup Windows sections).

edit: oh, and I think this resolves the yet-unresolved parts of #1069 (labelled 2 in first post, 1 was non-issue and 3 was fixed already)


Sappharad:
No concerns about the crappy Intel CPU's with onboard graphics that don't support OpenGL? Or is that problem past us now that you're dropping Windows 7?

YoshiRulz:
If it comes up again we can write a proper check for the chip model.

PR #1801, opened 2020-01-18
Retargetting non-executable projects to .NET Standard

YoshiRulz:
If existing code works with the APIs in Standard, then it can target that while still being referenced in Framework projects. It's more restrictive targeting Standard, but separating these classes from the rest of the codebase is helpful because it's a clear marker of portability—there won't need to be any more refactoring as Core is a superset of Standard. (If it's not obvious, the end goal is Core: see #1415.)

I've been updating this [dependency graph](https://gitlab.com/TASVideos/BizHawk/snippets/1886666) to show progress.

That brings me to the actual code changes in this PR, starting with `BizHawk.Common`:
* ~~`BizHawk.Common`~~
	* After extracting `BizInvoke` into its own project (c7801b8ce), this was straightforward, so I did it in `master` as 97cb9b47f.
* `BizHawk.Emulation.*`
	* Straightforward.
	* The new `BizInvoke` project *does* target Framework, but it's okay to reference it in `Cores` even though it's Standard, as long `Cores` is itself referenced by `BizHawk.Client.Common`, which is targeting the same Framework version as `BizInvoke`. Unfortunately, you can't silence the single warning it creates: NuGet/Home#8797.
* `BizHawk.Bizware.BizwareGL`
	* This mostly involves moving files to `BizHawk.Client.EmuHawk`.
	* `GDIPlusGuiRenderer` only used one method specific to `IGL_GdiPlus`, so I loosened its type to `IGL` and used the dynamic type for that call. The alternative would be to move it to `BizHawk.Client.EmuHawk`.
	* I put the `IGL.SetViewport(System.Windows.Forms.Control)` overload in `#if false` as it was never called.

Roadblocks in the way of proceeding (with `BizHawk.Client.Common`):
* NLua maybe, we might be able to use the same bypass as with `BizInvoke`.
* `Ionic.Zip`—the only version on nuget.org (which is the one we use) doesn't explicitly support Standard. If NLua works, so will this.
* `System.Data.SQLite`—the only version of this on nuget.org is 7 versions ahead of our `.dll`, and unfortunately doesn't work as a drop-in replacement. If anyone wants to try updating it, replace the `<Reference/>` with this: `<PackageReference Include="System.Data.SQLite" Version="1.0.112" PrivateAssets="All" />`
* WinForms. gl;hf


adelikat:
I tested several in-house cores and external cores, including potentially problematic ones like N64, and PSX.

I tried text vs binary states, multiple config options, tools like tastudio and hex editor

Everything looks fine, I'm approving this PR

YoshiRulz:
Necro'd because I just rebased [`clcommon-netstandard`](https://github.com/TASVideos/BizHawk/tree/clcommon-netstandard), `BizHawk.Client.Common` being the last* Framework project after 1ee131ed7. We can safely merge that branch into `master` after testing where we call into a Framework assembly:
* [ ] Ionic.Zip—used only in [`IonicZipWriter`](https://github.com/TASVideos/BizHawk/blob/f95bcf0e1466156ea51ba12d4d334684d9effdf3/BizHawk.Client.Common/IonicZipWriter.cs#L4). This package has been deprecated in favour of `DotNetZip`, which is available for Standard. It may also be possible to use `SharpCompress` or `SharpZipLib` instead.
* [ ] NLua—used in Lua.
* [ ] SQLite—used only in [`SqlApi`](https://github.com/TASVideos/BizHawk/blob/f95bcf0e1466156ea51ba12d4d334684d9effdf3/BizHawk.Client.Common/Api/Classes/SqlApi.cs#L3), see my comments in this PR's desc.

PR #1816, opened 2020-01-26
Convert SMS settings to use GenericCoreConfig

adelikat:
Instead of custom dialogs and menu items, pass in the SMS settings into the GenericCoreConfig.

Note that this is a config breaking change


hegyak:
It is nice to unify settings.  It should be noted in the release notes that this will break old Config Files.

YoshiRulz:
Every release breaks old config files...

PR #1834, opened 2020-02-03
Bulkread

vadosnaprimer:
Ability to read a bunch of addresses in one go (not necessarily consecutive). 3 widths. Delegates.

### DIFF
--- a/BizHawk.Client.EmuHawk/tools/HexEditor/HexEditor.cs
+++ b/BizHawk.Client.EmuHawk/tools/HexEditor/HexEditor.cs
@@ -602,12 +602,38 @@ namespace BizHawk.Client.EmuHawk
 				}
 			}
 
-			// TODO: other sizes
-			var vals = _domain.BulkPeekByte(addresses).ToList();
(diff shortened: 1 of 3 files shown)


YoshiRulz:
Can this be extended to poke as well?

adelikat:
It could absolutely be extended to poke, but I think it is out of scope for this PR.  The problem we are solving here is that typical integrations with a core (which are reads) are unseasonably slow, and it is a blocker for getting a core like mame out the door.

Bulk poking is a much less common use case by comparison.  It would still be a valuable thing, and this approach could be applied for sure.

adelikat:
Comments by Zeromus in chat:
i think a bulk read defined by addresses is not a great idea... should be a range. 
this is going to make it more realistic for cores to specially implement it at light-speeds

also even with a collection, I would not use ICollection. 
this will stop cores from locking an array and doing it at light speed

i would use an array of ranges, actually.
and I would have it pass in the preallocated return array as well
rather than return IEnumerable
if youre coding BULK stuff, you can't use fluffy OO stuff

PR #1839, opened 2020-02-07
Fix Macro Functionality (fixes #1838)

### LINKED ISSUE #1838: Macro functionality does not work.
This affects versions 2.3.2 up until latest master f9d29a0d982908714cec732d27883f0c188979d8 . This bug is reproducible with the Octoshock core specifically, but should occur with any controller configuration that contains float controls.

Trying to create or place macros, using either the menu options inside of TAStudio or via the Macro tool throws array out of bounds exceptions. The reason this happens is because float control ranges are not properly populated in MovieZone objects when created by macro code.

Fixing this issue however reveals several other problems preventing macros from working. See the related pull request for this issue for more information on what all is preventing macros from working.

Zinfidel:
All macro tools will fail due to the `MovieZone` class not populating the `FloatRanges` property of the `ControllerDefinition` objects it creates (if the current core has float inputs). If this issue is fixed, more issues arise however.

The second major issue is that the `MovieZone` constructor that is used for macros uses fields (`_emulator`, `_tools`) does not supply these fields.

Related to the last problem is that the menu items in TAStudio that perform macro functionality basically bypass actually showing the macro tool form, but call functions on the form. Since the form is not instantiated correctly via a standard `Load` call, some more stuff breaks. Creating the form with the `Load` calls however causes the macro tool form to show when clicking on menu items.

All of this is to say that this PR in its initial form, though it does fix the problems, is not a very good solution, because a proper solution would involve separating the macro functionality out of the macro tool form into its own class. I am creating it as a draft as a kind of extension to the related issue, so that the main devs can weigh in on if they have specific plans for how to fix this problem or if this is good enough. I'm happy to do a better fix if there is agreement that my proposal is a good one.

See issue #1838

### DIFF
--- a/BizHawk.Client.EmuHawk/tools/Macros/MacroInput.cs
+++ b/BizHawk.Client.EmuHawk/tools/Macros/MacroInput.cs
@@ -268,7 +268,7 @@ namespace BizHawk.Client.EmuHawk
 
 		private void DummyLoadMacro(string path)
 		{
-			MovieZone loadZone = new MovieZone(path);
+			MovieZone loadZone = new MovieZone(path, Emulator, Tools);
 			_zones.Add(loadZone);
 			ZonesList.Items.Add($"{loadZone.Name} - length: {loadZone.Length}");
 		}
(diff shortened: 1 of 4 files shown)


adelikat:
I think this is fine for now.  Refactoring the logic into its own class can be done down the road, there's a lot of work that needs to be done to the macro tool anyway.  This fixes functionality though, so I think ti should be merged.

adelikat:
Thanks for the fix :)

PR #1841, opened 2020-02-09
Fix ffmpeg race condition.

### LINKED ISSUE #1787: Octoshock crashes when dumping a movie that's being replayed.
Present on latest dev build as well.

How to crash BizHawk using Octoshock:
- Load any PlayStation game ([that has a BizHawk TAS](http://tasvideos.org/Movies-PSX-RatingU.html))
- Change PlayStation Options to Hardcore Debug Mode
- Pause BizHawk
- Load the movie to the chosen game
- Create an AVI Dump (Alt Sync Disabled if NTSC)
- Unpause BizHawk
- Crash.

Note: All the other screen modes also crash when doing the same steps.
2nd Note: fsvgm also has this exact same issue as me.

Zinfidel:
There is a known bug in the `Process.WaitForExit(time)` overload ([see remarks here](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netframework-4.8#System_Diagnostics_Process_WaitForExit_System_Int32_)). This method will wait for the process to exit but it will _not_ wait for the message pump for that process to finish dealing with events. This was causing a race condition for ffmpeg processes that were trying to write to stderr around the same time that the process variables are being nulled in `CloseFileSegment()`. 

This problem occurs much more frequently with hardcore debug mode due to switching resolutions causing file segments to get opened/closed frequently. The bug is still frustratingly inconsistent to reproduce though.

To my understanding, a second call to `WaitForExit()` after receiving true from `WaitForExit(time)` _should_ only be waiting for any remaining events to be handled on a process after it has been closed by the OS (ie not forever), but I can't confirm this, so there is risk.

Other solutions involve giving each ffmpeg process that is spawned its own instance variables, making the stderr handler just check for nulls when accessing stderr, or the nuclear option of manual thread handling with kill timers.

Submitting as a draft because this fix may not be good enough / needs independent testing.

### DIFF
--- a/BizHawk.Client.EmuHawk/AVOut/FFmpegWriter.cs
+++ b/BizHawk.Client.EmuHawk/AVOut/FFmpegWriter.cs
@@ -94,7 +94,7 @@ namespace BizHawk.Client.EmuHawk
 
 				_commandline = $"ffmpeg {_ffmpeg.StartInfo.Arguments}";
 
-				_ffmpeg.ErrorDataReceived += new DataReceivedEventHandler(StderrHandler);
+				_ffmpeg.ErrorDataReceived += StderrHandler;
 
 				_stderr = new Queue<string>(Consolebuffer);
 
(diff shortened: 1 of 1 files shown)


nattthebear:
Looks good to me. I've actually used the `WaitForExit()` after `WaitForExit(time)` hack before.  I assume I didn't use it here because I wasn't familiar with it at the time.

I'm not heavily involved in BizHawk at the moment, so I'll let someone else merge it.

Zinfidel:
Changed commit/PR message to not link to issue. The ffmpeg race condition coincidentally happened to also be causing a crash when dumping with ffmpeg.

PR #1843, opened 2020-02-10
Fix Octoshock crash when playing back movies with different input settings (fixes #1842)

### LINKED ISSUE #1842: Octoshock crashes when playing and dumping movie with different input configuration
Affects at least version 2.4 and latest master commit 0075693ec79793a9751edadc86927fbcb9586ffe

This issue is directly related to issue #1787 

Playing and dumping a movie that has a different input configuration than the current configuration at the time of playing the movie causes Octoshock to crash. For example, if the current configuration is for a Dualshock controller for P1 and a Dualshock controller for P2, loading and dumping a movie with a Dualshock controller for P1, but **no controller** for P2 will cause the crash. Sometimes no AVI dumping is needed, and the core will crash with just the movie, but this behavior is inconsistent.

Both NES and both SNES cores have also been tested for this behavior, and none of them exhibit it. Only the Ocotoshock core thus far.

For reproducing the bug, you need a movie file that is configured to use a Dualshock controller for P1 w/ memcard, and no controller or memcard for P2. My file started from a save state, and was a tasproj movie. Not sure if those are relevant.

Reproduction steps are very finicky, follow carefully:
1. Launch BizHawk
1. Load PSX game
1. PSX > Controller / Memcard Configuration
1. Change both players to use Dualshock controllers w/ memcards. If this is how it is already set up, click stuff to get the "reboot required" message to appear post config.
1. Reboot Octoshock core
1. Pause the emulator
1. File > Movie > Play Movie
1. Select the movie with different input configuration
1. File > AVI/WAV > Record (or configure)
1. Set up dump (avi-lossless used for testing)
1. Unpause emulator
1. Access violation / crash to desktop

The crash is occurring in the `UpdateInput()` function in **psx.cpp**:
image
It appears that the ports memory is getting clobbered or is otherwise corrupt at the time the movie is resumed. It will crash when trying to dereference the no-controller device at index 1.

A possibly relevant observation is that when BizHawk successfully loads and starts dumping movies, upon closing and restarting the emulator, the input settings will have been modified to match the previous movie. When the AV crash occurs however, BizHawk will frequently start up with the input configuration that causes the crash.

I can furnish all of the files necessary to reproduce this bug if requested.

Zinfidel:
During movie loading, the input data structures in Octoshock are recreated. This did not initially involve nulling the device objects of the ports. If a port had a controller in it (psx config had two dual shocks), and a movie was getting loaded that had only one dual shock configured, the second dualshocks type and buffer got cleared, but not the device. When assigning to the port from FIO, if it was getting assigned none, then the device was just skipped, leaving the old, disposed device hanging around.

It gets called later and crashes Octoshock during movie playback.

This seems suspiciously simple to me, but it does fix the issue for me in debug and release builds. Also I'm not sure why it required AVI dumping in 2.4 to get the crash to happen but in debug builds just trying to play the movie at all would result in a crash.

### DIFF
--- a/psx/octoshock/psx/psx.cpp
+++ b/psx/octoshock/psx/psx.cpp
@@ -1094,6 +1094,7 @@ struct {
 		for(int i=0;i<10;i++)
 		{
 			ports[i].type = ePeripheralType_None;
+			ports[i].device = nullptr;	// devices are freed in reconstruct
 			memset(ports[i].buffer,0,sizeof(ports[i].buffer));
 		}
 		reconstruct(FIO);


Zinfidel:
I believe this is implemented by 28ae6d44134c46ce13b2fb335b9a581f2c6f8a46, closing.

PR #1847, opened 2020-02-14
Remove Text states as an option

adelikat:
This PR remove Text savestates which significantly simplifies the demands on a core to implement states.

Textstates are theoretically useful, but in the past 6 years or so, I've heard nobody talk about using them, and I suspect nobody cares about this feature .

Some consequences of this PR are that previous text states and text-state-anchored movies will no longer load.  This isn't so bad except for 2 cores: NESHawk, and PCEngine, which unfortunately has BinaryStatesPreferred = false for some reason.  Essentially we broke savestate compatibility for those 2 cores.  I think it is worth it though.


adelikat:
Abandoning this PR in favor of https://github.com/TASVideos/BizHawk/pull/1848
It leaves text states as an option but significantly reduces complexity and burdeon on core implementations

PR #1848, opened 2020-02-15
Make Binary states default and Textstates opt-in for cores

adelikat:
Due to some IRC feedback about https://github.com/TASVideos/BizHawk/pull/1847, I created this more mature handling of text states.

zeromus wanted text states as a feature and thinks it is valuable for core devs and other advanced users (I agree, though I don't think they are utilizing it).

This PR instead:
1) Removes BinaryStatesPreferred and makes Binary the default for all cores, the "Default" config option is removed as well and binary is the default setting
2) Create a default implementation of text states that's just binary as text
3) Create an ITextState interface a core can use if they want to opt into text states (and presumeably implement them properly as readable text)
4) all cores that use the Serializer opt into ITextState since this class properly does text states
5) rip out all the junk from cores that did not use Serializer since none of them produced readable text states

The advantages of this approach are that we retain backwards compatibility with the cores that had text states as the default previously, NesHawk (and consequently SubNesHawk), and PCEHawk, while still retaining the freedom of not requiring cores to implement text state methods

### DIFF
--- a/BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
+++ b/BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
@@ -1,4 +1,5 @@
 ﻿using System.IO;
+using BizHawk.Common.BufferExtensions;
 
 namespace BizHawk.Emulation.Common
 {
@@ -19,13 +20,54 @@ namespace BizHawk.Emulation.Common
 	/// </summary>
 	public interface IStatable : IBinaryStateable, IEmulatorService
 	{
-		void SaveStateText(TextWriter writer);
-		void LoadStateText(TextReader reader);
-
 		/// <summary>
 		/// save state binary to a byte buffer
 		/// </summary>
 		/// <returns>you may NOT modify this.  if you call SaveStateBinary() again with the same core, the old data MAY be overwritten.</returns>
 		byte[] SaveStateBinary();
+	}
+
+	/// <summary>
+	/// Allows a core to opt into text savestates.
+	/// If a core does not implement this interface, a default implementation
+	/// will be used that doesn't yield anything useful in the states, just binary as text
+	/// </summary>
+	public interface ITextStatable : IStatable
+	{
+		void SaveStateText(TextWriter writer);
+
+		void LoadStateText(TextReader reader);
+	}
+
+	public static class StatableExtensions
+	{
+		public static void SaveStateText(this IStatable core, TextWriter writer)
+		{
+			if (core is ITextStatable textCore)
+			{
+				textCore.SaveStateText(writer);
+			}
+
+			var temp = core.SaveStateBinary();
+			temp.SaveAsHexFast(writer);
+		}
+
+		public static void LoadStateText(this IStatable core, TextReader reader)
+		{
+			if (core is ITextStatable textCore)
+			{
+				textCore.LoadStateText(reader);
+			}
+
+			string hex = reader.ReadLine();
+			if (hex != null)
+			{
+				var state = new byte[hex.Length / 2];
+				state.ReadFromHexFast(hex);
+				using var ms = new MemoryStream(state);
+				using var br = new BinaryReader(ms);
+				core.LoadStateBinary(br);
+			}
+		}
 	}
 }
--- a/BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
+++ b/BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
@@ -19,11 +19,6 @@ namespace BizHawk.Emulation.Common
 	/// </summary>
 	public interface IStatable : IBinaryStateable, IEmulatorService
 	{
-		/// <summary>
-		/// Gets a value indicating whether the core would rather give a binary savestate than a text one. Both must function regardless
-		/// </summary>
-		bool BinarySaveStatesPreferred { get; }
-
 		void SaveStateText(TextWriter writer);
 		void LoadStateText(TextReader reader);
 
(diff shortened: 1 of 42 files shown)


BizHawk.Emulation.Cores/Consoles/WonderSwan/WonderSwan.IStatable.cs
```diff
@@ -39,25 +36,6 @@ private void SaveTextStateData(TextStateData d)
 			d.Frame = Frame;
 		}
 
-		public void SaveStateText(TextWriter writer)
```
nattthebear: These are actual text states, no?  Looks like we're calling `BizSwan.bizswan_txtstatesave` as opposed to `BizSwan.bizswan_binstatesave`.  I don't think they should be removed.

BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IStatable.cs
```diff
@@ -88,46 +72,5 @@ private void NewSaveCoreSetBuff()
...
-			public uint frameOverflow;
-			public bool IsCgb;
-		}
-
-		internal TextState<TextStateData> SaveState()
-		{
-			var s = new TextState<TextStateData>();
-			s.Prepare();
-			var ff = s.GetFunctionPointersSave();
-			LibGambatte.gambatte_newstatesave_ex(GambatteState, ref ff);
```
nattthebear: Same comment as other core:  Why are we removing text states from cores that actually implement them?

BizHawk.Emulation.Cores/Consoles/Nintendo/GBA/VBANext.IStatable.cs
```diff
@@ -1,41 +1,11 @@
...
 	public partial class VBANext : IStatable
 	{
-		public bool BinarySaveStatesPreferred => true;
-
-		public void SaveStateText(TextWriter writer)
-		{
-			var s = new TextState<TextStateData>();
-			s.Prepare();
-			var ff = s.GetFunctionPointersSave();
-			LibVBANext.TxtStateSave(Core, ref ff);
```
nattthebear: As others

BizHawk.Emulation.Cores/Consoles/Atari/lynx/Lynx.IStatable.cs
```diff
@@ -1,41 +1,12 @@
...
-
-			////Console.WriteLine(BizHawk.Common.BufferExtensions.BufferExtensions.HashSHA1(SaveStateBinary()));
-		}
-
-		public void LoadStateText(TextReader reader)
-		{
-			var s = (TextState<TextStateData>)_ser.Deserialize(reader, typeof(TextState<TextStateData>));
-			s.Prepare();
-			var ff = s.GetFunctionPointersLoad();
-			LibLynx.TxtStateLoad(Core, ref ff);
```
nattthebear: As others

BizHawk.Emulation.Cores/Computers/AppleII/AppleII.IStatable.cs
```diff
@@ -88,16 +86,6 @@ private void InitSaveStates()
 
 		private readonly JsonSerializer _ser = new JsonSerializer();
 
-		public void SaveStateText(TextWriter writer)
```
nattthebear: This one is even worse... Apple II _only_ has text states, and the binary states are actually text states stuffed in more or less.  Why remove this?

BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
```diff
@@ -20,17 +21,55 @@ public interface IBinaryStateable
...
+		{
+			if (core is ITextStatable textCore)
+			{
+				textCore.LoadStateText(reader);
+			}
+
+			string hex = reader.ReadLine();
+			if (hex != null)
+			{
+				var state = new byte[hex.Length / 2];
```
nattthebear: I know most cores don't pool their buffers for this, but given this one thing will be used in lots of different places, might be nice to pool buffers here

BizHawk.Emulation.Cores/Computers/AppleII/AppleII.IStatable.cs
adelikat: The next thing on my list is to refactor apple II states to binary

BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IStatable.cs
adelikat: ditto to wonderswan

BizHawk.Emulation.Cores/Consoles/Atari/lynx/Lynx.IStatable.cs
adelikat: ditto

BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
adelikat: I like the idea, and now we have a singular place to do it, I think that's for another PR down the road though

BizHawk.Emulation.Cores/Consoles/Nintendo/GBA/VBANext.IStatable.cs
adelikat: I'll assume ditto here, if not, oh well, we are a few mGBA memory hooks from deleting this core anyway

PR #1858, opened 2020-02-21
Creating a new branch AltirraHawk

ViGadeomes:
Hello, I would like to try to implement the Altirra emulator into BizHawk. I already created the branch. Thanks.


alyosha-tas:
Ported emulators should have their own folder, not be in emulation/cores. I'm not sure in this case but it might make more sense to make a submodule of this.

I strongly recommend you go to IRC and discuss how to make this happen in a structured way.

PR #1868, opened 2020-03-01
Remove unused sound files

adelikat:
We had ISynchronizingAudioBuffer with 3 implementations, Nitsuja, Zeromus, and Vecna's.

Turns out only the Vecna one is being used (in the PCE core) and the others haven't been used in quite some time.  And async sound providers have been going away for the type of cores we currently have.

This removes both impelmentations and simplifes resulting code


zeromus:
These might still be useful in accommodating mismatches due to selection of sync to video, audio, or clock, aka DRC; but that should be done at one point in the frontend rather than on cores. if the old methods are valuable, they can be resurrected from the history.

PR #1876, opened 2020-03-07
Deduplicate code common to CPCHawk and ZXHawk

YoshiRulz:
Net -6325 LOC for free*. @Asnivor please double-check 5196b9745 as well.

\* genericised types excepted


Asnivor:
I'm in two minds about this. CPCHawk is still very much WIP and the stuff you have deduplicated was just copied across for dev purposes and is subject to change.

If anything, the disk controller stuff should probably be abstracted out into its own folder like I did with the onebitbeeper (in BizHawk.Emulation.Cores/Sound). Likewise the AY sound chip should have this done as well and be re-written to be reusable. But this requires some rewrites.

Having a bunch of stuff shared between ZXS and CPC in the way that you have done just to deduplicate some code actually makes it more complicated going forward.

Can I suggest we leave this untouched for now? In 6 months or so when I have time I would like to decouple the various hardware devices that both systems use anyway, so we don't have to have unweildy interfaces living in an a separate '2systems' folder.

PR #1883, opened 2020-03-18
Remove cbuffer

adelikat:
I did some testing with Benchmark .NET, and got some pretty reliable cpu usage statistics, that showed that the CBuffer class is actually a bit slower than standard array usage.  Not by much, but the point of having a custom array class is for speed. 

Given that it is hurting more than it is helping, and that it adds extra complexity to the code base (homebrew implementation vs standard array usage, needs to be disposed), I think it is adding negative value.  This PR removes it and replaces it with byte[] and int[] instead.

Note, I tested, and this does not break savestate compatibility.  A savestate made with a buffer class, can be loaded into a standard array.


BizHawk.Emulation.Cores/Consoles/Atari/2600/Mappers/m4A50.cs
```diff
@@ -89,17 +89,11 @@ private byte[] RomImage
...
-		{
-			_ram.Dispose();
-			base.Dispose();
-		}
+		public override byte[] CartRam => _ram;
 
 		public override void SyncState(Serializer ser)
 		{
-			ser.Sync("cartRam", ref _ram);
+			ser.Sync("cartRam", ref _ram, false);
```
nattthebear: Does it make sense to add another overload for `Sync(string, ref T[])` so we don't need to add this `false` parameter in many places?

adelikat: Yeah, I don't think that's in scope for this PR, but it's something I could look at, I think it would also reduce the number of overloads we have for this method

nattthebear: Maybe it should be in scope for this PR, because if done now it would make this PR feel more natural/smooth?  Since diffs like this line wouldn't even exist, there'd be less noise in the git history and you'd convey the intent of this PR better:  There should be no change in savestate functionality.

Your call though, I have no strong feelings about it.

adelikat:
Just an excerpt from BenchMark .NET.  This is a method that simply iterated through a 100,000 length array (allocated before the benchmark method ran) and set values.  Then it iterated again and read values (setting those values to another pre-allocated array, to make sure logic didn't get compiled or JIT'ed away).

|            Method |       Job |              Toolchain |     Mean |     Error |    StdDev |
|------------------ |---------- |----------------------- |---------:|----------:|----------:|
|  ByteArrayMetrics | InProcess | InProcessEmitToolchain | 2.166 ms | 0.0017 ms | 0.0015 ms |
| ByteBufferMetrics | InProcess | InProcessEmitToolchain | 2.379 ms | 0.0067 ms | 0.0063 ms |
|  ByteArrayMetrics | OutOfProc |                         Default | 2.018 ms | 0.0010 ms | 0.0010 ms |
| ByteBufferMetrics | OutOfProc |                         Default | 2.253 ms | 0.0048 ms | 0.0042 ms |

I ran this several types for all 3 CBuffer types and got very consistent results like this.

YoshiRulz:
Again, benchmarking an array of 100000 elements is **very different** to benchmarking an array of a dozen elements. If it's at all possible to split the arrays into multiple fields, that will be of much greater impact.

PR #1884, opened 2020-03-19
DS Hawk

SuuperW:
Add MelonDS by Arisotura. (Emulation of Nintendo DS)
There are a few known issues, but from my testing it's certainly usable. Examples:
1. The only memory domains are system bus and RAM
2. Some graphical bugs, but I'm hoping these will be fixed by Arisotura eventually.
3. Power button isn't supported. (May also need work from Arisotura, it isn't clear to me how to do this.)

My fork of MelonDS: https://github.com/SuuperW/melonDS/tree/DSHawk

Note that while I did port/copy DeSmuME's HLE bios for this, not all games will work without actual NDS bios+firmware files.


InfamousKnight:
How exciting! I prefer this over desmume. Currently theres some building problems Ive been having with dshawk branch but Ill just wait until its fixed. The build I have now seems like enough.

adelikat:
So, I'm ready to merge this PR.  There is still some odds and ends to do to make this release ready but nothing that would get in the way of testing.

1) add MelonDS icon to put in the statusbar
2) add a DS controller image for Controller config
3) map some of hte unknown buttons to mnemonics in bk2 mnemonics

PR #1898, opened 2020-03-24
Fixes joypad.set() method in Lua Console

campbell000:
See https://github.com/TASVideos/BizHawk/issues/1897

At some point from 2.2.1 to now, joypad.set() no longer works properly. The following no longer works when it used to:

```
-- Example #1
local input =  {["A"] = true}
joypad.set(input, 1) -- Does nothing when it used to work in 2.2.1

-- Example #2
local input = {["A"] = 1}
joypad.set(input, 1) -- Throws exception when it used to work in 2.2.1

local input = {["P1 A"] = true}
joypad.set(input) -- Interestingly, this still works
```

This pull request fixes the first two scenarios. The main issue was with this block of code:
```
public void Set(Dictionary<string, bool> buttons, int? controller = null)
{
  foreach (var button in Global.InputManager.ActiveController.Definition.BoolButtons)
  {
    Set(button, buttons.TryGetValue(button, out var state) ? state : (bool?) null, controller);
  }
}
```
I've observed that `button` will ALWAYS be prefixed with "P{n}" (not sure why, but I assume this is the way the emulator is supposed to work), but the keys in `buttons` may not be (as in Example #1 above). Therefore, in the case of Example #1, `buttons` will contain {A: true} but the logic above will be looking for "P1 A", "P2 A", "P3 A", etc. This means that A's state will be set to null when it should be set to true.

Therefore, my solution was to change the Set() method to iterate over button names WITHOUT prefixes if a controller is specified. This (hopefully) preserves the existing `joypad.set({["P1 A"] = true})` syntax, but also fixes the bug that prevents `joypad.set({["A"] = true}, 1)` from working

### DIFF
--- a/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
+++ b/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
@@ -44,6 +44,8 @@ namespace BizHawk.Client.Common
 
 		public void Set(Dictionary<string, bool> buttons, int? controller = null)
 		{
+			// If a controller is specified, we need to iterate over unique button names. If not, we iterate over
+			// ALL button names with P{controller} prefixes
 			foreach (var button in Global.InputManager.ActiveController.ToBoolButtonNameList(controller))
 			{
 				Set(button, buttons.TryGetValue(button, out var state) ? state : (bool?) null, controller);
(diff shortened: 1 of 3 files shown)


adelikat:
Thanks for this fix!

And this is a nice PR, I like this use of an IController extension method, and that file is a good place to keep it.

For clarity yes, it is how it the emulator is "supposed" to work that a button is "P1 A" for consoles that can have multiple controllers.  joypad.set("P1 A", true) therefore is the preferred way to handle it.

The joypad.set("A", 1, true) style was actually put in for legacy reasons many years ago since that's how previous emulators did it, and it has kind of hung around ever since.  It is slightly more convenient though.

PR #1917, opened 2020-04-08
Fix the Gambatte Palette

TiKevin83:
New GBC LCD correction for Gambatte palette from documented shader research https://forums.libretro.com/t/real-gba-and-ds-phat-colors/1540

Extrems uses this same correction in GBI and helped port it to BizHawk-Gambatte.

I would suggest that this should at least be a new default if not replacing the Gambatte palette completely as in my repo.  We don't really know what Sinamas' Gambatte palette is trying to do, it doesn't match documented behavior of the GBC LCD.


nattthebear:
I hate this, so I'm merging it in the hopes that someone else hates it as well and it will piss them off.

zeromus:
You just changed the gambatte palette. Only the gambatte developer gets to do that.
Add a new palette instead.

PR #1926, opened 2020-04-13
Reading gameDB include files on background threads. 

drakewill-CRL:
Client starts up a little faster this way. (saves 350-500ms on my machine, likely more on slower PCs).

### DIFF
--- a/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/BizHawk.Client.EmuHawk/MainForm.cs
@@ -31,6 +31,7 @@ using BizHawk.Emulation.Common.Base_Implementations;
 using BizHawk.Emulation.Cores.Nintendo.SNES9X;
 using BizHawk.Emulation.Cores.Consoles.SNK;
 using BizHawk.Emulation.Cores.Consoles.Nintendo.Gameboy;
+using System.Threading.Tasks;
 
 namespace BizHawk.Client.EmuHawk
 {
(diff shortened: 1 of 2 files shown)


zeromus:
Decent idea, but unless there's also a line to match that somewhere else that waits synchronously to acquire access to the game database when it's needed, this is broken.

drakewill-CRL:
I was seeing a populated dictionary when I tested it, but just to make sure it's not a timing fluke, I switched it to use ConcurrentDictionary to handle write-locking without any additional changes.

zeromus:
That's not what I said needed to be done. To understand the problem with your code, add a "sleep" call for a second or so between loading each of the 1000s of records of the game DB. Now while that is busy taking a millennium to run, load a game that depends on it. Actually your new code is WORSE since it will potentially the bug by making it "work" (but read from incomplete data) rather than crash due to a fundamentally unsafe concurrent access

zeromus:
Nope, not accepting a busy loop as a solution to this. Find a proper synchronization method. Also, it's an okay idea to do it near the end of the many boot steps so that it did succeed in working in the background, but while we're doing this, it would be BETTER to wait until it was actually needed, so that the time the player spends going through the menus and picking a game can be used for loading it in the background

drakewill-CRL:
you know what? fuck it.

zeromus:
Thanks for playing, but threading is hard.

YoshiRulz:
:( It's their first PR ever, zeromus, why not just tell us how to `await` it properly?

zeromus:
I don't know how to `await` anything. I use ad-hoc methods appropriate to the situation. In this case I would probably make a manual reset event and wait on it when the DB is needed, and set the event when the DB is done being loaded, but I havent checked the code

PR #1954, opened 2020-04-27
GBA as GBC mode is only relevant on GBC

### LINKED ISSUE #1953: CGB in GBA mode effects GB console mode
This setting shouldn't have any effect on a GB console, but apparently it's still taking effect because the initial timer is wrong when the setting is turned on.

@TiKevin83

TiKevin83:
merging resolves #1953


alyosha-tas:
@TiKevin83 thank you!

Issue #1777, opened 2019-12-30
Bring mGBAHawk to feature parity with VBANext


endrift:
I've been told that mGBAHawk is missing a few features that are keeping it from being in-line feature-wise with the VBA-Next core, namely:

- [ ] ROM pokes (which the mGBA core does support, using mCore's rawWrite interface)
- [x] Memory hooks
- [x] Trace logging

It would be nice to add these features to mGBAHawk to remove one remaining reason to use VBA-Next.

NarryG:
If I'm not mistaken, ROM pokes are already wired up but just don't work because the domain is marked as read-only.

RetroEdit:
Implementing `event.oninputpoll` for the mGBA core would be useful for me. I want to avoid a situation where an input is entered at the beginning of a frame and then somehow changes before the poll occurs, which could potentially desync the script I'm working on.

vadosnaprimer:
Trace logging is implemented in dev.

adelikat:
We have mem hooks, input callbacks, cpu setting.  We officially have the same features in mGBA as VBA-next :)

PR #1969, opened 2020-05-02
Remove vba-next

adelikat:
This PR remove the vba-next core.
Now that the mGBA core has feature parity (#1777) and is more accurate there's no reason to keep this core laying around.


InfamousKnight:
Good idea. All the desync issues I had with it have been fixed.

PR #1980, opened 2020-05-05
Add testing w/ MSTest

YoshiRulz:
edit: I've force-pushed to replace a useless test with 6 useful ones, testing the SubstringExtensions class I recently added. It remains true that, in general, the likelihood of regressions decreases as we write more tests.

---

This is the command to run all tests (writes to `test_output/*.coverage.xml`):
https://github.com/TASVideos/BizHawk/blob/4bf3c71a49eb25d89d25c6f5af971a1288fc97a1/Dist/BuildTestRelease.sh#L3

GitLab CI on the mirror runs that command and parses the `.xml` results, producing [this](https://gitlab.com/TASVideos/BizHawk/pipelines/testing-boilerplate/latest) ("Tests" tab). I imagine you could do the same on GitHub, maybe with the new GitHub Actions thing?

---

You might notice the test project is empty save for a single `<Import/>`. I made it easy to add test projects for the rest of the assemblies by putting all the logic in a common `.props`, which drops the `BHTest.` prefix from the project's name to import the correct main project.

If you're not familiar with MSTest, neither was I until today. Just read [this](https://www.automatetheplanet.com/mstest-cheat-sheet).

---

Comments welcome.


alyosha-tas:
I don't get it, what is it testing?

Sappharad:
> I don't get it, what is it testing?

I agree. Unit tests are for your own code, and @YoshiRulz is testing the .NET Framework instead to make sure it can convert numbers to hex.

Sappharad:
> I don't get it, what is it testing?

The extension method in BizHawk.Common.NumberExtensions that allows him to call .ToHexString() on type int.

Which could be accomplished with $"{i:X}" using fewer characters.

alyosha-tas:
Oh, ok I thought it was running BizHawk somehow and getting a result. Thanks.
This seems like it would take an enormous amount of work to get practical results, yikes.

YoshiRulz:
I'd love to have tests like:
1. load savestate
2. frame advance for n frames
3. hash memory and assert

...but that's probably not going to happen.

Issue #1401, opened 2018-12-18
Gameboy cheat converter does not parse codes that use bank switching


BitwiseX:
Entering a gameboy Gameshark/Action replay code that starts with a 9 results in an error saying that only codes that start with a 0 or 1 are valid. 

Cheats that start with 9X do a write with bank switching. 91 would write the byte in Bank 1 whereas 92 would write the byte in bank 2. Code 934204D2 would write 0x42 to address 0xD204 in bank 3.

hegyak:
Can you submit a sample Cheat and Game and Effect?  Preferably something obvious.

BitwiseX:
Pokemon Silver, random encounter will be a Bulbasaur **910104D2** (taken from https://www.supercheats.com/gameboy/pokemon-crystal/4715/here-are-all-gameshark-for-pok/). I have no idea if this works, but most of the codes I find for Pokemon Silver seem to be 9XXXXXXX format.

I'm pretty used to making AR/GS codes so I knew beforehand that the first character is usually an instruction (like write byte or write word). I'd never seen a 9X code before today and it was surprisingly difficult to work out what they did. I only worked it out when I found this code converter/builder here: https://gamehacking.org/system/gb

hegyak:
I tend to use Gamehacking.Org myself when I need a cheat sample or some help with understanding what a cheat is doing.  Pokemon Crystal has several language ROMs.  I am assuming you are using these Cheats on the English ROM?  And I will need some time to figure this out as I haven't done coding for BizHawk in a while.

What do you mean by "Bank 1" and "Bank 2"  Are those Memory Domains?  What are they called in BizHawk (if you know their names).

BitwiseX:
Indeed it's the english version I'm talking about. I've no idea if the cheats are for the US or International versions - but the safe money is usually US when it comes to random codes on the internet.

In regards to memory domains... um, maybe? I'm not familiar with bizhawk teminology. Here's what I'm referring to:

My (limited and possibly mis-remembered) understanding of the gameboy is that it can't actually address all that much memory. In order to have higher capacity game carts memory is split into "banks". One bank would be mapped into a certain address range and then the game's code could make the gameboy switch which bank those addresses mapped to by writing values to a magic address.

P.S: Thanks for taking the time to look into this. I can edit the validation easily enough (it's in the confusingly named Gameshark.cs), but everything beyond that is beyond me.

NarryG:
[...]

PR #1999, opened 2020-05-12
NESHawk: make game genie compare cheats work

alyosha-tas:
Here is a PR that gets Game Genie compare cheats working properly in NESHawk.

It's applicability is limited to this specific case. Other compare cases (ex greater then) still do not work, as do other cheats outside of this scope (ex RAM.)

But having said that, I think the implementation is reasonably clean for what it does. I needed to make a new memory domain builder class specifically for NESHawk sys. bus so that no other cores are effected by this change, but other then that not many changes were needed.

Also performance isn't so bad. Even though it checks every ROM access, it only loses about 5% performance, so still can run comfortably at at least double speed for a small number of cheats (at least on my laptop.)

I purposely made the scope of this very narrow, but I think that's fine as other cores are way behind NESHawk in terms of cheats anyway.

Feedback welcome.

### DIFF
--- a/src/BizHawk.Client.Common/tools/Cheat.cs
+++ b/src/BizHawk.Client.Common/tools/Cheat.cs
@@ -1,4 +1,5 @@
 ﻿using BizHawk.Emulation.Common;
+using System;
 
 namespace BizHawk.Client.Common
 {
@@ -229,6 +230,20 @@ namespace BizHawk.Client.Common
 						case WatchSize.DWord:
 							_watch.Poke(((DWordWatch)_watch).FormatValue((uint)_val));
 							break;
+					}
+				}
+
+				// This will take effect only for NES, and will pulse the cheat with compare option directly to the core
+				// Only works for byte cheats currently
+				if (_watch.Size == WatchSize.Byte && _watch.Domain.Name == "System Bus")
+				{
+					if (Compare.HasValue)
+					{
+						_watch.Domain.SendCheatToCore((int)Address.Value, (byte)Value, Compare.Value, (int)ComparisonType);						
+					}
+					else
+					{
+						_watch.Domain.SendCheatToCore((int)Address.Value, (byte)Value, -1, 0);
 					}
 				}
 			}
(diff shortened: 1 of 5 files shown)


zeromus:
Other cores are no further behind than neshawk, because NO cores have proper cheat device support. All cores are equally far behind, 0.0 meters from the starting point. Actually, some cores (perhaps mgba for instance) may already have proper cheat support internally and thus be further ahead than neshawk.

I dislike more technical debt if the method for communicating cheats is insufficient. IMO, we should not be merging any cheats related work unless MULTIPLE platforms gain functional game genie cheats (AND multi-code AR cheats) in order to PROVE that the interface can hold up. This includes making sure cheats can move into unmanaged cores, in case that proves tricky for some reason (although the nature of cheats is such that strings can communicate them, so it shouldn't be hard)

I _definitely_ do not like "SendCheatToCore". First of all, any new addresses we make should be longs, not ints. After that, we shouldn't privilege our half-baked ill-defined "pulse cheats with compares" as being "Cheat". It should be done as a virtual cheat device, call it BizPulseCheat type, and sent as a string, like all the other cheats need to be. Adding a new pile of code specially for our pulse cheats is a no-go from me.

alyosha-tas:
You don't like SendCheatToCore? Man, I thought that was the best part. It kept me from having to manage anything on the core side since the tool already keeps track of them. (Changing addresses to long is easy enough though.)

I could maybe expand this to include #1401, but all that other stuff isn't for me, I just wanted NES game genie cheats to work.

adelikat:
Does this code actively get in the way of an ideal system that's easy to implement in any core?  Because it does give us some desired functionality in a core.  And we aren't going to come up with an ideal cheat system anytime soon.

nattthebear:
Ugggg ugggg uggg.  I don't want to impede this, but I don't see a good way to make it work either.  SendCheatToCore on memory domain feels weird and it is a highly NES-focused API going all over MemoryDomain.

alyosha-tas:
I purposely separated this into something that only effects NESHawk Sys. bus domain. If a more complete system ever comes along you can revert this and it won't effect any other core.

PR #2007, opened 2020-05-14
Revert 4a5ece2076d3570c13350db4b4

nattthebear:
Because of intervening commits, there may be some other incidental changes.

While well intentioned, the refactoring was just a mess when it came to actually groking this low level memory block shuffling code.


src/BizHawk.BizInvoke/MemoryBlockBase.cs
```diff
@@ -142,7 +151,8 @@ public override long Position
 
 			private void EnsureNotDisposed()
 			{
-				if (_owner.AddressRange.Start == 0) throw new ObjectDisposedException(nameof(MemoryBlockBase)); //TODO bug?
```
nattthebear, line 145: Regarding the TODO:  Yes, this does indicate a bug, but not here; in the calling code.  So nothing to change here.

src/BizHawk.Emulation.Cores/Waterbox/PeRunner.cs
```diff
@@ -430,13 +430,13 @@ public UIntPtr Brk(UIntPtr _p)
...
 
-				var start = heap.Memory.AddressRange.Start;
+				var start = heap.Memory.Start;
 				var end = start + heap.Used;
-				var max = heap.Memory.AddressRange.EndInclusive;
+				var max = heap.Memory.End;
 
 				var p = (ulong)_p;
 
-				if (p < start || max + 1 < p) //TODO bug?
```
nattthebear, line 439: I think we went over this one in IRC a few months back.  If you read the specification of the brk syscall thoroughly, there's no bug here.  It returns a pointer to the first invalid byte, not the last valid byte.

Elohimuel:
2020-05-13-162535-#01089645cc7275e267cae27a12afa8a3a8fe83ec
Was this the base build? ^-^ I picked this one up really and I had a question about it. = thank you very much as to what was included.

src/BizHawk.BizInvoke/MemoryBlockBase.cs
YoshiRulz: I was unclear: before I changed it, it accessed `_owner.Start`, which I assumed would return `default(int)` if `_owner` was GC'd. Now it's accessing `_owner.AddressRange`, which I assumed would return `default(Range<int>) == null` if `_owner` was GC'd. In both cases an exception is thrown by the method.

YoshiRulz:
@nattthebear I strongly disagree with reverting, especially because of how long ago that commit was made. As was discussed on IRC, it would be better practice (and probably easier) to change the existing code. This PR is only ~100 LoC and I doubt you need most of it. I suggest adding a single helper:
```cs
protected MemoryBlockBase(ulong start, ulong size)
{
	// ...
	Size = WaterboxUtils.AlignUp(size);
	EndExclusive = start + Size;
	AddressRange = start.RangeToExclusive(EndExclusive);
	// ...
}
// ...
public readonly ulong EndExclusive;
```

---

@Elohimuel That's a 2.5 dev build at 01089645c, yes, but it's not related to this PR at all. If you mean to ask about 2.5, there's no changelog yet, but you can look through the commit history to get some idea. It might be easier for you to ask questions in [IRC](https://github.com/TASVideos/BizHawk#support-and-troubleshooting).

src/BizHawk.BizInvoke/MemoryBlockBase.cs
nattthebear: What in the seven hells are you talking about?  `_owner` cannot possibly be GCed.  That's how C# works.  References don't disappear randomly; if it's observable by any code, it is not GCed.

YoshiRulz: Not GC'd sorry, *disposed*. TBH I don't fully understand either mechanism.

PR #2027, opened 2020-05-20
Revamp waterbox build environment and runtime.

nattthebear:
Create an all new waterbox build environment:
WSL2 + Ubuntu 20.04 LTS (Other linuxes may work)
Musl libc with waterbox customizations
LLVM's libclang-rt, libunwind, libcxxabi, libcxx
Static linking to elf files

Compared with the old system, this is easier to set up a dev env for and easier to update in the future.  The executables are larger but produce smaller savestates due to static linking.  The modern toolchain means advanced library features and language features that sometimes appear in some upstream cores will be reusable.


.gitmodules
```diff
@@ -19,3 +19,6 @@
 	path = melonds
 	url = https://github.com/TASVideos/melonDS.git
 	branch = SuuperW_DSHawk
+[submodule "waterbox/musl"]
```
nattthebear, line 22: @adelikat @vadosnaprimer Do we want to sobmodule LLVM?  We have not made any modifications to it, nor do we plan to, and it would increase the size of a full recursive checkout massively.

src/BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
```diff
@@ -18,11 +18,8 @@ public interface IBinaryStateable
 	/// If unavailable these options will not be exposed
 	/// Additionally many tools depend on savestates such as TAStudio, these will only be available if this service is implemented
 	/// </summary>
-	public interface IStatable : IEmulatorService
+	public interface IStatable : IBinaryStateable, IEmulatorService
```
nattthebear, line 21: @adelikat `IStatable` used to inherit `IBinaryStateable` but then you changed it not to and instead copy+paste the method signatures.  I changed it back here without thinking too much; any idea what was up with that?  The two interfaces do have the exact same semantic meaning.

src/BizHawk.Emulation.Cores/BizHawk.Emulation.Cores.csproj
```diff
@@ -5,6 +5,7 @@
 		<TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
 	</PropertyGroup>
 	<ItemGroup>
+		<PackageReference Include="ELFSharp" Version="2.10.0" />
```
nattthebear, line 8: Get rid of penet

src/BizHawk.Emulation.Cores/Consoles/Nintendo/SNES/LibsnesApi.cs
```diff
@@ -34,7 +34,7 @@ static LibsnesApi()
 			}
 		}
 
-		private PeRunner _exe;
+		private Waterbox.Waterbox _exe;
```
nattthebear, line 37: Uggg, waterbox is now `Waterbox.Waterbox`?  I forgot how awful naming a namespace and class the same thing was.

src/BizHawk.Emulation.Cores/Consoles/SNK/DualNeoGeoPort.cs
```diff
@@ -28,9 +28,10 @@ public class DualNeoGeoPort : IEmulator
 		[CoreConstructor("DNGP")]
 		public DualNeoGeoPort(CoreComm comm, byte[] rom, bool deterministic)
 		{
+			// this will throw on construct:  we need to compile two different copies of ngp at different starts
```
nattthebear, line 31: The comment is wrong, since we did change the start parameter.  Dunno what it will do, just deadlock or something?  Maybe the core should be axed entirely.

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+					(s.Flags & SectionFlags.Executable) != 0 ? "X" : " ",
+					_savedSections.Contains(s) ? "V" : " ",
+					s.Name,
+					s.Size);
+			}
+		}
+
+		private void PrintTopSavableSymbols()
+		{
+			Console.WriteLine("Top savestate symbols:");
```
nattthebear, line 132: If there are no symbols, don't print this line

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+				Console.WriteLine(text);
+			}
+		}
+
+		/// <summary>
+		/// Returns true if section is readonly after init
+		/// </summary>
+		private bool IsSpecialReadonlySection(Section<ulong> sec)
+		{
+			return sec.Name.Contains(".rel.ro")
```
nattthebear, line 151: Revisit why this section exists.  Is there any reason to compile libc with fPIC?

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+			{
+				if (s.Size == 8)
+				{
+					var p = syscalls.GetProcAddrOrThrow(s.Name);
+					tmp[0] = p;
+					Marshal.Copy(tmp, 0, Z.US(s.Value), 1);
+				}
+				else
+				{
+					if ((s.Size & 7) != 0)
```
nattthebear, line 194: bitwise is too cute here, change to mod / div

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+						Marshal.Copy(tmp, 0, Z.US(s.Value + (ulong)(i * 8)), 1);
+					}
+				}
+			}
+
+			Protect();
+		}
+
+		[UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+		private delegate void ElfEntryDelegate();
```
nattthebear, line 210: Zap this and just use `Action`

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+			CallingConventionAdapters.Waterbox.GetDelegateForFunctionPointer<ElfEntryDelegate>(Z.US(_elf.EntryPoint))();
+		}
+
+		public void SealImportsAndTakeXorSnapshot()
+		{
+			if (_everythingSealed)
+				throw new InvalidOperationException($"{nameof(ElfLoader)} already sealed!");
+
+			// save import values, then zero them all (for hash purposes), then take our snapshot, then load them again,
+			// then set the .idata area to read only
```
nattthebear, line 223: Comment misleading because there's no `.idata` section.  Use the right name

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+			byte[] impData = null;
+			if (_imports != null)
+			{
+				impData = new byte[_imports.Size];
+				Marshal.Copy(Z.US(_imports.LoadAddress), impData, 0, (int)_imports.Size);
+				WaterboxUtils.ZeroMemory(Z.US(_imports.LoadAddress), (long)_imports.Size);
+			}
+			else
+			{
+				throw new Exception("Call natt (_imports??)");
```
nattthebear, line 233: fixme

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+
+		public IntPtr GetProcAddrOrThrow(string entryPoint)
+		{
+			if (_visibleSymbols.TryGetValue(entryPoint, out var sym))
+			{
+				return Z.US(sym.Value);
+			}
+			else
+			{
+				throw new InvalidOperationException($"Couldn't find {nameof(entryPoint)} {entryPoint} in {ModuleName}");
```
nattthebear, line 288: @YoshiRulz Once upon a time, `GetProcAddrOrThrow` was an extension method and not actually in the interface.  It's a lot of noise reimplementing the exact same thing the exact same way every time.  On the other hand, the individual implementations provide more relevant exception messages for each usecase.

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
```diff
@@ -0,0 +1,356 @@
...
+			}
+		}
+
+		const ulong MAGIC = 0x6018ab7df99310ca;
+
+		public void SaveStateBinary(BinaryWriter bw)
+		{
+			if (!_everythingSealed)
+				throw new InvalidOperationException(".idata sections must be closed before saving state");
```
nattthebear, line 297: As above, don't say `.idata`

src/BizHawk.Emulation.Cores/Waterbox/EmuLibc.cs
```diff
@@ -0,0 +1,42 @@
...
+		[BizExport(CallingConvention.Cdecl, EntryPoint = "__walloc_plain")]
+		public IntPtr AllocPlain(UIntPtr size)
+		{
+			return Z.US(_parent._plainheap.Allocate((ulong)size, 16));
+		}
+
+		[BizExport(CallingConvention.Cdecl, EntryPoint = "__w_debug_puts")]
+		public void DebugPuts(IntPtr s)
+		{
+			Console.WriteLine("_debug_puts:" + Marshal.PtrToStringAnsi(s));
```
nattthebear, line 39: If this is supposed to behave like puts, should it really adorn its output with `_debug_puts`?  Alternatively, should this function even exist?  It's an artifact from when waterbox didn't even have a functioning stdio

waterbox/common.mak
```diff
@@ -0,0 +1,104 @@
...
+DOBJS := $(patsubst $(ROOT_DIR)%, $(DOBJ_DIR)%, $(_OBJS))
+
+$(OBJ_DIR)/%.c.o: %.c
+	@echo cc $<
+	@mkdir -p $(@D)
+	@$(CC) -c -o $@ $< $(CCFLAGS) $(CCFLAGS_RELEASE)
+$(OBJ_DIR)/%.cpp.o: %.cpp
+	@echo cxx $<
+	@mkdir -p $(@D)
+	@$(CC) -c -o $@ $< $(CCFLAGS) $(CXXFLAGS) $(CCFLAGS_RELEASE)
```
nattthebear, line 49: At least one of our cores needs CCFLAGS that aren't CXXFLAGS and currently spews 9 billion errors as a result.  Revisit

src/BizHawk.Emulation.Cores/Waterbox/ElfLoader.cs
YoshiRulz: Extension method sounds like the better option, I'll put it on my to-do list.

src/BizHawk.Emulation.Cores/BizHawk.Emulation.Cores.csproj
YoshiRulz: Is this a TODO, or is PeNet made redundant by this PR?

nattthebear: PeNet is made redundant by this PR, so I am going to yoink it.  PeWrapper is gone.

src/BizHawk.Emulation.Common/Interfaces/Services/IStatable.cs
adelikat: It was my attempt to detangle them (eventually).  IStatable should be a BinaryStatable.  But I can detangle later.  If this makes for a cleaner waterbox class, go for it

nattthebear: I don't see how anything is "entangled"?  We have one interface which does one universal thing throughout Bizhawk.

.gitmodules
vadosnaprimer: I care about having a fork we'd independently work with, in this organization. If it's problematic to have the entire thing submoduled, releasing our fork's binaries and using them here may be a way to go?

zeromus: I would rather see a separate repository, which has bizhawk as a submodule and also contains all the ginormous build parts. This way the build parts can count on emitting dlls into the appropriate location in the bizhawk directory for testing/

src/BizHawk.Emulation.Cores/Consoles/Nintendo/SNES/LibsnesApi.cs
nattthebear: Renamed

src/BizHawk.Emulation.Cores/Consoles/SNK/DualNeoGeoPort.cs
nattthebear: Broke off into another issue

.gitmodules
nattthebear: Just got an even better idea:  Let's see what happens when other people try to recreate the environment.

PR #2034, opened 2020-05-21
Fix VRAM end of line testROM and a desync in GB Wario Land II

### LINKED ISSUE #2033: Gambatte: Wario Land 2 Desync (compared to real console)
Logging an issue for the Wario Land 2 console desync investigation

I'll be sending a PR which adds a corner case to VRAM reads, where the first time VRAM is accessed below 0x9000 on the exact cycle of the switch to mode0, the read should return 0x00 instead of the block 0xFF or the expected VRAM read.  This behavior has been verified in a testROM with mattcurrie and fixing the testROM also fixes the console desync.

GBP result
image

Before EMU
image
After EMU
image

Console Behavior compared to before in EMU
[WL2ConsoleLeftEmuRight.zip](https://github.com/TASVideos/BizHawk/files/4663285/WL2ConsoleLeftEmuRight.zip)

EMU behavior after fixing the testROM
image

TiKevin83:
Fixes #2033


nattthebear:
I'll click the button on this if no one with more GB knowledge chimes in

alyosha-tas:
I'm kind of skeptical of this fix. It's good enough for now I guess, but I thought a more rigorous test ROM was in the works. And what about GB mode, I thought different stuff happened there?

TiKevin83:
I don't think there's anything more rigorous to do with the testROM, it seems to cover all known edge cases for both GB and GBC/GBP, however you're correct that this only handles the edge case from the GBC/GBP perspective, and more would need to be done to wire in the ppu_bus value for GB.

TiKevin83:
I have a pass solution that accounts for both the new testROM and the GSR test suite, waiting to commit it until I have confirmation of the exact range of VRAM addresses affected by the WL2 bug.

alyosha-tas:
What test does it break?

I'm just skeptical that it really is just a one time read. I'm no hardware expert but that doesn't seem too likely to me. And the test ROM only tests it a few times. If it's something more abstract (like maybe it aligns with RAM refresh or something) it might not get all the details. (Take a look at sc_change_freq_gbc.gbc sometime, there's some crazy stuff happening.)

TiKevin83:
This is the test that breaks, it uses VRAM at 0x8000 which appears to not be affected by the bug
[prewrite_ds_1_cgb04c_out1.zip](https://github.com/TASVideos/BizHawk/files/4663874/prewrite_ds_1_cgb04c_out1.zip)

alyosha-tas:
That test uses double speed mode, so that might have more to do with it then the address.

TiKevin83:
Yeah the existing VRAMReadable code would certainly indicate an off by 1 in double speed mode that avoids the quirk, I've adjusted for that now and am passing both tests.

We think it's less that it only can happen once and more that you get the quirk until the first successful read in similar conditions (the quirk happens when SCX=3 and a success can happen in the same conditions when SCX=2).  GBHawk's change certainly doesn't match the testROM for GBC, on console it gets 0x00 only on the first pass and then 0x11 for the 2nd two, consistently across several power off/ons, but GBHawk is always returning 0x00.

alyosha-tas:
Yeah I haven't updated the implementation in GBHawk since I first conjectured what it might be. I think at this point I'd still rather fail the test then implement it as a one-off.

But that's just me, at least now the Warioland II TAS can work correctly in emulator now.

PR #2037, opened 2020-05-23
Remove infinity movies

adelikat:
This was a gimmick feature a long time ago, and has only been used twice to my knowledge.  It's a feature that would never lead to a tasvideos publication. 

It isn't that much code, but it does complicate the api (such as frame count being a double).  If we were to expand capabilities to include subframe input, this feature would actively complicate the implementation.

It was fun, but I think maybe it is time to go, what do people think?

### DIFF
--- a/src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
+++ b/src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
@@ -46,12 +46,12 @@ namespace BizHawk.Client.Common
 
 		/// <summary>
 		/// Gets the total number of frames that count towards the completion time of the movie
-		/// Possibly (but unlikely different from InputLogLength (could be infinity, or maybe an implementation automatically discounts empty frames at the end of a movie, etc)
 		/// </summary>
-		double FrameCount { get; }
+		int FrameCount { get; }
 
 		/// <summary>
-		/// Gets the actual length of the input log, should only be used by code that iterates or needs a real length
+		/// Gets the actual length of the input log, should only be used by code that needs a the input log length
+		/// specifically, not the frame count
 		/// </summary>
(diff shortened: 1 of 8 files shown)


src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
```diff
@@ -46,12 +46,12 @@ public interface IMovie
 
 		/// <summary>
 		/// Gets the total number of frames that count towards the completion time of the movie
-		/// Possibly (but unlikely different from InputLogLength (could be infinity, or maybe an implementation automatically discounts empty frames at the end of a movie, etc)
 		/// </summary>
-		double FrameCount { get; }
```
adelikat, line 51: So I pondered InputLogLength vs FrameCount and wanted to document my thinking.  I considered dropping one of them since they are the same. But we could anticipate they would not if we had a notion of "subframe" input.  

Normally I do not like prebuilding since in my experience it is often done wrong (you never fully anticipate future needs).  However, the work has already been done here and the calling code already has sorted out whether it cares about frame count or actual log length so I kept it in there.

Also, a double is arguably a good value for FrameCount if subframes are a thing.  But I decided that's nonsense and an implementation should define 100.5 frames as 101 frames.  Or at the very least, it deserves a future nerd fight

src/BizHawk.Client.EmuHawk/MainForm.FileLoader.cs
```diff
@@ -25,7 +25,7 @@ private enum LoadOrdering
 			LegacyMovieFile
 		}
 
-		public struct FileInformation
```
adelikat, line 28: I did not mean to include this useless commit with this PR, sloppy branch management

PR #2052, opened 2020-05-26
fix console garbage

nattthebear:
Add EmuHawkConsole.exe which forces EmuHawk.exe to run with a console.  Switch output type back to winexe


nattthebear:
@adelikat @vadosnaprimer I know you two were most interested in this

vadosnaprimer:
Doesn't seem to launch when running hawk from VS. When I run EmuHawkConsole.exe, it launches hawk, but it doesn't run through VS, so breakpoints don't work. Closing hawk crashes the console.

nattthebear:
Right, you'd have to attach yourself.  Didn't think of that one too much; all of my debugging right now is attach

nattthebear:
> Closing hawk crashes the console.

Yeah, have to fix exactly what handles are closed by what when

PR #2093, opened 2020-06-04
More cleanup to the new tasproj filename code

TiKevin83:


YoshiRulz:
While you can be bothered, could you write a short unit test? Bonus points if you commit the test before the refactor and it passes both times. [example](https://github.com/TASVideos/BizHawk/blob/master/src/BizHawk.Tests/Client.Common/Movie/LogGeneratorTests.cs)

TiKevin83:
Good suggestion, I'll see what I can do

TiKevin83:
I created a test for the scenario I fixed in the cleanup where extensions were duplicated on bk2s.  I like your idea to test the already working behavior with .tasproj extensions before/after the change, not sure how to work that in a platform-agnostic way

PR #2111, opened 2020-06-08
update msx bios filenames and description

LodanZark:
Updated the msx bios filename based on mame 0.221 and cbios0.29a, this way it fixes the filename conflict while using the feature "Organize" through Firmware Manager.
Couldn't figure it out about the bios_test origin, so I kept that one unchanged.


alyosha-tas:
Might change again in the future, but if it fixes a problem now good enough for me.

PR #2155, opened 2020-06-20
Create a generic core menu

adelikat:
Make a generic core menu that includes a Settings menu item that opens a GenericCoreConfig dialog.  Remove a lot of hardcoded menus that only did this.  Default to this, if system id is not recognized.

This reduces a lot of boilerplate code that is necessary to spin up a new core.  And we can make this more clever later on and remove other hardcoded menu items (such as dynamically adding core specific tool dialogs)


vadosnaprimer:
Does/will it include core switches for systems with several cores?

adelikat:
We already have a dynamic core picker.  If there is a core with a specialized menu and a core that isn't, for the same system, currently an if is necessary.  We can make that more clever though, if we try hard enough

PR #2164, opened 2020-06-22
Fix handling of duplicate Ram Watches

scrimpeh:
Fixes two issues relating to duplicate Ram Watches

- Because equality is checked only by address, domain and size, if there is multiple Ram Watches present, deleting one would always delete the first one
- The "Duplicate Watch" dialog would silently discard a ram watch if its address was identical to the previous one. This changes the behaviour of AddRange of WatchList, but as the method is used only here, this should be safe.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.cs
@@ -350,14 +350,6 @@ namespace BizHawk.Client.EmuHawk
 
 			if (SelectedWatches.Any())
 			{
-				foreach (var sw in SelectedWatches)
-				{
-					if (sw.Domain != SelectedWatches.First().Domain)
-					{
-						throw new InvalidOperationException("Can't edit multiple watches on varying memory domains");
-					}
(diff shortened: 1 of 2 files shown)


src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.cs
```diff
@@ -350,14 +350,6 @@ private void EditWatch(bool duplicate = false)
 
 			if (SelectedWatches.Any())
 			{
-				foreach (var sw in SelectedWatches)
```
adelikat, line 353: The reasoning behind this check was range considerations.  If you have a nes wram and system bus domain, and add 0x1000 for isntance, that's an invalid wram address but valid system bus.  So to not let the user add out of range addresses we restricted to a single comain

adelikat: After some thought and discusison, we agreed this is okay.  There isn't much you can do in the editor with multiple watches anyway (addresses is grayed out).  And we gracefully handle out of range addresses anyway.

Issue #1212, opened 2018-06-10
Missing audio and crashes when changing output device


Amphitryon0:
When playing a game through headphones, removing them from the computer sometimes (about 70 percent of the time) results in no game audio playing through my speakers, and no game audio plays through my headphones when I plug them back in.  I can re-enable sound by opening the sound configuration dialog and clicking OK with headphones in, but this crashes the program if I do it without headphones in when the sound is not working. This crash results in the following dialog.
![2018-06-09t23_20_22-bizhawk_error_cropped](https://user-images.githubusercontent.com/17864580/41198647-5a6ca0b8-6c48-11e8-97b7-9d98ab8b57de.png)
I am using DirectSound with the sound device set to \<default\> and a buffer size of 100 ms.
This issue happens on both the latest stable version (2.2.2) and the latest development version I compiled (173bb2a50).

Asnivor:
It sounds like your soundcard maybe has two disparate logical sound devices internally. When you unplug the headphones, the soundcard disables the one that is 'wired' to the headphone jack socket, and enables the the one 'wired' to your speaker/lineout jack socket. Or the soundcard is just temporarily re-initialising its buffer when this happens.

Regardless, SlimDX is initialised when a core loads. Unless someone can figure out how to catch this exception in the bizhawk audio-subsystem and re-init the sound provider (if this is even possible/recommended), then I'm not sure how to fix it.

Are you able to try the other output method options in Bizhawk (config -> sound configuration) and see if the error still happens?

![image](https://user-images.githubusercontent.com/7911038/44195159-1c4b5000-a130-11e8-84af-3d57d67eea4e.png)

Asnivor:
I wouldn't know if it were a common problem or not. But there are certainly many machines with onboard sound where this is a possibility (for whatever unknown reason).

Asnivor:
@Amphitryon0 did you get any further with this?

YoshiRulz:
It would be very easy to catch a `DirectSoundException` in [here somewhere](https://github.com/TASVideos/BizHawk/blob/master/BizHawk.Client.EmuHawk/Sound/Output/DirectSoundSoundOutput.cs), but is this bug our problem? I can't even follow the stacktrace.

zeromus:
actually what probably needs to be done is proper handling of the return errors to look for the device lost status. the most annoying part of directx programming, by far

PR #2166, opened 2020-06-23
Handling for BufferLost

diddily:
Fixes issue #1212 where adding/removing headphones would lead to a crash by restoring the sound device when it's detected and being defensive with exception handling.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/Sound/Output/DirectSoundSoundOutput.cs
+++ b/src/BizHawk.Client.EmuHawk/Sound/Output/DirectSoundSoundOutput.cs
@@ -51,9 +51,21 @@ namespace BizHawk.Client.EmuHawk
 
 		public void ApplyVolumeSettings(double volume)
 		{
-			// I'm not sure if this is "technically" correct but it works okay
-			int range = (int)Volume.Maximum - (int)Volume.Minimum;
-			_deviceBuffer.Volume = (int)(Math.Pow(volume, 0.1) * range) + (int)Volume.Minimum;
+			try
+			{
+				// I'm not sure if this is "technically" correct but it works okay
+				int range = (int)Volume.Maximum - (int)Volume.Minimum;
+				_deviceBuffer.Volume = (int)(Math.Pow(volume, 0.1) * range) + (int)Volume.Minimum;
+			}
+			catch (DirectSoundException ex)
+			{
+				_deviceBuffer.Restore();
+			}
+		}
+
+		public bool SoundLost()
+		{
+			return (_deviceBuffer.Status & BufferStatus.BufferLost) != 0 || _deviceBuffer.Status == BufferStatus.None;
 		}
 
 		public void StartSound()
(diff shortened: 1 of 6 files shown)


zeromus:
I'm a little skeptical. For instance, we have essentially this:
```
try { play() } catch { restore(); } 
```
so... if it fails... is it playing, or not? It isn't obvious how to handle each of these things. Unless you have synthesized a fake error during testing from each place that can throw a devicelost exception, it's essentially _impossible_ to be sure it's been done right. I would rather the code obviously make no attempt to get it right than to make a messed-up attempt at getting it right.

diddily:
So if it fails on the particular line you called out, it'll be not playing, and thus it's status will be None, so next time around it will try again to start up (it might also be BufferLost which will also be handled). The restores in each catch block are mostly to attempt to get the device up faster, in my testing I found that if the device actually switches, restore was insufficient to bring the sound back (until you brought it back to the initial state, i.e. headphones->speaker->headphones). We ultimately will fall back to the call of StopSound()/StartSound() which 100% resolves the issue.

I was able to hit exceptions in each location I wrapped in my testing (I am blessed with having this bug local hence my desire to see it fixed) with the exception of Play, and I know each of them end up resolved next frame. I am confident that if it were to fire on Play, it would fail for the current frame and then the next time we try to output sound it would retrigger and resolve.

nattthebear:
I object to `ISoundOutput` having a `bool SoundLost();` property.  That's an implementation detail of some sound output devices.  If DSound needs to reestablish some internally managed native handle, let it deal with that without leaking to the outside.

diddily:
Fair, I figured it might apply to other devices in the future and kept the code a bit cleaner (the idea of calling StopSound()/StartSound internally was unpleasant to me). I can refactor it to keep it entirely internal.

nattthebear:
Yes, my request is a bit sticky to solve, as you'll either need to add some layer of encapsulation, duplicate code, or call StopSound() / StartSound() internally.  I'll eat my words if we end up having similar semantics in other ISoundOutputs.

One possible way to do this would be something like this:

diddily:
My current plan is to keep the try/catch blocks where I have them, then add a check like SoundLost to the top of WriteSamples and do the recovery there as that seems the most appropriate. Unfortunately the first place that will throw an exception in the common case is in ApplyVolumeSettings, and I don't really want the recovery code to be there and I'd rather not have every exception block try to do the recovery and have a bunch more edge cases.

I was fully expected a reaction to adding to the interface, I don't know you're coding standards so I figured I'd do what felt best and ask forgiveness afterwards. I don't mind the stickiness, it's a very solvable problem.

nattthebear:
> I don't know your coding standards

Oh, there aren't any

diddily:
Take 2. Anything that could throw an exception is wrapped so we don't crash, all the handling for restoring state is in one place.

src/BizHawk.Client.EmuHawk/Sound/Output/DirectSoundSoundOutput.cs
```diff
@@ -90,55 +124,77 @@ public void StartSound()
...
-			_lastWriteCursor = 0;
+			while (_deviceBuffer == null)
+			{
+				try
+				{
+					_deviceBuffer = new SecondarySoundBuffer(_device, desc);
+				}
+				catch (DirectSoundException)
+				{
+					System.Threading.Thread.Sleep(10);
```
nattthebear, line 135: This seems very hazardous.  If, for any `DirectSoundException` reason, the buffer cannot be acquired at all, we'll endlessly loop.

diddily: I can have it try once per frame instead, it might take a little longer to reacquire but the rest of the code should now be tolerant to being in a non-recovered situation. Or I can change it to break out of the loop after N retries, which would you prefer?

nattthebear: I assume there are real world situations where the buffer will fail to create but it will succeed after Sleep(10)?  DirectSound is such a drag.

How about N retries?

diddily: I'll have to add a check for null _deviceBuffer in StartPlaying and attempt to create it there as well as currently there is no other attempt to create the buffer other than StartSound which is called on start and window resizing. Otherwise sound would not be able to return.

If there is a persistent exception being thrown, performance is going to tank with the sleep. I'll have the N retry counter fall off to the point where if it's consistent, it'll only try once per frame. Start at 5 and drop by one each time it  doesn't succeed.

PR #2173, opened 2020-06-27
N64 Plugin Config - Use Property Grid

adelikat:
Pros:
*lot's less code
*less maintenance burden for adding new versions of plugins with new options
*less maintenance burden for adding new plugins

Cons:
*The previous UI had a tab for game specific hacks.  Now it isn't as clear which can be per game and which can't. However, why can't any value be per game? They probably can, or it would be an easy fix


vadosnaprimer:
Haven't looked or tested but this sounds brilliant! Too bad I don't want to touch n64 anymore, but still an improvement.

PR #2174, opened 2020-06-27
RAM Watch Domain Fixes

scrimpeh:
Various fixes relating to RAM Watch Domains and Display

* Added an `IsValid` property to `Watch`, which is used to determine if a RAM watch is within range
* If a RAM watch is out of range, "-" is displayed
* Pokes on out of range watches no longer do anything
* `IMiniWatch` and `IMiniWatchDetails` have been adjusted in a similar way, however, out of range behaviour cannot be tested because loading from .wchs is broken.
* Display Type Binary on 32-bit watches was the same as Unsigned, this was fixed
* Changing a watch's size to one where the display type is incompatible (e.g. 32 bit float -> 16 bit unsigned) resulted in a exception, this was also fixed

Possible extensions:

* Disable Poke dialog entirely on out of range watches
* Always filter Out of Range addresses with any Ram Search criteria.

Considering that out of range watches aren't supposed to be the normal case anyway, I wouldn't invest too much time into that, however.

### DIFF
--- a/src/BizHawk.Client.Common/tools/RamSearchEngine/IMiniWatch.cs
+++ b/src/BizHawk.Client.Common/tools/RamSearchEngine/IMiniWatch.cs
@@ -11,6 +11,7 @@ namespace BizHawk.Client.Common.RamSearchEngine
 		long Address { get; }
 		long Previous { get; } // do not store sign extended variables in here.
 		void SetPreviousToCurrent(MemoryDomain domain, bool bigEndian);
+		bool IsValid(MemoryDomain domain);
 	}
 
 	internal sealed class MiniByteWatch : IMiniWatch
@@ -21,14 +22,29 @@ namespace BizHawk.Client.Common.RamSearchEngine
 		public MiniByteWatch(MemoryDomain domain, long addr)
 		{
 			Address = addr;
-			_previous = domain.PeekByte(Address % domain.Size);
+			_previous = GetByte(Address, domain);
 		}
 
 		public long Previous => _previous;
 
 		public void SetPreviousToCurrent(MemoryDomain domain, bool bigEndian)
 		{
-			_previous = domain.PeekByte(Address % domain.Size);
+			_previous = GetByte(Address, domain);
+		}
+
+		public bool IsValid(MemoryDomain domain)
+		{
+			return Address < domain.Size;
+		}
+
+		public static byte GetByte(long address, MemoryDomain domain)
+		{
+			if (address >= domain.Size)
+			{
+				return 0;
+			}
+
+			return domain.PeekByte(address);
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


src/BizHawk.Client.Common/tools/RamSearchEngine/IMiniWatch.cs
```diff
@@ -21,14 +22,29 @@ internal sealed class MiniByteWatch : IMiniWatch
...
+		}
+
+		public bool IsValid(MemoryDomain domain)
+		{
+			return Address < domain.Size;
+		}
+
+		public static byte GetByte(long address, MemoryDomain domain)
+		{
+			if (address >= domain.Size)
```
adelikat, line 42: This technically works but isn't it better to use IsValid here?

src/BizHawk.Client.Common/tools/RamSearchEngine/IMiniWatch.cs
```diff
@@ -40,14 +56,29 @@ internal sealed class MiniWordWatch : IMiniWatch
...
+		}
+
+		public bool IsValid(MemoryDomain domain)
+		{
+			return Address < (domain.Size - 1);
+		}
+
+		public static ushort GetUshort(long address, MemoryDomain domain, bool bigEndian)
+		{
+			if (address >= (domain.Size - 1))
```
adelikat, line 76: IsValid here too, don't want to duplicate logic

src/BizHawk.Client.Common/tools/RamSearchEngine/IMiniWatch.cs
```diff
@@ -59,14 +90,29 @@ internal sealed class MiniDWordWatch : IMiniWatch
...
+		}
+
+		public bool IsValid(MemoryDomain domain)
+		{
+			return Address < (domain.Size - 3);
+		}
+
+		public static uint GetUint(long address, MemoryDomain domain, bool bigEndian)
+		{
+			if (address >= (domain.Size - 3))
```
adelikat, line 110: IsValid()

src/BizHawk.Client.Common/tools/Watch/ByteWatch.cs
```diff
@@ -170,6 +170,7 @@ public string FormatValue(byte val)
 		{
 			return Type switch
 			{
+				_ when !IsValid => "-",
```
adelikat, line 173: Nice!

src/BizHawk.Client.DiscoHawk/BizHawk.Client.DiscoHawk.csproj
```diff
@@ -70,13 +70,13 @@
 		<Content Include="discohawk.ico" />
 	</ItemGroup>
 	<ItemGroup>
-		<Compile Update="About.cs" SubType="Form" />
```
adelikat, line 73: Did you mean to change this file?

scrimpeh: Nope, this must've been VS. I reverted it back to the previous state.

PR #2178, opened 2020-06-29
Extract interface from MainForm

YoshiRulz:


src/BizHawk.Client.EmuHawk/tools/ToolManager.cs
```diff
@@ -247,7 +247,7 @@ private void RefreshSettings(Form form, ToolStripItemCollection menu, ToolDialog
 			form.TopMost = settings.TopMost;
 
 			// do we need to do this OnShown() as well?
-			form.Owner = settings.FloatingWindow ? null : _owner;
+			form.Owner = settings.FloatingWindow ? null : (Form) _owner;
```
adelikat, line 250: If you have to assume an interface is a form, it's a highly suspicious use of an interface

YoshiRulz: Giving up and using MainForm (class) in ToolManager removes these casts, and as bonus it led me to split up the interface, where I'd failed to before.

PR #2180, opened 2020-06-30
Split BizHawk.Emulation.Cores into multiple projects

YoshiRulz:
This, like 3a3b22c03, will split the commit history on GitHub.

~~TODO: change the namespaces to match the projects~~ that would break loading sync settings from movies; if we want this it should be per core, and after a breaking change to sync settings
~~TODO: fix ext. tools in repo~~ predicated on above TODO


nattthebear:
Is more projects better?  What's the ideal number of projects; are we splitting for compile time or organization or something?

adelikat:
I don't think this improves the quality of the code base

YoshiRulz:
I don't know if more projects is necessarily better, or how many is the "right amount". But in this case, adding more projects should increase build speed (at the very least, for those using VS and working on cores). Having the Cores and EmuHawk projects being > 100 kLOC and most < 10k LOC creates bottlenecks. For EmuHawk, it's not important, as it's the last to be built.

adelikat:
There is some reflection magic this will break if you aren't careful.  We look for non-standard emulator services by typeof(somethingInCores) that no longer can make that assumption.

YoshiRulz:
On branch [`reflection-cache`](https://github.com/TASVideos/BizHawk/tree/reflection-cache) there's a commit to make the assembly explicit when getting defined types by reflection. I think that should be checked and merged into `master` before revisiting this.

YoshiRulz:
`Emulation.Cores.ReflectionCache.Types` is used 5 times: 3 are to enumerate cores (this usage will need to be updated to search all cores assemblies—I'm hoping to consolidate the calls into `CoreInventory` first); the other 2 are in NesHawk. The latter doesn't necessarily need to change as NesHawk types will stay together. For reference, they're in finding peripherals (I passed over this for d170972e3), and finding boards.

`typeof` is used 178 times in the Cores project plus 3 times in commented-out code. It's used for `IEmulatorServices`, controller decks, and CPUs, and for `[DefaultValue]`, and a few times with primitive types.

I don't have an estimate for how much `typeof` is used with a type from Cores in one of its dependent projects, but a quick look at Client.Common and EmuHawk suggests it would be with settings and sync settings types. I don't think anything but `using`s would need to change for those.

adelikat:
This has been here a long time with no progress, and it isn't an approach we want to take anyway

PR #2190, opened 2020-07-02
Rewrite WaterboxHost in rust.

nattthebear:
Why rewrite WaterboxHost in unmanaged code?
* Right now, this fixes some nondeterminism and potential OS interaction+crash issues in bsnes when a cothread tries to allocate memory.  Memory allocation means syscalls which means in the old waterbox transition to managed, while on a costack.  No good.  Now, those calls stay native, so we do have host thread running inside the guest costack, but it's not as bad.
* In the near term, I want to try some stack marshalling to get all guest execution onto saveable stacks, and I really need fully unmanaged code to do this.
* In the long term, I want to have more options to implement true threading in the box.
* I can easily integrate needed host native functionality and desirable host native functionality without having lots of different parts like winguard and linguard.
* I'd like to experiment with ABI changes that would be difficult with a managed host; using SysV inside the box might be a nice speedup.

Why rewrite WaterboxHost in rust?
* C and C++ are garbage languages and I can barely stand them.
* This was an opportunity for me to learn something new.

What does this mean for the maintainability of Waterbox?
* As far as I know, from the beginning of time to now we have had zero contributions (namespace shuffling and reformatting excluded) to the Waterbox host code by anyone but me.  So, no worse off.


nattthebear:
@adelikat @zeromus Particularly interested in your review of the long term effects of this on the BizHawk ecosystem.  Don't want to drop languages willy-nilly.

zeromus:
"C and C++ are garbage language" I do not really accept as a rationale. Yes, they are, but that doesn't mean we don't have to use them sometimes. 
All the other rationales are fine.
I am not really happy with this, but I am at peace with it, so long as:
1. you help me go through the exercise of building the rust parts in WSL and I can write docs on it.
2. you affirm that the rust parts could be written in c++ later if needed without it being a complete nightmare (suppose the person doing this work doesn't hate c++ as much as you so that having to use c++ isn't the monster of the nightmare). For instance if it depends on a rust part or language feature that nobody could ever rewrite in c++, that would be bad. But I doubt that's the case.

nattthebear:
I've added some documentation that includes basic build instructions.  I'm only building the Linux version of the host in WSL; I'm building Windows host in Windows.

InfamousKnight:
What is crusty? Some surprise core for us?

YoshiRulz:
@InfamousKnight Read the PR title. In terms of functionality:
> &lt;natt> It's funny
> &lt;natt> If you want to make an end user release note out of that PR, it's just "fix a rare crash in the bsnes core"
> &lt;natt> That's all it does. Oh well

nattthebear:
> you affirm that the rust parts could be written in c++ later if needed without it being a complete nightmare (suppose the person doing this work doesn't hate c++ as much as you so that having to use c++ isn't the monster of the nightmare). For instance if it depends on a rust part or language feature that nobody could ever rewrite in c++, that would be bad. But I doubt that's the case.

No, nothing like that is going on here.  The essential requirement is the ability to efficiently call low level OS functions and to operate on a non-OS thread stack without asploding.  Future improvements will use assembly for context switches, so there needs to be an ability to get that in, which of course C++ has.

PR #2197, opened 2020-07-04
Merge redundant copy-pasted movie loading code.

RetroEdit:
Split off from my ill-constructed #2193, which had its purpose muddied by the differently-motivated changes presented here.

Before I implemented my direct goal of altering emuVersion, I found a significant redundancy between Bk2Movie and TASMovie's Load() that seemed necessary to correct before adding new behavior. Along with significantly reducing the amount of code to maintain (removing almost 100 lines), the first change also makes Bk2 and TASMovie behavior more consistent:

- Duplicate keys in the header now use the first value given ([previously Bk2Movie-specific](https://github.com/TASVideos/BizHawk/commit/dd28c386de642e7858d910b7f53e9144f2bc2cbb)). Incidentally, testing exposed a race condition between rerecords and rendering related to this, which could have caused movie's rerecord count to be overwritten by 0 during the loading process.
- Subtitles are now sorted when loaded ([previously Bk2Movie-specific](https://github.com/TASVideos/BizHawk/commit/35a07f42eb2f023306e022f03a8a9cf9932735ed))
- Movies missing Comments, Subtitles, or SyncSettings will no longer be aborted during loading (previously Bk2Movie-specific [1](https://github.com/TASVideos/BizHawk/commit/21ae27dd9486653982f911715d1dd01a146eb861), [2](https://github.com/TASVideos/BizHawk/commit/5a408f9321fe99da518bb19e3452cfb763812aea))
- Savestate-anchored tasprojs will now load the framebuffer ([previously Bk2Movie-specific](https://github.com/TASVideos/BizHawk/commit/daf74eb91be9832e9588ea1b3b186d621346cd64)); this may be unnecessary for tasprojs, but it won't hurt anything because it's ignored if it's missing.

Deleting redundant code did not mean TASMovie's differences weren't taken into consideration. The main functionality of Load() had some differences, so these were split into separate overridable methods ClearBeforeLoad() and LoadFields(), with these methods overriden to call both the original code and Tasproj-specific code.

### DIFF
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
@@ -32,7 +32,8 @@ namespace BizHawk.Client.Common
 			{
 				if (!Header.ContainsKey(HeaderKeys.Rerecords))
 				{
-					Header[HeaderKeys.Rerecords] = "0";
+					// Modifying the header itself can cause a race condition between loading a movie and rendering the rerecord count, causing a movie's rerecord count to be overwritten with 0 during loading.
+					return 0;
 				}
 
 				return ulong.Parse(Header[HeaderKeys.Rerecords]);
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.IO.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.IO.cs
@@ -42,7 +42,121 @@ namespace BizHawk.Client.Common
 			}
 
 			ClearBeforeLoad();
+			LoadFields(bl, preload);
 
+			Changes = false;
+			return true;
+		}
+
+		public bool PreLoadHeaderAndLength(HawkFile hawkFile)
+		{
+			var file = new FileInfo(Filename);
+			if (!file.Exists)
+			{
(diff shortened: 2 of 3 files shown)


alyosha-tas:
The unused 'preload' parameter is supposed to be used to scan movie files in the 'play movies' dialog. there should be something like:

`if (preload) { return true; }`

after the `bl.GetLump(BinaryStateLump.Movieheader, true, delegate(TextReader tr)` segment. 

I noticed just the other day that this dialog is loading very slowly due to this, because it is loading the entire movie. I was actually going to commit something today but noticed your PR.

Can you add in this line and test on your end? Probably good to test before merging. Otherwise looks good to me.

RetroEdit:
@alyosha-tas There's one main barrier to your proposal: preload is supposed to get the movie length, but it needs to read the entire movie file to get that info, because the frame count isn't saved to the header currently. So another field would have to be added to the header to store the frame count, and corresponding code to keep it up-to-date.

If we move forward on that, I'm also just thinking we might consider scrapping the `preload` parameter entirely and just moving the needed functionality to PreLoadHeaderAndLength(), since that simplifies the method definition for Load, and Load does a lot more than preload should need to do.

alyosha-tas:
Oh, you're right, well that's a separate issue then I guess.

PR #2204, opened 2020-07-06
Firmware Rename

hegyak:
Rename All the Firmware Files to have a more Unified Naming Convention

### DIFF
--- a/src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs
+++ b/src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs
@@ -48,22 +48,22 @@ namespace BizHawk.Emulation.Common
 			Option("GBA", "Bios", gbaNormal);
 			Option("GBA", "Bios", gbaJDebug);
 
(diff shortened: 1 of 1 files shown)


RetroEdit:
Changing the bios names may be annoying for people who store firmware for many separate BizHawk versions in one folder. Doing this would mean they would have to start from scratch on getting the bios names correct for 2.5 and beyond. Even people who just copy over their bioses to new versions still would have to contend with renaming.

There could be a renaming tool within BizHawk, but that wouldn't really work for people who still want to keep older BizHawks around to use them (e.g. judges); the downsides of renaming probably outweigh the upsides.

An alternative to outright renaming them and throwing away backwards-compatibility is to retain aliases with the previous names are also searched for. This may be the best compromise between disrupting existing users and also making sure the bios names are comprehensible.

hegyak:
@RetroEdit This only impacts/changes what BizHawk's Organize Function would rename function would rename the Firmware files to.

BizHawk uses the SHA-1 for the Files to determine what is available and what to use.  This change has no impact on Older versions of BizHawk's Firmware as well.  As the SHA-1 is not dependent on the File's Name.

This is an aesthetic change for BizHawk not a functional change.

RetroEdit:
I see. In that case, I don't think anything here is objectionable.

PR #2206, opened 2020-07-06
Retype IMovie.TimeLength to TimeSpan; cleanup frame rate code

RetroEdit:


src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
```diff
@@ -216,6 +216,11 @@ protected string CommentsString()
 			return sb.ToString();
 		}
 
+		public bool IsPal
```
adelikat, line 219: better is public bool IsPal => , rather than a get with a =>

PR #2209, opened 2020-07-07
Waterbox:  Stack Marshalling

nattthebear:
Waterbox guest code now runs on a stack inside the guest memory space.  This removes some potential opportunities for nondeterminism and makes future porting of libco-enabled cores easier.


waterbox/waterboxhost/src/context/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+			gs = Box::into_raw(Box::new([0usize; 4]));
+			arch_prctl(ARCH_SET_GS, gs);
+		}
+	}
+}
+
+/// Get the currently loaded ActivatedWaterboxHost
+/// If called outside call_guest, will fail in various wonderful ways?
+/// Unsafe:  The lifetime is really only good in the current function
+pub unsafe fn access_context() -> &'static mut ActivatedWaterboxHost<'static> {
```
nattthebear, line 94: Can't `guest_syscall` just pass this as an 8th parameter?

waterbox/waterboxhost/src/context/thunks.rs
```diff
@@ -0,0 +1,81 @@
...
+
+// each generated thunk should look like this:
+// 49 ba .. .. .. .. .. .. .. ..             mov r10, <host_context_ptr>
+// 49 bb .. .. .. .. .. .. .. ..             mov r11, <guest_entry_point>
+// 48 b8 .. .. .. .. .. .. .. ..             mov rax, <call_guest_impl>
+// ff e0                                     jmp rax
+
+const THUNK_SIZE: usize = 32;
+
+// every time we remount, the thunks need to be updated with the new r10 value
```
nattthebear, line 14: Outdated comment:  the context ptr doesn't move

waterbox/waterboxhost/src/context/thunks.rs
```diff
@@ -0,0 +1,81 @@
...
+				bin::writeval::<u8>(dest, 0xff)?;
+				bin::writeval::<u8>(dest, 0xe0)?;
+
+				self.lookup.insert(guest_entry_point, p);
+				Ok(p)
+			}
+		}
+	}
+	/// updates context value for all previously created thunks
+	pub fn update_context_ptr(&mut self, context: *mut Context) -> anyhow::Result<()> {
```
nattthebear, line 67: Since the context ptr doesn't move, should this just be eliminated?

waterbox/waterboxhost/src/host.rs
```diff
@@ -51,24 +64,17 @@ impl WaterboxHost {
...
-			tag: TAG,
 			h,
 			b,
-			sys
 		});
-		res.sys.syscall.ud = res.as_mut() as *mut ActivatedWaterboxHost as usize;
-		res.h.elf.connect_syscalls(&mut res.b, &res.sys);
 		res.h.active = true;
+		res.h.context.host_ptr = res.as_mut() as *mut ActivatedWaterboxHost as usize;
+		// not needed as it hasn't moved
```
nattthebear, line 76: Correct, but maybe we should just axe this whole comment and idea

waterbox/waterboxhost/src/lib.rs
```diff
@@ -2,12 +2,14 @@
 
 #![feature(try_trait)]
 #![feature(core_intrinsics)]
+#![feature(asm)]
+#![feature(naked_functions)]
```
nattthebear, line 6: Not actually using this feature anywhere

src/BizHawk.Emulation.Cores/Waterbox/NymaCore.cs
```diff
@@ -181,7 +182,7 @@ public abstract unsafe partial class NymaCore : WaterboxCore
 					// if (deterministic)
 					// 	throw new InvalidOperationException("Internal error: Core set a frame thread proc in deterministic mode");
 					Console.WriteLine($"Setting up waterbox thread for {_frameThreadPtr}");
-					_frameThreadStart = CallingConventionAdapters.Waterbox.GetDelegateForFunctionPointer<Action>(_frameThreadPtr);
+					_frameThreadStart = CallingConventionAdapters.WaterboxDepartureOnly.GetDelegateForFunctionPointer<Action>(_frameThreadPtr);
```
nattthebear, line 185: This is actually very sneaky.  The way `_frameThreadPtr` was passed back, it doesn't get wrapped by `thunks.rs` and so it runs on a host stack and any attempt for it to call one of the registered callbacks, or make a syscall, will crash.  This is fine because it's just a consequence of our fake not-at-all-real "threading" support.  Maybe add a comment...

src/BizHawk.Emulation.Cores/Waterbox/WaterboxMemoryDomain.cs
```diff
@@ -140,7 +140,8 @@ public StubResolver(IntPtr p)
 
 		public static MemoryDomainAccessStub Create(IntPtr p, IMonitor monitor)
 		{
-			return BizInvoker.GetInvoker<MemoryDomainAccessStub>(new StubResolver(p), monitor, CallingConventionAdapters.Waterbox);
+			return BizInvoker.GetInvoker<MemoryDomainAccessStub>(
```
nattthebear, line 143: This has the same issue as nyma frame procs, but shouldn't be that way.  `WaterboxDepartureOnly` is probably a super broken loaded footgun, revisit

Issue #2023, opened 2020-05-19
Should the 'Power On' marker be a special-case?


RetroEdit:
I question the usefulness of having a special-case 'Power On' marker that can't be deleted or moved.

If it's always on frame 0 as the standard game start, then it's kind of redundant. I also can't recall any specific instance where it has any usefulness whatsoever; generally it's just something that takes up space and I ignore. One potential advantage I can imagine is clicking on the Power On marker in the sidebar list to seek to frame 0, but you can always scroll back up to frame 0, or retain the Power On marker while still having the option to delete it; it doesn't need to be indelible.

Specifically, the current implementation for the 'Power On' marker has several active bugs, and seems to present a future maintenance burden disproportionate to the potential usefulness of retaining its special-case status:

- You can rename the marker by creating a new marker at frame 0
- If you delete frames from the beginning until another marker and have Bind Markers to Input on, two markers will occupy the same frame. This only occurs for the 'Power On' marker because other markers would have been deleted by merit of deleting the frames they occupy. (This occurred both before and after my changes to the delete frames logic).
- If you insert frames at frame 0, the Power On marker will be moved. This was introduced by my changes to the insert frames logic.

I believe the current solution is to haphazardly check for frame 0 in certain situations where markers could affect Power On, but this is not very robust and requires updates every time a new method for modifying markers is created. A better solution might be to entirelyremove the 'Power On' marker from the standard markers list at all, but just refer to it in the appropriate places (when displaying the sidebar list and the ). However, this would amount to 1-indexing our markers list somewhere in the code, which also seems inideal.

In particular, the first bug I presented in my above list is an interesting conceptual problem: what is the ideal resolution? Are we going to silently delete a marker when it overlaps with the Power On marker? Pop up a dialog box warning the user one of their markers could be deleted frames at frame 0? Get rid of marker binding entirely to avoid such a situation (I jest)? Sure, this is perhaps a niche case in practice, but I believe it's an illustrative concrete demonstration of why having a special-case marker complicates all of the marker handling code.

RetroEdit:
This has more depth than I initially realized. I think the maker was partially intended to denote a frame that couldn't be seeked to by nature of being the initial frame (i.e. a frame such that seeking to it means that nothing has happened to the movie). That itself has some bugs, including some related to this particular marker, but offers a bit more insight into the intended purpose of the marker, at least in my interpretation. I still don't necessarily agree with the necessity of having such a marker, though.

[...]

PR #2212, opened 2020-07-07
TAStudio: allow users to delete marker 0

RetroEdit:
In relation to #2023.

~~I would like to look over this a bit and clean it up before considering merging; this is old code I threw together while I was originally examining the issue, but I forgot about it. The changes also touch a bit more than simply removing the marker's immutability. There were some questionable things in seeking that I still want to look at. Hopefully that doesn't dilute this issue too much.~~ All cleaned up.

~~TODO: Are savestates a problem, specifically for savestate-anchored movies? I would guess no, but it's worth checking.~~ These aren't an issue; they're handled separately.

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/TasMovieMarker.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/TasMovieMarker.cs
@@ -201,7 +201,7 @@ namespace BizHawk.Client.Common
 
 		public new void Remove(TasMovieMarker item)
 		{
-			if (item == null || item.Frame == 0) // TODO: Don't do this.
+			if (item == null) // TODO: Don't do this.
 			{
 				return;
 			}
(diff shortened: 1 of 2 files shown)


RetroEdit:
@adelikat You raised the concern that users would not be able to tell the initial state if they happened to delete the marker?

Firstly, I would say, this isn't *that* important when a user can simply verify the starting state for their movie by playing from the beginning.

My second thought was that perhaps the info could appear in the TAStudio's `Metadata > Header...` window as a readonly field.

Finally, I find it odd that I can't seem to find a `Movie info` anywhere. This could be shared between traditional `.bk2`s and TAStudio projects, and display basically every piece of info about the movie: the game name, along with the rest of the header fields, buttons to open further menus.

InfamousKnight:
O

adelikat:
Find a way to put the start type (power-on, sram, savestate) visibly in tastudio, and I'll approve and merge this PR.  Because it could show sram, it is actually a better ui than what the marker is trying to do.

Now that zwinder is merged, the marker is doing nothing for navigation since frame 0 is guaranteed to exist

adelikat:
Author isn't going to finish this, and this isn't the main work.  The real problem is how we would indicate the movie start type since marker 0 is the only indicator of this. Closing this PR

PR #2220, opened 2020-07-10
Hex editor: disable 0-length domains (b/c mGBA SaveRAM)

RetroEdit:


src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBA/MGBAHawk.IMemoryDomains.cs
```diff
@@ -31,7 +31,7 @@ private void CreateMemoryDomains(int romsize)
 			mm.Add(_vram = new MemoryDomainIntPtr("VRAM", le, IntPtr.Zero, 96 * 1024, true, 4));
 			mm.Add(_oam = new MemoryDomainIntPtr("OAM", le, IntPtr.Zero, 1024, true, 4));
 			mm.Add(_rom = new MemoryDomainIntPtr("ROM", le, IntPtr.Zero, romsize, true, 4));
-			mm.Add(_sram = new MemoryDomainIntPtr("SRAM", le, IntPtr.Zero, 0, true, 4)); // size will be fixed in wireup
+			mm.Add(_sram = new MemoryDomainIntPtr("SaveRAM", le, IntPtr.Zero, 0, true, 4)); // size will be fixed in wireup
```
nattthebear, line 34: Reason for this change?

RetroEdit: 'SRAM' as short for savedata seems misleading to me, especially since GBA uses separate save types, one of which is SRAM (static RAM in that context, of course).

zeromus: SRAM means save ram and static ram both. That ship has sailed. To make matters worse, WRAM is usually DRAM which may or may not be SRAM depending on whether it's battery backed, and may be called SRAM in any event. The words have no clear meaning so I don't care which one is used, but if there's a reason not to change it then it's better than a reason for changing it, unless it's to make it match other cores more closely in case this is an odd man out

adelikat: Changing it will break pre-existing .wch files, and maybe some other files I can't think of

src/BizHawk.Client.EmuHawk/Extensions/ToolExtensions.cs
```diff
@@ -242,7 +242,10 @@ public static IEnumerable<ToolStripItem> MenuItems(this IMemoryDomains domains,
 				var item = new ToolStripMenuItem
 				{
 					Text = name,
-					Enabled = !(maxSize.HasValue && domain.Size > maxSize.Value),
+					// 0-length domains cause several problems, so they're disabled.
```
adelikat, line 245: We have many cores with conditional domains that only show in certain (undocumented) situations. 

I think better is to not add the memory domain at all, at the core level if it is 0 bytes.  

If for some reason that isn't feasible, they should just be filtered when grabbing a list of domains from the core.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBA/MGBAHawk.IMemoryDomains.cs
nattthebear: I'm pretty sure that it was my (possibly poor) decision to use "SRAM" here when originally brining in our first GBA core, and then I carried through the names identically for back compat since then.

RetroEdit: I don't think backwards compatibility with `.wch` files is a huge concern, since recent changes have very likely broken backwards compatibility for a few versions; feos's change that exposed the implementation detail that mGBA's internal  SaveRAM domain is not always defined (and zeromus's fix that really just worked around this fact).

adelikat:
After some discussion we all concluded that the best way to handle the GBA SaveRam domain is the following:
1) create a domain that is a wrapper around the save data that is always 128kb
2) always return 0 for values outside of the actual savedata array

This is preferred over this approach.  As such, I am closing this PR

PR #2221, opened 2020-07-10
Yoshi's UI experiment

YoshiRulz:
POC that could potentially form an abstraction over WinForms, which we could later implement using another UI framework without needing to rewrite EmuHawk. It doesn't work with the Designer, but we'd need to ditch that when switching UI frameworks anyway.

Pick yer poison:
"procedural in ctor" (as opposed to a method in the Designer file):
https://github.com/TASVideos/BizHawk/blob/229c31e4192a22cc2f9159804a75fbfadbfcab8a/src/BizHawk.Client.EmuHawk/config/AutofireConfig.cs#L32-L55
"procedural + builders"
https://github.com/TASVideos/BizHawk/blob/879fdc8f432d484cececd8c0cbeef68dc2c294c3/src/BizHawk.Client.EmuHawk/config/AutofireConfig.cs#L36-L62
"DSL"
https://github.com/TASVideos/BizHawk/blob/ce8498da976e80cad1157d8784a0e56e35d20c2c/src/BizHawk.Client.EmuHawk/config/AutofireConfig.cs#L47-L72


tommai78101:
> I'd like feedback on which coding style people are most comfortable with for UI design.

From an outsider's point of view, I prefer code styles that "pack" information together without too many indentations, that are related to one another of an UI object, and being presentable/readable to the developers who are modifying its properties "on a whim" to suit the needs that are "in demand".

That means, I would be more inclined to go with the "imperative + builders" style a bit more.

PR #2223, opened 2020-07-11
Fix issue with settings adapter

nattthebear:
This would fail to work on any core that used the serviceprovider to provide a settings definition that wasn't on the iemulator itself?  Guess we never did that

### DIFF
--- a/src/BizHawk.Emulation.Common/Interfaces/Services/ISettable.cs
+++ b/src/BizHawk.Emulation.Common/Interfaces/Services/ISettable.cs
@@ -69,20 +69,18 @@ namespace BizHawk.Emulation.Common
 	/// </summary>
 	public class SettingsAdapter
 	{
-		public SettingsAdapter(IEmulator e)
+		public SettingsAdapter(IEmulator emulator)
 		{
-			_emu = e;
-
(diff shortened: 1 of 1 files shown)


src/BizHawk.Emulation.Common/Interfaces/Services/ISettable.cs
```diff
@@ -69,46 +69,46 @@ public enum PutSettingsDirtyBits
...
 	{
-		public SettingsAdapter(IEmulator e)
+		public SettingsAdapter(IEmulator emulator)
 		{
-			_emu = e;
-
-			Type impl = e.GetType().GetInterfaces()
-				.FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition() == typeof(ISettable<,>));
-			if (impl == null)
+			var settableType = emulator.ServiceProvider.AvailableServices
```
nattthebear, line 74: @adelikat Does this look right?  It seems like such an obvious bug that I can't believe it's been this way for so long.

adelikat: Pretty sure it is an obvious bug, gone unnoticed since it was always the core class implementing the interface

Issue #2144, opened 2020-06-16
Painting input on TAStudio's tiles should disable recording mode


RetroEdit:
When the user paints an input, it is almost never desirable for that input to be immediately overwritten by recording mode input when the emulator is unpaused. One theoretical exception is accidentally painted input, but even then, the user will potentially overwrite more input than they intend to because the accidental painting will invalide the greenzone afterwards and force recording to continue from that point. I think having to re-check one checkbox for a destructive function (Recording mode) is a vastly more desirable behavior than the many mishaps that can occur when trying to paint input while recording mode is on.

This is exacerbated by #2143 making it difficult to undo the input that recording mode overwrites.

adelikat:
@vadosnaprimer, is this a good idea?

vadosnaprimer:
Rec mode doesn't erase any inputs during position restore (auto or manual). If you want to freely edit and then freely unpause without losing input, you don't want the recording mode at all. If you switch to editing, then untick the checkbox. If you want to be switching back and forth, use the Toggle Read-Only hotkey. Everything is properly set up, just decide what you want to be doing :D

RetroEdit:
Yes, "using the tools the proper way" is all nice and fine in theory, but imo it's easy to accidentally leave recording mode on when one doesn't want to.

I just think when one painting inputs, it's desirable to turn recording mode off; I can't see why one would want painted inputs to be immediately overwritten by recording mode.

I could potentially see an argument being made that this pollutes the purpose of the recording mode checkbox by turning it off without the user's explicit direct action, but I see this more as a safety catch than an overriding of the user's will. I've seen others indirectly mention this issue before in terms of painting input but not realizing recording mode was still on and having input overwritten (I could dig this up if you're curious).

For my own case, I would say it's probably true that I can eventually develop the habits of using recording mode "the proper way", but my use is intermittent enough it's difficult to have any guarantee of consistency (though I think it will probably be a while before I make the same mistake again). I don't have a huge amount of personal stake if placing the burden on the user to use the tools correctly is the decision here, but I do think having this suggestion implemented would benefit users generally with minimal downsides.

vadosnaprimer:
@adelikat you've tried real rec mode, what do you feel regarding the above? I'd probably be fine with having an option, but again it seems to only mean rec mode wasn't the real intent if one is now back to full-on editing.

adelikat:
[...]

PR #2228, opened 2020-07-12
Make tastudio automatically switch to read-only mode when input painting

adelikat:
The approach is simple, remove all the code that was try really hard to not make this happen automatically. This is to address the feature request in #2144

I'd like some feedback here. What situations am I not considering?


adelikat:
Well, that's unfortunate, that's a lot of nuance that is required.

dancrimpirate:
great !
sorry for bump 
could you add the feature to change the color so instead of black pixel become another color 
also : this one is hard add the feature to load a picture from the files instead of real picture

PR #2237, opened 2020-07-19
Simple interface between Lua and .Net ClientWebSocket class

davidackroyd99:
This class exposes some Lua functions that allow you to easily connect to Websocket servers. It's incredibly basic at the moment but I have a Lua script which 'proves that it works'. If you build my fork and then run the following script, you will receive the message 'hello world' back from the websocket.org test echoing service:

```
local ws = bizsocket.open('wss://echo.websocket.org')
bizsocket.send(ws, 'hello world', true)
local msg = bizsocket.receive(ws, 1024)
print(msg)
bizsocket.close(ws, 1000, 'close')
```

I understand that this is not a core feature that people are dying for, but I wanted to mess around with it and see what I could come up with for a personal project. I also understand that core contributors will want me to change a lot of this around and I'm happy to make any modifications that people suggest, especially around error handling and so on.

### DIFF
--- a/src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using NLua;
+
+
+namespace BizHawk.Client.Common.lua.LuaHelperLibs
+{
+	[Description("A library exposing standard .NET string methods")]
+	public sealed class WebSocketLuaLibrary : LuaLibraryBase
+	{
+		private Dictionary<string, ClientWebSocket> activeSockets;
+
+		public WebSocketLuaLibrary(Lua lua) : base(lua)
+		{
+			activeSockets = new Dictionary<string, ClientWebSocket>();
+		}
+
+		public WebSocketLuaLibrary(Lua lua, Action<string> logOutputCallback) : base(lua, logOutputCallback)
+		{
+			activeSockets = new Dictionary<string, ClientWebSocket>();
+		}
+
+		public override string Name => "bizsocket";
+
+		[LuaMethodExample("local ws = bizsocket.open('wss://echo.websocket.org');")]
+		[LuaMethod("open", "Opens a websocket and returns the id so that it can be retrieved later.")]
+		public string OpenSocket(string url)
+		{
+			var ws = new ClientWebSocket();
+			var id = new Guid().ToString();
+
+			ws.ConnectAsync(new Uri(url), CancellationToken.None).Wait();
+
+			activeSockets[id] = ws;
+
+			return id.ToString();
+		}
+
+		[LuaMethodExample("local ws = bizsocket.send(ws_id, 'some message', true);")]
+		[LuaMethod("send", "Send a message to a certain websocket id (boolean flag endOfMessage)")]
+		public void SendToSocket(string id, string content, bool endOfMessage)
+		{
+			var ws = activeSockets[id];
+			var msg = new ArraySegment<byte>(Encoding.ASCII.GetBytes(content));
+
+			ws.SendAsync(msg, WebSocketMessageType.Text, endOfMessage, CancellationToken.None);
+		}
+
+		[LuaMethodExample("local ws = bizsocket.receive(ws_id, max_read);")]
+		[LuaMethod("receive", "Receive a message from a certain websocket id and a maximum number of bytes to read.")]
+		public string ReceiveFromSocket(string id, int maxRead)
+		{
+			var ws = activeSockets[id];
+
+			var rcvBytes = new byte[maxRead];
+			var rcvBuffer = new ArraySegment<byte>(rcvBytes);
+
+			var result = ws.ReceiveAsync(rcvBuffer, CancellationToken.None).Result;
+			byte[] msgBytes = rcvBuffer.Take(result.Count).ToArray();
+			string rcvMsg = Encoding.UTF8.GetString(msgBytes);
+
+			return rcvMsg;
+		}
+
+		[LuaMethodExample("local ws_status = bizsocket.getstatus(ws_id);")]
+		[LuaMethod("getStatus", "Get a websocket's status")]
+		public int GetSocketStatus(string id)
+		{
+			var ws = activeSockets[id];
+
+			return (int)ws.State;
+		}
+
+		[LuaMethodExample("local ws_status = bizsocket.close(ws_id, close_status);")]
+		[LuaMethod("close", "Close a websocket connection with a close status, defined in section 11.7 of the web sockets protocol spec.")]
+		public void CloseSocket(string id, int status, string closeMessage)
+		{
+			var ws = activeSockets[id];
+
+			ws.CloseAsync((WebSocketCloseStatus) status, closeMessage, CancellationToken.None);
+		}
+	}
+}
--- a/src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
@@ -38,7 +38,7 @@ namespace BizHawk.Client.Common.lua.LuaHelperLibs
 
 			activeSockets[id] = ws;
 
-			return id.ToString();
+			return id;
 		}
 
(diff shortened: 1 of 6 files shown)


YoshiRulz:
Fork to PR in 20 minutes? Nice.

I'll look into merging this as soon as I get around to refactoring comms in ApiHawk. Looks like this part of the BCL is [on NuGet](https://www.nuget.org/packages/System.Net.WebSockets.Client/) for .NET Core.

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
```diff
@@ -0,0 +1,88 @@
...
+		public string OpenSocket(string url)
+		{
+			var ws = new ClientWebSocket();
+			var id = new Guid().ToString();
+
+			ws.ConnectAsync(new Uri(url), CancellationToken.None).Wait();
+
+			activeSockets[id] = ws;
+
+			return id.ToString();
```
nattthebear, line 41: `id` is already a string

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		[LuaMethodExample("local ws = bizsocket.send(ws_id, 'some message', true);")]
+		[LuaMethod("send", "Send a message to a certain websocket id (boolean flag endOfMessage)")]
+		public void SendToSocket(string id, string content, bool endOfMessage)
+		{
+			var ws = activeSockets[id];
+			var msg = new ArraySegment<byte>(Encoding.ASCII.GetBytes(content));
```
nattthebear, line 49: ASCII on send but UTF8 on receive?

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
```diff
@@ -0,0 +1,88 @@
...
+		[LuaMethod("receive", "Receive a message from a certain websocket id and a maximum number of bytes to read.")]
+		public string ReceiveFromSocket(string id, int maxRead)
+		{
+			var ws = activeSockets[id];
+
+			var rcvBytes = new byte[maxRead];
+			var rcvBuffer = new ArraySegment<byte>(rcvBytes);
+
+			var result = ws.ReceiveAsync(rcvBuffer, CancellationToken.None).Result;
+			byte[] msgBytes = rcvBuffer.Take(result.Count).ToArray();
```
nattthebear, line 64: Don't make another byte buffer here; there are various overloads of Encoding.GetString that can be instructed to not use the entire buffer

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
```diff
@@ -0,0 +1,88 @@
...
+			return (int)ws.State;
+		}
+
+		[LuaMethodExample("local ws_status = bizsocket.close(ws_id, close_status);")]
+		[LuaMethod("close", "Close a websocket connection with a close status, defined in section 11.7 of the web sockets protocol spec.")]
+		public void CloseSocket(string id, int status, string closeMessage)
+		{
+			var ws = activeSockets[id];
+
+			ws.CloseAsync((WebSocketCloseStatus) status, closeMessage, CancellationToken.None);
```
nattthebear, line 85: Do we need to remove this from the dictionary now?

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
```diff
@@ -0,0 +1,88 @@
...
+		}
```
nattthebear, line 86: I'm pretty ignorant on our lua system, so forgive the questions, but what is our general story for not leaking stuff like this?

My normal expectation for an interop system like would be that you pass an opaque handle to the guest language, so you can be notified when it GCs and free the host resource.  A cursory examination of some documentation suggests that we should be passing the "userdata" type to lua and then using what lua calls "finalizers" to accomplish this.

Every high level GCed language I can think of (even C#) provides guarantees that foreign resources will be eventually released if you throw exceptions / forget objects.

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
davidackroyd99: suggestion noted, will rectify in another commit

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
davidackroyd99: You're right, I think it should be UTF-8 all round, will check docs and amend

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
davidackroyd99: suggestion noted, will rectify in another commit

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
davidackroyd99: No because you can still do get status to see if it's properly closed. If you're going to remove from the dictionary then you should also wait on the close task.

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
davidackroyd99: I've never done interop before I literally just looked at what else was already in there and tried to copy the established patterns, which seemed to me to be returning simple value types.

I will do some research and make some amendments.

Thanks for all your comments btw, I've never contributed to open source before and you have been insanely helpful. Makes me want to do more! :)

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
davidackroyd99: image

yup it should be utf-8

src/BizHawk.Client.Common/lua/LuaHelperLibs/WebSocketLuaLibrary.cs
nattthebear: The established patterns probably aren't great.  If existing code isn't using lua userdata+finalizers, I bet it's a hard problem and not worth worrying about for something like this.

YoshiRulz:
I've done the comms refactor as promised, go ahead and rebase. You should be able to add this to the .NET API too as I've simplified it quite a bit.

davidackroyd99:
> You should be able to add this to the .NET API too as I've simplified it quite a bit.

To CommsApi? and should I refactor so that the comms api points to methods that return the .NET objects instead of the guids that the Lua methods use?

YoshiRulz:
Depends on how safe they are. From your previous comments, `ClientWebSocket` needs to be kept around when closed, so you should create a wrapper class/struct for it. For .NET, you need to copy all the method signatures from `WebSocketLuaLibrary` into another new class, and hook that up like the existing three. So something like `((ICommApi) _comms).WebSocketServer.Open("wss://echo.websocket.org").Status`, where `Status` is defined in the wrapper. For Lua, use the code you have now but put it in the existing `CommLuaLibrary`, and prefix the function names with `websocket_`. The implementations for Lua can delegate like everything else in `CommLuaLibrary`, but you should continue using a `Guid` or simply `GetHashCode()` in the API surface. If you go with `GetHashCode`, I think the dict can become a `HashSet`?

davidackroyd99:
Will continue to work on this and let you know when it can be merged. Will mark as draft for now, thanks for the pointers.

YoshiRulz:
I need this in `master` for a... secret project that I didn't know I was starting. If you can finish this ASAP, or just rebase+push what you have, that'd be much appreciated.

davidackroyd99:
> I need this in `master` for... a ~secret~ project that I didn't know I was starting. If you can finish this ASAP, or just rebase+push what you have, that'd be much appreciated.

I have got it into a building state, put the methods in CommApi but commented them out as I couldn't add them to the main form without everything exploding. If there's anything you want me to do, let me know.

YoshiRulz:
See 166c50b6c, hopefully everything's still working as expected.

PR #2242, opened 2020-07-19
clone x times

davidackroyd99:
This works pretty nicely.

Notes:

- Might want to consider renaming FramesPrompt to something more generic like IntPrompt as I'm using it to get a number for something else
- Haven't done the keyboard shortcut because I don't know how, might do it tomorrow night if I get chance

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/FramesPrompt.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/FramesPrompt.Designer.cs
@@ -61,7 +61,7 @@
 			// 
 			this.label1.Location = new System.Drawing.Point(12, 20);
 			this.label1.Name = "label1";
-			this.label1.Text = "Insert number of frames:";
+			this.label1.Text = bodyMessage;
 			// 
 			// NumFramesBox
 			// 
@@ -91,7 +91,7 @@
 			this.Name = "FramesPrompt";
 			this.ShowIcon = false;
 			this.StartPosition = System.Windows.Forms.FormStartPosition.CenterParent;
-			this.Text = "Insert Frames";
+			this.Text = headMessage;
 			this.Load += new System.EventHandler(this.FramesPrompt_Load);
 			this.ResumeLayout(false);
 			this.PerformLayout();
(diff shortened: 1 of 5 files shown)


YoshiRulz:
```
git checkout feature/clone-x-times
git branch bak
git rebase -i master # assuming master is tracking TASVideos/BizHawk:master and you've not messed with it
# (delete the irrelevant commits)
git push -f origin feature/clone-x-times # assuming origin is your fork
```

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -500,6 +500,7 @@ public void AddColumn(string columnName, string columnText, int columnWidth, Col
 		public void InsertNumFramesExternal() => InsertNumFramesMenuItem_Click(null, null);
 		public void DeleteFramesExternal() => DeleteFramesMenuItem_Click(null, null);
 		public void CloneFramesExternal() => CloneFramesMenuItem_Click(null, null);
+		public void CloneFramesXTimesExternal() => CloneFramesXTimesMenuItem_Click(null, null);
```
adelikat, line 503: This isn't necessary, it's only if we need to run this functionality elsewhere, such as lua

vadosnaprimer: Isn't it used for hotkeys?

adelikat: that too, but none of that is wired up in this PR, so I don't think this line should be added, until it is needed

src/BizHawk.Client.EmuHawk/tools/TAStudio/FramesPrompt.Designer.cs
```diff
@@ -91,7 +91,7 @@ private void InitializeComponent()
 			this.Name = "FramesPrompt";
 			this.ShowIcon = false;
 			this.StartPosition = System.Windows.Forms.FormStartPosition.CenterParent;
-			this.Text = "Insert Frames";
+			this.Text = headMessage;
```
adelikat, line 94: Does this work in the designer?

YoshiRulz: It won't, this assignment needs to be the empty string and another assignment can be done in the ctor after `InitializeComponent`.

davidackroyd99: I've resolved this, thanks for the input

PR #2248, opened 2020-07-21
Experiment to speed up libco and hence libsnes

nattthebear:
The compiler now can fully inline the co_switch, and with most registers being specified as clobbers and not saved explicitly, the compiler can choose to save only what it needs to (we don't have to defensively save everything).

Practically speaking, the co_switch calls are usually inlined, but the functions they're in don't seem to be that big and don't make direct use of r12..r15 too much anyway, so (push r12..r15, switch, pop r12..r15) is a common emit.  But I see a miniscule FPS increase.


nattthebear:
For general technical review, tagging @zeromus .

I'm also interested in general perf regression the libsnes core if anyone wants to do that.  My numbers show slightly (~1%?) faster.

If there's a measurable speedup for others, I'm willing to merge this, but otherwise, no.

nattthebear:
Technical review in IRC was well received, so moving forward with this

PR #2256, opened 2020-07-28
Move some more stuff to Client.Common

YoshiRulz:
natt wanted to know why this code should be moved, and why EmuHawk needed to be common-ised when it's our only frontend. This PR's motivations are to have less code targeting Framework (see also #1801), and to reduce the amount of non-UI code in the EmuHawk project. It's also uncertain if we'll be able to keep one executable that runs both on Windows and on Linux via Mono; though should EmuHawk split, most files can be included in both Core- and Framework-targeting projects without any modification or preprocessor guarding.

### DIFF
--- a/src/BizHawk.Client.Common/DisplayManager/DisplaySurface.cs
+++ b/src/BizHawk.Client.Common/DisplayManager/DisplaySurface.cs
@@ -4,7 +4,7 @@ using System.Drawing.Imaging;
 using System.Runtime.InteropServices;
 using BizHawk.Common;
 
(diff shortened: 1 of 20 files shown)


YoshiRulz:
rebased, included `IMainFormForApi` in the misc. commit

YoshiRulz:
I now also think that `ArgParser`, the `MainForm` interfaces, and some of the other miscellany need special attention. If I removed the third commit (and rebased), would this be acceptable?

adelikat (approved):

Issue #2181, opened 2020-06-30
[Regression] TAStudio doesn't restore cursor to the frame it was saved


RetroEdit:
A user commented on Discord that the latest version of BizHawk doesn't properly restore the cursor to the frame it was saved. Instead, the cursor always starts on frame 0.

From my own testing, TAStudio started loading at frame 0 in 2.4.1. In 2.4, it loads at the frame the cursor was at when saving.

adelikat:
Duplicate of #2063

RetroEdit:
Whoops, my bad. Forgot to search before filing for that one.

PR #2267, opened 2020-08-03
TAStudio cleanup

SuuperW:
I cleaned up some movie loading code in TAStudio.
I'd like other users to test for any issues these changes might cause.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
@@ -552,7 +552,8 @@ namespace BizHawk.Client.EmuHawk
 				return false;
 
 			_engaged = true;
-			Settings.RecentTas.Add(newMovie.Filename); // only add if it did load
+			if (!newMovie.Filename.Equals(DefaultTasProjName()))
+				Settings.RecentTas.Add(newMovie.Filename); // only add if it did load
 
 			if (gotoFrame > -1)
 				GoToFrame(gotoFrame);
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
@@ -529,7 +529,7 @@ namespace BizHawk.Client.EmuHawk
 			MainForm.SetMainformMovieInfo();
 		}
 
-		private bool LoadFile(FileInfo file, int gotoFrame = 0)
+		private bool LoadFile(FileInfo file, int gotoFrame = -1)
 		{
 			if (!file.Exists)
 			{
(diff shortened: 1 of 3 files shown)


adelikat:
need to readdress this after the zwinderstatemanager branch merge

Did you test not saving default.tasproj?  It was doing that for a reason, otherwise how do you do a core reboot when they pick new project?

SuuperW:
You're right, not saving the file is causing problems. I could have sworn I tested that, though...

adelikat:
I committed the same change before too, swore I tested too, and got so many bugs on the tracker :D

SuuperW:
From new commit's code comment:
		We want to create a new file for the movie, unless the original movie was the default movie.
		Cluttering up the user's movie folder with default 1.tasproj, default 2.tasproj, etc. isn't nice.
		We also want to promt the user for a file name if we haven't saved a file here, so we use the default name.
		Now, playing a movie loads the movie from the file; which would mean playing the new movie actually loads the default movie.
		However if the movie's file doesn't exist, then it simply skips that loading process!
		But I find it strange that it doesn't fail to play the movie without a file, so maybe this isn't an ideal solution?

I do not know if this skipping of the load process for non-existent files is meant to actually work. If it is, then a better solution is probably to make that explicit by including this code at the top of [Bk2Movie.IO.cs -> Load](https://github.com/TASVideos/BizHawk/blob/a4e3dbb8855868062f93b49c3415d177a9cd924a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.IO.cs#L30):
`if (!File.Exists(Filename)) return true;`
The we could give the movie a filename of null after conversion, which looks like it would (along with a change to auto-save code to check for null filename) would also have the desired effect of prompting the user for a file name upon saving.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -529,39 +529,32 @@ private void ConvertCurrentMovieToTasproj()
...
-				GoToFrame(0);
-			}
-			else if (gotoFrame > 0)
-			{
-				GoToFrame(gotoFrame);
-			}
-			else
-			{
-				GoToFrame(CurrentTasMovie.TasSession.CurrentFrame);
-			}
```
RetroEdit, line 564: Would removing this prevent the tasproj from loading and returning the current frame cursor to the old position? (see also: #2181)

(Haven't tested it.)

SuuperW: Probably. I couldn't think of any possible use case where that behavior would be desired, and saw no code or comments that appeared to utilize or explain it.

Why would you ever want to re-load the current movie, and want it to automatically seek to the previously loaded frame?

(I certainly can add this back in, it just seemed like a messy bit of code with no explanation whatever so I cleaned it.)

RetroEdit: I meant when you opened the movie for the first time: the cursor is supposed to remember what frame the movie was on the previous time the movie was saved.

I do think the way it was previously implemented looked a bit messy and could be improved.

SuuperW: Oh, that makes sense. I will revert that change.

YoshiRulz:
New state manager was merged and bugs squished, is any of this still relevant?

SuuperW:
No, I don't think any of this is still needed.

RetroEdit:
Then this PR should be closed. I'm not keen on releasing code that was written right before a major refactor of a related component without

PR #2282, opened 2020-08-10
I don't understand why we're clearing the rewind buffer...

nattthebear:
... but if we want to do that, just do it, and don't deallocate + reallocate many megs of ram on every loadstate

### DIFF
--- a/src/BizHawk.Client.Common/rewind/IRewinder.cs
+++ b/src/BizHawk.Client.Common/rewind/IRewinder.cs
@@ -19,5 +19,7 @@ namespace BizHawk.Client.Common
 
 		void Suspend();
 		void Resume();
+
+		void Clear();
 	}
 }
--- a/src/BizHawk.Client.Common/rewind/Zwinder.cs
+++ b/src/BizHawk.Client.Common/rewind/Zwinder.cs
@@ -85,5 +85,10 @@ namespace BizHawk.Client.Common
 		{
 			// this possess no resources to dispose of, but other IRewinder impls might
 		}
+
+		public void Clear()
+		{
+			_buffer.InvalidateEnd(0);
+		}
 	}
 }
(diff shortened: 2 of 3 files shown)


nattthebear:
I don't use movies and stuff enough to test this

adelikat:
I can't tell you why we clear there.  As you know the rewinder is linear, in fact it will rewind through load states.
Because of this, I used to disable the rewinder while movie recording since it will very easily and quickly destroy sync.  Some people wanted it (TAStudio was bad back then), so the compromise is that on loadstate we clear the rewinder to protect the user from major desyncs.

nattthebear:
I don't follow.  If I take the following set of actions, you're saying my movie will desync?!?!?!?!

1. Start on frame 200.
2. Advance to frame 201.
3. Load state on frame 100.
4. Load state on frame 200.

Bizhawk sounds completely broken and utterly unusable for any situation at all !?!??!?!

adelikat:
No, That scenario will work.  That feature is called "bullet proof savestates".

The problem is if you combine our ring buffer rewinder into this process:
Advance from 1 to 100
Make a Savestate
Advance to frame 200
Loadstate to go back to frame 100
Advance 5 frames
Rewind 6 frames

You will now be on frame 200, with input for 101 frames

nattthebear:
Not so bulletproof, evidently.  Sorry if it wasn't clear, but I was imagining my scenario from two different perspectives.  How about this slightly modified version of the same text:

1. Start on frame 200, with a previously created state on frame 100 available.
2. Advance to frame 201.
3. The rewinder captures a frame 201 savestate
4. Load state on frame 100.
5. Command the rewinder to go back one, which will cause it to load the 201 savestate.

Same exact sequence of actions, but now it doesn't work.

adelikat:
Correct, that will not work.  Our bullet proof savestates can not work when working with the rewinder this way.
It's because to do this, we must have the old input log for your 201 savestate.  Rewinder states have no input log

nattthebear:
OK, I understand now.  Thanks for explaining.  So long as this PR passes tests, then I think we should merge it because it will increase the performance of loading states while movies are going.

PR #2290, opened 2020-08-12
Unify setting of window titles, and add static title option

### LINKED ISSUE #1996: OBS Studio BizHawk Detection
### Summary
OBS Studio Can pick up on the TAStudio window, but you have to choose the window again each time you open TAStudio.
Somewhat related to #1529 but specific to TAStudio

### Repro
Open TAStudio to a specific TAS.  In OBS Studio create a new window capture, choose TAStudio, capture method automatic, any Window Match Priority method.  Close EmuHawk and reopen to the same tastudio project, the tastudio window will switch to the emulator.  EmuHawk will stick fine if window match priority is set to "find window of same executable" which covers #1529 but TAStudio has the same executable name as emuhawk.exe so OBS can't consistently find it.

### Host env.
- BizHawk 2.4.1; Win10 Pro 1909; AMD

YoshiRulz:
From commit desc.:
> The only forms that did anything special with their window title were Basic Bot, CDL, Hex Editor, Log Window, MainForm, and TAStudio. That behaviour is in `WindowTitle`, while the rest use `WindowTitleStatic`.
> The implementations of `WindowTitleStatic` in those six forms are new. There's a checkbox in `Config` > `Display...` > `Misc` to disable their special behaviour and use the new static titles.

fixes #1996

### DIFF
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -313,5 +313,7 @@ namespace BizHawk.Client.Common
 		public string LastWrittenFromDetailed { get; set; } = VersionInfo.GetEmuVersion();
 
 		public EHostInputMethod HostInputMethod { get; set; } = OSTailoredCode.IsUnixHost ? EHostInputMethod.OpenTK : EHostInputMethod.DirectInput;
+
+		public bool UseStaticWindowTitles { get; set; }
 	}
 }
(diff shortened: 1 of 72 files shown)


nattthebear:
70 files of diffs and more new in house winforms base classes to address an issue that's already closed?  I'm inclined to say no, but as I don't work on the spaghetti mess that is our winforms system anymore, someone else can make that decision.

adelikat:
After the 2.5 release, I'd like to discuss this approach, and learn more about specific OBS problems, and consider merging or improving this approach

adelikat:
This is kind of gross, but it also unifies some things with Config handling too.  I think the feature is important enough to justify going this far out of our way to support OBS

YoshiRulz:
Rebased. Another immediately realisable use for this is to run all the window parenting through it (`ToolFormBase` does some of this already). I was recently trying to iterate all child windows of `MainForm` and the best I got was `Application.OpenForms.Cast<Form>()`.

adelikat (approved):

PR #2291, opened 2020-08-12
zwinder: Make GetStateClosestToFrame actually get closest state.

RetroEdit:


src/BizHawk.Client.Common/movie/tasproj/IStateManager.cs
```diff
@@ -33,7 +33,7 @@ public interface IStateManager
 		void Clear();
 
 		/// <summary>
-		/// Get a nearby state.  The returned frame must be less (but not equal to???) the passed frame.
+		/// Get a nearby state.  The returned frame must be less then or equal to the passed frame.
```
nattthebear, line 36: "than"

nattthebear:
LGTM.  In light of https://github.com/TASVideos/BizHawk/issues/2288, @adelikat would be the other one to chime in (because he thinks this rewinder can be salvaged)

RetroEdit:
I think this change is beneficial whether we end up using zwinder or not. Ideally, `GetStateClosestToFrame` should perform like it suggests and not have weird edge cases of using frame 1 to actually retrieve the savestate closest to frame 0. Also, moving all the functionality for getting earlier frames for TAStudio's screen buffer in `GetPriorStateForFramebuffer()` makes the intentions of the code clearer (as opposed to having unexplained code in four separate places).

PR #2299, opened 2020-08-13
Remove default joypad bindings for fast-forward/rewind

### LINKED ISSUE #2297: Controller buttons are mapped to fast-forward and rewind by default
Joypad buttons are mapped to fast-forward and rewind by default.

https://github.com/TASVideos/BizHawk/blob/ad07eb85742c6ef89c74642f1363369758d611fa/src/BizHawk.Client.Common/config/Binding.cs#L108-L110

I might be the only one who feels this way, but this annoys me incessantly, and deleting them from the config is quite annoying every time. I understand the idea (probably shoulder buttons for navigating forward and backwards), but there's a few problems here:

1. Not all controllers map these buttons the same. On my controller, buttons `J1 B7` and `J1 B8` are Start and Select.
2. These buttons could conflict with actual console's controller mappings.

In my opinion, joypad buttons should not be bound to non-controller functionalities by default. The user can easily add these if they're wanted. A controller's primary purpose should be restricted to actual in-game controls.

For the Left Trigger and Right Trigger, I care slightly less because it doesn't personally affect me at the moment, but I think the same philosophy applies, if a bit less so because they aren't arbitrary numbered buttons at least.

This annoys me so much I went to the trouble of digging up the commit that added these in, way back in `1.5.1-approximate`: ac25f3f3e62abff9c22e2213f70611eda9aee92e

@adelikat J'Accuse!

(Just kidding.)

RetroEdit:
fixes #2297

### DIFF
--- a/src/BizHawk.Client.Common/config/Binding.cs
+++ b/src/BizHawk.Client.Common/config/Binding.cs
@@ -105,9 +105,9 @@ namespace BizHawk.Client.Common
 					_defaultValues = new List<Binding>
 					{
 						Bind("General", "Frame Advance", "F"),
-						Bind("General", "Rewind", "Shift+R, J1 B7, X1 LeftTrigger"),
+						Bind("General", "Rewind", "Shift+R"),
 						Bind("General", "Pause", "Pause"),
-						Bind("General", "Fast Forward", "Tab, J1 B8, X1 RightTrigger"),
+						Bind("General", "Fast Forward", "Tab"),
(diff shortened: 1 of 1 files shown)


adelikat:
In the simple days when bizhawk only had some 8 and 16 bit systems to worry about, these defaults were so great.  You had no other use for those buttons, and having rewind/fast forward out of the box was helpful.

Days aren't so simple anymore, so I guess these defaults haven't aged well

PR #2317, opened 2020-08-19
Fix GBC-AGB bios issues

### LINKED ISSUE #2100: Gambatte: cgb in gba doesn't do anything?
Expected to see indication on the title screen of Shantae, saw nothing.

This is a regression, but I wouldn't prioritize it very high; don't think any games do anything interesting with this mode.

nattthebear:
All six combinations (bios, no bios) X (dmg, cgb, cgb in agb) should work now

Closes #2100


TiKevin83:
This is exactly what I had in mind!  Does the hash check for the AGB BIOS matcb the result of gambatte's edit or the more recent information?
https://github.com/ISSOtm/gb-bootroms/commit/dd0809d60ec167a70c49d690bee7533bc43d34df

nattthebear:
No, this was a like for like change.  The data exactly matches what Gambatte uses, so is neither a regression nor an improvement.

Looking at that repository, seems like there are a lot of additional bootroms in there that we don't have listed.  Are they correct and are we just out of date?

TiKevin83:
yeah those are all up to date, attached some explanations of them
[GB Bootrom Explanations.txt](https://github.com/TASVideos/BizHawk/files/5099655/GB.Bootrom.Explanations.txt)

RetroEdit:
As I understand, this uses the hash for gambatte's patched GBC in GBA instead of the recently acquired official hash for GBC in GBA. We should probably be able to detec that as well, since it's the real boot rom.

nattthebear:
Replaced all firmware infos with those in ISSOtm's repo

TiKevin83:
That's a great improvement, tests syncing with the AGB-CGB BIOS from ISSOtm's repo.

There's still a back compatibility issue with the EnableBIOS sync setting, having to edit in EnableBIOS=true for many published GBC movies.

PR #2321, opened 2020-08-23
Majorly refactor ZwinderStatemanager to address # 2287

adelikat:
A lot of fixes and refactors here.
1) Rename ancient to reserved and make it a dictionary
2) roll frameZeroState into the reserved section
3) roll bookmark states into reserved
4) add a reserved callback that the client code can pass in, that will check this and put things in reserved rather than evict them,
client code will check if marker or branch state
4) refactor assumptions about reserved always being states before current or recent
5) add a gapFiller, that will do the job highPriority was badly doing, and do it better
6) add a state cache that keeps track of all frames that have a state, use this in HasState, for a huge performance boost, compared to both the previous zwinder code and the old state manager, both had a O(n) problem, where performance slowed down greatly as the size of the movie increased
7) add unit tests to cover a variety of scenarios

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/IStateManager.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/IStateManager.cs
@@ -23,12 +23,6 @@ namespace BizHawk.Client.Common
 		void Capture(int frame, IStatable source, bool force = false);
 
 		/// <summary>
-		/// Guarantee's a state will be captured and "reserved" for future use
-		/// Reserved states do not get evicted from memory when decay logic is applied
-		/// </summary>
-		void CaptureReserved(int frame, IStatable source);
-
-		/// <summary>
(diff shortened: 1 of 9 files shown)


adelikat:
Added in the hacky logic of reserving the state before, at least the code is is a lot simpler and better documented this time around

alyosha-tas:
Tested and seems to be working well. Thanks!

adelikat:
Thanks everyone for helping work out some kinks.  This is by no means perfect (not state manager is), and there's some I intend to do, but I think this is in a much better state and fixes major issues, so I'm going to go ahead and merge

YoshiRulz:
> Can we get the bot to build PRs again? Testing things like this by more people is exactly why it's needed.

GitLab builds every branch. It stops pulling changes if a branch is force-pushed, though.

PR #2331, opened 2020-08-24
Add warning for loading config.ini from an older or newer version

YoshiRulz:
edit: Force-pushed with bugfixes and extra cleverness for 2.4.1 and 2.4.2, which identify themselves as 2.4.0.

edit 2: If you want to test this, set `DeveloperBuild` to `false` in `VersionInfo`, and edit the strings near the bottom of `config.ini` (you can also use `Config` > `Load Config From...`). If you want to see all the messages, set `MainVersion` to `2.5.1`. Then { `` (delete the keys), `2.3.3`, `2.4.0`, `2.5`, `2.5.2` } will trigger each of the 5 cases.

### DIFF
--- a/src/BizHawk.Client.Common/config/ConfigService.cs
+++ b/src/BizHawk.Client.Common/config/ConfigService.cs
@@ -1,7 +1,11 @@
 ﻿using System;
 using System.IO;
 using System.Reflection;
+
+using BizHawk.Common;
+
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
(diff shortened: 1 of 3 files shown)


adelikat:
So what's the plan here? to add more logic for every release that breaks config??

YoshiRulz:
If we keep `VersionInfo.MainVersion` up-to-date, this code will detect each version as older/newer correctly. If two releases have the same value, each will think the other is the same as itself and no warning will be displayed. In that case, the other warning for corrupted config will display as always.

PR #2333, opened 2020-08-24
GBA: allow light sensor to reach 255 and default it to 0

### LINKED ISSUE #2323: No input method allows for full range of solar sensor input
It appears that no input method allows for the full solar sensor input range of 0 to 255.
- Analog input allows a range of ~80 to 200 (100-200 with default settings)
- TAStudio input roll allows for 0 to 200
- The auto-hold tool in TAStudio allows for 0-100

Manually modifying a .bk2 or .tasproj file, a value of 255 is accepted and handled correctly.
So far i have tested that in Boktai 1. I'll update this issue, with screenshots and how it affects Boktai 2 and 3, soon.

mGBA Core in BizHawk 2.4.2 on Windows 10 (v2004) with a Xbox 360 Controller.

RetroEdit:
Rename AxisSpec.Mid to AxisSpec.RestingValue to better reflect its purpose

The goal here is to fix #2323.

There's still a problem where TAStudio recording mode (and general gameplay, probably?) defaults to 128 for the light sensor value instead of 0. This is due to `Controller.NormalizeAxes` setting it to the midpoint between `min` and `max`.

### DIFF
--- a/src/BizHawk.Client.Common/Controller.cs
+++ b/src/BizHawk.Client.Common/Controller.cs
@@ -14,7 +14,7 @@ namespace BizHawk.Client.Common
 			Definition = definition;
 			foreach (var kvp in Definition.Axes)
 			{
-				_axes[kvp.Key] = kvp.Value.Mid;
+				_axes[kvp.Key] = kvp.Value.RestingValue;
 				_axisRanges[kvp.Key] = kvp.Value;
 			}
 		}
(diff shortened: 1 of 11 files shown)


RetroEdit:
As an alternative solution to Light Sensor issues, zeromus has suggested constructing the Light Sensor as `range (-255,255)` with mid of `0`. That may be a better approach than the one pursued in this branch.

YoshiRulz:
Obsoleted by #2334 (will probably rename `Mid` later).

PR #2335, opened 2020-08-25
Make movie hash check failure spawn MessageBox

### LINKED ISSUE #2328: Give a warning pop-up when opening a .bk2 with a different ROM
When you open a .bk2 movie, BizHawk compares the ROM hash with the one stored in the Header.txt of the .bk2 file. If they don't match, a hud warning message appears for a moment in the lower left: "Warning: Movie hash does not match the ROM". The issue is that this message is very easy to miss, due to its short duration and the presence of other messages at the same time. My request is to have a warning pop-up window for this specific event, so that it can't possibly be missed by the user.

RetroEdit:
Fixes #2328.

Also annotate some Hash-related code that needs to be looked at.

### DIFF
--- a/src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
+++ b/src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
@@ -90,6 +90,11 @@ namespace BizHawk.Client.Common
 		bool StartsFromSaveRam { get; set; }
 		string GameName { get; set; }
 		string SystemID { get; set; }
+		// TODO: This should be further defined in some manner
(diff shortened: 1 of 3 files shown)


adelikat:
This would pop-up 100% of the time for things like multi-disc movies made with the .xml bundler.  That seems to make the user situation WORSE in those cases.

Also, I'm not sure if the hashes are even reliable for disc based games

adelikat:
Rejecting this PR.  Our hashing isn't nearly reliable enough to cause such a disruptive notification.  Especially, since all .xml bundles will get this pop-up, everytime you load a movie.  Rejecting this.

PR #2339, opened 2020-08-26
TAStudio: Fix deletion crashing when deleting invalid indexes

RetroEdit:
The problem occurs when trying to delete a selection where some of the frames were valid, but the last displayed line (which is currently an extra non-input frame) was in the selection.

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
@@ -114,7 +114,8 @@ namespace BizHawk.Client.Common
 			{
 				// Separate the given frames into contiguous blocks
 				// and process each block independently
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
```diff
@@ -114,7 +114,8 @@ public void RemoveFrames(ICollection<int> frames)
 			{
 				// Separate the given frames into contiguous blocks
 				// and process each block independently
-				List<int> framesToDelete = frames.OrderBy(fr => fr).ToList();
+				List<int> framesToDelete = frames.Where(fr => fr >= 0 && fr < InputLogLength)
```
adelikat, line 117: super nitpick but can you put the LINQ each on separate lines?
framesToDelete = frames
    .Where()
    .OrderBy()
   .ToList()

PR #2342, opened 2020-08-28
Technical changes for dropping VB.NET dep on Unix

YoshiRulz:
I'd like get the dep removed for 2.5. If someone has a better solution to this (other than having separate releases and using conditional compilation) I'd be all for it.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/AVOut/IVideoWriter.cs
+++ b/src/BizHawk.Client.EmuHawk/AVOut/IVideoWriter.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+
+using BizHawk.Common;
 using BizHawk.Emulation.Common;
 
 namespace BizHawk.Client.EmuHawk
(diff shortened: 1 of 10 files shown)


adelikat:
If a C# only solution was found, would you still want this code merged?

YoshiRulz:
(continuing convo from IRC)
> &lt;YoshiRulz> Is that directed at me? I don't care about the single-instance feature at all
> &lt;YoshiRulz> If you can keep it without depending on that assembly, I'm all for it
> &lt;adelikat_> yes, you
> &lt;adelikat_> not do you want the single instance code, would you still want all this assembly checking code
> &lt;YoshiRulz> No, it's only there to workaround the "assembly not found" exception
> &lt;adelikat_> if you don't want the code and are just working around a solvable prolbem
> &lt;adelikat_> then let's just solve the problem

I'm not sure there's any other way forward. We either have the dependency on both OSes, and the feature works on both; or we have some way of making the VB.NET assembly only load on Windows (this PR is one solution), and the feature only works there; or we reimplement the feature without VB.NET (it probably uses Windows-only APIs, which Mono hacked together a replacement for). Giving Linux users/packagers an extra annoying dependency for what is IMHO an anti-feature isn't acceptable. I'm guessing that ditching the feature on both OSes isn't acceptable either? Of those 5 paths, keeping the status quo on Windows and using a quick hack on Linux seemed the most reasonable to me.

zeromus:
I'll take care of this. Shitcan the dep and anything designed to accommodate it. Unhook the feature and I'll hook it back up asap with pure c# code

adelikat:
The feature is useful.  In conjunction with associating rom file extensions with BizHawk.  You can open a folder and double click to switch bizhawk to that game.

I don't use it much, but I have before.  And it is there because someone requested it for that purpose, a long time ago.

Issue #2235, opened 2020-07-19
Telemetry


YoshiRulz:
Reminded about this by natt and feos joking about crash reports in IRC. I'd briefly discussed it with adelikat earlier and came across [[1]](https://www.meziantou.net/use-application-insights-in-a-desktop-application.htm), which looks promising technically but might not be suitable functionally.

Here are all the unanswered questions of telemetry in EmuHawk:
* Should crash reports be collected, or hardware/OS specs, or app usage, or everything?
	* The first is generally useful. The other two may be helpful with diagnosing certain problems (#1820), and with choosing to drop features, respectively. --yoshi
* How should users opt-in?
	* We don't currently do much (visibly) on first run. I don't think a telemetry opt-in dialog would be much of a burden, and they've become ubiquitous. --yoshi
* Where will data be sent?
	* `tasvideos.org` seems the obvious choice, though it ~~doesn't have HTTPS for some reason~~ has issues with HTTPS. Using [[1]](https://www.meziantou.net/use-application-insights-in-a-desktop-application.htm) would have Microsoft do all the work, including collection. --yoshi
* Are we *in practice* able to anonymise telemetry data e.g. file paths?
	* leaving this here: Stacktraces are localised in at least one configuration (net48 on Win10), so collecting the "language" datapoint may be unavoidable. IMO we should be collecting system language(s) anyway. natt at some point mentioned disabling localised stacktraces. --yoshi
* How much of the dataset will be public?
* Should users be able to *de*-anonymise themselves?
	* (with a TASVideos forum account, for example)
* Will EmuHawk 2.6 (i.e. the first version w/ telemetry) be able to call home in 5 years' time? Should it try?

lonks
* \[1] [unofficial blog post *Use Application Insights in a desktop application* from 2017-03-29](https://www.meziantou.net/use-application-insights-in-a-desktop-application.htm)

PR #2359, opened 2020-08-31
Add Discord Rich Presence as an option

ijre:
This PR adds Discord Rich Presence to the MainForm as an opt-in setting inside the Config -> Customize settings, displaying the current game name and console  
  
Shown below is how it looks:
Discord_ 07, 10, le 31 sur Monday

### DIFF
--- a/src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj
+++ b/src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj
@@ -85,6 +85,7 @@
 	</ItemGroup>
 -->
 	<ItemGroup>
+		<PackageReference Include="DiscordRichPresence" Version="1.0.150" />
 		<PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="all" />
 		<PackageReference Include="Microsoft.VisualBasic" Version="10.3.0" />
 		<PackageReference Include="System.Net.Http" Version="4.3.4" />
--- a/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
@@ -17,2407 +17,2413 @@ namespace BizHawk.Client.EmuHawk
 		/// </summary>
 		private void InitializeComponent()
 		{
-			this.components = new System.ComponentModel.Container();
-			this.MainformMenu = new BizHawk.WinForms.Controls.MenuStripEx();
-			this.FileSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.OpenRomMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.RecentRomSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.toolStripSeparator3 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
-			this.OpenAdvancedMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.CloseRomMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.toolStripMenuItem1 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
-			this.SaveStateSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState1MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState2MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState3MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState4MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState5MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState6MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState7MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState8MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState9MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SaveState0MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.toolStripSeparator6 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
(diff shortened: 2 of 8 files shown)


YoshiRulz:
Clean your diff please, you don't need +2k/-2k LOC of whitespace changes in Designer files. If you recreate the changes on a new branch it'd fix that plus there wouldn't be merge commits. Force push with `git push -f origin new-branch:master` (origin being your fork) to overwrite this PR.

ijre:
Ah, my bad; I'll fix that yeah

nattthebear:
What if anything does this do on linux builds?

adelikat:
if someone can fix the merge conflicts, I will merge this feature

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -49,6 +51,8 @@ public partial class MainForm : Form, IMainFormForApi, IMainFormForConfig, IMain
 			(new[] { "PCE", "PCECD", "SGX" }, new[] { CoreNames.HyperNyma, CoreNames.PceHawk, CoreNames.TurboNyma })
 		};
 
+		private static DiscordRpcClient Client = new DiscordRpcClient("749686424248385716");
```
YoshiRulz, line 54: Is this a token generated for EmuHawk? Did you need to register as a developer/partner/whatever to get one?

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -4461,6 +4476,36 @@ private bool Rewind(ref bool runFrame, long currentTimestamp, out bool returnToR
...
+					Assets = new Assets { LargeImageKey = "corphawk" },
+					Details = Emulator.IsNull() ? "No ROM loaded"
+					:
+					$"{Client.CurrentUser.Username} is playing {Game.Name} for the {Emulator.System().DisplayName}"
+				});
+		}
+
+		private void SetupPresence()
+		{
+			if (Process.GetProcessesByName("Discord").Length == 0) // check to make sure discord is actually open
```
YoshiRulz, line 4493: ew

src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
```diff
@@ -17,2407 +17,2407 @@ partial class MainForm
 		/// </summary>
 		private void InitializeComponent()
 		{
-			this.components = new System.ComponentModel.Container();
```
YoshiRulz, line 20: I still want don't want this checked in. If `git diff -w filename` is blank, don't stage the file.

src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj
```diff
@@ -85,6 +85,7 @@
 	</ItemGroup>
 -->
 	<ItemGroup>
+		<PackageReference Include="DiscordRichPresence" Version="1.0.150" />
```
YoshiRulz, line 88: This is a third-party wrapper over the first-party C/C++ lib. Not necessarily a bad thing, I'm just personally wary of adding more technical debt. Also, it targets .NET Standard 2.0.

---

Not sure where to leave this comment, but it'd be nice if this feature was shared between front-ends. Sadly we don't have a good abstraction for that yet, so if this was all moved to `Client.Common`, there wouldn't be an obvious place for it.

src/BizHawk.Client.EmuHawk/config/EmuHawkOptions.cs
```diff
@@ -69,6 +69,7 @@ private void GuiOptions_Load(object sender, EventArgs e)
 			RunInBackgroundCheckbox.Checked = _config.RunInBackground;
 			AcceptBackgroundInputCheckbox.Checked = _config.AcceptBackgroundInput;
 			AcceptBackgroundInputControllerOnlyCheckBox.Checked = _config.AcceptBackgroundInputControllerOnly;
+			DiscordRPC.Checked = _config.DiscordRPC;
```
YoshiRulz, line 72: I think I mentioned this in either IRC or Discord when you were there discussing this feature: there's no precedent for this in the codebase (see #2235). I don't think we should hide this setting in `Config` > `Customize...` with the miscellany. It needs its own `TabPage` at the very least, and probably deserves its own dialog with some warning text in bold.

YoshiRulz (changes requested):
Yo @ijre if you're still AWOL come Monday I'm adding presence myself and closing this PR

YoshiRulz:
I said I'd do this myself, but when I went to find docs I was greeted with:
> The SDK that this documentation references, Discord-RPC, has been deprecated in favor of our new Discord GameSDK.

So this needs to be implemented from scratch. At least there's a first-party library for Unity we can use. When adding Discord integration, we should consider how downstream projects such as multiplayer randomisers could use ApiHawk for networking and be automatically integrated with Discord's "party" UX. All things considered, I think those features can wait until NextHawk.

src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj
CasualPokePlayer: It isn't, this library re-implements the Discord IPC in pure managed C#. It is completely third party in that regard, and not deprecated unlike the C++ library.

I've personally used it (well, [a fork](https://github.com/CasualPokePlayer/GSE/tree/ee07f2b/externals/DiscordRPC)) to implement Discord Rich Presence in my own emulator frontend. It works well enough, the actual IPC protocol is quite simple, other than Discord having the genius decision of using JSON for IPC messages.

Its JSON usage also comes with a minor wrinkle: this library uses Newtonsoft.Json, which of course has issues for trimming/NativeAOT in the future, along with known footguns (but that probably doesn't matter too much). My fork specifically swaps it out with System.Text.Json, which isn't that hard to do, but would be trickier to add to upstream due to some of their wants (compat with ancient .NET versions / some ancient Unity crap which needs Newtonsoft.Json).

CasualPokePlayer:
> The SDK that this documentation references, Discord-RPC, has been deprecated in favor of our new Discord GameSDK.

Discord-RPC was the old official implementation of the Discord IPC protocol (which mind you, the DiscordRichPresence library did not use). This "Discord GameSDK" is another official implementation, having "more features" which also ended up mostly becoming deprecated, so in practice it's not that different of a same feature set compared to the older implementation.

The old Discord-RPC implementation was made in C++ with a C API, and was MIT licensed with full source code available, so emulators generally used it. The Discord GameSDK actually just has functionality locked behind native libraries (probably C++ again), without any available source code and no FOSS license, just API headers. As such, pretty much all emulators have kept to the old Discord-RPC implementation, as many of them cannot use the newer Discord GameSDK as it's just not GPL compatible. Likewise, I don't think we should be using the Discord GameSDK at all. Using the DiscordRichPresence library seems better considering its wide usage and not deprecated status and not needing to pull in native libraries.

src/BizHawk.Client.EmuHawk/MainForm.cs
CasualPokePlayer: The client ID is something associated with an "app" on Discord, which can be a bot, have webhooks, do OAuth2 verification, and is used as a front for Discord IPC usage. Any Discord account can become a "developer" for this "app" (although if we want this we'd probably want to create a "team" for BizHawk developers).

PR #2372, opened 2020-09-03
Fix new bugs from painting during auto-restore

### LINKED ISSUE #2371: In 2.5, painting while a position is being auto-restored will terminate auto-restore
- Open TAStudio for any game
- Check `Auto-restore last position`
- Seek to frame 30 or roundabout
- Set the emulation speed to 12% and uncheck `Turbo seek` (these steps are not strictly necessary, because it isn't inherent to the bug, it just makes testing more convenient)
- Paint frame 0, auto-seek will begin and go towards frame 30. Unpaint it before the cursor reaches frame 30, and then it will resume playback again, but it won't stop at a specific frame (this is because of my pause fix for #2324)

However, even if my fix is undone, painting will still just seek to the frame it happened to stop at mid-restore if painting occurs mid-restore. This did not occur in 2.4.2.

I think this is probably related to how painting initiates a seek operation in 2.5, which would explain why it's clobbering the last cursor position when it didn't previously. There's probably some nuance here, though, as it only happens when mid-restore, so there's hopefully a relatively painless way to avoid overriding the last cursor position without building in more complex infrastructure (which would probably just introduce more bugs).

RetroEdit:
fixes #2371

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
@@ -14,7 +14,7 @@ namespace BizHawk.Client.EmuHawk
 				if (frame <= Emulator.Frame)
 				{
 					if ((MainForm.EmulatorPaused || !MainForm.IsSeeking)
-						&& !CurrentTasMovie.LastPositionStable)
+						&& !CurrentTasMovie.LastPositionStable && !_playbackInterrupted)
 					{
 						LastPositionFrame = Emulator.Frame;
 						CurrentTasMovie.LastPositionStable = true; // until new frame is emulated
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
@@ -1018,6 +1018,13 @@ namespace BizHawk.Client.EmuHawk
 
 		private void DoTriggeredAutoRestoreIfNeeded()
 		{
+			// Disable the seek that could have been initiated when painting.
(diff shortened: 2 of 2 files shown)


src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
```diff
@@ -14,7 +14,7 @@ private void GoToLastEmulatedFrameIfNecessary(int frame)
 				if (frame <= Emulator.Frame)
 				{
 					if ((MainForm.EmulatorPaused || !MainForm.IsSeeking)
-						&& !CurrentTasMovie.LastPositionStable)
+						&& !CurrentTasMovie.LastPositionStable && !_playbackInterrupted)
```
vadosnaprimer, line 17: This fixes the main bug.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -1018,18 +1018,24 @@ private void SetSplicer()
 
 		private void DoTriggeredAutoRestoreIfNeeded()
 		{
+			// Disable the seek that could have been initiated when painting.
+			// This must done before DoAutoRestore, otherwise it would disable the auto-restore seek.
+			if (_playbackInterrupted)
+			{
+				MainForm.PauseOnFrame = null;
```
vadosnaprimer, line 1025: Desired behavitor (the way taseditor does it):
Disable autorestore, pause, edit, unpause to start seeking: then no matter how much you edit _while it's still seeking_, it will pause on the seek frame.

With this edit, it won't pause there in this situation, but I'm not sure when this started or if it ever worked like in taseditor.

adelikat (approved):

PR #2390, opened 2020-09-12
Re-add M3U capability

nattthebear:
Only supports optical disks.


nattthebear:
If someone familiar with it has time to test this, that would be great

YoshiRulz:
alternative implementation pushed to `master` as 8da552581

PR #2391, opened 2020-09-13
Update mGBA to fix a savestate bug.

### LINKED ISSUE #2355: mGBA - important savestate deserialization fix
~~This commit is important:~~

~~https://github.com/mgba-emu/mgba/commit/27b31024cee96991735069a72d5d709dcb9ebb9e~~

Currently, this means savestate loading in mGBA will imperfectly restore the exact state that was saved, which could cause desyncs. We'll want to cherry-pick or otherwise include this commit before the next release.

EDIT: The linked commit doesn't specifically change any code, but there's probably an issue in upstream mGBA that endrift is looking into.

RetroEdit:
fixes #2355


endrift (approved):

RetroEdit:
Still needs @YoshiRulz to build for Linux, but this is pretty cut-and-dry and should be part of 2.5.1. No reason to wait on pushing easy fixes to known savestate bugs, since those could be extremely annoying for users.

(force pushed because something went wrong with the first DLL build, but this one worked fine when I tested it)

YoshiRulz:
The output binary seems to be the same as the checked-in one? I also noticed the build script outputs to `output/dll` instead of to `Assets/dll` (or to both).

RetroEdit:
YoshiRulz is right; this update to the submodule doesn't actually affect BizHawk, because it's Game Boy-specific code, as confirmed by endrift. Therefore, this PR can be closed.

YoshiRulz:
Are we gonna force push `bizhawk-0.9` in the fork to remove the cherry-picked commit or nah?

endrift:
I would recommend doing so to make merging easier in the future.

PR #2414, opened 2020-09-19
Added Pop_OS support for library paths

ethanarns:
Pop_OS is a Linux distro based on Ubuntu. I added it to the Unix launch shell script in order to remove an error message I received.

### DIFF
--- a/Assets/EmuHawkMono.sh
+++ b/Assets/EmuHawkMono.sh
@@ -9,7 +9,7 @@ winepath=""
 if [ "$(command -v lsb_release)" ]; then
 	case "$(lsb_release -i | cut -c17- | tr -d "\n")" in
 		"Arch"|"ManjaroLinux") libpath="/usr/lib";;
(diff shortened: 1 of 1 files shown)


YoshiRulz:
Thank you! Can you check if the workaround is necessary?
```sh
...
case ... in
	"Arch"|"ManjaroLinux") libpath="/usr/lib";;
	"Debian"|"LinuxMint"|"Ubuntu") libpath="/usr/lib/x86_64-linux-gnu"; export MONO_WINFORMS_XIM_STYLE=disabled;; # see https://bugzilla.xamarin.com/show_bug.cgi?id=28047#c9
	"Pop") libpath="/usr/lib/x86_64-linux-gnu";;
esac
...
```

ethanarns:
It doesn't seem necessary, as it works for me anyway, though I think it might be on select systems that this occurs. Pop_OS is Ubuntu, so I suggest it stays.

PR #2422, opened 2020-09-23
Fix SGB core picker which was broken

nattthebear:


YoshiRulz (approved):
With the special case removed this can be simplified further. Good enough for now though

PR #2430, opened 2020-09-27
Minor GB/C and PCFX firmware detection revisions

RetroEdit:
Change the status of alternate GB/C firmware options to unacceptable. Quoting from CasualPokePlayer "Gambatte is definitely not emulating those cpus [dmg0 and cgb0]. So you would end up with quirks of newer cpus being combined with bootroms only present in older cpus."

Also fix what looks to be a longstanding typo where the PCFX BIOS v1.01 was not being properly detected.


nattthebear (approved):
LGTM

PR #2432, opened 2020-09-28
Fallback to Patching the GBA version of the GBC BIOS in Memory

TiKevin83:
This avoids some of the hacks mentioned by natt on the last commit.

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/GBC_GB_PPU.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/GBC_GB_PPU.cs
@@ -146,7 +146,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.GBHawk
 					// note that their is no stat interrupt bug in GBC
 					STAT = (byte)((value & 0xF8) | (STAT & 7) | 0x80);
 
-					if (((STAT & 3) == 0) && STAT.Bit(3)) { HBL_INT = true; } else { HBL_INT = false; }
+					if (((STAT & 3) == 0) && STAT.Bit(3) && !glitch_state) { HBL_INT = true; } else { HBL_INT = false; }
 					if (((STAT & 3) == 1) && STAT.Bit(4)) { VBL_INT = true; } else { VBL_INT = false; }
 					// OAM not triggered?
 					// if (((STAT & 3) == 2) && STAT.Bit(5)) { OAM_INT = true; } else { OAM_INT = false; }
@@ -515,6 +515,8 @@ namespace BizHawk.Emulation.Cores.Nintendo.GBHawk
 					// we exit vblank into mode 0 for 4 cycles 
 					// but no hblank interrupt, presumably this only happens transitioning from mode 3 to 0
 					STAT &= 0xFC;
+					glitch_state = true;
+					LY_inc = 1;
 
 					// also the LCD doesn't turn on right away
 					// also, the LCD does not enter mode 2 on scanline 0 when first turned on
@@ -606,6 +608,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.GBHawk
 								STAT &= 0xFC;
 								STAT |= 0x03;
 								OAM_INT = false;
+								glitch_state = false;
 
 								OAM_access_read = false;
 								OAM_access_write = false;
@@ -1854,6 +1857,8 @@ namespace BizHawk.Emulation.Cores.Nintendo.GBHawk
 			for (int i = 0; i < OBJ_bytes.Length; i++) { OBJ_bytes[i] = 0xFF; }
 
 			pal_change_blocked = false;
+
+			glitch_state = false;
 		}
 	}
 }
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/GBC_PPU.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/GBC_PPU.cs
@@ -144,7 +144,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.GBHawk
 					// note that their is no stat interrupt bug in GBC
 					STAT = (byte)((value & 0xF8) | (STAT & 7) | 0x80);
 
-					if (((STAT & 3) == 0) && STAT.Bit(3)) { HBL_INT = true; } else { HBL_INT = false; }
+					if (((STAT & 3) == 0) && STAT.Bit(3) && !glitch_state) { HBL_INT = true; } else { HBL_INT = false; }
 					if (((STAT & 3) == 1) && STAT.Bit(4)) { VBL_INT = true; } else { VBL_INT = false; }
 					// OAM not triggered?
 					// if (((STAT & 3) == 2) && STAT.Bit(5)) { OAM_INT = true; } else { OAM_INT = false; }
@@ -510,6 +510,8 @@ namespace BizHawk.Emulation.Cores.Nintendo.GBHawk
 					// we exit vblank into mode 0 for 4 cycles 
 					// but no hblank interrupt, presumably this only happens transitioning from mode 3 to 0
 					STAT &= 0xFC;
+					glitch_state = true;
+					LY_inc = 1;
 
 					// also the LCD doesn't turn on right away
 					// also, the LCD does not enter mode 2 on scanline 0 when first turned on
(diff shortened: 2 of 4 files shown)


src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -101,7 +101,27 @@ public Gameboy(CoreComm comm, GameInfo game, byte[] file, Gameboy.GambatteSettin
...
+							bios = comm.CoreFileProvider.GetFirmware(biosSystemId, "World", true, "BIOS Not Found, Cannot Load.  Change SyncSettings to run without BIOS.");
+							byte[] agbOverride = { 0xFF, 0x00, 0xCD, 0x03, 0x35, 0xAA, 0x31, 0x90, 0x94, 0x00, 0x00, 0x00, 0x00 };
+							for (int i = 0xF3; i < 0x100; i++)
+							{
+								bios[i] = (byte)((agbOverride[i - 0xF3] + bios[i]) & 0xFF);
+							}
+						}
+						else
+						{
+							bios = comm.CoreFileProvider.GetFirmware(biosSystemId, biosId, true, "BIOS Not Found, Cannot Load.  Change SyncSettings to run without BIOS.");
```
YoshiRulz, line 122: For this line to be hit, the first lookup has to fail when looking not for an AGB BIOS. I think this should be `throw;` as it's only going to try the same lookup and fail.
If that's true, you'll then be able to invert the conditional to `if (...) throw; else ...;` and remove the redundant `else`. (I think there's a new syntax to make it even shorter but I can't remember it so ¯\\\_(ツ)_/¯.)

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -101,7 +101,27 @@ public Gameboy(CoreComm comm, GameInfo game, byte[] file, Gameboy.GambatteSettin
...
+					catch (MissingFirmwareException)
+					{
+						if (_syncSettings.ConsoleMode == GambatteSyncSettings.ConsoleModeType.GBA)
+						{
+							//implicitly throws MissingFirmwareException if GBC BIOS is also missing
+							bios = comm.CoreFileProvider.GetFirmware(biosSystemId, "World", true, "BIOS Not Found, Cannot Load.  Change SyncSettings to run without BIOS.");
+							byte[] agbOverride = { 0xFF, 0x00, 0xCD, 0x03, 0x35, 0xAA, 0x31, 0x90, 0x94, 0x00, 0x00, 0x00, 0x00 };
+							for (int i = 0xF3; i < 0x100; i++)
+							{
+								bios[i] = (byte)((agbOverride[i - 0xF3] + bios[i]) & 0xFF);
```
YoshiRulz, line 117: Are you able to use XOR instead of addition? (And aren't the last 4 0x0 bytes of the patch redundant?)

YoshiRulz (commented):
Just 2 style nitpicks.

Can you push the "completed" version in a second commit so zeromus is able to look at that too?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
zeromus: we should do whatever gambatte does

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
TiKevin83: thanks for the tip, learned some stuff about exception syntax

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
nattthebear: Does gambatte (upstream) do anything?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -101,7 +101,27 @@ public Gameboy(CoreComm comm, GameInfo game, byte[] file, Gameboy.GambatteSettin
 
 				if (_syncSettings.EnableBIOS)
 				{
-					bios = comm.CoreFileProvider.GetFirmware(biosSystemId, biosId, true, "BIOS Not Found, Cannot Load.  Change SyncSettings to run without BIOS.");
+					try
+					{
+						bios = comm.CoreFileProvider.GetFirmware(biosSystemId, biosId, true, "BIOS Not Found, Cannot Load.  Change SyncSettings to run without BIOS.");
```
nattthebear, line 106: Use `required: false` here if you're going to have a fallback instead of `catch (MissingFirmwareException)`

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -101,7 +101,27 @@ public Gameboy(CoreComm comm, GameInfo game, byte[] file, Gameboy.GambatteSettin
...
+					catch (MissingFirmwareException)
+					{
+						if (_syncSettings.ConsoleMode != GambatteSyncSettings.ConsoleModeType.GBA)
+						{
+							throw;
+						}
+						else
+						{
+							//implicitly throws MissingFirmwareException if GBC BIOS is also missing
+							bios = comm.CoreFileProvider.GetFirmware(biosSystemId, "World", true, "BIOS Not Found, Cannot Load.  Change SyncSettings to run without BIOS.");
```
nattthebear, line 117: Are we concerned that this leaves the "wrong" hash in the movie file?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -101,7 +101,27 @@ public Gameboy(CoreComm comm, GameInfo game, byte[] file, Gameboy.GambatteSettin
...
+							byte[] agbOverride = { 0xFF, 0x00, 0xCD, 0x03, 0x35, 0xAA, 0x31, 0x90, 0x94, 0x00, 0x00, 0x00, 0x00 };
```
nattthebear, line 118: Would it make more sense to override to the real bios and not the gambatte one?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
TiKevin83: in upstream it edits the BIOS in the core, which we removed in our fork to support DMG BIOS and the fake BIOSes

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
TiKevin83: Yes, there's some discussion about moving this in-memory fallback into the firmware manager to reduce duplicate code between GBHawk and Gambatte

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
nattthebear: OK, I was not aware if whether the code I removed was an actual upstream thing.

In that case, we are doing exactly what upstream does.  I don't know why it does the extra 4 0x00 bytes, but they have no negative effect.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
CasualPokePlayer: Well, there are some issues for that.
1) Copyright (which yes you could argue fair use but eh, RE has a way better argument for that)
2) Trade secret laws (our knowledge of the code from the agb-cgb bios literally comes from illegal leaks)
3) Overall would make the code more complicated (since there's another change elsewhere in the BIOS... which doesn't change cycle counts and its only practical effects would be hardware only).

The Gambatte one is a bit more simple and comes from RE work so it would be far more legal to use than to deal with putting leak content in the fallback.

nattthebear: Those sound like nasty consequences for either choice.  Maybe it's better to just not do this.

Issue #2428, opened 2020-09-26
TAStudio slows down noticably after the project becomes larger


RetroEdit:
Something I've noticed and several other users have reported is that TAStudio will become very sluggish once you get far enough into the project. Dragging inputs can take a few seconds once you get to 60k frames.

This is a result of the new state manager. If you configure the Savestate History Settings to allocate less space for the Current buffer and Recent buffer, ~~it will no longer be an issue~~ *EDIT*: it can still happen even at the smallest settings. However, it is still troubling that a project will slow down noticeably and likely indicates some underlying inefficiency.

~~If this is not fixable by reworking logic, I would recommend the defaults be set to be smaller, but this shouldn't be necessary.~~ This needs to be fixed by reworking the logic.

RetroEdit:
Okay, this is unfortunately not straightforward to reproduce. I'll need to see if I can come up with a better set of reproduction steps at some point.

RetroEdit:
I have a straightforward way of reproducing this now, thanks to Samsara.

1. Open Super Mario Bros. in NESHawk and open TAStudio
1a. Click and drag the mouse over the input roll to get a sense of the baseline speed for drawing the inputs.
2. Just unpause emulation and let around 40000 frames be emulated in TAStudio (41000 is a good target)
3. Click and drag the mouse around the 40000 frame mark. It will be noticeably slower than before, and probably take a few seconds to draw the letters corresponding to the inputs.

nattthebear:
I probably can't do anything about this as is, but feel free to assign to me if you can't straighten it out and can explain to me exactly what calls are slow  @adelikat @RetroEdit

RetroEdit:
@nattthebear One major issue seems to be that painting calls `ZwinderStateManager.InvalidateAfter()` several times in quick succession, once for each frame that is changed. The `StateCache.RemoveWhere(s => s > frame);` call seems to be somewhat expensive. If the state manager internally kept track of the last valid frame in a separate variable, then it could skip doing invalidation in many cases when it would be redundant.

nattthebear:
Seems like it would also help to turn those into a single `InvalidateAfter` call

RetroEdit:
Yeah, turning those into a single `InvalidateAfter` call might be better, but it also might be a bit complicated, due to the many convuluted layers of architecture involved in painting. The choice to call `InvalidateAfter` for each frame that was changed was made 5 years ago, so it's a bit more set-in.

adelikat:
[...]

PR #2433, opened 2020-09-29
StateManager - increase performance when there are a lot of states, fixes issue #2428

adelikat:
The data structure for the StateCache was the culprit here.  The datastructure meant calls to Invalidate got slow at large numbers, even if no frames are determined to need to be invalidated, (the RemoveWhere() is still costly), causing painting to be laggy.

Using a custom SortedInt<int> allows for quick removals as well as fast duplicate check + additions.

Maybe a better structure could be conceived down the road but this seems to address the primary performance bottlenecks

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using  BizHawk.Common;
 using BizHawk.Emulation.Common;
 
 namespace BizHawk.Client.Common
@@ -11,7 +12,7 @@ namespace BizHawk.Client.Common
 		private static readonly byte[] NonState = new byte[0];
 
 		private readonly Func<int, bool> _reserveCallback;
-		internal readonly List<int> StateCache = new List<int>();
+		internal readonly SortedList<int> StateCache = new SortedList<int>();
 		private ZwinderBuffer _current;
 		private ZwinderBuffer _recent;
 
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using  BizHawk.Common;
```
nattthebear, line 5: Two spaces between `using` and the namespace?

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -458,7 +469,7 @@ public bool InvalidateAfter(int frame)
 			var b1 = InvalidateNormal(frame);
 			var b2 = InvalidateGaps(frame);
 			var b3 = InvalidateReserved(frame);
-			StateCache.RemoveWhere(s => s > frame);
+			StateCache.RemoveAll(s => s > frame);
```
nattthebear, line 472: Seems like this could be replaced by a `RemoveAfter` method on the sorted list, which would have better runtime characteristics.

But if this is fast enough, maybe we don't care more.

nattthebear (approved):
lgtm

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
RetroEdit: Done in e050938.

src/BizHawk.Tests/Common/CustomCollections/CustomCollectionTests.cs
```diff
@@ -27,5 +27,30 @@ public void TestSortedListContains()
...
+
+		[TestMethod]
+		public void TestSortedListRemoveAfter_None()
+		{
+			RemoveAfterTestcase(new[] { 2, 3 }, new[] { 2, 3 }, 5);
+		}
+
+		[TestMethod]
+		public void TestSortedListRemoveAfter_All()
+		{
```
YoshiRulz, line 45: This is not how you write test parameters in MSTest but w/e

PR #2440, opened 2020-10-06
Confine SlimDX to a single project

YoshiRulz:
Future work:
* create new project for OpenTK too
* ...except I can't, because OpenTK types are used as a "base" or "default", in graphics and in input; those will need abstracting or perhaps something more clever
* with the usual Attribute shenanigans, instantiate `DirectSoundSoundOutput`/`XAudio2SoundOutput`, `IGL_SlimDX9`, and `DirectInputAdapter` by reflection
* instantiate GDI+ and the OpenTK wrappers by reflection in the same way


nattthebear:
I think this refactoring is a good step forward on cleaning up our platform code

adelikat (approved):

PR #2486, opened 2020-11-23
Update libgambatte to r729+

### LINKED ISSUE #1987: Gambatte: emulate un-mapped memory in GBA mode
@TiKevin83 
The region 0xFEA0 - 0xFEFF behaves differently on GBA then on GBC, so this might be relevant for console verification.

see: https://github.com/AntonioND/gbc-hw-tests
memory -> oam_echo_ram_lcd_on.gbc

I just went through this with GBHawk, it should return (addr & 0xF0) | ((addr & 0xF0) >> 4) and writes should have no effect, not too hard. Could throw off a run though since Gambatte currently doesn't do anything with GBA mode in this case.

### LINKED ISSUE #2436: Sprites do not appear in Mortal Kombat in GBC mode
Sprites are used for score and the player icon on the ladder. They do not appear in Gambatte, appear as normal on GBP.

See: https://github.com/pokemon-speedrunning/gambatte-speedrun/issues/66

EDIT: This has been fixed upstream

TiKevin83:
fixes #1987 and several other bugs, mostly related to CGB-DMG mode
fixes #2436


alyosha-tas:
Looks good to me.
It looks like this also addresses #2436 ?
(There's an extra semi colon in memory.cpp line 106)

There's still an issue with gambatte with sprite priority (I opened an issue in gambatte speedrun but not here) that I'm still hoping to see fixed for 2.5.3 since it's pretty annoying, but looks like the fix is still WIP and this is still a big improvement at least.

alyosha-tas (approved):

nattthebear:
LGTM

PR #2490, opened 2020-11-24
GB layers

brunovalads:
- Fixed OBJ layer toggle message;
- Added WIN (window) layer toggle hotkey option;
- GPU Viewer add-on to display sprite tiles on the real position, just like what you have for the bsnes core. Has several issues:
  - Drawing is identical to the current sprite display, so it doesn't have the transparency it should have;
  - Drawing doesn't take priority into account, so overlapping is usually wrong;
  - The drawing flickers, probably the way of refreshing (with `.Clear()`) is bad.

### DIFF
--- a/src/BizHawk.Client.Common/config/Binding.cs
+++ b/src/BizHawk.Client.Common/config/Binding.cs
@@ -240,6 +240,7 @@ namespace BizHawk.Client.Common
 
 						Bind("GB", "GB Toggle BG"),
 						Bind("GB", "GB Toggle Obj"),
+						Bind("GB", "GB Toggle Window"),
 
 						Bind("Analog", "Y Up Small", toolTip: "For Virtual Pad"),
 						Bind("Analog", "Y Up Large", toolTip: "For Virtual Pad"),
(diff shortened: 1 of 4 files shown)


nattthebear:
Wish we handled layer toggle keybinds in a less ad-hoc manner, but this isn't the place for that

brunovalads:
> layer toggle keybinds

I was thinking in the Hotkeys menu instead of multiple tabs for each console (for some reason only SNES, GB and NDS have), you could have a "Console specific" tab, then inside it it would be divided into the different consoles that benefit from hotkeys (like layer toggling, audio channel toggling, etc).

brunovalads:
[Image reference for the GB sprite object viewer](https://i.imgur.com/bpMjyAd.png)

nattthebear:
> The drawing flickers, probably the way of refreshing (with .Clear()) is bad.

Try clearing directly in the draw routine (just write whatever backdrop color you want to every pixel of dest)

adelikat (approved):

PR #2493, opened 2020-11-26
Cleanup initialisation of GlobalWin.GL

YoshiRulz:
Mupen is still able to load on my Win10 machine with the D3D display method selected.

### DIFF
--- a/src/BizHawk.Bizware.BizwareGL/OpenTKConfigurator.cs
+++ b/src/BizHawk.Bizware.BizwareGL/OpenTKConfigurator.cs
@@ -1,0 +1,22 @@
+using System;
+
+using OpenTK;
+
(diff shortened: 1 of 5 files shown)


zeromus:
1. Don't care
2. Obviously it wasn't in use any more.... you can remove this if you want, but be ready to be maligned if anyone ever needs it again. Moving the framebuffer into CPU and then again into a texture is a big speed hit right now in what would most likely be pcsx2 or dolphin, but it isn't even clear how that would work out for a lot of other reasons. 
3. I don't want to say yes to a goto removal and a local function, but I won't say no.

nattthebear:
>Mupen is still able to load

Eww.  Could you make it so Mupen is not able to load?

PR #2508, opened 2020-12-07
[PSX] Make Octoshock more debuggable

TwelveBaud:
I'm working on making an external tool for second-screening a PSX game, and as part of that I'm trying to dig into what makes the game tick. This pull request lights up the debugger (as much as I can).

* 28b36930 adds support for a "System Bus" memory domain to match the memory callback. All the normal-ish memory on the bus is represented, with MMIO and empty spaces represented by `00`. This prevents a crash when trying to add a breakpoint, and read and write breakpoints function.
* d686dc5d calls into the core's own disassembler to drive the BizHawk disassembler. This causes disassembly to appear instead of just "".
* _Commits after this point modify the core and include a new binary I built._
* 67342a6d adds support for execute callbacks to the core. Note that these *dramatically* reduce performance, and because of how BizHawk's debugger is structured do not stop immediately; it's recommended not to set any breakpoints that'd be hit multiple times per frame.
* 372a32ff relocates where read and execute callbacks occur to just after the fetch completes. This allows callback code to know what value was fetched. Note that for read callbacks this means it's too late to change what value is being read at that point.
* 89287639 changes the interop layer to use `SafeHandle`s rather than `IntPtr`s. It's not necessary, especially since the core only pretends to support multiple instances, but it'll be safer when that support is added.
* 5311d275 reverts the change from 28b36930 and exposes the core's own `PeekMemory`/`PokeMemory` functions as the System Bus domain. It uses the ones from the CPU's perspective so that `DCache`/scratchpad is also within view; the actual system bus (and the ones from the perspective of the core) doesn't map that region.
* a9971bc9 moves all of the `MemoryDomain`-related code to `Octoshock.IDebuggable.cs` to live next to the CPU registers and `MemoryCallbackSystem`.


TwelveBaud:
**Next Steps:**

I really want to be able to have breakpoints stop the core *immediately* when hit, but the way the debugger works is that it sets a flag so that EmuHawk pauses at end of frame; that's beyond my ability to safely change. Similarly, I want to add support for `Step Into` (advance one instruction), `Step Over` (advance until `newPC == oldPC + 4` for regular instructions and `newPC == oldPC + 8` for branch instructions), and `Step Out` (advance until `newPC == oldLR`), but the core's atom of work ([`PS_CPU::RunReal`](https://github.com/TASVideos/BizHawk/blob/a9971bc96a91f0aba95db579d48f556ea3f8bf1e/psx/octoshock/psx/cpu.cpp#L557)) runs several instructions until another part of the core needs servicing, and I'm not competent to detangle the part of the loop that executes one instruction from the stuff surrounding it (something about hot locals and backing globals), nor was I able to figure out a way to turn the loop into a coroutine based on the available-but-unused `CPUHook` function (since you "can't" `longjmp` into a function that was deeper on the stack when its execution ended).

vadosnaprimer:
> I really want to be able to have breakpoints stop the core _immediately_ when hit, but the way the debugger works is that it sets a flag so that EmuHawk pauses at end of frame; that's beyond my ability to safely change. Similarly, I want to add support for `Step Into` (advance one instruction), `Step Over` (advance until `newPC == oldPC + 4` for regular instructions and `newPC == oldPC + 8` for branch instructions), and `Step Out` (advance until `newPC == oldLR`), but the core's atom of work ([`PS_CPU::RunReal`](https://github.com/TASVideos/BizHawk/blob/a9971bc96a91f0aba95db579d48f556ea3f8bf1e/psx/octoshock/psx/cpu.cpp#L557)) runs several instructions until another part of the core needs servicing, and I'm not competent to detangle the part of the loop that executes one instruction from the stuff surrounding it (something about hot locals and backing globals), nor was I able to figure out a way to turn the loop into a coroutine based on the available-but-unused `CPUHook` function (since you "can't" `longjmp` into a function that was deeper on the stack when its execution ended).

Not sure if feasible in the corrent state of the project, but one person successfully implmented that for our n64 core some time ago.

https://github.com/TASVideos/BizHawk/issues/676#issuecomment-255526749
https://github.com/TASVideos/BizHawk/issues/676#issuecomment-407103055

> Being able to export to the core what specific address are of interest would greatly increase performance when callbacks are enabled (since it'd avoid transitioning into and out of managed code for _every_ bus access and only require it when something interesting needs to happen), but the `MemoryCallbackSystem` interface doesn't support that.

We do this in mGBA as well as n64, you might wanna check those out and do something similar. For such a heavy-weight core indeed it makes sense to host the breakpoints on the c(++) side.

zeromus:
SafeHandle is just noise. This isn't an appropriate application. Most applications of SafeHandle aren't appropriate. I disagree with the entire philosophy. There's nothing important in octoshock to clean up. If a handle is leaked, that's a bug. We don't need it automatically finalized. I don't mind having a typed handle, but the only special finalizing logic I want is to THROW AN EXCEPTION if it hasn't been manually released. Failing to manually release a handle is a bug that needs fixing. Cleaning up when the appdomain tears down is pointless. A plain boring typed handle can be done by a struct containing an Intptr.

Those are my thoughts on the matter. If it's important to you, leave it. For all the other work you've done, you're entitled to it.

Everything else looks fine. Good job, and thanks. Sorry the debugger is in such a shoddy state.  If you add step-over and such, make it a new eShockStep.

TwelveBaud:
> Why not near [the trace callback] though?
Definitely fits better there, fixed!

> I disagree with the entire philosophy. [...] Failing to manually release a handle is a bug that needs fixing.
Since I work almost exclusively with managed code, my default expectation is that I don't have to manually clean up anything; I can trust the compiler to nag me about `IDisposable`s where appropriate. Using safe handles helps reinforce that for me, especially when I was about to make that handle actually represent unmanaged resources. That said, I didn't follow through with that, the paradigm isn't used in other cores either, and ultimately it's not important, so I reverted it.

> Sorry the debugger is in such a shoddy state.
Having any kind of debugger is helpful, and I'm glad it's there to tie into!

zeromus:
Ahhh hm. Well, if your compiler can help you nag you about IDisposables, then that's cool. I just object to finalizers. Finalizers should not exist (or should be reworked to not be terrible). Any code relying on finalizers can't be relied upon.

adelikat (approved):

PR #2526, opened 2020-12-19
add a Lua method to get inputs from the movie

SuuperW:
It seems we didn't have a way for Lua scripts to get the inputs created by movies.

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
@@ -25,6 +25,11 @@ namespace BizHawk.Client.Common
 			return _inputManager.AutofireStickyXorAdapter.ToDictionary(controller);
 		}
 
+		public IDictionary<string, object> GetWithMovie(int? controller = null)
+		{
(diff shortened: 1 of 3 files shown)


YoshiRulz (approved):
LGTM, though I feel like a better name than `GetWithMovie` exists.

Does anyone remember why we're removing entries from tables in this library?

YoshiRulz:
I tracked down the commit which added the `table[...] = null;` nonsense: 78d8effcf. There's no explanation. It was so long ago that we should just assume it wass a workaround for a problem that has since been fixed, and remove it.

adelikat:
Why is this a joypad method and not a movie method?

YoshiRulz:
I think it makes sense where it is. Its semantics aren't specifically about movie input, the inputs it's reading are the sum of the whole stack including autofire/autohold and Virtual Pads (IIRC).

adelikat:
Fair enough.

adelikat (approved):

ThunderAxe31:
I think there was already one Lua function for getting inputs from a movie file: `string movie.getinputasmnemonic(int frame)`

PR #2527, opened 2020-12-19
add an option to capture the Lua without capturing the full OSD, when…

SuuperW:
It's a bit sloppy, do we want this? Should be easy to extend this to screenshots if so.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
@@ -136,7 +136,6 @@ namespace BizHawk.Client.EmuHawk
 			this.toolStripMenuItem4 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.DisplayStatusBarMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.DisplayMessagesMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.toolStripSeparator8 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.DisplayLogWindowMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.ConfigSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.ControllersMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
@@ -375,6 +374,7 @@ namespace BizHawk.Client.EmuHawk
 			this.ShowMenuContextMenuSeparator = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.ShowMenuContextMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.timerMouseIdle = new System.Windows.Forms.Timer(this.components);
+			this.toolStripSeparator8 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.MainformMenu.SuspendLayout();
 			this.MainStatusBar.SuspendLayout();
 			this.MainFormContextMenu.SuspendLayout();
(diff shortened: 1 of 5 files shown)


src/BizHawk.Client.EmuHawk/MainForm.Events.cs
```diff
@@ -538,6 +539,15 @@ private void SynclessRecordingMenuItem_Click(object sender, EventArgs e)
...
 			Config.AviCaptureOsd ^= true;
+			if (Config.AviCaptureOsd)
+				Config.AviCaptureLua = true;
+		}
+
+		private void CaptureLuaMenuItem_Click(object sender, EventArgs e)
+		{
+			Config.AviCaptureLua = !CaptureLuaMenuItem.Checked;
+			if (!Config.AviCaptureLua)
+				Config.AviCaptureOsd = false;
```
YoshiRulz, line 550: Is this trying to re-implement the behaviour of a checkbox with a child checkbox? I can't understand what the changes in this file do.

src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
```diff
@@ -382,28 +383,28 @@ private void InitializeComponent()
...
+            this.ColecoSubMenu,
+            this.N64SubMenu,
+            this.DGBSubMenu,
+            this.AppleSubMenu,
+            this.C64SubMenu,
+            this.IntvSubMenu,
+            this.zXSpectrumToolStripMenuItem,
+            this.GenericCoreSubMenu,
+            this.amstradCPCToolStripMenuItem,
+            this.HelpSubMenu});
```
YoshiRulz, line 407: Fix this

YoshiRulz (changes requested):

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
SuuperW: "checkbox with a child checkbox" .. I guess? It's to avoid having Capture OSD still turned on when Capture Lua is turned off (because I haven't implemented a way to capture OSD elements without also capturing Lua).

YoshiRulz: To allow OSD + Lua, ~OSD + Lua, and ~OSD + ~Lua, but not OSD + ~Lua, try:
```csharp
private void AVSubMenu_DropDownOpened(object sender, EventArgs e)
{
	// ...
	CaptureOSDMenuItem.Checked = Config.AviCaptureOsd;
	CaptureLuaMenuItem.Checked = Config.AviCaptureLua;
	// ...
}

private void CaptureOSDMenuItem_Click(object sender, EventArgs e)
{
	Config.AviCaptureOsd = ((CheckBox) sender).Checked;
}

private void CaptureLuaMenuItem_Click(object sender, EventArgs e)
{
	var c = ((CheckBox) sender).Checked;
	Config.AviCaptureLua = c;
	CaptureOSDMenuItem.Enabled = c;
	if (!c) CaptureOSDMenuItem.Checked = false; // on negative edge only; setter triggers event handler which sets config
}
```

SuuperW: With this code, if config is currently ~OSD + ~Lua and the user checks CaptureOSD, then Config will have OSD + ~Lua.

Also, imo disabling the OSD menu item when ~Lua is less user-friendly than current implementation (which is to automatically turn on Lua when the user turns on OSD).
(We would also need to add code to conditionally en/disable CaptureOSDMenuItem inside AVSubMenu_DropDownOpened or on init if we go that route, for consistency.)

Perhaps some comments added to current implementation would help?
`			Config.AviCaptureLua = !CaptureLuaMenuItem.Checked;
			if (!Config.AviCaptureLua) // Logic to capture OSD w/o Lua does not currently exist, so disallow that.
				Config.AviCaptureOsd = false;
`

YoshiRulz: If you used the CheckBox' state instead of writing to and then immediately reading from config, then added comments, that would also work.

YoshiRulz (approved):

PR #2528, opened 2020-12-19
Input roll scrolling

SuuperW:
A couple of fixes. I have not tested this outside of TAStudio, though I see no reason it would break anything.
See commit messages for details.


vadosnaprimer:
Tested, looks good. 13 there certainly doesn't look intended.

PR #2533, opened 2020-12-25
Drive states

SuuperW:
Add support for putting TAStudio's state history buffers and/or ancients in files instead of keeping it all in RAM.

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Diagnostics;
 using BizHawk.Common;
 using BizHawk.Emulation.Common;
 
@@ -9,7 +10,8 @@ namespace BizHawk.Client.Common
 {
 	public class ZwinderStateManager : IStateManager, IDisposable
 	{
-		public const string baseStatePath = "tastudiostates/";
+		private const string _path = "tastudiostates/";
+		public static string baseStatePath = _path;
 
 		private bool disposed = false;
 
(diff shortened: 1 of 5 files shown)


nattthebear:
Would it be simpler and more effective to simply allow the backing address space for ZWinderBuffer to be a mmapped file?

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -41,24 +49,41 @@ public ZwinderStateManager(Func<int, bool> reserveCallback)
...
 		{
+			SetId();
 			_current = current;
 			_recent = recent;
 			_gapFiller = gapFiller;
 			_ancientInterval = ancientInterval;
 			_reserveCallback = reserveCallback;
 		}
 		
+		private void SetId()
```
nattthebear, line 76: This needs to be threadsafe -- the usual expectation in C# is that while single objects aren't threadsafe, multiple threads can independently create independent objects without conflict.  But here the internal implementation detail hits a static resource without synchronization.  Try `Interlocked.Increment`.

src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
```diff
@@ -18,6 +18,13 @@ public class ZwinderBuffer : IDisposable
 		*/
```
nattthebear, line 18: The general problem I have in this file is that ZwinderBuffer is made to do one thing: abstract on top of a ring buffer that's mapped to memory.  With the UseDrive stuff, a bunch of important methods have completely unrelated separate implementations for the two cases.  I don't see that this should all be handled in one file.  If there is commonality, it should be brought out to a base class or base interface or something,

zeromus:
I spent a full 2 hours thinking about this and have a number of ideas for how it could be done. But the lowest-impact, safest, and simplest idea: Can we have SaveStateStream and LoadStateStream use an underlying stream indiscriminately after doing the ring buffer logic? Then we can make the underlying stream a filestream or a new slim little stream that simply sits on top of a raw pointer, if there isn't already such a thing. I think I could do this in about 1 more hour.

nattthebear:
> I spent a full 2 hours thinking about this and have a number of ideas for how it could be done. But the lowest-impact, safest, and simplest idea: Can we have SaveStateStream and LoadStateStream use an underlying stream indiscriminately after doing the ring buffer logic? Then we can make the underlying stream a filestream or a new slim little stream that simply sits on top of a raw pointer, if there isn't already such a thing. I think I could do this in about 1 more hour.

I like that.  Take a look at `MemoryBlock.GetStream`, that should do what we need.

SuuperW:
I also like the idea of accessing the memory block as a stream. I had considered making the two share an interface, but for some reason I only thought of trying to make the file accessed more like the memory block, but that didn't seem very clean either. Going the other way and turning the memory block into a stream should be super easy and clean, since we already have MemoryBlock.GetStream.

nattthebear:
I would like to see this refactored to use the capability added in https://github.com/TASVideos/BizHawk/pull/2536.  It's OK if some additional changes are needed to ZwinderBuffer above and beyond what that does, but hopefully most of that is already taken care of.

PR #2536, opened 2020-12-26
Refactor ZWinderBuffer to allow use of arbitrary backing Streams

nattthebear:
A future commit may expose this functionality for use in ZWinderStateManager.

Should we expose this directly to the user, so the rewinder can be disk based?

### DIFF
--- a/src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
+++ b/src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
@@ -42,7 +42,7 @@ namespace BizHawk.Client.Common
 				case IRewindSettings.BackingStoreType.TempFile:
 				{
 					var filename = TempFileManager.GetTempFilename("ZwinderBuffer");
-					var filestream = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite);
+					var filestream = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite, FileShare.None, 4096, FileOptions.DeleteOnClose);
 					filestream.SetLength(Size);
 					_backingStore = filestream;
 					_disposables.Add(filestream);
--- a/src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
+++ b/src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
@@ -1,6 +1,8 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using BizHawk.BizInvoke;
 using BizHawk.Common;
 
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
```diff
@@ -26,18 +28,42 @@ public ZwinderBuffer(IRewindSettings settings)
...
+					buffer.Protect(buffer.Start, buffer.Size, MemoryBlock.Protection.RW);
+					_disposables.Add(buffer);
+					_backingStore = new MemoryViewStream(true, true, (long)buffer.Start, (long)buffer.Size);
+					_disposables.Add(_backingStore);
+					break;
+				}
+				case IRewindSettings.BackingStoreType.TempFile:
+				{
+					var filename = TempFileManager.GetTempFilename("ZwinderBuffer");
+					var filestream = new FileStream(filename, FileMode.Truncate, FileAccess.ReadWrite);
```
SuuperW, line 45: I attempted to run this, with a quick modification to force it to use a file, and I got a System.IO.FileNotFoundException exception on ZwinderBuffer line 45.

SuuperW (commented):
I attempted to run this, with a quick modification to force it to use a file, and I got a System.IO.FileNotFoundException exception on ZwinderBuffer line 45.

src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs
nattthebear: Thanks, I didn't actually test this mode yet.  Will do a bit later today and make whatever corrections are needed.

nattthebear: I had used the wrong FileMode; looking at the docs, `Create` is what I want (or maybe `CreateNew`).  The unit tests now pass with `Create` used and the ZWinderBuffer modified to always use TempFile.

SuuperW:
It seems to work now, from my quick test, although it isn't very good about deleting the files when closing the application. That's probably a problem to be fixed in MainForm/Shutdown logic though.

zeromus:
We have a temp file manager with a bunch of logic in it for dealing with this kind of thing. At least it can use windows to open a file that's automatically deleted when it's closed, among other things. It's kind of hard to figure out how to use, but it's been there for a long time

nattthebear:
> It's kind of hard to figure out how to use

That's my fault then, I assumed it wasn't a pile of shit.  I'll roll my own NattTempFileManager I guess...

zeromus:
Oh never mind, I mis-remembered. Mine is designed for a special case where FileOptions.DeleteOnClose won't work.
For this we should use FileOptions.DeleteOnClose

PR #2542, opened 2020-12-30
Drive states2

SuuperW:
A re-work of #2533, following #2536.
In addition, saving/loading now works properly with the new settings.


nattthebear (approved):
Looks good as far as I'm concerned.  @adelikat might have other feedback on the state manager side.

adelikat:
I won't hold this up.  I keep thinking I'm going to deep dive the state manager changes but I'm not getting around to it.  A cursory glance it looks okay

PR #2563, opened 2021-01-12
Fix for state gap when frame advancing and painting input

alyosha-tas:
Painting inputs in the piano roll can cause gaps in the greenzone and desyncs in movies if done while frame advance is happening. This PR prevents frame advance from any source while inputs are being painted.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -768,6 +768,17 @@ namespace BizHawk.Client.EmuHawk
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -768,6 +768,17 @@ private set
 			}
 		}
 
+		private bool _block_frame_advance;
```
adelikat, line 771: There's no reason for this backing value.  You can just have public BlockFrameAdvance { get; set; }

adelikat (approved):

PR #2566, opened 2021-01-14
Rework GB GPU memory areas API

### LINKED ISSUE #2564: Improve the IGameboyCommon.GetGPU API
It started as a Gambatte only API, then was minimally worked over to support waterbox.  It doesn't really feel nice implementing on a managed core at all.  Here's what I'm thinking of.

```cs
interface IGameboyCommon
{
  // ... other methods omitted

  // Acquire GPU memory for inspection.  The returned object must be disposed as soon as the frontend
  // tool is done inspecting it, and the pointers become invalid once it is disposed.
  IGPUMemoryAreas LockGpu();
}

interface IGPUMemoryAreas : IDisposable
{
  IntPtr Vram { get; }
  IntPtr Oam { get; }
  IntPtr Sppal { get; }
  IntPtr Bgpal { get; }
}
```

* The waterbox (Sameboy) implementation could acquire the waterbox lock in `LockGpu` return the pointers directly to guest memory, and release the lock in `Dispose`.
* The Gambatte implementation could just return the pointers to unmanaged memory, no problem.
* The GBHawk implementation could pin the managed arrays in `LockGpu`, return those pointers, and then unpin in dispose, removing the need for the copy that we have now.

So all three cores would be able to do this with 0 copy and in a reasonably ergonomic way.

I'm willing to do this work myself, but I want to see if it will fit our needs.  @alyosha-tas let me know what you think.

nattthebear:
All cores can easily implement it now with no copying or awkward garbage

Fixes #2564

### DIFF
--- a/src/BizHawk.Common/IsExternalInit.cs
+++ b/src/BizHawk.Common/IsExternalInit.cs
@@ -1,0 +1,4 @@
+namespace System.Runtime.CompilerServices
+{
+	public static class IsExternalInit {}
+}
--- a/src/BizHawk.Client.EmuHawk/tools/GB/GBGPUView.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/GB/GBGPUView.cs
@@ -32,13 +32,24 @@ namespace BizHawk.Client.EmuHawk
 		// g' = 8.25g
 		// b' = 8.25b
 
-
-		private GPUMemoryAreas _memory;
-
 		private bool _cgb; // set once at start
 		private int _lcdc; // set at each callback
 
-		private IntPtr _tilesPal; // current palette to use on tiles
+		/// <summary>
+		/// Whether the tiles are being drawn with the sprite or bg palettes
+		/// </summary>
+		private bool _tilesPalIsSprite;
+		/// <summary>
+		/// How far (in bytes, I guess?) we should offset into the tiles palette
+		/// </summary>
+		private int _tilesPalOffset;
+
+		private IntPtr ComputeTilesPalFromMemory(IGPUMemoryAreas m)
+		{
+			var ret = _tilesPalIsSprite ? m.Sppal : m.Bgpal;
+			ret += _tilesPalOffset;
+			return ret;
+		}
 
 		private Color _spriteback;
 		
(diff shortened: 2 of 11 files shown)


src/BizHawk.Common/IsExternalInit.cs
```diff
@@ -0,0 +1,4 @@
+namespace System.Runtime.CompilerServices
+{
+	public static class IsExternalInit {}
```
YoshiRulz, line 3: Is this kosher? The `init` auto-props you added aren't really an improvement over using named arguments w/ a ctor anyway.

adelikat (approved):

alyosha-tas (approved):

alyosha-tas:
Tested everything and works the same as current master, didn't get any other errors.

I did notice an error though that is also in master, in Sameboy when you click on 'scanline' in 'refresh control' it gives this error:
```
System.InvalidOperationException: All callback delegates must be registered at load
   at BizHawk.BizInvoke.CallingConventionAdapters.WaterboxAdapter.GetFunctionPointerForDelegate(Delegate d) in C:\Users\david\Desktop\BizHawk\src\BizHawk.BizInvoke\CallingConventionAdapter.cs:line 225
   at Bizhawk.BizInvokeProxyLibSameboy.SetScanlineCallback(ScanlineCallback , Int32 )
   at BizHawk.Client.EmuHawk.GbGpuView.UpdateBefore()
   at BizHawk.Client.EmuHawk.ToolManager.UpdateToolsBefore()
   at BizHawk.Client.EmuHawk.MainForm.StepRunLoop_Core(Boolean force)
   at BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop()
   at BizHawk.Client.EmuHawk.Program.SubMain(String[] args)
```

Can this be fixed now or should I make an issue for it?

nattthebear:
> I did notice an error though that is also in master, in Sameboy when you click on 'scanline' in 'refresh control' it gives this error:

That's my fault, but probably not a new bug.  In any event, I can and will fix it for this PR.

> I also noticed that Gambatte's timing for refresh on 'frame' is slightly inconsistent, this is apparent on the language select screen in Wacky Racers. This is also in master though, maybe it has always been this way.

Changing that would be too deep for this PR, I think.  I recommend a separate issue.

alyosha-tas:
Never mind about the second part, I looked at trace logs and it's just an unfortunate consequence of frame timing, the game is changing palette regs right as VBL starts. GBHawk frames end in a different spot and avoid the issue just by luck.

Issue #2567, opened 2021-01-15
Movie Importer imports a blank LogKey


alyosha-tas:
As the title says, this has been an issue since 2.5, it worked in 2.4.2.

At the time ProcessMovieImport is called in MainForm, MovieSession.MovieController is the null controller.

That's as far as I could get so far.

YoshiRulz:
🤔
https://github.com/TASVideos/BizHawk/blob/7f0947108c318c6d1851720c5e63ba49583ddc9e/src/BizHawk.Client.Common/movie/bk2/Bk2LogEntryGenerator.cs#L14
https://github.com/TASVideos/BizHawk/blob/7f0947108c318c6d1851720c5e63ba49583ddc9e/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.InputLog.cs#L17

alyosha-tas:
Just some additional info here. I tried importing a one player .fm2 movie with NESHawk loaded and 2 zappers in NEShawk in 2.4.2. The resulting imported movie logKey had the 2 zappers, so this apparently was never accurate.

YoshiRulz:
At c2d5a9c93, printing out `MovieSession.MovieController.Definition.OrderedControlsFlat` here:
https://github.com/TASEmulators/BizHawk/blob/c2d5a9c9316a4aa3af2c7265cae13b3e366185fd/src/BizHawk.Client.EmuHawk/MainForm.cs#L4130-L4132
...shows it's still blank, recent changes haven't affected this bug.

PR #2568, opened 2021-01-15
Fix for #2567 LogKey missing in input log of .bk2 made by importer

YoshiRulz:
#2567

### DIFF
--- a/src/BizHawk.Client.Common/movie/MovieSession.cs
+++ b/src/BizHawk.Client.Common/movie/MovieSession.cs
@@ -239,7 +239,7 @@ namespace BizHawk.Client.Common
 
 		public void RunQueuedMovie(bool recordMode, IEmulator emulator)
 		{
-			MovieController = new Bk2Controller(emulator.ControllerDefinition);
+			SetControllerFrom(emulator);
 
 			Movie = _queuedMovie;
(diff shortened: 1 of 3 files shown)


alyosha-tas:
Do we want the key to come from the emulator? It could be way different then that of the imported file in some cases (like if you have 2 zappers currently plugged in on NES.)

YoshiRulz:
The emulator should be restarted with the correct sync settings when importing a movie. Though I'm not sure if that's the case now.

adelikat:
This isn't the right way to do this.  This causes the import to change the MovieSession state!  Also, does this not put the logkey as the current state of the current emulator? Which isn't guaranteed to be the configuration of the imported movie. I'd rather have no log key than the wrong one.

The importer already has to create a controller object in order to import key presses, it should use that controller state to generate a log key.

YoshiRulz:
I have no idea then. The code which serialises the `LogKey` reads `MovieSession.MovieController` which, without this change, can only ever be set by `MovieSession.RunQueuedMovie`, which in turn is only ever called in `MainForm.StartNewMovie`. (This latter method *can* be called when using the importer, but not with `File` > `Movie` > `Import Movies...`, and I think it's executed in the wrong order for this purpose.)

PR #2579, opened 2021-01-18
Unification of modal dialogs for errors and such

YoshiRulz:
Done for now. There are a couple I couldn't convert because they used something WinForms-specific which I felt did not belong in our abstraction.

The part of this that I think would be the most objectionable is 07eeca8d1, where I move the root and leaf interfaces to `BizHawk.Common` and use extension methods to call into WinForms.

Note that even with 1c5bdc0a9, the AV stuff still transitively depends on WinForms because they have config `Form`s which, though not shown directly, are shown by static helper methods. It may be possible to create an abstraction for those `Form`s but I don't think it will look nice.

edit 2021-02: ~~when rebasing, add a new method `IDialogController.AskSwitchToAccurateCore` afterwards (to replace existing usage, and to be used in movie importers which currently write config without notice)~~
edit 2021-03: nevermind that won't work; also rebased with suggested changes


adelikat (changes requested):
Put that interface in BizHawk.Client.Common instead of Bizhawk.Common and I will approve.  But it is a client specific concern.  Common is reserved for completely agnostic code.

YoshiRulz:
I'm using it in `CoreComm`, in the `BizHawk.Emulation.Common` project. Shall I revert that, move the interface, and merge the rest?

adelikat:
Unfortunately, yeah.  At least for now.  It doesn't feel right in CoreComm either, it is such a client ui specific thing, I'd like to see it avoided in CoreComm as well, and keep it to generic callbacks

adelikat (approved):

PR #2586, opened 2021-01-20
Rename Alternate Sync

### LINKED ISSUE #1398: Alternative sync description

TiKevin83:
Alternate sync syncs video frames to audio instead of stretching audio to fit video, so rename it to describe that.

closes #1398


nattthebear:
LGTM.  I think there was an issue tracker item about this?

TiKevin83:
yeah this should help #1398 - I wonder if a tooltip could be added with even more description

vadosnaprimer:
I thought we enabled CI for PRs?

TiKevin83:
might work differently because the PR is from a fork and not a branch?

PR #2589, opened 2021-01-20
Use SharpZipLib for lsmv import instead of hawkfile

nattthebear:
This is the same library and coding style we use daily to load every single savestate and movie, so it can't be that bad

### DIFF
--- a/src/BizHawk.Client.Common/movie/import/LsmvImport.cs
+++ b/src/BizHawk.Client.Common/movie/import/LsmvImport.cs
@@ -6,6 +6,7 @@ using BizHawk.Common.IOExtensions;
 using BizHawk.Emulation.Common;
 using BizHawk.Emulation.Cores;
 using BizHawk.Emulation.Cores.Nintendo.SNES;
+using ICSharpCode.SharpZipLib.Zip;
 
 namespace BizHawk.Client.Common.movie.import
 {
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.Common/movie/import/LsmvImport.cs
```diff
@@ -14,20 +15,25 @@ namespace BizHawk.Client.Common.movie.import
...
+				if (!data.SequenceEqual(Zipheader))
+				{
+					Result.Errors.Add("This is not a zip file.");
+					return;
+				}
 			}
 
+			using var zip = new ZipFile(SourceFile.FullName);
+
+			Result.Movie.HeaderEntries[HeaderKeys.Core] = CoreNames.Bsnes;
```
YoshiRulz, line 36: Does the Core header need to be set before failure?

The rest looks good, though I wish you'd used SharpCompress.

YoshiRulz (approved):

src/BizHawk.Client.Common/movie/import/LsmvImport.cs
nattthebear: I've restored the original ordering of the header setting.

Regarding SharpCompress, this was not an exercise in innovation or imagination; I used the library I did because I simply copied working code from other systems.

PR #2590, opened 2021-01-22
Remove icsharp

nattthebear:
Since .NET 4.5, there's been a good framework alternative, and it gets lots of eyes on it.  I've heard that it's quite fast in .NET Core too.


nattthebear:
I wasn't able to test the NES music ripper, and did only cursory testing of LsmvImport and JMDWriter

adelikat (approved):

PR #2618, opened 2021-02-12
Use builders for FirmwareDatabase

YoshiRulz:
Building upon recent commits to `master`, I've redone `FirmwareDatabase`'s helpers from scratch to be better encapsulated and easier to read at the callsites.

Do others agree that this reads better?
https://github.com/TASVideos/BizHawk/blob/62566da2ae10d4a3e55b698a0e19d6d7f8799458/src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs#L272-L288
Was:
https://github.com/TASVideos/BizHawk/blob/dc3bd050da9212f215ca02b88c3ec07b4ec1db22/src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs#L187-L199

### DIFF
--- a/src/BizHawk.Client.Common/fwmanager/FirmwareManager.cs
+++ b/src/BizHawk.Client.Common/fwmanager/FirmwareManager.cs
@@ -13,6 +13,8 @@ namespace BizHawk.Client.Common
 {
 	public sealed class FirmwareManager
 	{
+		private static FirmwareDatabase DB => FirmwareDatabase.Instance;
+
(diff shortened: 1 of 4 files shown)


adelikat:
I agree it reads better

YoshiRulz:
In its current form, I don't think all this extra complexity is worth it for a slightly more structured init block. `FirmwareDatabase` is immutable enough as-is. I may revisit the scoping function for grouping by system.

PR #2623, opened 2021-02-14
Confine OpenTK 3 to a single project

YoshiRulz:
...as I did SlimDX in #2440. This changeset should make it easier to add OpenTK 4 as an alternative graphics/input method without having as many namespace collisions to deal with.

First I removed what unnecessary references I could, then I copied a few types into our codebase from OpenTK 3 (decompiled). I don't think licensing is an issue because of the decomp, plus I heavily refactored them, with the exception of enums which I assume are from the OpenGL APIs.

edit 2021-03-18: rebased, dropping some commits as per @zeromus suggestion


adelikat (approved):

zeromus:
What do you think I will think about this? I'm not sure what to be looking for in here.

YoshiRulz:
IDK maybe you were planning on using one of the methods I removed. I'll just merge it after testing on Windows.

YoshiRulz:
A lower-level rendering lib for Lua? I don't know if that's actually a good idea, there's always ext. tools if Lua is too slow for a particular use case.

zeromus:
It's a great idea. External tools could use this to draw faster, also.

YoshiRulz:
Can't that stay unimplemented until we have OpenTK 4/5? If you're set on keeping the current interface, could you throw a dummy class on `master` which calls the methods that need to stay? This whole PR is about removing references to the OpenTK assembly, and I may need to rework it all if the first few commits are dropped.

zeromus:
I'm not set on keeping the current interface. Only the current features. If the interface has some dependency on something youre getting rid of, then change the interface. I don't really want to dig into this. We can take an example: why get rid of CreateBlendState
()?

YoshiRulz:
It was either delete the method, or copy 3 more enums into our codebase. I've looked at the commits again and I think there are only 12 types total that I'll need to copy.

zeromus:
yes, copy the enums

PR #2631, opened 2021-02-20
Waterbox build refactorings (round 1 of many)

YoshiRulz:
Nix build stops with:
```
cc emulibc.c
make: /tmp/nix-build-bizhawk-wbox-combined-2.6.1-dev.drv-0/waterbox/sysroot/bin/musl-gcc: No such file or directory
make: *** [../common.mak:53: /tmp/nix-build-bizhawk-wbox-combined-2.6.1-dev.drv-0/waterbox/emulibc/obj/release/emulibc.c.o] Error 127
```
This would be the main problem I've been seeing from the start: reading and writing to the same dir. Unfortunately, the setup for that seems to be entirely within the Makefile, which is opaque to me.

With regards to this changeset:
* Am I using `$LLVMDIR` correctly? It was unused before, the relative paths were hardcoded.
* I changed `libcxx/do-everything.sh` to use short-circuiting (`a; b` -> `a && b`), but I don't think the working directory will be correct if one fails, BUT I don't think THAT matters because of `set -e`?


nattthebear (approved):
Seems good enough.  So long as you'll mop up whatever breakages come from this, have at it

Issue #1895, opened 2020-03-23
Handle RTC through hard reset in Gambatte core


TiKevin83:
### Summary
MrWint's update which added cycle accuracy to RTC didn't handle the corner case of hard resets during games that support RTC

### Repro
TAS Pokemon Crystal and hard reset sometime after setting the time and saving

### Output
The core behaves non-determinstically after hard resetting, tracking RTC from the system clock instead of the cycles of the RTC on the cartridge.

This should be fixable somewhat trivially by both confirming that the new cycle-accurate cart RTC time is appended to the end of .SaveRAM and getting BizHawk to continue using that value through hard reset instead of using the system clock

alyosha-tas:
This should be handled internally by the core, Can the RTC simply not be effected by hard reset?

alyosha-tas:
I think the best solution here is to add a bool here:

`void gambatte::setInitState(SaveState &state, const bool cgb)`

to check if the RTC should be reset or not. BizHawk hard reset calls this upon reset the same as it would upon first loading the core. Which in turn calls the system clock function to set the RTC bas time.

I'm actually not a fan of using the system clock at all, but it works I guess.

TiKevin83:
@alyosha-tas I looked at this one briefly and wasn't understanding what was going on at all, are you able to take a look or expand on your previous note?

alyosha-tas:
I tried (and failed) to fix this about a month ago. I'll give it a fresh look (and leave notes this time.)

alyosha-tas:
`	if (reset_RTC)
	{
		state.time.seconds = 0;
	}`

Making this deterministic seems to be just a matter of doing something similar to the above in initstate.cpp (after adding the bool condition reset_RTC.) However, this is technically not correct as it does not account for the 8 cycles of CPU time that should be taken for the cpu to power back up. In reality this doesn't really matter since the hard reset is not instant anyway, so missing a couple cycles is irrelevant compared to the un-quantifiable time it takes to toggle power.

What really should happen is a reset operation that takes 8 cycles should be added to the cpu so we can do away with the constantly starting from cycle 8 instead of 0 initial state stuff. (this is what GBHawk does already, it turned out much cleaner this way.)

CasualPokePlayer:
I took a stab at this issue and found some potential issues:

[...]

Issue #2423, opened 2020-09-24
[🐧] Core build/rebuild checklist for Linux port


YoshiRulz:
Looking to get these to "built recently" status or higher by 2.5.2:

Cygne (`libbizswan.dll.so`)
* [x] built
* [ ] built recently
* [ ] automated

Gambatte (`libgambatte.dll.so`)
* [x] built
* [ ] built recently
* [ ] automated

Handy (`libbizlynx.dll.so`)
* [x] built
* [ ] built recently
* [ ] automated

mGBA (`libmgba.dll.so`)
* [x] built
* [x] built recently
* [ ] automated

QuickNes (`libquicknes.dll.so`)
* [x] built
* [ ] built recently
* [ ] automated

And now for the moonshots:

Mupen64Plus
* [ ] built
* [ ] built recently
* [ ] automated

Octoshock
* [ ] built
* [ ] built recently
* [ ] automated

Waterbox cores
* [ ] built
* [ ] built recently
* [ ] automated

Sappharad:
Why is Octoshock a moonshot? I had that building fine for macOS in 1.x
Did it get significantly more complicated than 1.x, or is it just harder on Linux for some reason?

nattthebear:
Every waterbox core was built recently, so that can be scratched off (except for the automated part).

YoshiRulz:
Our Octoshock build setup requires VS. In fairness, Gambatte is the same and I shouldn't separate them.

YoshiRulz:
I finished setting up CI for the four unmanaged libraries: [GitLab repo](https://gitlab.com/TASVideos/BizHawk-ports-CI). The binaries I commited in c0b755a33 were the build artifacts from [this Pipeline](https://gitlab.com/TASVideos/BizHawk-ports-CI/-/pipelines/194833466).

InfamousKnight:
I looked at the repo and found you attempted to build melonds? If it helps, here are the commits where suuper attempted to do it himself https://github.com/Arisotura/melonDS/pull/527/commits/0747e459d194e3869a98ce42363948489e23db7a

InfamousKnight:
I tested gambatte, and I don't know if it works because under config-cores-gb i have gambatte enabled but when a rom is loaded gbhawk shows up in the status indicator instead. I can't tell if its gambatte or gbhawk but I do notice a low framerate so it must not be loading gambatte. I was testing Pokémon red

YoshiRulz:
[...]

PR #2649, opened 2021-02-28
Gambatte-Speedrun: Wisdom Tree Mapper Support, Doc Update, Output to Assets, Fix RTC

### LINKED ISSUE #1260: [FEATURE REQUEST] Wisdom Tree unl mapper in Gambatte core support
they used custom 32K banking mapper so does not work in bizhawk

TiKevin83:
fixes #1260 , #1895 
Concerned because 0567623 appears to not completely handle the libgambatte.dll here, I had to move it manually for this PR. Probably need to also update the build script to drop the libgambatte dll there and update the dll import to pull from there if the dll actually needs to be moved to the assets folder.

Edit: I changed the output directory to Assets for this PR


YoshiRulz:
Yes, all unmanaged cores' build scripts will need to be updated. (I hadn't documented that anywhere though, so it's now in #2423.)

---

Did you check in the correct binary? GitHub seems to think it's an "empty file".

nattthebear:
The changes in the submodule seem to be fine as best as I can tell.

TiKevin83:
not sure what happened with the dll "empty file" - going to mark this WIP for a bit, we're working on fixing some outstanding bugs in RTC

TiKevin83:
CPP finished fixing RTC to work through hard reset, core reboot, and the real time RTC feature if anyone wants to take another look before merging.

TiKevin83:
Not sure why the dll says "empty file" but it appears to be building to the correct location now and contains the mapper and RTC updates

PR #1277, opened 2018-08-16
Fix movie header SystemID

Asnivor:
From ticket #1246 

Small change to MovieConversionExtensions.PopulateWithDefaultHeaderValues()

Just changes SystemID to "GG" or "SG" if Global.Emulator .IsGameGear or IsSG1000 is true.

Without this, the play movie dialog shows 'Platform' as SMS for game gear and SG1000 movies (and highlights this in RED as it is incorrect).

Issue #1246, opened 2018-07-22
Incorrect platform label in header file for Sonic Labyrinth Movies


MemoryTAS:
In all current Bizhawk versions, if you record a movie file for Sonic Labyrinth, the header file will list the platform as SMS, when the game is in fact Game Gear. When you go to Play Movie and look at the header information, SMS will be marked in red because it is incorrect. I'm able to reproduce this in both 1.13.2 and 2.3. This may affect other Game Gear games as well but I am not certain.

Asnivor:
PR #1277 

@MemoryTAS This actually affects all GG (and I think SG) games.

vadosnaprimer:
It doesn't affect anything. For whatever reason, when system info is dumped to the header, for GG it's written as SMS, with IsGameGear flag set. Similarly for SG1000 - SMS and IsSG1000 flag. But when these are looked up in the DB, their platform is GG and SG respectively, while the header has SMS. They are loaded just fine, and even the emuhawk caption tells the system correctly. Just this tiny check in PlayMovie is wrong.

vadosnaprimer:
Here's the proper fix that I'm not applying, because there's no official decision to dump GG/SG to movie header.
[proper-gg-sg-fix.zip](https://github.com/TASVideos/BizHawk/files/2406890/proper-gg-sg-fix.zip)

PR #2650, opened 2021-02-28
Fix SMSHawk's system ID

YoshiRulz:
see #1246, #1277, #1324, 57fd6051e, and https://github.com/TASVideos/BizHawk/commit/57fd6051e957a2fa4c33a63c5d06bf07b9bac321#commitcomment-47540438

I'm setting `SMS.SystemId` (`IEmulator` impl.) *after* this compat mode check. Should it be before?
https://github.com/TASVideos/BizHawk/blob/a529b7d4ae07631da879022eafd8374b36c9559e/src/BizHawk.Emulation.Cores/Consoles/Sega/SMS/SMS.cs#L94-L104


adelikat (approved):
This doesn't interfere with adding flags such as IsGGMode into movies, so it will not break the tasvideos site parser, so this is fine

Issue #1810, opened 2020-01-24
[BSNES core] crash when load state after restart


g0me3:
I'm not sure if this is because of some specific settings combinations, or I have some problems here with my pc, but

when you play snes game with BSNES core (the same lawnmower again lol), the save states works just fine as soon as you does not close the emulator.

when you close it and then reopen the game again and try to reload the last save, emu crashes silently.

zeromus:
Worked fine for me.
Are you sure youre not loading a savestate taken before the waterbox core was updated? That should trigger an error message, but maybe you're not getting exception error messages for some reason.
Can you try debugging emuhawk in VS and finding out what's going on?
@nattthebear did this ever happen to you? Sorry to nameflash you but if there's something wrong with my waterbox builds we had better find out ASAP

g0me3:
savestates are created just now with newest build.

g0me3:
sorry, I tried to build latest version of bizhawk but I failed lol build tool crashes, sln project does not load with my 2017. i give up.

YoshiRulz:
You need VS 2019 now, changed about 3 months ago.

g0me3:
ok, i did it after all. problem occurs at BSNES core somewhat.
as I checked carefully, if I use my own compiled version, all is fine actually, all states are load no matter when.
however, those states that were produced by one of previous appveyor builds does not load and cause crash even in my own build. so there are some save corruption were happens with a particular build but not in my current fresh build...

crash appears somewhere inside external compiled code (waterbox for BSNES i presume).

if you want to dig it for some reason, here are states. state2 is good one, state1 is a bad one.
[State.zip](https://github.com/TASVideos/BizHawk/files/4139146/State.zip)

g0me3:
happens to me again, see more corrupted savestates.
First I was thought it is because using cheats, but European version of clay fighter crashes the same way without any cheats.
[State.zip](https://github.com/TASVideos/BizHawk/files/4142737/State.zip)

under debug crash occurs somewhat outside the main emu code, so it seems it is inside bsnes waterbox core as I said before...

can anybody confirm if my saves crashes on other pcs? also, note, i use goodsnes romset with gamedb disabled in emulator.

g0me3:
something has changed. another time got this crash message

`System.InvalidOperationException: Memory consistency check failed.  Is this savestate from different SyncSettings?
[...]

PR #2662, opened 2021-03-12
Probably fix SNES core crash on load with CDL enabled, or formerly enabled

nattthebear:
The crash @g0me3 sent me was very simple to debug:  `cdlInfo.blocks` contains absolute garbage.  It appears that these are foreign pointers brought into the core, in which case they must be refangled on every state load.


nattthebear:
```
(gdb) p ((void*[23])cdlInfo)
$6 = {0x4, 0x207a890aca8, 0x207a8b0ace0, 0x207a8d0ad18, 0x0, 0x207a7b46b98, 0x20799b9db68, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
  0x0, 0x0, 0x0, 0x20000000200000, 0x400000, 0x1000000020000, 0x0, 0x0, 0x0}
```
(In the current session, 0x0207:----:---- points to unallocated space.)

nattthebear:
With this commit, I can continue play after loading the savestate in #1810 

image

zeromus:
THANK YOU SO MUCH!!!!!!!! WHAT A GREAT MYSTERY SOLVED!!

g0me3:
RRRR-EEAAT!

PR #2672, opened 2021-03-23
Waterbox build refactorings (round 2 of many)

YoshiRulz:
I figured using patches to add the `waterbox` arch to musl would be easier to maintain, as comparing our fork to upstream shows a whole lot of *added* files, but only a few *changes*.

After copying the upstream source and the added files into a new dir `src-combined`, I'm using a `sed` call and a patch file for said changes:
https://github.com/TASVideos/BizHawk/blob/9e03991c2957d42dcfe540cf56b3ddf5cc22babd/waterbox/musl/do-local-build.sh#L14-L19
The build then proceeds as before in the `src-combined` dir.

Nix derivation isn't useful yet, but it's nice to have something concrete showing my progress on that front.

There's an unmerged branch [`threading`](https://github.com/nattthebear/musl/commits/threading) on the fork which is 2 ahead, 9 behind the main branch—is that unfinished work or was it recreated on the main branch already?


zeromus:
What's the rationale here? We already have musl in a fork. We can manually manage merges at our leisure.

YoshiRulz:
First line of OP:

> I figured using patches to add the `waterbox` arch to musl would be easier to maintain, as comparing our fork to upstream shows a whole lot of *added* files, but only a few *changes*.

It's the "no code is good code" idea again. Why have a repo full of code that's copied verbatim from elsewhere?

zeromus:
That's how it's done, man. That's _the best case scenario_. It isn't easier to maintain. You are basically recreating git by setting up a bunch of patches and a patching apparatus in the build steps. Stop.

vadosnaprimer:
Yeah git is supposed to solve the need for patches by how it manages branches and forks.

nattthebear:
> as comparing our fork to upstream shows a whole lot of added files, but only a few changes.

Yes, I did that to keep it easy to reconcile upstream changes when they happen, not harder.

> I figured using patches to add the waterbox arch to musl would be easier to maintain

Strange that you didn't submit this PR as a patch serial, hmmm?  If that's easier than using git...


I personally find git infinitely easier to use than futzing around with patch serials.  So much so that when I made the nyma project, I put unmodified mednafen in a git repo, and then to take later mednafen updates, I updated that "fake upstream" branch and then three way merged it with our fork.  And you know what?  It worked!  I've been able to keep up with changes so far, and I don't regret that I had to make a whole extra repo to do it.

This situation is even more favorable to git, as the official upstream for musl is a git repo.

Not just anyone can integrate upstream changes here -- the musl innards are some of the most technical parts of waterbox guest code, as they have to handle all the custom syscall stuff.  So long as I'm doing it, I'll be gitting it.

PR #2680, opened 2021-03-26
N64 Rumble

### LINKED ISSUE #1756: [F.Rq.] Pass gamepad rumble/haptics through to host (Mupen/Octoshock)
Hello there. I've been using this emulator to test some games, and I realized that while there's an option for the controller pak to be set to Rumble Pak, there's no rumble on the controller whatsoever. I'm playing on Windows 10 64-bit, and I'm using an Xbox One controller through USB. The games don't seem to detect the Rumble Pak as connected either, even if the option is checked. 

Could an option for configuring it be added? Mostly, because Mupen64plus appears to support rumble if the pak is configured to be Rumble Pak.

Meerkov:
Now N64 games support rumble correctly. Works in both DirectInput and OpenTK.

**Known Issues**
Pausing the game while the vibration is on does not stop the vibration. However, if the user doesn't want this, they can simply not play with a RumblePak.

partially resolves #1756


libmupen64plus/mupen64plus-input-bkm/mupen64plus-input-bkm/mupen64plus-input-bkm.vcxproj
```diff
@@ -72,15 +73,15 @@
...
+    <OutDir>$(SolutionDir)..\..\..\output\dll\</OutDir>
     <PostBuildEventUseInBuild>true</PostBuildEventUseInBuild>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <OutDir>$(SolutionDir)..\..\..\..\output\dll\</OutDir>
     <PostBuildEventUseInBuild>true</PostBuildEventUseInBuild>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <OutDir>$(SolutionDir)..\..\..\..\output\dll\</OutDir>
+    <OutDir>$(SolutionDir)..\..\..\output\dll\</OutDir>
```
YoshiRulz, line 84: Should be `Assets\dll`—and please change this across the rest of Mupen too.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/NativeApi/mupen64plusInputApi.cs
```diff
@@ -70,14 +70,12 @@ public mupen64plusInputApi(mupen64plusApi core)
...
 
-			mupen64plusApi.m64p_error result;
 			InpSetInputCallback = GetInputDelegate<SetInputCallback>("SetInputCallback");
 			InpSetRumbleCallback = GetInputDelegate<SetRumbleCallback>("SetRumbleCallback");
 			InpSetControllerPakType = GetInputDelegate<SetControllerPakType>("SetControllerPakType");
 			InpSetControllerConnected = GetInputDelegate<SetControllerConnected>("SetControllerConnected");
 
-			m64pRumbleCallback = new RumbleCallback(FireOnRumbleChange);
-			result = InpSetRumbleCallback(m64pRumbleCallback);
+			SetM64RumbleCallback(new RumbleCallback(FireOnRumbleChange));
```
YoshiRulz, line 78: I didn't do this (712dbc041) in the old branch, what's its purpose?

src/BizHawk.Client.Common/input/HostInputAdapter.cs
```diff
@@ -20,6 +20,8 @@ public interface IHostInputAdapter
 
 		IEnumerable<KeyEvent> ProcessHostKeyboards();
 
+		void ProcessVibrationAll(IReadOnlyCollection<(string name, int strength)> haptics);
```
YoshiRulz, line 23: The way you've done this means the code is duplicated between OpenTK and DirectInput. (It also seems to merge all channels into one, sending the result to all gamepads?) Try adding a parameter to `IHostInputAdapter.ProcessHostGamepads` instead.

YoshiRulz (changes requested):

src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/NativeApi/mupen64plusInputApi.cs
Meerkov: I think it's cleaner this way.

libmupen64plus/mupen64plus-input-bkm/mupen64plus-input-bkm/mupen64plus-input-bkm.vcxproj
Meerkov: This change will only fix this file. Others are out of scope for this pull.

src/BizHawk.Client.Common/input/HostInputAdapter.cs
Meerkov: While I agree that some code is duplicated, unfortunately, that's simply how these two classes are setup.

DirectInput and OpenTK duplicate code for Buttons and Axis controls too.

Meerkov:
Output folder updated, please take another look.

YoshiRulz:
One thing at a time—please test w/ OpenTK on branch [`rumble-202103`](https://github.com/TASVideos/BizHawk/tree/rumble-202103), and open a new PR *against that branch* adding DirectInput support. I'm not actually getting anything out of my gamepad with OpenTK, on either Windows or Linux.

Meerkov:
That branch changes the layout of the code, which is fine. But please consult this change for the _logic_ used. Open TK doesn't take INT_MAX, it takes a float from 0.0 to 1.0.

Meerkov:
https://github.com/TASVideos/BizHawk/pull/2680/commits/01cb240cece381d56d92a86ae5753043130ad361

Please check your branch against the full changes.

You don't have this step:

// Convert Int32 to 0.0 to 1.0 range float.
					pad.SetVibration((float)leftStrength / Int32.MaxValue, 
									 (float)rightStrength / Int32.MaxValue);

YoshiRulz:
That's [right here](https://github.com/TASVideos/BizHawk/commit/087c7903314da427b7651d30c20c7094b422be5a#diff-259490b469fc51dbff93d25ef58b866180ed59788c2ce50f5fa9b033f5eff266R380-R381). I didn't like that the int->float conversion wasn't at the end of the call chain, so I moved it.

Meerkov:
Ah yeah, I see it. I'm not sure why your branch doesn't work. Were you able to get vibration from my branch? Yours has a bug, and debugging this again seems backwards. I think it makes more sense to take the working feature and re-organize it, than to start from scratch.

Meerkov:
Did you remember to actually fix the plugin code and recompile the dll? If you didn't do that, it doesn't matter what changes you make to the bizhawk side.

YoshiRulz:
> Were you able to get vibration from my branch?

no

> Did you remember to actually fix the plugin code and recompile the dll?

Don't worry about Mupen for now, I'm focused on getting working haptic feedback. My hack for debugging is to hold `Tab` to fake rumble. If you put a print [here](https://github.com/TASVideos/BizHawk/blob/087c7903314da427b7651d30c20c7094b422be5a/src/BizHawk.Client.EmuHawk/MainForm.cs#L701-L702) you should see that working, and if you put a print/breakpoint [here](https://github.com/TASVideos/BizHawk/blob/087c7903314da427b7651d30c20c7094b422be5a/src/BizHawk.Bizware.OpenTK3/OpenTKInputAdapter.cs#L32-L33) to look at `hapticsDict` you should see it has the correct values.

Meerkov:
OpenTK was discovered to not have any vibration implementation. So this only works for DirectInput, but something is better than nothing.

Meerkov:
I opened a new PR, which hopefully addresses concerns around the number of commits.

PR #2683, opened 2021-03-28
N64 Vibration for DirectInput

### LINKED ISSUE #1756: [F.Rq.] Pass gamepad rumble/haptics through to host (Mupen/Octoshock)

Meerkov:
This commit is cleaner than the previous attempt. This makes vibration work for N64 and DirectInput. I think this means that Linux users are out of luck for now.

Known Issues:
The controller will continue to rumble while paused.
The controller will rumble even if a memory card or no peripheral is set on the controller (is it a bug or a feature?)

---

partially resolves #1756


YoshiRulz:
I ended up committing most of this to `master` myself. `master` now has haptic feedback in a working state, albeit with the Fast Forward hack, and branch [`mupen-rumble-finally`](https://github.com/TASVideos/BizHawk/tree/mupen-rumble-finally) has some of the required changes to Mupen. Please checkout that branch, commit the remaining changes to Mupen, and `git push -f  mupen-rumble-finally:rumble2` to re-use this PR.

Meerkov:
:/ It seems like you want to submit the changes yourself instead of accepting my PR. Please don't stop half way.

YoshiRulz:
bump

adelikat:
@YoshiRulz : please fix all the way, or undo your changes.  We want this contribution and you've blocked the ability for anyone else to complete this feature.  Not cool.

YoshiRulz:
Merged and pushed an additional commit to effectively revert all the non-Mupen changes. Query whether the binary:
* works
* reflects the changeset (the only changes were to `<OutputPath/>` so it should be good)
* is built in Release and not Debug configuration

GhostlyDark:
#2680 works after setting a Rumble Pak in the Controller Settings. This PR doesn't. Any thoughts on why?

YoshiRulz:
No idea, but we'd welcome any contribution. I've updated branch [`mupen-rumble-finally`](https://github.com/TASVideos/BizHawk/tree/mupen-rumble-finally) for you.

Meerkov:
https://github.com/TASVideos/BizHawk/pull/2683#issuecomment-879934288
My original suggestion was that we should just merge my PR that I tested pretty thoroughly...

I'll try to take one stab at what the bug might be:
It is outputting to the ../../../../Assets/dll folder (4 levels up) but mine outputs to ../../../Assets/dll (3 levels up). Which would mean that the correct dll folder isn't actually getting the updated version with the bugfix.

If that doesn't fix it, then honestly, the correct engineering solution is to roll all of Yoshi's changes back, and resubmit the working version from my PR. Not rolling this back seems to be delaying this feature by several weeks, and as far as I understand, the changes Yoshi made with regards to refactoring the N64 controller were not necessary.

GhostlyDark:
> [#2683 (comment)](https://github.com/TASVideos/BizHawk/pull/2683#issuecomment-879934288)
> My original suggestion was that we should just merge my PR that I tested pretty thoroughly...
> 
> I'll try to take one stab at what the bug might be:
> It is outputting to the ../../../../Assets/dll folder (4 levels up) but mine outputs to ../../../Assets/dll (3 levels up). Which would mean that the correct dll folder isn't actually getting the updated version with the bugfix.
> 
> If that doesn't fix it, then honestly, the correct engineering solution is to roll all of Yoshi's changes back, and resubmit the working version from my original PR. Not rolling this back seems to be delaying this feature by several weeks, and as far as I understand, the changes Yoshi made with regards to refactoring the N64 controller were not necessary.

It is true that the output directories are wrong, but fixing them does nothing to get Rumble working. I still fail to get any feedback with this PR while the previous approach works right out of the box and without issues.

Meerkov:
Ok thanks for checking and confirming that is a bug... Not sure what else might be wrong with Yoshi's changes here. It's not good practice to submit broken features, because now I have to debug these other changes before I can submit the working one...

I think the path forward is to rollback 373d5e0 421ccd5 and 4c5e5b5 and other haptics related changes going back to ~Mar 29 of affected files? Then you can roll forward your changes on top of the working version.

YoshiRulz:
I can't simply undo or remove those commits to get back to the same state as when the PR was opened, the commit tree has changed. As of recently, I should be able to build Mupen, so I'll take another look at this changeset.

> I still fail to get any feedback with this PR while the previous approach works right out of the box and without issues.

To be clear: it's not meant to work out-of-the-box. You need to enable the Rumble Pak in the gamepad sync settings, then bind the feedback channel to a host gamepad.

GhostlyDark:
> To be clear: it's not meant to work out-of-the-box. You need to enable the Rumble Pak in the gamepad sync settings, then bind the feedback channel to a host gamepad.

The working PR requires one to enable the Rumble Pak, which makes perfect sense.
I don't know why you'd need to bind the feedback to a gamepad first. It doesn't sound user friendly, nor does any other emulator that I know require such a thing. Regardless, binding feedback with this PR while having the Rumble Pak enabled still does absolutely nothing to enable vibration.

PR #2686, opened 2021-03-30
Fix some TAStudio branch view issues

### LINKED ISSUE #2645: Save state doesn't work properly in TAStudios after 2.4.2
Repro

1 -  Open up a rom
2 -  Open TAStudios
3 -  Any point in the game use savestate more than one (ex: "Shift+F1" and "Shift+F2")
4 -  Use F1 to reload your branch 1
5 - Now if you wanna save or reload your branch 2 with "Shift+F2" or "F2" it saves or reload from branch 1

This problem does not exist in 2.4.2.

Morilli:
- rightclicking a branch when the contextmenu was already open would throw an exception and not work properly (I don't understand how 764f4ad81e744ce2b1949c6e97a46c0cb40debc8 tried to fix this very issue and somehow failed)
- branch view will now scroll to the updated branch when using hotkeys to update a branch as I feel like that makes sense
- fix #2645 (was broken in a2414f2b4ea41c18ad52b29e13f1a363b4ff425f)

### DIFF
--- a/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/InputRoll.cs
+++ b/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/InputRoll.cs
@@ -317,7 +317,7 @@ namespace BizHawk.Client.EmuHawk
 		public bool FullRowSelect { get; set; }
 
 		/// <summary>
-		/// Gets or sets a value indicating whether multiple items can to be selected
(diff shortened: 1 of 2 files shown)


Morilli:
You're right, I missed the `LoadBranch` thing. I've made it behave the same as `UpdateBranch` now, in that it deselects all but the current branch before loading / updating it to avoid conflicts with the current logic logic that relies on the first selected branch.

Issue #2640, opened 2021-02-24
TAStudios not saving branches on projects


lunjesse:
[//]: # "This issue description supports Markdown syntax (this is what comments look like). There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "You can leave these comments here or delete them. Also, please remember to search for similar issues before writing anything, including in closed issues!"
[//]: # "One more thing: if you're on Linux, please comment on #1430 instead of opening an issue so we don't annoy the other devs."

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
Making a branch in TAStudios does not get saved to the tasproj file, so it disappears on reload.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Open up a ROM (I used GBA)
2. Open TAStudios
3. Add a new branch by clicking on the green + button
4. Save the project.
5. Reopen the project
Now the branch you made in step 3 is gone.

### Host env.
[//]: # "List the computers you've found the bug with. If there's a version that doesn't have the bug, please put that in too. Here are some examples:"
- BizHawk 2.6.1; Win10 Home; Intel

[//]: # "(screenshots, if applicable)"

[//]: # "That's it! If you'd like to help more, you could try a dev build (see Testing in the readme) or an older release. Click submit now and you can edit it later."

vadosnaprimer:
If you open your .tasproj in 7zip, do you see the Branches folder?

Zimond47:
> If you open your .tasproj in 7zip, do you see the Branches folder?

Yes, all branches are saved in the branches folder.

YoshiRulz:
@Zimond47 The question was directed at OP specifically. Unless that's your alt account?

Zimond47:
> @Zimond47 The question was directed at OP specifically. Unless that's your alt account?

Uhm, no. sorry. I just thought I can answer this to help.

YoshiRulz:
@ThunderAxe31 reproduced this. Does anyone know if 2.6 is affected or if it was only broken in the interim and 2.6.1?

vadosnaprimer:
Bisected! It's caused by adca19c30a816c1627a7c436cc8776a918ba5f36

[...]

PR #2689, opened 2021-03-31
Support branches in all "old" .tasproj files

Morilli:
a91fa2754e21b0b2323bd06593ebcb57f24468e0 broke all old .tasproj files in that all branch info was not loaded from them. I don't think that is a wise choice? That's just asking for another #2640.

Some more info in hopes that it's helpful: all "old" tasproj files were saved with `\` as path separator, which was changed to `/` in the aforementioned commit. Without my fix here, the new logic will fail to understand `\` as path separator and ignore all branch etc. info in .tasproj files.

### DIFF
--- a/src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
+++ b/src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
@@ -62,7 +62,7 @@ namespace BizHawk.Client.Common
 					name = name.Substring(0, i);
 				}
 
-				_entriesByName.Add(name, z);
+				_entriesByName.Add(name.Replace('\\', '/'), z);
 			}
 		}
 


nattthebear:
The ZIP spec requires `/` as the path separator.  There is no such thing as an alternate path separator.  I'll start digging around into what various ZIP libraries we are and were using do when fed with malformed filenames / malformed zipfiles when I have a chance, but I don't want another hack in the code if at all avoidable.

adelikat:
How about a small hack until you find something better?

nattthebear:
Something is deeply wrong here, because as mentioned before there is no such thing as a legal zip file with a `\` path separator in it.  I need to take zip files, input and output, with various old versions of bizhawk, and get them into hex editors and see what's going on.  Because we also support reading and writing our zip files with normal zip file editing tools as a regular user scenario, I need to look at what happens in those cases as well.

I advise against releasing without understanding what the problem is here.

Morilli:
As far as I understand it, this matter is pretty simple. All currently creates .tasproj (zip format) files use `\` as "path separator". Whether this zip is technically invalid or not doesn't really matter much, as all zip programs support the backslash. 

In regards to its invalidity I can see why you would change that to actual valid zip files using `/` only, but no longer supporting all old .tasproj files is wrong imo. I don't know how SharpZipLib handles this matter, but from my experience with C#'s native zip library, they absolutely do not care about the separator; aka they don't convert or normalize at any point, which results in ``string name = z.FullName;`` potentially containing forward slashes, backward slashes or even both, depending on how the zip file was set up.

vadosnaprimer:
It's common that people can't open .tasproj files with regular archivers too, so maybe this invalid path is the reason? 7zip opens them fine, but even that had some old versions that failed.

nattthebear:
>Whether this zip is technically invalid or not doesn't really matter much, as all zip programs support the backslash.

Not with an attitude like that it doesn't.  I'll trust feos's evidence here.

YoshiRulz:
Just confirming `.tasproj` files have caused no problems with libarchive-based programs on Linux either.

PR #2693, opened 2021-03-31
Add initial marker in frame 0 when loading older tasprojs

Morilli:
Before this would load old "incompatible" .tasprojs without any marker, making the first created marker take slot 1 and be uneditable and undeletable, seemingly without reason.
I have looked a bit into the discussion around this (#2023), but I think a quick fix is better for now.

Not sure if this line of code should be duplicated in the code as often as it is now (3 times total), so I'm open for suggestions on whether this could be handled better.


adelikat:
Will merge for now.  In theory this shouldn't be replicated 3 times before probably best is a better solution like #2023 suggests.  So I'm not too worried about the duplicated code

PR #2697, opened 2021-04-02
Add version metadata to all assemblies (read from VersionInfo by build script)

### LINKED ISSUE #2696: Consider re-adding accurate AssemblyVersion or AssemblyFileVersion information to generated binaries
Currently, the `EmuHawk.exe` and `Discohawk.exe` display their version like this when viewing detail in windows explorer, which isn't exactly descriptive of their actual version.
image
Previously, this would (more or less lol) show their actual version, like here for `2.4.0`:
image

This affects all versions from 2.5+ (after ebd9645c1e7dbb55dc725014415b783115e4e7e8). I would consider finding a way to re-add this information here to make it easier to view Bizhawk's actual version without having to open it and look in the `About`  section to find the required info.

I tried to experiment with the code logic a bit and haven't found a good way that didn't break anything yet, so I'm just throwing this out here in form of an issue.

Morilli:
A working solution that I found that adds version information to the `EmuHawk.exe` and `DiscoHawk.exe`.

Unsure about the verbosity of the info currently, should definitely be reduced for release builds (probably just keeping `VersionInfo.MainVersion`) and potentially a bit even for debug builds (I'm not sure what exactly the `SVN_REV` is even referring to).

I may be missing something; I'm not really too well versed with the entire build system and stuff.

- will close #2696

### DIFF
--- a/Dist/.BuildInConfigX.sh
+++ b/Dist/.BuildInConfigX.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+set -e
+if [ ! -e "BizHawk.sln" ]; then
+	printf "wrong cwd (ran manually)? exiting\n"
+	exit 1
+fi
+config="$1"
+shift
+Dist/.InvokeCLIOnMainSln.sh "build" "$config" "$@"
--- a/Dist/.BuildTestInConfigX.sh
+++ b/Dist/.BuildTestInConfigX.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+if [ ! -e "BizHawk.sln" ]; then
+	printf "wrong cwd (ran manually)? exiting\n"
+	exit 1
+fi
+config="$1"
+shift
+Dist/.InvokeCLIOnMainSln.sh "test" "$config" -a . -l "junit;LogFilePath=$PWD/test_output/{assembly}.coverage.xml;MethodFormat=Class;FailureBodyFormat=Verbose" "$@"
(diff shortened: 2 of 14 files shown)


YoshiRulz (changes requested):
This is a thin layer of automation over the "old way", which I'm very much against. The [blog post from Cezary Piątek](https://cezarypiatek.github.io/post/setting-assembly-and-package-metadata/) that I linked in #2696 describes the "new way" and should be detailed enough for you to get it working after a bit of experimentation.

Morilli:
Hmm, if I'm seeing it correctly the only ways to set those version attributes are to 
1. pass them in the dotnet build command, or
2. set them as environment variables,

I assume I want to set the appropriate variables in the dotnet build command, but how do I do that in a way that both commandline (script) builds and visual studio builds have them set correctly? 

Also, what's the current workflow for creating a release package? I would have assumed `Dist/QuickTestBuildAndPackage_Release.bat`, but the `DeveloperBuild` replacement commands are completely outdated in there.

YoshiRulz:
TBH you really don't need to bother with local builds or AppVeyor, only GitLab CI. (Artifacts for releases are taken from there.) The analogues for `Dist/QuickTestBuildAndPackage*.bat` are `Dist/*Build*.sh`. I committed 84e195659 so the `dotnet` invocations from the Linux build scripts are centralised in one file, which hopefully saves you some copy-pasting.

Morilli:
How about this? Haven't cared about adding the version info to the .bat files; the .sh scripts seem better anyway

PR #2700, opened 2021-04-05
Add method to IEmuClientApi for getting approximate framerate

YoshiRulz:
(and a delegated Lua function too)

I spent no time on this so I probably overlooked something


adelikat:
Can you please spend some time on this then? You are touching the main frame loop and admitting you didn't test, why should I merge this?

PR #2706, opened 2021-04-08
Add rgbds syntax for GB/C disassembly

CasualPokePlayer:
rgbds is the main toolchain for use in the gbdev community and one of the few currently maintained gb/c toolchains. It's also the most used toolchain in use for various game disassemblies. Thus in my opinion, the tracelogger should use its syntax too.


alyosha-tas:
I disagree on this one. I stare at a lot of trace logs and much prefer the style we have now, especially the caps.

adelikat:
There's a lot to be said for following standards

CasualPokePlayer:
Apologies on breaking standards here; I admit some changes were stylistic changes I made. Mostly done to follow common stylistic usage within disassemblies and the gbdev community in general. And outright replacing it wasn't my original intent, I first intended to make it options (but then I got lazy so I did this mostly to see how rgbds syntax would be reacted to). Would it be viable to make these stylistic (e.g. uppercase/lowercase) and syntax (e.g. # vs $, () vs []) changes options within the tracelogger?

vadosnaprimer:
It can become an option in SyncSettings of the core, since the core generates disasm, not tracelogger.

nattthebear:
>SyncSettings

Shirley this could be a non-sync setting?

vadosnaprimer:
Yeah sorry that's what I meant.

adelikat:
@CasualPokePlayer , Oh, I meant, there's something to be said for _changing_ BizHawk to conform to a community standard

CasualPokePlayer:
> @CasualPokePlayer , Oh, I meant, there's something to be said for _changing_ BizHawk to conform to a community standard
Could you elaborate on that? Is it just due to this PR replacing the syntax (which I am open to making it an option) or is it just any change at all to the current syntax (at the very least anyways, that JP (HL) is bad since that opcode does not do any dereferencing and should be replaced with JP HL anyways)

nattthebear:
I personally am partial to assemblers:  If the syntax matches that of a toolchain that can actually compile code for the target, that's worth something.

At the same time there's one person going deep in tracelogs of this core all day who is well served by the current syntax, so I can't complain too strongly for something I don't actually use.  Maybe both syntaxes is the way forward.

Would it make sense to add syntax as an option on our debugger interfaces?

alyosha-tas:
Fixing JP HL is of course fine, and if you can add an option for the syntax sure go for it,

CasualPokePlayer:
I've gotten RGBDS syntax to be a non-sync setting, although I've only gotten it to work with Gambatte's tracelogger. I'm not sure how to send that sync setting through for lua/GBHawk's tracelogger.

YoshiRulz:
[I think I got it](https://github.com/YoshiRulz/BizHawk_BranchBackup/tree/tracelogger-rgbds-syntax)

CasualPokePlayer:
Thanks!

alyosha-tas:
I can't really test it right now but it looks ok. I'd prefer that the setting not be passed through ExecuteOne() but I can always change that later once I can focus on it again.

CasualPokePlayer:
I've added in a fix to some opcodes that were not disassembled at all correctly, seems like somebody didn't realize these opcodes used sp as the base rather than pc, and didn't realize putting in the direct address calculated doesn't make sense for explicit adds (jr is more implicit with its addition, so it still makes sense to put in the direct address rather than the actual value used for the addition). Figured it should be in this PR too so that is handled along with the other disassembler changes.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/GBHawk.cs
```diff
@@ -213,13 +215,13 @@ public GBHawk(CoreComm comm, GameInfo game, byte[] rom, /*string gameDbFn,*/ GBS
 			ser.Register<ISoundProvider>(audio);
 			ServiceProvider = ser;
 
-			_settings = (GBSettings)settings ?? new GBSettings();
+			_ = PutSettings(settings ?? new GBSettings());
```
nattthebear, line 218: Why is this being changed?

YoshiRulz: `PutSettings` updates the dissassembler.

TiKevin83:
Since this is enabled as an option and not replacing the syntax and it looks like discussion has died down I think we're good to go ahead and merge

PR #2719, opened 2021-04-18
Simplify IToolForm and remove winform specific notions

adelikat:
replace some winforms specific values and ram watch chacks with IToolForm properties.  Removes hacks, and some winforms dependencies in tools, and allows for easier implementation of closeable tools

Note: this is a breaking change to existing external tools


YoshiRulz (approved):
New API looks sound.

PR #2732, opened 2021-05-07
[Gambatte] Add User-Defined Frame Length Type For Subframe TASing

CasualPokePlayer:
Doesn't create a new core for sub-frame input, rather it just sets it as a sync setting. Gambatte is perfectly capable of doing sub-frame stuff already, all it needed was some frontend work.

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IEmulator.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IEmulator.cs
@@ -9,16 +9,15 @@ namespace BizHawk.Emulation.Cores.Nintendo.Gameboy
 	{
 		public IEmulatorServiceProvider ServiceProvider { get; }
 
-		public ControllerDefinition ControllerDefinition => ((int)_syncSettings.FrameLength == 2) ? SubGbController : GbController;
+		public ControllerDefinition ControllerDefinition => (_syncSettings.FrameLength == GambatteSyncSettings.FrameLengthType.UserDefinedFrames) ? SubGbController : GbController;
 
 		public bool FrameAdvance(IController controller, bool render, bool rendersound)
 		{
 			FrameAdvancePrep(controller);
 			uint samplesEmitted;
-			switch ((int)_syncSettings.FrameLength)
+			switch (_syncSettings.FrameLength)
 			{
-				// VBlank Driven Frames
-				case 0:
+				case GambatteSyncSettings.FrameLengthType.VBlankDrivenFrames:
 					// target number of samples to emit: always 59.7fps
 					// runfor() always ends after creating a video frame, so sync-up is guaranteed
 					// when the display has been off, some frames can be markedly shorter than expected
@@ -35,8 +34,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.Gameboy
 						ProcessSound((int)samplesEmitted);
 					}
 					break;
-				// Equal Length Frames
-				case 1:
+				case GambatteSyncSettings.FrameLengthType.EqualLengthFrames:
 					while (true)
 					{
 						// target number of samples to emit: length of 1 frame minus whatever overflow
@@ -63,8 +61,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.Gameboy
 						}
 					}
 					break;
-				// User Defined Frames
-				case 2:
+				case GambatteSyncSettings.FrameLengthType.UserDefinedFrames:
 					while (true)
 					{
 						// target number of samples to emit: input length minus whatever overflow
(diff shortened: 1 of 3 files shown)


nattthebear:
As I've stated before, Bizhawk is fundamentally unready for this.  This pull request must be rejected.  The SubGbHawk core must be removed from Bizhawk.

CasualPokePlayer:
By this token, you might as well just remove Gambatte. It already does some sub-frame shenanigans by default (non Equal Length Frames), and the Equal Length Frames code is largely what I used to even do this truer sub-frame functionality (quite literally it's mostly copy paste, the Equal Length Frames code is just as hacky as my sub-frame code). Unless it's just a problem of using an axis to define input length? Even still, that seems rather odd as a reason to call this fundamentally unready and for the removal of SubGBHawk.

YoshiRulz:
No-one will be removing cores just because the code is ugly.

This small changeset brings Gambatte closer to feature-parity with Sub/GBHawk, so I don't see why it shouldn't be merged if its movies dump correctly. @CasualPokePlayer Can you and @TiKevin83 coordinate a resync and verification of a SubGBHawk movie (maybe [this one](http://tasvideos.org/4329M.html)) using Gambatte with this new subframe mode?

nattthebear:
> By this token, you might as well just remove Gambatte. It already does some sub-frame shenanigans by default (non Equal Length Frames), and the Equal Length Frames code is largely what I used to even do this truer sub-frame functionality (quite literally it's mostly copy paste, the Equal Length Frames code is just as hacky as my sub-frame code). 

Agreed, yes.  Note that it's not the frame length code in Gambatte that's particularly bad; it's that the core has to provide these various frame length ideas because the frontend has no idea what to do with variable timing data.

nattthebear:
It depends on the AVsync method; if you sync to video then the dumper ASS-umes that every video frame corresponds to exactly CoreComm.Framerate samples of audio, and will squish your audio to match.

Almost all of our cores are actually VFR, and can output varying counts of samples, so sync to video is mostly useless.

CasualPokePlayer:
> It depends on the AVsync method; if you sync to video then the dumper ASS-umes that every video frame corresponds to exactly CoreComm.Framerate samples of audio, and will squish your audio to match.
> Almost all of our cores are actually VFR, and can output varying counts of samples, so sync to video is mostly useless.

I used Sync to Audio, so that probably explains it.

CasualPokePlayer:
I converted my geg Gold movie over to Gambatte: https://cdn.discordapp.com/attachments/583805335660265474/840316251589836860/pokesilver_geg_1_1_1.bk2

And it was able to be console verified: https://youtu.be/-IZIN2x5RNE

YoshiRulz:
@adelikat It's going to be *such a pain* to split `Gameboy` into two classes because the type parameters on `ISettable` are invariant. Can you reconsider?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IEmulator.cs
```diff
@@ -9,56 +9,94 @@ public partial class Gameboy : IEmulator, IBoardInfo
...
 
-		public ControllerDefinition ControllerDefinition => GbController;
+		public ControllerDefinition ControllerDefinition => ((int)_syncSettings.FrameLength == 2) ? SubGbController : GbController;
 
 		public bool FrameAdvance(IController controller, bool render, bool rendersound)
 		{
 			FrameAdvancePrep(controller);
-			if (_syncSettings.EqualLengthFrames)
+			uint samplesEmitted;
+			switch ((int)_syncSettings.FrameLength)
```
YoshiRulz, line 18: Don't need to cast to `int` for equality or for `switch`.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.ISettable.cs
```diff
@@ -255,14 +255,21 @@ public int LatchedRTCSeconds
...
-			public bool EqualLengthFrames
+			[DisplayName("Frame Length")]
+			[Description("Sets how long an emulation frame will last.\nVBlank Driven Frames will make emulation frames sync to VBlank. Recommended for TASing.\nEqual Length Frames will force all frames to emit 35112 samples. Legacy, not recommended for TASing.\nUser Defined Frames allows for the user to define how many samples are emitted for each frame. Only useful if sub-frame input is desired.")]
+			[DefaultValue(FrameLengthType.VBlankDrivenFrames)]
+			public FrameLengthType FrameLength
 			{
-				get => _equalLengthFrames;
-				set => _equalLengthFrames = value;
+				get => _frameLength;
+				set => _frameLength = value;
```
YoshiRulz, line 272: I think this can be an auto-prop (`public FrameLengthType FrameLength { get; set; }`). Elsewhere in this file, a backing field is used because the property's setter has extra validation, but not here.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IEmulator.cs
CasualPokePlayer: https://prnt.sc/135ohdh
https://prnt.sc/135oit9

adelikat:
Any objections to merging this?

alyosha-tas:
> 
> 
> Any objections to merging this?

nope, go for it.

CasualPokePlayer:
Any progress on this?

PR #2734, opened 2021-05-09
Add Serilog

YoshiRulz:


vadosnaprimer:
Works if I `Console.WriteLine()` in https://github.com/TASVideos/BizHawk/blob/f67de7a23b422508df94156113c2403f6939eab1/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/NativeApi/mupen64plusCoreApi.cs#L1057
but not from here https://github.com/TASVideos/BizHawk/blob/f67de7a23b422508df94156113c2403f6939eab1/src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs#L314
Both cores run in separate threads.

nattthebear:
The big meta-problem with what we have right now is lack of understanding of requirements.  Both the commit to add the bad log window, and the commit to remove the option of win32 console entirely, were completely undocumented and unjustified.

Here are my use cases and needs:

1 The window needs to capture all unmanaged stdout writes.
2 Ideally, the window should still work (be scrollable and copyable) even when the entirety of Bizhawk is halted in a debugger.

That combination of requirements seems to rule out anything but a win32 console.  Maybe my requirements are unfair?  I do definitely use both 1) and 2) extensively when working.

YoshiRulz:
2\. is dumb, the only reason to use print-debugging is when you're not able to step through execution.

I'd add these to the list:
3\. Anything printed to stdout from managed code must be (perhaps with a toggle) copied into a log file, so lay users can provide logs in bug reports.
4\. Must be portable to .NET 5 with few modifications.

And the way you've worded 1\. is specific to Windows. From what you've said on IRC, it seems you only care that unmanaged prints to stdout are *in the same place as* managed prints, and aren't silently dropped.

zeromus:
Your use case is to support natt doing the things nobody else can do, not doing things nobody but you asked for

vadosnaprimer:
> 2. is dumb, the only reason to use print-debugging is when you're not able to step through execution.

Or when something unknown happens and you want to look up what might have went wrong during the previous steps? It's not about `print("HI")`, it's about reading legitimate logs.

Jonny2579:
I'd like to request RetroAchievement information be tracked in the emulator's log output

PR #2740, opened 2021-05-14
Implement bsnes version 115 as a new core

### LINKED ISSUE #1797: SNES emu core updates (BSNES)
The Bsnes and SNES 9x have lot of update to the core type emu that Boost performance need to review these core and update in bizhawk one of these update Fix some games can have scanline filter problem

Morilli:
This aims to port the current version of the bsnes snes core to BizHawk. This should hopefully resolve all issues people have been having about emulation inaccuracies (haven't verified it myself because I am not going to play through a whole game to find out whether the ending works or not).

I've used the existing implementation of bsnes v087 as a basis, but have heavily modified the entire construct, so the code should actually look decent now.
Also, I have only modified the bsnes source code to a necessary minimum for bizhawk integration and to support features such as lag detection or trace logging.

Because this is pretty much a fresh implementation, I can basically modify anything about it without issues, so if anything is missing or broken or should be done differently; it can all be changed.
Some notable (currently) missing things that the 087 version has: No debugging support, no graphics debugger and no `CodeDataLog` implementation, whatever that even is.
Some thing both implementations don't have but I would like to see in this new one eventually: 
- using bsnes's own serialization, which is small, about as fast as the waterbox one and can be deterministic if we are able to get libco updated
- support for sufamiturbo and bsmemory cards.. I honestly have no idea about all of that but I saw it and I'm pretty sure this is currently unsupported in bizhawk's frontend, but could be easily done in the core
- runahead? bsnes basically supports that natively, but the frontend probably isn't ready for it right now (oh and it crashes rn but that's just a minor issue)
- frame skip support; as bsnes has frameskip capabilities
- probably some more fun stuff, like mode7 scaling for higher resolutions

I have left some comments and TODO's scattered around the new implementation so if anyone wants to look at those feel free to.

closes #1797


nattthebear:
In the short term, I think this makes the most sense as a fourth SNES core option, probably flagged as unreleased.  Power users who are interested can try out all the games on it, see what's working that wasn't before, and do free beta testing for us.  Long term, I'm not sure where we stand.  Four cores for a single system is pretty depressing, and gives the user nothing but confusion.  But I'm not sure what the alternative is...

adelikat:
I agree with Natt, I approve of this

YoshiRulz:
Suggestion: super interface `IBSNES` for old and new BSNES, and while you're at it a superinterface for `IBSNES` plus Faust and Snes9x. I think a lot of this code duplication in the front-end can go away if you do that.

Morilli:
>super interface IBSNES for old and new BSNES

I thought of that too, but at this point there are many small differences, like added configuration options for the new core, enums having a different length / different order; inputs having a different order now compared to before, that I feel like it may not even be worth the effort to try and "generify" this to one interface, as there will still have to be differentiations made on whether it's the new or the old core.

YoshiRulz:
Uhh where's the binary?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.IVideoProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace BizHawk.Emulation.Cores.Nintendo.BSNES
+{
+	public partial class BsnesCore : IVideoProvider
+	{
+		// TODO: This should probably be different for PAL?
+		public int VirtualWidth => (int) Math.Ceiling((double) BufferHeight * 64 / 49);
+
+		public int VirtualHeight => BufferHeight;
+
+		public int BufferWidth { get; private set; } = 256;
```
nattthebear, line 13: Is support for hires games a future endeavor?  That's fine with me so long as this doesn't absolutely crash out or anything on hires.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.IMemoryDomains.cs
```diff
@@ -0,0 +1,38 @@
...
+			foreach (int i in Enum.GetValues(typeof(BsnesApi.SNES_MEMORY)))
+			{
+				void* data = Api.core.snes_get_memory_region(i, out int size, out int wordSize);
+				if (data == null) continue;
+				if (Enum.GetName(typeof(BsnesApi.SNES_MEMORY), i) == BsnesApi.SNES_MEMORY.CARTRIDGE_RAM.ToString())
+				{
+					_saveRam = (byte*) data;
+					_saveRamSize = size;
+				}
+				mm.Add(new MemoryDomainIntPtr(Enum.GetName(typeof(BsnesApi.SNES_MEMORY), i), MemoryDomain.Endian.Little, (IntPtr) data, size, true, wordSize));
```
nattthebear, line 23: These raw core pointers are in waterbox space, so you need to Enter / Exit when using the memory domain.  `MemoryDomainIntPtrMonitor` should take care of a lot of the work.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs
```diff
@@ -0,0 +1,391 @@
...
+			for (int color = 0; color < 32768; color++) {
+				int r = (color >> 10) & 31;
+				int g = (color >>  5) & 31;
+				int b = (color >>  0) & 31;
+
+				r = r << 3 | r >> 2; r = r << 8 | r << 0;
+				g = g << 3 | g >> 2; g = g << 8 | g << 0;
+				b = b << 3 | b >> 2; b = b << 8 | b << 0;
+
+				palette[color] = r >> 8 << 16 | g >> 8 <<  8 | b >> 8 << 0;
```
nattthebear, line 318: Back in the day, some of our frontend code required the top 8 bits of this to be 1.  That may or may not be the case now?  I forget.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.IVideoProvider.cs
Morilli: This value is just a default. The actual value gets set every frame in `snes_video_refresh`.

PR #2742, opened 2021-05-16
[Gambatte] Many Updates

CasualPokePlayer:
- Fix MBC5 RAM Gate register (behavior confirmed via ACE payload)
- Add more proper no BIOS support (restoring to original Gambatte behavior of hacking in the post BIOS state)
- Misc so Gambatte builds with GCC + testrunner works
- Add a workflow to Gambatte-Speedrun for the testrunner
- Fix LY 0 Mode 1 to 2 transition in GBA
- Expose isCgbDmg + add isCGBDMGMode instead of isCGBMode to movie headers if active
- Fix LY in tracelogger, closes https://github.com/TASVideos/BizHawk/issues/2710
- Fix nontrivial reads from the tracelogger, closes https://github.com/TASVideos/BizHawk/issues/2701
- Switch the F register to be next to the A register in the tracelog (mimics GBHawk, and makes more sense anyways)

https://github.com/TASVideos/gambatte-speedrun/pull/14


YoshiRulz:
Please fix the Gambatte build script. Only the `.dll` needs to be copied.

vadosnaprimer:
I vaguely remember something about this in relation to the tasvideos movie parser, but to not rely on assumptions, please provide rationale for this change. Because it _looks_ counterintuitive.

CasualPokePlayer:
So far, GB games ran in GBC mode are marked as GBC by the movie parser. This was confusing and made so all these movies got marked as GB instead, which was more just manually adjusting the movie platform manually for each movie and doing so every time a GB in GBC movie is submitted (the latter this PR should get rid of in effectively most cases).

CasualPokePlayer:
An idea that popped into my mind, would something like this be better for this:
```
if (gb.IsCGBMode())	
{
movie.HeaderEntries.Add(gb.IsCGBDMGMode()) ? "IsCGBDMGMode" : "IsCGBMode", "1");
}
```

vadosnaprimer:
> The way I did it wasn't exactly the best, I would prefer just adding in a "isCGBDMGMode" to the movie header, but I have no idea how easy that would be for site admins to go and get the movie parser to actually consider such a flag.

I missed this because it was an edit :P

As long as that flag means it's a DMG game, it should be fine for the parser to ignore it.

CasualPokePlayer:
> > The way I did it wasn't exactly the best, I would prefer just adding in a "isCGBDMGMode" to the movie header, but I have no idea how easy that would be for site admins to go and get the movie parser to actually consider such a flag.
> 
> I missed this because it was an edit :P
> 
> As long as that flag means it's a DMG game, it should be fine for the parser to ignore it.

Ah, when I said that, I meant adding that flag alongside the isCGBMode flag. Although placing that flag in instead of isCGBMode in (as my next reply talks about) would probably be better anyways.

CasualPokePlayer:
Decided to combine this PR with a ton of other updates I had ready to go, should let this also close out 2 issues I had open, and put Gambatte by in line with upstream.

alyosha-tas:
Lot's of cool updates here, nice work. I think it can be merged now.

PR #2757, opened 2021-05-28
Commonise `DisplayManager`

YoshiRulz:
This wasn't the big refactor I'd been planning to do post-release, but I was drawn into it when I couldn't easily take a raw screenshot without using `MainForm`, `DisplayManager`, or any of the other globals.

Anyway, the aggregate diff sucks, read each commit.


zeromus:
Rationale?

YoshiRulz:
> [...] easily take a raw screenshot without using MainForm, DisplayManager, or any of the other globals

With this changeset, I can instantiate `DisplayManagerBase` and use that.

Also I would like it to be re-usable in a hypothetical non-WinForms frontend.

zeromus:
All fine with me except 1787d7b9c89d6d8cc38bcfaddc6eea6a81c63979. please redo as a rename and edit commit separately to increase the odds of blame data staying useful. IOW, erase that commit and instead rename DisplayManager.cs to DisplayManagerBase.cs in one commit and then add DisplayManager.cs and modify DisplayManagerBase.cs in another commit. Never rename files and edit them in the same commit. And never squash merge... In fact what is this "not enabled for this repository" nonsense on the merge commit option.........

PR #2774, opened 2021-06-04
Unthread tarsestudio seek progress bar

nattthebear:


src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -98,15 +98,12 @@ private void StartSeeking(int? frame, bool fromMiddleClick = false)
 			TastudioPlayMode(); // suspend rec mode until seek ends, to allow mouse editing
 			MainForm.UnpauseEmulator();
 
-			if (!_seekBackgroundWorker.IsBusy && diff > TasView.VisibleRows)
```
vadosnaprimer, line 101: This line was used to disable the progressbar if we're seeking through some short distance, because that usually gets edited the most, and the progressbar becomes distractive. And if we're seeking to a frame that's within the view, there's no useful info in the progressbar. Can you bring this back?

nattthebear: Added back the relevant part of the conditional.  With that, I can't get the progress bar to ever show up in trivial testing scenarios now, but maybe you have bigger movies that you can still verify it with.

vadosnaprimer:
Did you enable auto-restore while testing? If not, you'll need to press mouse wheen after editing a cell to make it restore playback position.

Also found another thing. The progressbar was updating during turbo-seek, but here it only appears as already full and stays full until playback restores its position. Not sure which peace of code was responsible for that tho.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
vadosnaprimer: Did you enable auto-restore while testing? If not, you'll need to press mouse wheen after editing a cell to make it restore playback position.

adelikat (changes requested):
Requesting changes, to fix the turbo seek issue feos mentioned in comments

vadosnaprimer:
<natt> I bet the frontend doesn't update tools during turbo seek
<natt> Which is probably the whole reason we have that hack

PR #2783, opened 2021-06-06
Fix crash when XInput controllers disconnect

### LINKED ISSUE #2782: BizHawk crashes when an xinput controller is disconnected

### Summary
If an xinput controller is disconnected, BizHawk immediately closes. This does not occur with OpenTK.

### Repro
1. Start EmuHawk with an xinput controller connected and EmuHawk configured to use XInput+DInput
2. Disconnect the controller
3. Observe the crash

### Output
Running a dev build through the debugger, the crash appears to be an unhandled exception in BizHawk.Bizware.DirectX.GamePad360.SetVibration(int left, int right) on line 239
```
Exception Unhandled: SlimDX.XInput.XInputException: 'ERROR_DEVICE_NOT_CONNECTED: n/a (-2147023729)'
```

### Host env.
- BizHawk 2.6.2; Win10 Pro 20H2; AMD/NVIDIA
- BizHawk 2.6.2 Developer Build (master#b62f4bc6a); Win10 Pro 20H2; AMD/NVIDIA

Klades:
Fixes #2782

Skips polling on disconnected Xinput controllers, wraps SetVibration call in a try/catch

### DIFF
--- a/src/BizHawk.Bizware.DirectX/DirectInputAdapter.cs
+++ b/src/BizHawk.Bizware.DirectX/DirectInputAdapter.cs
@@ -48,6 +48,8 @@ namespace BizHawk.Bizware.DirectX
 		{
 			foreach (var pad in GamePad360.EnumerateDevices())
 			{
+				if (!pad.IsConnected)
+					continue;
 				for (int b = 0, n = pad.NumButtons; b < n; b++) handleButton(pad.InputNamePrefix + pad.ButtonName(b), pad.Pressed(b), ClientInputFocus.Pad);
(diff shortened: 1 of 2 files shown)


zeromus:
please check it AND try/catch. try is only for emergencies. checking it and then using it has a problem where it could get disconnected between the check and use. in that case handling by try/catch is a proper emergency. for all other cases the initial check should work better

Klades:
It should be possible to skip the entire process loop for a disconnected xinput gamepad in ProcessHostGamepads, would that be preferable or should the scope be limited to SetVibration?

zeromus:
Since youre looking into it, I guess it would be better to skip polling disconnected pads to save some time

YoshiRulz (approved):
LGTM

PR #2785, opened 2021-06-06
Skip DirectInput initialization for device names containing "Xbox One…

### LINKED ISSUE #2784: Xbox One controllers are initialized twice

### Summary
Xbox One controllers and 8bitdo Xinput controllers (likely many generic xinput controllers) end up being registered as both GamePad objects and as GamePad360 objects.
This seems to cause fewer problems than one would expect, and if you rebind the controls everything is fine, but some games are unplayable with the default bindings.

### Repro
1. Use an 8bitdo Pro 2 controller in xinput mode, or similar
2. Start Super Metroid (for example)
3. Observe that the "Start Game" option can't be selected, because hitting A and B at the same time at that point returns to the save select screen.

### Host env.
- BizHawk 2.6.2 Developer Build; Win10 Pro 20H2; AMD/NVIDIA

Klades:
…" or "XINPUT"

Fixes #2784

No idea if this is an exhaustive check but it should at least pick up official controllers and hopefully a lot of generic ones (such as 8bitdo)

### DIFF
--- a/src/BizHawk.Bizware.DirectX/GamePad.cs
+++ b/src/BizHawk.Bizware.DirectX/GamePad.cs
@@ -26,7 +26,7 @@ namespace BizHawk.Bizware.DirectX
 				{
 					Console.WriteLine("joy device: {0} `{1}`", device.InstanceGuid, device.ProductName);
 
-					if (device.ProductName.Contains("XBOX 360"))
+					if (device.ProductName.Contains("XBOX 360") || device.ProductName.Contains("Xbox One") || device.ProductName.Contains("XINPUT"))
 						continue; // Don't input XBOX 360 controllers into here; we'll process them via XInput (there are limitations in some trigger axes when xbox pads go over xinput)
 
 					var joystick = new Joystick(_directInput, device.InstanceGuid);


YoshiRulz:
I was going to berate you for not simply checking if a `GamePad360` was made for the device in the `GamePad` ctor, but it's not actually possible to match up DirectInput (`GamePad`) and XInput (`GamePad360`) devices using either API. Yet another masterful play by MS. They have the nerve to tell you to copy-paste [this C++ snippet](https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput#xinput-and-directinput-side-by-side) into your game to get that functionality.

Knowing that, this seems to be the least bad solution. @zeromus?

zeromus:
Knowing this, any solution that limps by fine by me. It's a messy situation. Having a fully great solution would require months of dedicated work. 

But why does this matter? You don't want the default bindings for dinput and xinput to clash? There shouldnt even BE default bindings for dinput; it doesn't make sense. IMO only default bindings for xinput should exist.  Or does it pose a problem with input binding if the same device is feeding xinput and directinput both? That would be a good reason for this.

YoshiRulz:
I've seen the same xbone gamepad come up as `J#` and `X#`, seemingly at random, when binding all the buttons in turn. I'm assuming that's a symptom of #2784 too. If this PR prevents that then I'll merge it.

Issue #2585, opened 2021-01-19
Waterbox:  Improve handling of reentrant guest calls


nattthebear:
"Guest calls" is when code inside the waterbox is actually running; the emulation core itself.

"Reentrant guest calls" is when you are in host code (not guest code), that was _called back_ from guest code, for instance, a memory callback system, and you try to call more guest code.  An example would be having an input poll callback, and then trying to peek a system bus value while inside that callback, assuming peeking the system bus actually runs guest code.

As of 2.5.0, these all crash the emulator horribly with no possibility of recovery, and honestly not much possibility of debuggability:  I was faced with one of these crashes recently.  I knew what would happen on a reentrant call, but did not know that I was looking at a reentrant call, and I could not find out based on stacks, because everything was screwed++.

Bare minimum fix:  Kill the emulator with some more useful error message on a reentrant call attempt.
Better fix:  Support reentrant calls.  Usual caveat that the core probably doesn't support what you're trying to do with it, but in basic cases of reentrantly calling some memory domain access function, it may work out OK.

nattthebear:
Here's an example crash case:

```lua
-- This script is for genesis.  It will crash the emulator horridly whenever
-- advancing through any non-lag frame (ie, oninputpoll is actually called).

-- If "M68K BUS" is replaced by "VRAM", then this will not crash.  This is because VRAM domain
-- reads do not require executing code in the waterbox, but M68K BUS reads do.

function handle_input_poll()
	local val = memory.readbyte(0, "M68K BUS")
	console.log("InputPoll happened and we read the value: ", val)
end
event.oninputpoll(handle_input_poll)
```

This exact script is obviously super contrived, but is reading a domain in a callback like this something we think "should" work?

vadosnaprimer:
<natt> feos:  Does my demo script work in 2.4.2?

Yes. And dies since 2.5.

nattthebear:
Besides things that you can trigger in lua, this also affects the GPGX tracelogger (and possibly some other traceloggers?) which work by using memory callbacks plus state queries inside the memory callback.

PR #2797, opened 2021-06-13
Implement memory callbacks for the new bsnes core

Morilli:
Basically, everything "works" but the moment a callback gets executed bizhawk crashes. I can not pinpoint what is going on or what is crashing because apparently memory gets corrupted so badly the execution jumps into places it shouldn't be able to.

No idea what's going on or how to fix it, so I'm just putting this here for feedback because it would probably be nice to have this at some point.
Also it's likely libco.

If anyone wants to test, I'm using the `SMW-BizHawk.lua` script from https://github.com/brunovalads/smw-stuff (needs super mario world rom) with the `Lagmeter` option enabled (which executes get_registers on callback).


nattthebear:
Haven't looked yet, but is the crash just #2585?

Morilli:
>Haven't looked yet, but is the crash just #2585?

Hmm, you might be right about that. I guess whenever host code was calling core functions before it was never a reentrant call.

Morilli:
Have rebased onto #3007; should be in a working state now.

vadosnaprimer:
So this one is ready too?

vadosnaprimer:
```
[19:49:56] <natt> The idea seems fine
[19:51:02] <natt> In particular, with modern waterbox it is 100% legal and allowed to have the following call sequence:  1) Emuhawk calls waterbox frameadvance. 2) Transition to guest  code and start frame advance.  3)  Guest code switches to a different libsnes cothread 4) Guest code calls back into host space.
[19:51:32] <natt> 3 and 4 don't need a thread transition between them.  The host does not require the guest to have any particular stack pointer value for #4 (just that it has to be a legal memory address in the guest)
[19:51:53] <natt> I spent a lot of effort making that work and this is the payout
[19:53:17] <natt> (With classic native code, using cothreads like that would cause 4) to have C# managed code running on something which is not understood by win32 as a thread, and creates a giant mess sometimes)
```

PR #2798, opened 2021-06-14
[Gambatte] Open Bus Emulation

CasualPokePlayer:
Res0lves https://github.com/TASVideos/BizHawk/issues/2482 on the Gambatte side.

https://github.com/TASVideos/gambatte-speedrun/pull/18


alyosha-tas:
Is this inducing those test failures or were those already there?

CasualPokePlayer:
> Is this inducing those test failures or were those already there?

Oops, seems like I hit nontrivial reads when I wasn't supposed to. That should be easy to fix.

PR #2803, opened 2021-06-17
Add sa1_iram memory domain for new bsnes

### LINKED ISSUE #1097: bsnes - Watching / Searching SA-1 I-RAM memory possible?
Is watching / searching SA-1 I-RAM memory possible?

I think the I-RAM is not in the memory domains list. Also, I tried to watch it through the SNES System Bus 003000h-0037FFh, but the region is entirely filled by 0. I used Hoshi no Kirby Super Deluxe for the test.

Am I doing something wrong?

Morilli:
Adds a memory domain for SA1 IRAM (whatever that is) in the new bsnes core. 
Note that I didn't do the same for the old one because I don't want to touch the code because it's ugly and it would be more effort.
This addresses #1097 for the new bsnes core. Will leave the issue open for the sake of the core being experimental though.

Need any other memory domains btw? I can offer you `bwram`, `rom` and all the other coprocessors.


nattthebear:
LGTM.  Dunno what other memory domains people would want.

adelikat:
Fix conflict, and I'll merge

Morilli:
@adelikat have rebased and force-pushed to resolve conflicts; should be good now.

Issue #1567, opened 2019-05-18
Add ability to parse JSON to Lua Commands


Netbrian:
Bizhawk has been great at adding features like sockets, memory mapped files, and other features to allow it to be controlled with external programs, such as Python.

When doing this, we'd want the ability to have the external program command Bizhawk to do certain things.  One of the easiest ways I've found to do this is to use JSON to encode communications to and from the client.  However, you need to install an external library to encode and decode the JSONs on the Bizhawk end.

I'd like to request making this functionality built-in.  There are existing libraries under open-source licenses, such as JSON.lua ( https://github.com/rxi/json.lua ) that could be integrated, so we don't end up rolling our own.

I'm completely open to serialization formats (YAML, MessagePack, etc), but JSON seems like the simplest place to start.

zeromus:
No. bring your own json.lua. We don't include things which can be trivially included by the lua users, only things which are difficult for the lua users to bring themselves.

Netbrian:
Fair enough, thank you very much!

PR #2807, opened 2021-06-19
Add Middleclass lua library

HyagoOliveira:
[Middleclass](https://github.com/kikito/middleclass) is a simple and popular Object-Oriented Programming (OOP) library for Lua language. 

It allows to create better (and more readable) Lua scripts using the concept of Classes.

A Vector2 class was add as an example.  You can use it adding  `local Vector2 = require("Vector2")` on your script and instantiating by `local position = Vector2:new()`.

Middleclass is distributed under the MIT license.

### DIFF
--- a/Assets/Lua/Vector2.lua
+++ b/Assets/Lua/Vector2.lua
@@ -1,0 +1,54 @@
+--
+-- Author: HyagoGow
+-- A common Vector2 class
+--
+
+local class = require("middleclass")
+local Vector2 = class('Vector2')
+
+function Vector2:initialize(x, y)
+    self.x = x or 0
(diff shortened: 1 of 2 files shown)


zeromus:
I don't think we should be dropping in someone's "suggested super cool libraries". Bundling lua stuff is okay when it's really hard for lua users to incorporate it themselves (requiring native code or incompatible dlls), but these are just straight-up lua scripts so it's trivial for lua users to incorporate it

YoshiRulz:
The library is ~150 LOC of pure Lua so there's little benefit in shipping it with BizHawk. See also #1567.

PR #2813, opened 2021-06-22
Add regression tests for GB/C using various testroms

YoshiRulz:
edit: DONE FINALLY and with [･ﾟ✧ *documentation* ✧*･ﾟ](https://github.com/TASVideos/BizHawk/blob/testroms/src/BizHawk.Tests.Testroms.GB/README.md)

Things which are working:
* Frontend
	* Emulation, input, graphics (screenshot)
	* Firmware (from embedded file)
	* ApiHawk
	* Gambatte w/ BIOS
	* Gambatte w/o BIOS
	* GBHawk w/ BIOS
* Test suites
	* Gambatte suite (most of it)
	* Mealybug suite and 3 "acid" testroms
	* BullyGB suite
* Image comparison using Magick.NET

The things which aren't working:
* rtc3test (¯\\\_(ツ)_/¯)
* may be added post-merge:
	* SameSuite (no screenshots in repo so I didn't bother)
	* blargg's testroms (requires Nix to build locally)
	* Mooneye suite (couldn't get Nix build working)
	* CPP's testroms (couldn't get Nix build working)
	* some others I found in the SameBoy issue tracker

Thanks to @CasualPokePlayer for putting up with my dumb questions.

edit: When I asked where the tests for EmuHawk were a couple years ago (https://github.com/TASVideos/BizHawk/pull/1380#issuecomment-448089269), I never imagined I'd be the one to write the test framework, but here we are :)


alyosha-tas:
Very cool! There is a lot of scattered GB/C stuff out there, it's good to accumulate a lot of it in one place and have a testing frame work, great work! I haven't set up everything to run it, but looking over the code it looks good to me, a big step forward for sure.

I will say that linking to boot rom, and even licensed games, disassemblies (in other potentially volatile repos at that) and then assembling them is not my favorite approach, I'd prefer the simple 'find it yourself' that is usually used in these situations for those things. Not strictly related to BizHawk, but seems better to avoid it.

YoshiRulz:
Force pushed, changes from before:
- .NET 6
- renamed and split project
- Gambatte on Linux
- added human-readable portion to screenshot filenames
- made downloading from CI easier for Windows+WSL users
- improved docs
- misc. refactors

There have been a couple of regressions since:
- rtc3test passed on DMG in Gambatte before force push (CPP fixed this upstream); and
- `serial.start_wait_restart_read_if_2_dmg08_cgb04c_outE8` from Gambatte suite passed on DMG and CGB in GBHawk, but now shows `E0` on both.

The previous commit is backed-up on GitLab [here](https://gitlab.com/TASVideos/BizHawk/-/tree/testroms-202107).

alyosha-tas:
The link for documentation is currently broken.

I made a change to the linking code in GBHawk to make Perfect Dark linking work, this game expects setting the flag bit in the control register to not start a new transfer, but the regression in that test rom expects it to. They can't both be true, so something is amiss, but I haven't had a chance to look into the details yet.

Anyway, thanks for the further explanation wrt. copyrighted material, that seems sufficient to me. I think this can be merged if it's ready.

alyosha-tas (approved):

CasualPokePlayer:
> I made a change to the linking code in GBHawk to make Perfect Dark linking work, this game expects setting the flag bit in the control register to not start a new transfer, but the regression in that test rom expects it to. They can't both be true, so something is amiss, but I haven't had a chance to look into the details yet.

GambatteLink links up fine with that game, so I would presume this change isn't really needed per se.

Looking at GBHawk's code, you seem to not care whether bit 7 is set for starting a transfer, which would be wrong. That's probably why you fail that test and why Perfect Dark was broken.

---

This all looks good, LGTM

CasualPokePlayer (approved):

PR #2823, opened 2021-06-28
[Gambatte] Many Updates

CasualPokePlayer:
misc cleanup/fixes elsewhere

https://github.com/TASVideos/gambatte-speedrun/pull/20

Edit: Decided to combine this PR with an uninitialized AGB WRAM update, so instead of just using the same pattern for CGB (wrong for obvious reasons), it goes off based off a dump from a GB Player (which is the AGB revision targeted anyways).


alyosha-tas:
How consistent is that RAM state? Changing initial RAM may desync some TASes (oracle of ages / seasons), not sure it's worth it unless it's something consistent with high confidence.

CasualPokePlayer:
As far as I have tested, the RAM state is consistent 99.9% of the time on hard reset (power off doesn't seem long enough for corruption), so in any case for console verification forcing RAM to match the expected state should just work.

It's also effectively a provable state that the console can have (it's a raw dump, which is the exact same approach DMG and CGB initial RAM states have), and the previous state was effectively not a provable state (CGB is different obviously, if just mostly due to the AGB just uses its own IWRAM for the CGB WRAM, which itself likely explains the lack of easily noticable pattern).

CasualPokePlayer:
Ok I added some cleanup to the SetCpuRegister, but it seems appvendor is having problems? The code builds fine for me.

CasualPokePlayer:
TiKevin decided to go on a cleanup spree too and requested it was combined in this PR.

TiKevin83:
I can do an independent test of the RAM layout on my GBP and two GBA SPs but I need to have a different flashcart shipped so I won't have access to the equipment for ~2 weeks until after GDQ when I'm back in town.

alyosha-tas:
I'd like to wait on TiKevin83's results for the RAM part, I can merge everything else if you want to seperate it out again.

CasualPokePlayer:
Added some more updates to this PR; my latest seems to be having problems with the testrunner on github actions, although an offline run shows no regressions so far. I assume github is just acting up.

TiKevin83:
Confirmed that the initial WRAM pattern documented by CasualPokePlayer also appears on my GBP, and interestingly also my GBA SP systems. I'm good with merging

alyosha-tas:
Cool, that is unexpected, thanks for the verification.

PR #2837, opened 2021-07-09
Add support for more SGB firmware + Use homebrew bootroms instead of bsnes' illegally provided official ones

CasualPokePlayer:
Unsure if this is something really wanted. Added in all the SGB ROMs from nointro (SGB, SGB Rev A, SGB Rev B, SGB2, along with some SGB Rev A Beta ROM?). Also implemented optionally using official SGB bootroms in the newer BSNESv115 core since that has actual decent GB emulation (it's actually fairly up to date SameBoy) unlike whatever the heck the old BSNES core had.

~Also I have no idea if the waterbox part compiles with what I added /shrug~ compiles fine, everything works as intended, builds added to this PR

EDIT: And now I have found out the provided bootroms were just official bootroms, so I've replaced them with SameBoy's homebrew bootroms.


alyosha-tas:
Good catch, perhaps 2.6.2 should be re-released to not contain the original files as well.

CasualPokePlayer:
> 
> 
> Good catch, perhaps 2.6.2 should be re-released to not contain the original files as well.
With that train of thought... you would have to re-release every single version with the old BSNES with SGB stuff too, as that has official bootroms in there too (including the CGB bootrom which is way outside of any sort of triviality argument).

alyosha-tas:
> 
> 
> > Good catch, perhaps 2.6.2 should be re-released to not contain the original files as well.
> 
> With that train of thought... you would have to re-release every single version with the old BSNES with SGB stuff too, as that has official bootroms in there too (including the CGB bootrom which is way outside of any sort of triviality argument if just due to it responding to user inputs).
> 
> (although as far as re-releasing goes, I don't care, you guys do what you want to deal with old versions)

oh, didn't realize it was old version too. Well, I guess it's not hard to just replace the files directly in the dll by hex editor and then update the releases that way, I checked and at least in the recent builds they all show up directly adjacent to each other in the dll file.

zeromus:
Fun thought experiment, but don't bother. It's worth the trouble to work towards a more pure result but nobody actually cares about the old builds.

Morilli (approved):
Don't know much about the sgb rom stuff, but in general this looks and works fine for me 👍

PR #2840, opened 2021-07-11
N64 timing is bullshit

CasualPokePlayer:
This is an attempt to correct https://github.com/TASVideos/BizHawk/issues/2838 likely done horribly wrong. I am not an expert on N64, and hooking in audio for timing seemed to produce the most sane time (but it's still slightly off compared to sync to audio dumps, so I do not know what is differing).

EDIT: For testing I'm using the new 1 Key SM64 TAS, and the audio timing method I hooked up gives a time of 4:16.616. A Sync to Audio video dump however produces a 4:16.366 time, so the audio timing method hooked up appears off, and I do not know what would be differing between those times.

This PR is more just for feedback on what the fuck should be done regarding whatever the fuck Mupen is doing for timing, or whether I am just crazy.

### DIFF
--- a/src/BizHawk.Client.Common/movie/HeaderKeys.cs
+++ b/src/BizHawk.Client.Common/movie/HeaderKeys.cs
@@ -21,6 +21,7 @@ namespace BizHawk.Client.Common
 		public const string SyncSettings = "SyncSettings";
 		public const string VBlankCount = "VBlankCount";
 		public const string CycleCount = "CycleCount";
+		public const string SecondCount = "SecondCount";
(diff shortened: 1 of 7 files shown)


nattthebear (changes requested):
Looks like this includes changes from https://github.com/TASVideos/BizHawk/pull/2837 as well.  Either separate the branches or wait for that PR to settle first.

CasualPokePlayer:
This PR is less to be merged but more for feedback. Regardless, I'll see what I can do on separating that shit out.

PR #2842, opened 2021-07-11
add options for virtualboy to only show left or right screen

mooware:
On several occasions I've seen people ask how to configure BizHawk and/or OBS to only capture a single screen of the VirtualBoy, and it's usually a convoluted solution like cropping one screen away in OBS, or setting a high SideBySideSeparation. So I thought it would be nice to be able to configure the emulator to just show one screen, with a simple setting.

This change adds the new values OnlyLeft/OnlyRight to the ThreeDeeMode setting, which only display the left or right screen.


YoshiRulz:
This should be submitted upstream first.

It would be nice if we had a general solution for multiple screens, currently there's a hack for NDS. @zeromus

zeromus:
Did you actually build and test this as a waterbox core

mooware:
I did build and test the waterbox core. The VirtualBoy startup screen has icons in each corner which I used to confirm which screen is shown (top-left/bottom-right are on the left screen, the other two corners on the right screen, IIRC). The updated waterbox core is part of this pull request, as you can see in the changed files.

I suppose I can try to submit this to Mednafen. I found a thread there where somebody already suggested it in 2013 (https://forum.fobby.net/index.php?t=msg&th=859&start=0&), and they seemed rather dismissive, but maybe it helps that I'm bringing a patch for it. Although I guess it needs a bit more work to integrate into Mednafen, e.g. for configuration and documentation.

I also had a look at RetroArch's VB core, which is based on the same Mednafen emulator and is configured by default to show just one screen (i.e. the use case that 99% of users would probably want). It looks like they set it to anaglyph mode, but the anaglyph colors preset is off by default, and somehow that works out to it just showing one screen, and using the regular color palette instead of the anaglyph colors. Maybe that would be an alternative option if I can't get my changes into Mednafen, although it still seems less obvious to the user than just saying "give me only the left/right screen".

zeromus:
You will probably not have luck getting your patch merged into mednafen, but that it's actually a patch increases the odds from none to slim. I suggest you post on that forum thread and link to this PR. I also suggest you make a DIFFERENT commit which works it into the settings framework.

Then give it a month or til Administrator replies. VB core is not likely to ever get very big updates so forking it on our end is no big deal.

waterbox/vb/vip.cpp
```diff
@@ -1262,6 +1268,26 @@ static void CopyFBColumnToTarget_HLI(void)
...
+		else
+			CopyFBColumnToTarget_SideBySide_BASE(0, 1, 0);
+	}
+	else
+	{
+		if (!lr)
+			CopyFBColumnToTarget_SideBySide_BASE(1, 0, 0);
+		else
+			CopyFBColumnToTarget_SideBySide_BASE(1, 1, 0);
+	}
```
Morilli, line 1288: Is there a reason why this isn't just
```suggestion
  CopyFBColumnToTarget_SideBySide_BASE(DisplayActive, VB3DMode != VB3DMODE_ONLYLEFT, 0);
```

mooware: I would generally agree that it would be simpler and cleaner, but I am also strongly for sticking to existing conventions and coding style, and this is what the other functions in this source look like. I don't know why they're written like this, but I don't want my added function to look different.

adelikat:
The point of nyma cores is the ease of updating from upstream, let's focus on getting this feature there.  While it is a useful feature, easily and often updating our cores is more important

nattthebear:
I agree with all the advice given above about upstream, but do keep in mind that VBoy is not a nyma core; it is still in its own one-off port which has diverged somewhat from upstream.  It predated nyma, and once nyma came around, there was no compelling reason to switch it as there had been no major tech improvements upstream, but our core was moderately customized.

adelikat:
Based on @nattthebear 's comments, I think it makes sense to merge this

Issue #2844, opened 2021-07-11
memory.readbyterange returns 0 indexed lua table


CasualPokePlayer:
Noticed in #scripting, lua generally does not like the 0 indexed part and it's typically easier to work with a 1 indexed table (e.g. you can't assign the 0th index on initialization and have to use [0] to assign it). It should be returning a 1 indexed table.

~~Oh and it appears that even with matching inputs in the table there is some difference (revealable via tostring) compared to the table returned in memory.readbyterange which causes comparisions to fail. I don't know what this is, although interestingly it is workaroundable via table.concat it seems (but the 0th index is cut off, so issue above screws that over).~~

vadosnaprimer:
Please post an example that would check the resulting difference you're talking about.

CasualPokePlayer:
For testing I used Prehistorik Man (GB) for this script (script reads out from ROM)
```lua
local a = true
local f = io.open("lol.txt", "wb")
local g = io.open("lol1.txt", "wb")

while true do
	b = memory.readbyterange(0, 2)
	c = { }
	c[0] = 49
	c[1] = 254
        if a then
	f:write(tostring(b))
	g:write(tostring(c))
	f:close()
	g:close()
        a = false
        end
	emu.frameadvance();
end
```

alyosha-tas:
tostring just prints out the memory location of the tables, (just try running that script with different programs open to see the difference.) 

I think there is some misunderstanding here about how lua works. c[0] is not necessarily the first entry of the table, it is the entry associated with the key with (integer) value 0. Both b and c in the above produce correct tables representing memory addresses 0 and 1 in main memory. You just can't do things like `for i,line in ipairs(c) do` because then lua only starts looking for keys of (integer) value starting with 1.

But, it turns out the above is only correct by accident, the problem is that readbyterange is returning incorrect key data if the start address is not zero. In this case it returns initial index zero instead of whatever the base address is, contrary to what the documentation says should happen.

Should be an easy fix.

YoshiRulz:
In this case I think the documentation should be changed, not the behaviour, to keep it consistent with the .NET API.

CasualPokePlayer:
> tostring just prints out the memory location of the tables, (just try running that script with different programs open to see the difference.)

Ah, misunderstood that. Thought that was why equality checks against b and c were failing. Although I guess it might be something else? (or maybe I'm misremembering, idk) 

[...]

PR #2860, opened 2021-07-16
Always initialize MMF + Implement new LUA functions

E-Sh4rk:
This commit adresses two things:

1. Currently, the MMF LUA functions are only available if Bizhawk is started with the "--mmf=filename" flag.
I think this is not intuitive, because we should not have to specify a filename in order to active MMF functions
as the given "filename" is only used for "mmfScreenshot" (and even for this function, the filename can be set in the LUA).
I changed this behavior so that MMF functions are always available, and "mmfScreenshot" will fail with a log if
no filename has been specified by the command line nor by the LUA script.

2. I have added several LUA functions:

- "memory.writebyterangeataddress": This is because the function "memory.writebyterange" does not match with what "memory.readbyterange" returns...
"memory.readbyterange" returns a table with the keys going from 0 to len-1, while "memory.writebyterange" returns a table with each key being the full address (not starting at 0).
Actually, "memory.writebyterange" is not only for writing a range, by any set of addresses.
On the other hand, the new function "memory.writebyterangeataddress" has a behavior similar to "memory.readbyterange" or "comm.socketServerSendBytes":
the lua table given as argument is seen as an array instead of an (address, value) dictionnary.

- "mmfWriteBytes", "mmfReadBytes": to send/read a sequence of bytes to/from a memory mapped file

- "mmfCopyToMemory", "mmfCopyFromMemory": to copy the content of a memory mapped file to a given location in the emulator memory (and the other direction).
Note that this could be achieved directly in the LUA script with the combination of "memory.readbyterange" and "mmfWriteBytes", but is is A LOT slower
(probably because of the type conversions). Also note that I personaly need these functions in order to be able to efficiently read/write to the game memory
from an external tool.

### DIFF
--- a/src/BizHawk.Client.Common/Api/MemoryMappedFiles.cs
+++ b/src/BizHawk.Client.Common/Api/MemoryMappedFiles.cs
@@ -8,12 +8,11 @@ namespace BizHawk.Client.Common
 	public sealed class MemoryMappedFiles
 	{
 		private readonly Dictionary<string, MemoryMappedFile> _mmfFiles = new Dictionary<string, MemoryMappedFile>();
-
 		private readonly Func<byte[]> _takeScreenshotCallback;
 
-		public string Filename;
+		public string? Filename;
 
-		public MemoryMappedFiles(Func<byte[]> takeScreenshotCallback, string filename)
+		public MemoryMappedFiles(Func<byte[]> takeScreenshotCallback, string? filename)
 		{
 			_takeScreenshotCallback = takeScreenshotCallback;
 			Filename = filename;
@@ -21,13 +20,24 @@ namespace BizHawk.Client.Common
 
 		public string ReadFromFile(string filename, int expectedSize)
 		{
-			using var viewAccessor = MemoryMappedFile.OpenExisting(filename).CreateViewAccessor();
-			var bytes = new byte[expectedSize];
-			viewAccessor.ReadArray(0, bytes, 0, expectedSize);
+			var bytes = ReadBytesFromFile(filename, expectedSize);
 			return Encoding.UTF8.GetString(bytes);
 		}
+		public byte[] ReadBytesFromFile(string filename, int expectedSize)
+		{
+			if (!_mmfFiles.TryGetValue(filename, out var mmfFile))
+			{
+				mmfFile = _mmfFiles[filename] = MemoryMappedFile.OpenExisting(filename);
+			}
 
-		public int ScreenShotToFile() => WriteToFile(Filename, _takeScreenshotCallback());
+			var viewAccessor = mmfFile.CreateViewAccessor(0, expectedSize, MemoryMappedFileAccess.Read);
+			var bytes = new byte[expectedSize];
+			viewAccessor.ReadArray(0, bytes, 0, expectedSize);
+			viewAccessor.Dispose();
+			return bytes;
+		}
+
+		public int ScreenShotToFile() => WriteToFile(Filename!, _takeScreenshotCallback());
 
 		public int WriteToFile(string filename, byte[] outputBytes)
 		{
@@ -35,6 +45,7 @@ namespace BizHawk.Client.Common
 			{
 				using var accessor = m.CreateViewAccessor(0, outputBytes.Length, MemoryMappedFileAccess.Write);
 				accessor.WriteArray(0, outputBytes, 0, outputBytes.Length);
+				accessor.Dispose();
 				return outputBytes.Length;
 			}
 
--- a/src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
@@ -57,7 +57,7 @@ namespace BizHawk.Client.Common
 			return APIs.Comm.Sockets.SendString(SendString);
 		}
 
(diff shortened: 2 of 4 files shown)


src/BizHawk.Client.Common/Api/MemoryMappedFiles.cs
```diff
@@ -8,33 +8,44 @@ namespace BizHawk.Client.Common
...
+			{
+				mmfFile = _mmfFiles[filename] = MemoryMappedFile.OpenExisting(filename);
+			}
+
+			var viewAccessor = mmfFile.CreateViewAccessor(0, expectedSize, MemoryMappedFileAccess.Read);
 			var bytes = new byte[expectedSize];
 			viewAccessor.ReadArray(0, bytes, 0, expectedSize);
-			return Encoding.UTF8.GetString(bytes);
+			viewAccessor.Dispose();
+			return bytes;
```
YoshiRulz, line 37: Why replace `using` with `Dispose` here?

src/BizHawk.Client.Common/Api/MemoryMappedFiles.cs
```diff
@@ -8,33 +8,44 @@ namespace BizHawk.Client.Common
...
 		}
 
-		public int ScreenShotToFile() => WriteToFile(Filename, _takeScreenshotCallback());
+		public int ScreenShotToFile() => WriteToFile(Filename!, _takeScreenshotCallback());
```
YoshiRulz, line 40: This should probably have better error handling:
```suggestion
		public int ScreenShotToFile()
		{
			if (Filename is null) {
				Console.WriteLine("MMF screenshot target not set; start EmuHawk with `--mmf=filename`");
				return 0;
			}
			return WriteToFile(Filename, _takeScreenshotCallback());
		}
```

And then the filename doesn't need to be checked in `CommLuaLibrary.MmfScreenshot`.

src/BizHawk.Client.Common/Api/MemoryMappedFiles.cs
```diff
@@ -8,33 +8,44 @@ namespace BizHawk.Client.Common
...
 
 		public int WriteToFile(string filename, byte[] outputBytes)
 		{
 			int TryWrite(MemoryMappedFile m)
 			{
 				using var accessor = m.CreateViewAccessor(0, outputBytes.Length, MemoryMappedFileAccess.Write);
 				accessor.WriteArray(0, outputBytes, 0, outputBytes.Length);
+				accessor.Dispose();
```
YoshiRulz, line 48: This is superfluous, the local is declared with `using` already.

src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
```diff
@@ -133,43 +133,55 @@ private bool CheckSocketServer()
...
+		}
 		[LuaMethod("mmfRead", "Reads a string from a memory mapped file")]
 		public string MmfRead(string mmf_filename, int expectedSize)
 		{
-			CheckMmf();
-			return APIs.Comm.MMF?.ReadFromFile(mmf_filename, expectedSize);
+			return APIs.Comm.MMF.ReadFromFile(mmf_filename, expectedSize);
 		}
-
-		private void CheckMmf()
```
YoshiRulz, line 167: This is hard to read, can you split the changeset into two commits? One to make `--mmf` optional, and one to add extra functions.

src/BizHawk.Client.Common/lua/CommonLibs/MemoryLuaLibrary.cs
```diff
@@ -47,12 +47,23 @@ public MemoryLuaLibrary(IPlatformLuaLibEnv luaLibsImpl, ApiContainer apiContaine
...
+		[LuaMethod("writebyterangeataddress", "Writes the given values as a contiguous block of unsigned bytes to a specific address")]
+		public void WriteByteRangeAtAddress(LuaTable memoryblock, long addr, string domain = null)
+		{
+			int i = 0;
+			foreach (var v in _th.EnumerateValues<double>(memoryblock))
+			{
+				APIs.Memory.WriteByte(addr+i, (uint)v, domain);
+				i++;
+			}
+		}
```
YoshiRulz, line 63: This is out of scope, and we haven't reached a consensus re: #2844 yet.

YoshiRulz (changes requested):
I have no problems with this conceptually. The MMF stuff is niche enough that you can add whatever you need.

src/BizHawk.Client.Common/lua/CommonLibs/MemoryLuaLibrary.cs
E-Sh4rk: Ok, I am gonna remove this function!

src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
E-Sh4rk: Sure. I dont know if I can do that in this pull request, so maybe I will close this one and open a new one.

vadosnaprimer: You can force-push whatever set of commits you consider proper.

E-Sh4rk:
I'll make a new pull request with the changes you requested.

PR #2872, opened 2021-07-21
Add new GLideN64 settings

GhostlyDark:
Needs the updated version of GLideN64 that is commited to master to be useful first:
https://github.com/TASVideos/GLideN64/tree/Branch_bizhawk

That version will not work unless threaded video is turned off.
@vadosnaprimer Can you please update the .dll file of the repo with a working 2020-02 build of GLideN64?

These settings are added:
- BackgroundsMode: Sets the rendering mode of large background images, like the JFIF backgrounds of Ocarina of Time. Defaults to stripped as it is more accurate. A user may want to change this to one piece mainly for texture pack usage or performance.
- txEnhancedTextureFileStorage: Use HTS cache instead of HTC for enhanced (upscaled) textures. Defaults to HTC.
- txHiresTextureFileStorage: Use HTS cache instead of HTC for HD textures. Textures are streamed from disk, eliminating the need to prefetch the textures at load. Boots the game near instantly and decreases RAM usage. Defaults to HTC.

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64SyncSettings.GLideN64.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64SyncSettings.GLideN64.cs
@@ -11,7 +11,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.N64
 			{
 				UseDefaultHacks = true;
 
-				BackgroundsMode = 1;
+				BackgroundsMode = BackgroundsRenderingMode.Stripped;
 				MultiSampling = 0;
 				AspectRatio = AspectRatioMode.FourThree;
 				BufferSwapMode = SwapMode.OnVIUpdateCall;
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64SyncSettings.GLideN64.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64SyncSettings.GLideN64.cs
@@ -11,6 +11,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.N64
 			{
 				UseDefaultHacks = true;
 
+				BackgroundsMode = 1;
 				MultiSampling = 0;
 				AspectRatio = AspectRatioMode.FourThree;
 				BufferSwapMode = SwapMode.OnVIUpdateCall;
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
2020-02 build https://github.com/TASVideos/GLideN64/commit/2894d54da2786bb4b816a7f44b65913997994972 is committed to bizhawk since 2020-02 https://github.com/TASVideos/BizHawk/commit/8ff2ca2cd19d3280a45ab498c72f5d99785adc1d

src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64SyncSettings.GLideN64.cs
```diff
@@ -11,6 +11,7 @@ public N64GLideN64PluginSettings()
 			{
 				UseDefaultHacks = true;
 
+				BackgroundsMode = 1;
```
CasualPokePlayer, line 14: BackgroundsMode is type BackgroundsRenderingMode, not type int. This is causing the build error. Replace with this:
```suggestion
				BackgroundsMode = BackgroundsRenderingMode.Stripped;
```

CasualPokePlayer (changes requested):
Build error is just an implicit cast being assumed (C# doesn't allow this), simple fix.

GhostlyDark:
My bad about the build error. I changed how the BackgroundsMode is displayed last minute and forgot to rebuild.

Maybe I'm just really dumb here or still super confused and if that's so I am really sorry, but the .dll file that BizHawk gets distributed with, both the latest releases and the latest builds, use a GLideN64 dll that is around 2 MB big. **It is not up-to-date**. Loading HTS cache with this build is impossible, the Owl in Ocarina of Time is buggy, something the newer build fixed. The one in the Assets folder isn't the new one either. What ships is an old binary file.

vadosnaprimer:
Check https://github.com/TASVideos/BizHawk/commit/11b1e328a6751bf947ad27a5e2e97f60cfe1fbe1

GhostlyDark:
@vadosnaprimer Thank you very much!
This can me merged then.

vadosnaprimer:
Not sure what went wring initially. I even remember setting the output path to Assets but can't see that part committed. So you confirm everything works now?

GhostlyDark:
Everything works. GLideN64 is updated for real now and the options I added are recognized by it.

PR #2878, opened 2021-07-24
Fix #2877

### LINKED ISSUE #2877: RewindFrequency > 1 often causes rewinds to drift 1 frame forward when loaded.
### Summary
With RewindFrequency > 1, whenever you rewind once and then unpause / frame advance, the just loaded rewind drifts 1 frame forward. This is caused by the fact that the emulator advances by 1 frame after every rewind, so the rewinder can't re-capture the frame that was just loaded.

Due to some special cases in the implementation of `rewind()`, this doesn't happen if you rewind 1 frame after a frame that was captured by the rewinder. In particular, this issue doesn't occur if RewindFrequency = 1.

Happens both with and without delta compression.

### Repro
1. Open Bizhawk
2. Emulation > Pause (pauses the emulator)
3. File > Open Rom > open a Super Mario World rom.
4. Config > Rewind & States.
   - Enable rewinds
   - Set max buffer size to 2^6 = 64 MB (smallest value to get larger RewindFrequency).
   - Set "Desired frame length" to 50000 (to get RewindFrequency > 1).
   - Click "OK" to close the configuration window.
5. Repeat the following steps.
   - Press frame advance (f by default) twice.
   - Press rewind (shift+R by default) once.
   - Note that for every repetition, the frame counter increases by 1, but no additional rewind states have been captured, so the last rewind state just drifted 1 frame forward.
6. Alternatively, you can also repeat the following steps:
   - Emulation > Pause (unpauses the emulator)
   - Let the emulator run for a few seconds.
   - Emulation > Pause (pauses the emulator)
   - Hold rewind (shift + R by default) and rewind all the way back.
   - Once again, for every repetition, we end up rewinding to a state 1 frame later.
7. You can also first let the emulator run for a bit and then repeat the steps of 5.
   - In particular, this issue also occurs with non-empty rewind buffers.

### Host env.
- BizHawk dev build at f1da6e6cd3835a418a3f3b64e1ac63c93f61bc5d; Win10 Pro 19041; Intel

dacin21:
Fix #2877 This fix affects the rewind function (both in Zwinder and Zeldawinder)

https://github.com/TASVideos/BizHawk/blob/fd0d5a38c780dc167496a459deb8ee31ca8b40b8/src/BizHawk.Client.Common/rewind/Zwinder.cs#L63-L78

The main difficulty here is that the emulator frame advances after loading a rewind, so we need to avoid loading a rewind from just 1 frame ago.
In this function, we need to handle 3 dinstinct cases (excluding the trivial case where we return false).

1. `state.Frame == frameToAvoid` and `Count = 1`.
2. `state.Frame == frameToAvoid` and `Count > 1`.
3. `state.Frame != frameToAvoid`.

Here's how I think these cases should be handeled:

1. `state.Frame == frameToAvoid` and `Count = 1`.
   - Load the 1 state we have, then invalidate the rewind buffer (then it's empty).
   - Note: This is then handeled by the emulator by not frame advancing. In particular, we need to invalidate the state to avoid capturing it again.
   - The current master also handles this case this way.
2. `state.Frame == frameToAvoid` and `Count > 1`.
   - Load the second-newest state, then invalidate the newest state.
   - The current master also handles this case this way.
3. `state.Frame != frameToAvoid`.
   - Load the newest state, but do *not* invalidate it.
   - This case can only happen if RewindInterval > 1.
   - The current master *does* invalidate the newest state, which leads to the linked issue.

### DIFF
--- a/src/BizHawk.Client.Common/rewind/ZeldaWinder.cs
+++ b/src/BizHawk.Client.Common/rewind/ZeldaWinder.cs
@@ -231,31 +231,27 @@ namespace BizHawk.Client.Common
 			if (!_active || _count == 0)
 				return false;
 
-			if (_masterFrame == frameToAvoid && _count > 1)
+			if (_masterFrame == frameToAvoid)
 			{
-				var index = _buffer.Count - 1;
-				RefillMaster(_buffer.GetState(index));
-				_buffer.InvalidateEnd(index);
-				_stateSource.LoadStateBinary(new BinaryReader(new MemoryStream(_master, 0, _masterLength, false)));
+				if (_count > 1)
+				{
+					var index = _buffer.Count - 1;
+					RefillMaster(_buffer.GetState(index));
+					_buffer.InvalidateEnd(index);
+					_stateSource.LoadStateBinary(new BinaryReader(new MemoryStream(_master, 0, _masterLength, false)));
+				}
+				else
+				{
+					_stateSource.LoadStateBinary(new BinaryReader(new MemoryStream(_master, 0, _masterLength, false)));
+					_masterFrame = -1;
+				}
 				_count--;
 			}
 			else
 			{
+				// The emulator will frame advance without giving us a chance to
+				// re-capture this frame, so we shouldn't invalidate this state just yet.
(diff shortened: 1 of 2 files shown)


nattthebear (approved):
Yes yes, this makes sense.  Great work.

PR #2882, opened 2021-07-27
Set firmwares declared with `FirmwareAndOption` to `Ideal` status instead of `Acceptable`

YoshiRulz:
(`FirmwareAndOption` being the helper which declares a record *with a single option*.) These have always received the default of `Acceptable` since the status enum was introduced in f43859f42, with the exception of the FDS "Nintendo" BIOS (the first declaration, see b9e784e15). I believe that was unintentional.

This PR changes the *default* param value in `FirmwareAndOption`; individual calls may still override it with `Acceptable` if that is desired.

see also https://github.com/TASVideos/BizHawk/issues/2825#issuecomment-871231323 (icons for `Ideal` and `Acceptable` swapped?)


zeromus:
makes sense to me and cant remember why it didnt make sense to me.

PR #2884, opened 2021-07-29
Add onclose event

HyagoOliveira:
Adds a `string event.onclose(nluafunc luaf, [string name = nil])` event that fires every time the emulator closes.


YoshiRulz:
Name and docs are misleading. (Granted, so is `event.onexit`.) Call it `event.on_console_close` or something.

HyagoOliveira:
Rename to `onconsoleclose` since almost all other events are in lowercase. But I've to admit lowercase decreases legibility.
I can change to pascalCase if you want.

PR #2893, opened 2021-08-02
Add config option for rewinds every fixed number of frames.

### LINKED ISSUE #2463: Add config option for rewinds every fixed number of frames.
The new rewind system doesn't allow you to have rewinds every fixed number of frames, say every 2 frames. Instead, you have to set a "Desired frame length" and the rewind length is dynamically calculated based on the current average state size. As the average state size changes over time, there is no way of avoiding the occasional 1-frame rewind or 3-frame rewind.

It would be nice to have an additional config option that allows you to have rewinds every x frames (as in the pre 2.5 rewind system) instead of setting a "Desired frame length". Having rewinds every x frames is very useful for tasing without TASudio, as it allows you to predict exactly how many frames you go back. For example, after pressing rewind once, pressing rewind 3 more times always moves you back 6 frames. (With the current system, you might go back 5 or 7 frames, which is very awkward.)

dacin21:
Fix #2463

No changes to TAStudio.
Outside of TAStudio, there's now an option to specify a fixed rewind interval instead of a desired frame length:

grafik

This is how the rewind frequency used to be specified in the pre 2.5 rewind system.
While the new way of specifying a "desired frame length" is great for TAStudio, it is very cumbersome outside of it: For one, with a buffer size of >20'000 frames, you need to specify a huge desired frame length of 100'000 to (approximately) get 5-frame rewinds.  And, more importantly, even if you set a desired frame length of 100'000, then the rewind intervals still vary (more or less randomly) between 4-6 frames, as the state size differs from frame to frame. This "randomness" makes them very awkward to use.

Don't get me wrong, TAStudio is an amazing tool for making highly optimized TASes, doing small tweaks to your inputs, re-adjusting lag frames, etc. However, for non micro-optimized runs or glitch hunting, 5-frame (or even 2-frame) rewinds are a lot faster to use.

### DIFF
--- a/src/BizHawk.Client.Common/config/RewindConfig.cs
+++ b/src/BizHawk.Client.Common/config/RewindConfig.cs
@@ -49,7 +49,7 @@
 		public bool UseDelta { get; set; }
 		public bool Enabled { get; set; } = true;
 		public long BufferSize { get; set; } = 512; // in mb
-		public bool UseFixedRewindInterval { get; set; } = true;
+		public bool UseFixedRewindInterval { get; set; } = false;
 		public int TargetFrameLength { get; set; } = 600;
 		public int TargetRewindInterval { get; set; } = 5;
 		public IRewindSettings.BackingStoreType BackingStore { get; set; } = IRewindSettings.BackingStoreType.Memory;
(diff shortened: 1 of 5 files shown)


nattthebear (approved):

InfamousKnight:
This would benefit rewind on higher tier cores like melonds.  I would use it paired with a gamepad.

PR #2896, opened 2021-08-05
Add automagic patching capability to `FirmwareManager`

YoshiRulz:
Replaces hack in Hawk-side Gambatte. (I recall seeing other hacks, but can't find any.) Would also work for NDS if the hashes of all N permutations were hardcoded... I have ideas for how to get around that limitation.

I also added a simple tool to `Help` > `Debug Utilities` for trying out this feature without needing to mess around with firmware customisations and sync settings.

edit 2021-09-18: Force-pushed because of #2917. The extra two patchsets added in that PR (which happen to be identical to existing ones) are now included here. I also realised it was stupid to hardcode the reverse of a patchset instead of computing it, so that's gone.

### DIFF
--- a/src/BizHawk.Client.Common/fwmanager/FirmwareManager.cs
+++ b/src/BizHawk.Client.Common/fwmanager/FirmwareManager.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Security.Cryptography;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


zeromus:
As long as there's a rationale given for why we can't patch in memory and have to drop the patched file on the disk, I guess this is as good as it could possibly get, and I don't mind the idea of clearing the patching crap out of the cores. In the long run we have 50 emulators for every core and the concept of firmware delivery is a frontend responsibility.

PR #2911, opened 2021-08-17
Deduplicate some controller deck code

YoshiRulz:
Generated code looks like:
```csharp
public partial class A7800HawkControllerDeck
{
	public const PeripheralOption DEFAULT_PERIPHERAL_OPTION = PeripheralOption.Standard;

	private static readonly IReadOnlyList<Func<int, IPort>> _controllerCtors = new Func<int, IPort>[]
	{
		static portNum => new UnpluggedController(portNum),
		static portNum => new StandardController(portNum),
		static portNum => new ProLineController(portNum),
		static portNum => new LightGunController(portNum),
	};

	private static Func<int, IPort> CtorFor(PeripheralOption option) => _controllerCtors[(Int32) option];
}
```
No reflection in sight! The callsites are very simple:
```csharp
// ctor in the non-generated file
public A7800HawkControllerDeck(PeripheralOption port1Option, PeripheralOption port2Option)
{
	Port1 = CtorFor(port1Option)(1);
	Port2 = CtorFor(port2Option)(2);
	// ...
}

// in A7800Hawk's ctor
_controllerDeck = new A7800HawkControllerDeck(_syncSettings.Port1, _syncSettings.Port2);
```
The magic is in the attributes and source generator, which I think are not possible to apply incorrectly because of all the validation and warnings. It should always produce clean, valid code. The code for the generator itself is another story... I tried to use descriptive names but I'm sure the nested loops don't read well.

There was one problem I didn't solve, and that's how to allow only a subset of the enum members. The GBHawk linked cores don't support the "+ tilt" option and that works fine, but the type is not part of their sync settings. It's when the enum is used in sync settings and a UI is generated that it becomes an issue. I think the `DescribableEnumConverter` might be able to read an attribute, say `[ExcludePeripheralOption(AnEnum.AMember)]`, from the property it's applied to, and limit what it sends to the PropertyGrid.


alyosha-tas:
Honestly, the auto generator stuff is too complicated for me to follow. I don't think it's better then what's there now, can any of the stuff you want to clean up be done without the added background complexity?

YoshiRulz:
I could take the generated code and copy it into the current codebase. That would get rid of the divergent implementations, but it doesn't solve the problem of needing to update 5 files to change the peripheral inventory. Maybe that could be covered by some tests?

alyosha-tas:
I gave this another look, and I honestly don't like this one. The tradeoff between some duplicated code and the complexity of the auto generator just doesn't seem worth it to me.

If someone else with more experience with these kind of structural things thinks this is a good idea then fine, but personally I'd have to give it a no.

Maybe @zeromus can give an opinion.

zeromus:
I will say, however painful it is, the current system has been used successfully many times. 
But as long as you don't HAVE to use this new generator, then it doesn't matter how complex it is. Maybe nobody will ever use it again. It's not hurting anything.  Is that the case?

alyosha-tas:
I suppose in this case it is, given how infrequently this particular code is changed. But I don't know, each new re-factor seems to add complexity, parts of the system are becoming inscrutable (at least to me.) I don't think it's always for the best, this case included, but you are right I can mostly just ignore it if it's merged.

zeromus:
I too agree that some parts of bizhawk are getting too complex, and would generally oppose this, but since you could opt out completely, and the wireup for this stuff was _already_ too complex for anyone who didn't already know how to do it, I'll let this one pass.

YoshiRulz:
If it's not going to be used then merging it would be adding unnecessary complexity, so I won't. Which isn't a huge loss as I learned a lot about Roslyn Analyzers.

New question: Should I use replace the current code with the generated output?

zeromus:
as long as it all STILL WORKS, having it "normalized" seems to have some value to me, since I assume it was considerably more haphazard before

alyosha-tas:
Yeah I can agree that the generated code itself is fine it if works in a stand alone form.

PR #2917, opened 2021-08-30
Implement HLE SGB Emulation in Gambatte

CasualPokePlayer:
This is an attempt to implement SGB emulation into Gambatte, aiming for feature parity with the Sameboy core.

Currently implemented:
Palettes
Sound
Border
Savestates
Multiplayer
Nearly everything an SGB core can ask for, save for a nice default border

Also going to go slap in all other upstream misc fixes in this PR (including fixing https://github.com/TASVideos/BizHawk/issues/2919, among some other critical state fixes)

EDIT: Also fixes https://github.com/TASVideos/BizHawk/issues/2921


alyosha-tas:
How does using a GB core for SGB compare to using the BSNES core for SGB?

CasualPokePlayer:
> How does using a GB core for SGB compare to using the BSNES core for SGB?

You mean our Sameboy core vs BSNES?

Well, it's slightly complicated given BSNESv115 exists, but ultimately this is going down to HLE vs LLE differences.

Our Sameboy core and Gambatte from this PR can implement most SGB commands a game would use, except for ones which require an entire SNES like JUMP (which granted only 1 game used that, so little lost). Even SNES audio can be HLE'd using blarrg's snes_spc, which is just an SNES APU emulator.

By contrast, an LLE core such as BSNES or BSNESv115 can do everything an SGB can do, as they are properly emulating both a full SNES and full GB.

There's also input timing differences since a real SGB locks inputs per SNES VBlank, which is at a different rate than the GB.

Of course anyways, BSNES and BSNESv115 have their own GB cores, BSNES having byuu's own GB emulator and BSNESv115 having a recent Sameboy version for its GB core.

It should be noted anyways our Sameboy core (which is based on a 4 year old Sameboy version) and BSNES have fairly inaccurate GB emulation. A case in point is just launching Prehistorik Man, which breaks horribly in those two cores. BSNESv115 doesn't break horribly and is fairly accurate due to using a recent version of Sameboy.

Still, overall the HLE SGB cores are meant more for casual users, with the LLE cores better for TASing and users aiming for full accuracy.

The main reason for this project anyways is to provide an HLE SGB core which is not horribly inaccurate and will be maintained (Gambatte will have active development as long as I am around at least), with Gambatte also chosen due to its speed (it's much faster than the Sameboy core) which is fairly valuable for a causal user.

nattthebear:
This is exciting!  I always thought there was more potential for a better HLE SGB (specifically, one with sound support) as you can handle everything in every official SGB game except for space invaders AFAIK.  It never took off though, because I'm lazy and I guess everyone else was still thinking HLE must be as bad as VBA?

alyosha-tas:
Cool stuff. I always ended up the obvious bugs whenever I tried to play something SGB previously. though I never really looked too much into it.

Do you think there could be a possible BSNES + Gambatte in the future built off this work?

Nice work either way though, maybe it can just replace the old Sameboy core which would simplify things a bit.

CasualPokePlayer:
BSNES+Gambatte is just lsnes, and quite frankly pointless for BizHawk. BSNESv115 provides a very accurate LLE core with its recent versions of BSNES and Sameboy.

As far as replacing Sameboy goes, I'll leave that question after this PR is finished.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.IEmulator.cs
```diff
@@ -9,7 +9,7 @@ public partial class Gameboy : IEmulator, IBoardInfo
 	{
 		public IEmulatorServiceProvider ServiceProvider { get; }
 
-		public ControllerDefinition ControllerDefinition => (_syncSettings.FrameLength == GambatteSyncSettings.FrameLengthType.UserDefinedFrames) ? SubGbController : GbController;
+		public ControllerDefinition ControllerDefinition => CreateControllerDefinition(IsSgb, _syncSettings.FrameLength is GambatteSyncSettings.FrameLengthType.UserDefinedFrames);
```
YoshiRulz, line 12: Is `FrameLength` allowed to change while the core is running? I assume not, so this only needs to be called once during initialisation.
```suggestion
		public ControllerDefinition ControllerDefinition { get; } = CreateControllerDefinition(IsSgb, _syncSettings.FrameLength is GambatteSyncSettings.FrameLengthType.UserDefinedFrames);
```

CasualPokePlayer: It technically can change, although it would probably be better anyways if it doesn't take affect until core reboot. Although that change doesn't actually build. ("A field initializer cannot reference the non-static field, method, or property 'Gameboy.IsSgb'")

CasualPokePlayer:
On some positive notes on this PR, issues https://github.com/TASVideos/BizHawk/issues/2477 and https://github.com/TASVideos/BizHawk/issues/2565 seem to not affect this SGB implementation in Gambatte.

CasualPokePlayer:
I believe everything should be okay in regards to functionality and such. There are some general todos which should come eventually:

SGB1 vs SGB2 options (currently I'm just forcing over SGB2 for simplicity/that's a better system anyways, but options would be nice here. Not sure how to handle this, there's already a stubbed console mode option, so I feel like this might want to be resolved with a UI update)
SGB2 in GambatteLink (since the SGB2 (not 1) did have a link port...) (Not sure how to approach this if mix matching border/borderless is a thing...)
Better API/approach towards SGB samples (I originally intended to append the samples to the end of the gb audio buffer, but that ended up not working as intended, might have been me just being dumb with how I coded it, idk. I would prefer the mixing to be put into the core anyways but I wouldn't know how to do that)

There is 1 thing however that seems to weird with the spc here. This [assertion](https://github.com/pokemon-speedrunning/gambatte-core/blob/b389ea65d1b7f170e0c106ea8cc184bdede61248/libgambatte/src/mem/snes_spc/SNES_SPC_misc.cpp#L334) seems to sometimes fail after a long period of use. I believe it might be a weird state issue (from some cursory debugging, although I'm not sure if it's the state's fault per se), and all I have for it is simply increasing the extra_buffer size to avoid it (although I don't know if this really fixes the issue per se, although I haven't ran into the assertion failing yet so might be enough to get around the issue?).

Oh I also found weird state integrity failures which some seem to be related at all to SGB stuff, of course these states are nearly impossible to parse so I don't know anything else about it (it at least didn't cause any desync so eh?)

nattthebear:
Not sure about the savestates; I never trusted them and so the sameboy sgb is waterboxen.

CasualPokePlayer:
With a complete hack I believe the assertion failure should now be "fixed".

YoshiRulz (approved):
LGTM. I can come up with a better UX for switching to SGB later.

PR #2934, opened 2021-09-21
Nuke SameBoy

CasualPokePlayer:
Nukes the HLE SGB SameBoy core; there was some talk about the new HLE SGB in the Gambatte core replacing SameBoy in https://github.com/TASVideos/BizHawk/pull/2917, although I left that out of that PR.

I'll let @nattthebear decide whether this is appropriate.


alyosha-tas:
I downloaded the most recent Sameboy release, and it currently does better then Gambatte on some mode 3 timing tests (our current build doesn't, but it's pretty old.) Do you think there will be any work on mode 3 timing? (I don't think this should hold up anything here as it is a good clean up and reduces the maintenance burden, and it's still a secondary option to BSNES 115 anyway.)

CasualPokePlayer:
> 
> 
> I downloaded the most recent Sameboy release, and it currently does better then Gambatte on some mode 3 timing tests (our current build doesn't, but it's pretty old.) Do you think there will be any work on mode 3 timing? (I don't think this should hold up anything here as it is a good clean up and reduces the maintenance burden, and it's still a secondary option to BSNES 115 anyway.)

There will likely be work on mode 3 in Gambatte at least. Largely have been focusing on adding new features at the moment, so I haven't really done too much on this front.

PR #2945, opened 2021-09-29
[melonDS] Attempt to update core, redo a bunch of shit

CasualPokePlayer:
Slapping this here until I finished this shit. Hopefully at the end it's not dogshit (already have SDL dep and file IO out of core) and can be more easily maintained.

Also opting to slap in Drastic's freebios as a resource to fall back on when ds bios files are missing. They're licensed under a fairly permissible license so should be a quick and easy option.


InfamousKnight:
Surprised to see how fast this got done. Thank you!

CasualPokePlayer:
> Wow, this beats desmume in every way now!
> 
> Is it ready for use yet?

No

CasualPokePlayer:
Everything should be (at least roughly) implemented. My own fork of melon would remain submoduled until this PR is handled upstream: https://github.com/Arisotura/melonDS/pull/1237

Morilli:
Getting `System.InvalidOperationException: SetFileOpenCallback was not exported from elf`.

CasualPokePlayer:
> 
> 
> Getting `System.InvalidOperationException: SetFileOpenCallback was not exported from elf`.

That function was removed several commits ago, you shouldn't be getting that error unless you're using an old build.

Morilli:
Right. forgot to rebuiild 😅 
Seems to work fine now

nattthebear:
Should file ` Assets/dll/libstdc++-6.dll` exist?  Seems like leftovers from old stuff.

What's up with the threading stuff?  Waterbox doesn't really support any of that.  It does provide a fake stub implementation; so if that's sufficient, that's fine, but let's make sure the core isn't doing anything we can't support.

nattthebear:
> The newer libstdc++6-6.dll was due to me adding std::thread in the dll and the old dll didn't have that for some reason (which makes little sense when std::thread is a C++11 feature).

Waterbox doesn't use that file, and it hurts to add if it breaks some other core that we aren't testing right now; so let's just axe it.

PR #2948, opened 2021-10-01
Implement missing part of DiscoHawk CLI

YoshiRulz:
@Meerkov


Meerkov:
I look forward to trying this out later, but I noticed in the change that `discInterface ` is unused. Is it needed?

zeromus:
Is there any particular reason why this should be so controversial or tricky that it requires a review that I should pay attention to?

YoshiRulz:
> Is it needed?
Yes, oops.

> Is there any particular reason why this should be so controversial or tricky that it requires a review that I should pay attention to?
Nope

zeromus (approved):

zeromus (approved):

Meerkov:
I've tested the HAWK function, and it looks like it works to me. Thanks!

PR #2955, opened 2021-10-05
Update gamedb_intv.txt with alt of Super Pro Football

Glitcholog:
According to INTV Funhouse, there are 2 versions of Super Pro Football, the later version fixing a bug about the quarterback not appearing until the ball is hiked when playing on Intellivision II system. As I have an INTV II, the personal dump seems to be the later version. Adding the sha1 generated by the dump solved the issue. Played a whole game without glitches or game crash.


Glitcholog:
No, is there a way to check this?

alyosha-tas:
I looked around a bit myself but didn't find anything helpful. I would recommend you contact the folks at AtariAge (they have an Intellivision forum) regarding your cart, they have the expertise for this kind of thing.

alyosha-tas:
@Glitcholog If you can test your dump in the current dev build (and check that the checksum given by bizhawk matches the raw checksum of the ROM, I'll go ahead and merge this.

Glitcholog2:
Hello! Sorry for the late reply, my GitHub account somehow was deleted, so I had to create another one.

Unfortunatly , it  doesn't match what Bizhawk has given me. I do not have a second copy of the game, so Id guess that it might have been a faulty dump I did back in October.

alyosha-tas:
It could still be correct, it's just that i recently changed how INTV games are interpreted in the DB. What's the new hash that bizhawk gives? That's the one that needs to be added,

Glitcholog2:
It give the following 
Super Pro Football
SHA1:383A774973DB126C1E17E88DE939BCEA29BBD220
MD5:3F050F4D2A6E976201E17910EB8C1A29

alyosha-tas:
@Glitcholog2  ok, I'll just add that to the DB and close this PR. Thanks for the update. I would also still urge you to contact other Intellivision people to get wider circulation of this.

PR #2956, opened 2021-10-06
Discohawk CLI for music extraction

### LINKED ISSUE #2953: DiscoHawk CLI missing MP3 extraction
Following up on the last bug: Now that you can HAWK a file from the command line, you can't perform the Audio Extraction step.

There should be a command to run this function
https://github.com/TASVideos/BizHawk/blob/253d532a63c7073068bec246db288dc6a84cc180/src/BizHawk.Client.DiscoHawk/AudioExtractor.cs#L14

Meerkov:
Intended to fix #2953

This adds a CLI to extract MP3s.

Just use
`./DiscoHawk.exe MUSIC ./myfile.ccd OVERWRITE`


YoshiRulz:
Rebased to include new changes, inlcuding mine from today. Double-check it actually works and then I'll merge this. Your original commit had a warning message telling the user to add `-overwrite`, but it looks like you're not supposed to use hyphens on CLI flags, so I changed it.

Meerkov:
It works

PR #2983, opened 2021-10-28
fix date time display

### LINKED ISSUE #2796: Generic Sync Settings Has DateTime Display Get Truncated

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end.
Whenever my cat sits on the left side of my keyboard, games run too fast for me to react to."

Sync Settings shows only Date in RTC Initial Time row

### Repro
1. Open any GBA game
2. GBA -> Settings
3. Sync Settings -> RTC Initial Time

### Output
image

### Host env.
- BizHawk 2.6.2 Win10 Pro, commit #7c17f31cd

CasualPokePlayer:
Somewhat hacky, fixes #2796, the issue stems from the default DateTimeConverter, so I just hacked in a new one which overrides the bad behavior described in that issue.

### DIFF
--- a/src/BizHawk.Common/BizDateTimeConverter.cs
+++ b/src/BizHawk.Common/BizDateTimeConverter.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace BizHawk.Common
+{
+	public class BizDateTimeConverter : DateTimeConverter
+	{
+		public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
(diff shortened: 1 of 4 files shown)


CasualPokePlayer:
> 
> 
> So I can't actually _edit_ the value by typing, even if I also implement `ConvertFrom`. Might only affect Mono.

It's Mono only, Windows works just fine here.

I am skeptical of this format being used, that T just looks ugly there. Might be best to use local format (for least user confusion).

Issue #2985, opened 2021-10-29
PS1 - Swagman - Freezes on level 1 (BizHawk only)


Meerkov:
### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."

Exiting "The Garden", by any means, including dying, backtracking to the previous level, or even pausing, is impossible. If you pause or die, the game freezes. If you exit any other way, the loading screen spins forever.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1) Start the game
2) Get to the title screen
3) Wait for the demo to start (then skip it)
4) Wait for the demo to start again (then skip it)
5) Wait for the demo a third time (this time is the Garden level).
6) Press any button to return to the main menu... but actually you are stuck on a loading screen forever.

Now try the same thing on Mednafen.
Mednafen returns to the title screen successfully.

### Host env.
[//]: # "List the computers you've found the bug with. If there's a version that doesn't have the bug, please put that in too. Here are some examples:"
- BizHawk 2.6.2

[//]: # "(screenshots, if applicable)"

[//]: # "That's it! If you'd like to help more, you could try a dev build (see Testing in the readme) or an older release. Click submit now and you can edit it later."

Issue #2806, opened 2021-06-19
Update Octoshock core for Mednafen 1.27.1 changes


nattthebear:
Split off from #2804.  Not sure how many of these changes are relevant to us.

```
	PSX: Initialize some missed state on reset/poweron.
	PSX: Removed incorrect filtering of CD-XA ADPCM sectors based on the file and channel numbers of the first ADPCM sector played,
	originally added in 0.9.24-WIP to fix the speech playback in "Yarudora Series Vol.1: Double Cast", but made unnecessary by later sector buffering
	accuracy improvements; fixes missing audio in the FMVs of "Blue's Clues: Blue's Big Musical".
	PSX: Adjusted which sectors are processed and reflected by GetLocP after SeekP completes to only those before the seek target(which is slightly different
	than SeekL behavior, which also includes the target and a few after, at least with a typical PS1 CD); fixes hangs at different points in
	some versions of "Tomb Raider", a problem initially fixed in 0.9.38.3, but regressed in 1.22.0-UNSTABLE.
```

zeromus:
yuck. I still have some desync in that core to fix.
I've been thinking about reverting all my savestate related changes and trying to use mednafen's logic as-is.  However I'm concerned that a number of changes I made are _improvements_ (additions of missing variables) so I'm not sure if that will really buy us anything.

Issue #2639, opened 2021-02-23
No Audio during FMV: PSX Blue's Clues - Blue's Big Musical


Meerkov:
### Summary
The FMVs in Blue's Clues - Blue's Big Musical have no audio

### Repro
1. Start Blue's Clues the Big Musical
2. The FMV before the main menu has no audio.
3. After starting a new game, the FMV also has no audio.

### Host env.
- BizHawk 2.5.2; Win10

zeromus:
test in mednafen

Spikestuff:
upstream issue.

Meerkov:
FWIW, does not reproduce in ePSXe.

Spikestuff:
That's not worth anything considering that ePSXe is a closed source emulator, that is also heavily dependent on plugin usage and is also extremely out of date and is a hackjob in terms of accuracy.

"But it's the highest % in accuracy" cause it cheated the results.
ePSXe should be forgotten in a ditch for all that care about it these days.

PR #2989, opened 2021-11-03
update octoshock to Mednafen 1.27.1

CasualPokePlayer:
closes https://github.com/TASEmulators/BizHawk/issues/2806, closes https://github.com/TASEmulators/BizHawk/issues/2985, and closes https://github.com/TASEmulators/BizHawk/issues/2639


zeromus:
well, i wasnt doing this for ages because we have some problems with the quality of that port and I wanted to start over entirely from scratch while making fewer changes to mednafen. But if you're willing to tackle this, then I won't stop you.. I will just warn you that it's really tricky work, sprinkling these changes in. I don't have time to really add any value reviewing here other than to say you've got guts and it seems like you used the appropriate caution.

PR #2997, opened 2021-11-11
Dual Melon

CasualPokePlayer:
Preliminary work on Dual NDS core. Seems to operate fine. Needs to actually have some sort of frame stepping and lan packet transferring hooked up, and upstream needs to implement other stuff in their implementation.


nattthebear:
We've never had a dual wbx core go to production that I can recall.  We might want to look into dual relocations for speed; as `linkscript.T` puts everything at 36f:0000:0000, the two copies of DS will have to swap each other out every time you `Enter()` one or the other.  It all works, it's just less than super performant.

CasualPokePlayer:
> 
> 
> We've never had a dual wbx core go to production that I can recall. We might want to look into dual relocations for speed; as `linkscript.T` puts everything at 36f:0000:0000, the two copies of DS will have to swap each other out every time you `Enter()` one or the other. It all works, it's just less than super performant.

That would explain why I was constantly at 10 FPS... yeah I'll look into that

nattthebear:
As a quick testing hack:
1. Chang the linkscript's instances of 36f00000000 to 36e00000000.
2. Build a second copy of the wbx file with a separate name
3. make the nds emu able to load either, and make the dual nds load one of each

We'll need to clean up the build scripts but that should get you started

PR #3002, opened 2021-11-17
Refactor GambatteLink

CasualPokePlayer:
This refactors GambatteLink to be more modular, and should allow for easily adding in GB3x/GB4x support and keeping everything within this single core. Keeping this draft until I actually do that and hook up the needed UI crap for that.


CasualPokePlayer:
> Cool! Do you think you can add IR support along with this work? Should be a pretty simple copy paste from GBHawk.
> 
> Once the Gambatte linked modes are up to speed I can probably kill off the GBHawk ones, they will be too slow to be relevant at that point.

I did see that you had IR stuff in there (which was actually a surprise as I only recall one obscure emu actually doing that), I imagine it wouldn't be too hard to copy the logic.

There is maybe some benefit for GBHawkLink being those do bitwise transfers while Gambatte is byte wise, although I imagine it wouldn't be impossible to do bitwise stuff in Gambatte (could optimally be a sync setting to switch between bytewise and bitwise?), but eh.

On another note this should allow just using a single core id for link cores (I think I'll go rename DGB to GBL or GBLink and change everything to account for that). The GB3x/GB4x crap is not needed.

Unrelated note, but a lot of this can probably be abstracted out to generic classes, a lot here is directly copy pasted from my mGBALink PR which uses similar logic for wrapping around multiple cores. I imagine that would reduce a lot clutter too.

PR #3007, opened 2021-11-23
Handle reentrant calls in waterbox

nattthebear:
Only allows one level of reentrancy.
Probably shouldn't use this with dobiestation.


vadosnaprimer:
IT WORKS!!! Tested on GPGX tracelogger and script that has memory hooks (those even work together).

Issue #1626, opened 2019-07-28
Revisiting Sameboy for GB/GBC core


TiKevin83:
This is a follow up to issue 1282.

EZGames and I are looking to resync TASes for the broader GB/GBC library.  We've seen only minor improvements to syncing Mickey's Dangerous chase on GBP with the improvements MrWint made to Gambatte for Pokemon Crystal, we now get past the first enemy and die on the 2nd.  GBHawk is still way off, doesn't even sync to console through the intro.

Seperately, DrD2K9 also messaged me about syncing Donkey Kong '94, and that game also does not appear to play near console on either Gambatte or GBHawk.

Gambatte is simply too tuned for Pokemon with little attention to other franchises. 
 GBHawk from what I gathered in the last post with Alyosha only really accounted for GB tests when there are numerous GBC only timing quirks.  GBHawk also runs almost an order of magnitude slower than Gambatte, which is a significant factor for TASing.  The inaccuracies in GBHawk were highlighted in the last issue and no progress has been made on them.

These incremental issues could be avoided entirely by using SameBoy as a GB core across all of GB, GBC, and SGB.  SameBoy has constant active development and rigorous testing from a whole community of GB emu devs.  It has no known timing inaccuracies on any game; any desync between the SameBoy core and console playback would be down to the game relying on initial SRAM state and being unsyncable, or a potentially revolutionary discovery about the GameBoy hardware.  These cases can only be properly identified at this rate if SameBoy support is expanded.

DrD2k9:
Specifically for Donkey Kong (aka DK'94):  GBHawk and Gambatte cores emulate drastically differently in regards to loading times.  This obviously affects timing as well as things like RNG.  GBHawk seems to run the game well enough, but certain sounds don't play correctly (for example, When starting the mid-world DK fight at stage 1-4 the stage begins before the stage-intro SFX completes playing.).

YoshiRulz:
Do you know how up-to-date our core is to upstream?

TiKevin83:
Yoshi I'm not sure what you're referring to as "our core."

alyosha-tas:
All the tests you sent me pass for both GB and  GBC mode in GBHawk, so certainly progress has been made. But yeah I haven’t worked on it in a while .

[...]

PR #3009, opened 2021-11-29
Add SameBoy as a GB/C core

CasualPokePlayer:
Resolves https://github.com/TASEmulators/BizHawk/issues/1626

At the moment this core has most things hooked up and is largely functional. However, I have seen non-deterministic behavior around disabled LCD frames (currently being investigated by @LIJI32), and there is some API updates I am waiting on (so I can get rid of some of the hacks I stuffed in to act as temporary API wrt tracing/execute cb/scanline cb), along with other misc issues which will be resolved in time.

Also on "why"? SameBoy is competitively accurate with Gambatte and is actively developed. Some places SameBoy is more accurate, but Gambatte also has places where it is more accurate. SameBoy is also decently fast, not as fast as Gambatte but 2x faster than GBHawk. I've also tweaked SameBoy's frame timing to roughly match Gambatte's (besides an occasional extra frame around disable lcd places), so it should be fairly simple to resync movies from Gambatte to SameBoy if needed for say console verification in the case Gambatte desyncs but SameBoy works, or vice versa, and improve each emulator accordingly.


getCursorsExe:
Normally, this would cause the SGB core to go back, but I don't know.

YoshiRulz (approved):
LGTM

PR #3010, opened 2021-12-01
BSNESv115+: Replace the 'snes_audio_sample' callback with a dynamic audio sample vector to reduce callbacks

Morilli:
A somewhat alternative to https://github.com/TASEmulators/BizHawk/commit/d6ba93150c111688e1638aa57a8bbd4b4ded815c I suppose?
Main idea is to get rid of
1. the thousands of audio sample callbacks per frame
2. the resampler (which is "quite inefficient" and "will be replaced")

PRing because I'm not sure on the safety and general coding of the implementation.


Morilli:
@nattthebear going to tentatively request feedback here; everything should work, but I'm not very sure about the entire sound system.

nattthebear:
So what's doing the 32k -> 44.1k resampling now?  I'm not familiar with all of the core internals here.

Morilli:
> So what's doing the 32k -> 44.1k resampling now?

I think it's https://github.com/TASEmulators/BizHawk/blob/master/waterbox/bsnescore/bsnes/nall/dsp/resampler/cubic.hpp... I can't really tell you how it's working or what it's doing though.

nattthebear:
I suspect the speex resampler has better quality, but this one sounds fine to me.  Seems to be slightly faster too.

nattthebear (approved):

PR #3017, opened 2021-12-05
Cache `ControllerDefinition.ControlsOrdered`

### LINKED ISSUE #3011: Bk2InputDisplayGenerator.Generate() is too slow
The [Generate()](https://github.com/TASEmulators/BizHawk/blob/34ec13635b914e1dd74e62b7e6e9f140a260c771/src/BizHawk.Client.Common/display/IInputDisplayGenerator.cs#L30) method of the `Bk2InputDisplayGenerator` class is called every frame and performs way too slowly for that. It iterates through all inputs and performs super slow C# operations to get a list of valid controls and displays them on screen if pressed (all this only happens if the `Display Input` option is checked, though).

I did a profile run on a snes game with rewind and throttling disabled over 50,000 frames using the `Payload` controller, which has ~35 virtual inputs and got a result of all `ControlsOrdered` calls combined (150k) taking around 19% of total running time, with the rest of the `Generate()` method taking another 11%. That is 30% of total running time spent in code that does nothing, because no buttons were pressed.

Caching at least the return of the `ControlsOrdered` function should provide a substantial speedup already.

YoshiRulz:
...and many prerequisite changes. Fixes #3011. With 18897b5ab, I think the only speed left would be from span-ification.

Contains part of PR #2911.


Morilli (changes requested):
The general cleanup/refactor changes seem good from what I can see and I like the deduplication efforts, although I'm not sure whether all the `MakeImmutable()` calls are actually useful.

In regards to #3011 however, I'm still not really satisfied.

The core issue is that while `ControlsOrdered` / now `_cachedInputSpecs` is now cached, it is only ever accessed once. (see [MakeStringFor](https://github.com/TASEmulators/BizHawk/blob/34ec13635b914e1dd74e62b7e6e9f140a260c771/src/BizHawk.Client.Common/DisplayManager/OSDManager.cs#L212), which is the main culprit here). From my testing, the `MakeStringFor` function is called 3 times per frame, twice with an `AndAdapter` and once with a `AutoFireStickyXorAdapter`. And every time it is called, it creates a new `Bk2InputDisplayGenerator` object (which is never again used afterwards), which then caches its controls. See image for some stats.
image
Ignoring the rest of the `Generate()` method (>95% of it are `IsPressed` calls), the fact that the `_cachedInputSpecs` list is still built 3 times per frame (instead of once ever and then cached) is something that needs to be fixed. 
One simple idea of mine would be to just cache the entire `Bk2InputDisplayGenerator` object in `OSDManager` and use it in all `MakeStringFor` calls, although I'm not sure whether anything can possibly go wrong there, considering the autofirecontroller supposedly changes states every other frame.

YoshiRulz:
I swear I was going to do that, but forgot. Let me know what effect that has.

> I'm not sure whether all the MakeImmutable() calls are actually useful

They're *required*. If `ControlsOrdered` is called (to generate input display) before `MakeImmutable`, it throws. That's to ensure the set of available inputs doesn't change during the core's lifetime—in practice, this was already the case.

Morilli:
>> I'm not sure whether all the MakeImmutable() calls are actually useful
>
> They're *required*. If ControlsOrdered is called (to generate input display) before MakeImmutable, it throws. That's to ensure the set of available inputs doesn't change during the core's lifetime—in practice, this was already the case.

👍 I suppose having that guarantee now instead of blindly relying on it is good.


> Let me know what effect eca63e3 has.

This does improve things by a lot by caching 2/3 of the calls to `MakeStringFor`, but [this call](https://github.com/TASEmulators/BizHawk/blob/eca63e3a2338e55f50c7c435c70da53eba3a10db/src/BizHawk.Client.Common/DisplayManager/OSDManager.cs#L309) is still not being cached. Was that intentional because complications or an oversight?

YoshiRulz:
I skipped it because it wasn't trivial, but the components are all readonly in `InputManager` so it might be possible. I'll take another look.

Morilli (approved):
Looks good enough to me (obviously still a little heavy but MUCH bettter than before)
image

ᴵ ˢʷᵉᵃʳ ᵗᵒ ʰᵃᵛᵉ ᵗʰᵒʳᵒᵘᵍʰˡʸ ᶜʰᵉᶜᵏᵉᵈ ᵃˡˡ ¹⁶⁰⁰ ˡᵒᶜ ᶜʰᵃⁿᵍᵉᵈ ᶦⁿ ᵗʰᶦˢ ᴾᴿ

YoshiRulz:
@alyosha-tas Are these GBHawkLink changes okay?

adelikat (approved):

alyosha-tas:
Looks fine to me, but you're making me awfully nervous deleting that line from inputmanager

YoshiRulz:
Force pushed to clean up commit history. Only one change, a bugfix because someone on the language design team thought it would be a good idea to make `x is not 0 or null` equivalent to `!(x is 0) || x is null` instead of `!(x is 0 || x is null)`.

I'll merge this tomorrow unless there are any objections.

PR #3023, opened 2021-12-08
Pick core ctor by linked console count w/ hack

YoshiRulz:


nattthebear:
Why can't we have one core that accepts 2-4 roms?

CasualPokePlayer:
iirc I was thinking of just removing the 3x/4x options for GBHawkLink as just selecting the 2x GBHawkLink will fallback on the 3x/4x GBHawkLink when having 3x/4x roms anyways. Although even then that relied on coincidences anyways, so I didn't do it, not sure if this hack is better than just doing that (it could be nice if there could be some wrapper core that just makes one of the actual link cores based on rom and registers their services accordingly, but idk how much better that is).

YoshiRulz:
I briefly tried the wrapper core idea, but dropped it for some reason I can't remember even though it was an hour ago...

nattthebear:
Wrapper core would be another layer of shart, agreed.  Why were these built as separate cores in the first place?

alyosha-tas:
I just made them separate for simplicity. Probably the 4x can be made the main one and do whatever it needs to based on the number of roms sent.

PR #3049, opened 2021-12-28
Add fast-forward rewinds and rewind speed multipliers.

dacin21:
No changes to TAStudio. Outside of TAStudio, there's now an option to specify two rewind speed multipliers: one for normal rewinds, and one for rewinds while fast-forwarding.

grafik

If the rewind speed is `k`, then a single press of the rewind key will go back `k` states in the rewind buffer instead of just `1`. In practice, this makes the overall rewind speed when holding down the rewind key around `k` times faster, at the cost of having less precision. (It also leads to less mashing if you want to rewind multiple times back-to-back.) Having two separate rewind speeds allows you to profit from this increased speed by holding fast-forward without affecting the precision of your usual rewinds.

I've gotten a lot of mileage out of these fast-forward rewinds while playing some SMW glitch puzzle home brew, where rewinds play a similar role to the "undo" feature from e.g. Baba is you. Rewinding back >4k frames because you messed up an earlier step is not uncommon and not having to worry about making manual save states allows you to really focus on the puzzles. (Loading a save state would also clear the rewind buffer.) I have hit the 16k rewind limit at some point and it really sucked, so I bumped that up. (I don't see why it was that low to begin with.)

### DIFF
--- a/src/BizHawk.Client.Common/config/RewindConfig.cs
+++ b/src/BizHawk.Client.Common/config/RewindConfig.cs
@@ -34,6 +34,12 @@
 		/// </summary>
 		int TargetRewindInterval { get; }
 
+		/// <summary>
+		/// Number of rewind states to go back per press of the rewind button, without and with fast forwarding.
+		/// </summary>
+		int speedMultiplier { get; }
+		int fastSpeedMultiplier { get; }
+
 		public enum BackingStoreType
 		{
 			Memory,
@@ -52,6 +58,8 @@
 		public bool UseFixedRewindInterval { get; set; } = false;
 		public int TargetFrameLength { get; set; } = 600;
 		public int TargetRewindInterval { get; set; } = 5;
+		public int speedMultiplier { get; set; } = 1;
+		public int fastSpeedMultiplier { get; set; } = 5;
 		public IRewindSettings.BackingStoreType BackingStore { get; set; } = IRewindSettings.BackingStoreType.Memory;
 	}
 }
--- a/src/BizHawk.Client.Common/rewind/IRewinder.cs
+++ b/src/BizHawk.Client.Common/rewind/IRewinder.cs
@@ -15,7 +15,7 @@ namespace BizHawk.Client.Common
 		/// <summary>
 		/// Rewind 1 or 2 saved frames, depending on whether the last frame is stale.
 		/// </summary>
-		bool Rewind();
+		bool Rewind(bool fastForward);
 
 		void Suspend();
 		void Resume();
--- a/src/BizHawk.Client.Common/rewind/IRewinder.cs
+++ b/src/BizHawk.Client.Common/rewind/IRewinder.cs
@@ -13,9 +13,9 @@ namespace BizHawk.Client.Common
 
 		void Capture(int frame);
 		/// <summary>
-		/// Rewind 1 or 2 saved frames, avoiding frameToAvoid if possible.
+		/// Rewind 1 or 2 saved frames, depending on whether the last frame is stale.
 		/// </summary>
-		bool Rewind(int frameToAvoid);
+		bool Rewind();
 
 		void Suspend();
 		void Resume();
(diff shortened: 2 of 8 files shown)


nattthebear:
Here's what I'm seeing with (presumably default?) settings in release versions of Bizhawk right now, with a 60fps nominal core:  If I hold down SHIFT+R, I rewind at 60fps.  If I hold down TAB+SHIFT+R, I rewind at 240fps.  What does this PR add that makes it needed on top of that already existing capability?

dacin21:
Here are two example situations where this PR applies (In my case it's Super Mario World running on the BSNES core at 60fps):
1)  I'm playing a glitch puzzle and have to rewind 4000 frames because I messed up an earlier step. With rewinds every 2 frames and 66 rewinds / second, this takes me around 30 seconds. 5x speed cuts this down to around 6 seconds. 
2) Let's say I'm TASing with rewinds every 2 frames and I want to go back 24 frames to re-do some jump. With 1x speed, I have to hit the rewind button 12 times, while with 5x and 1x speed, I can do 24 = 2*10 + 2*2, so I only have to hit it 2+2 = 4 times.

The current master doesn't allow for (2), and tries to do (1) by increasing the frame rate cap if you fast-forward while rewinding. The PR does both by going back multiple states at a time inside the rewinder.

While increasing the frame rate cap might work on default setting (inaccurate snes9x core + 2x window size), I've found it to be a lot less effective in a more practical setting (accurate BSNES core + 4x window size + sometimes a lua script). Even without lua, I can only get 66 rewinds / second, not the "advertised" 120 rewinds / second (aka. 240 fps). With lua, it's only 26 rewinds / second, so pressing fast-forward does actually nothing! With the PR, pressing fast-forward always makes rewinds (close to) 5x faster, without using 5x as many resources. In particular, this scales a lot better to even higher speed multipliers.

nattthebear:
This feels like a misuse of tools.  If you need to go back a large number of frames, use the piano roll.  If you need to go back to a specific instant in time because that instant is particularly significant, take a state of it and load that specific state.  @vadosnaprimer could you shed some more light on these use cases?

vadosnaprimer:
TAStudio has different problems (like not even having a state that far back), but I've never had the problem with having to hold rewind for too long. Because you just scroll up as much as you need (or jump to a marker), and point at that row to make emulation (almost) instantly go to that frame (if there is a savestate nearby).

For for actual rewind outside TAStudio, I don't see the harm in more flexibility.

nattthebear:
Fair enough.  The reason I'm hesitant here is that these rewinders are complicated, took a lot of debugging to get right, and are easy to break.  (The rewinders they replaced were also very complicated.)  How close do we get in performance if we simply have the frontend call rewind() N times?  With no frame advancing in between, we'd have to pay for extra loadstates, but maybe that's not too bad.

YoshiRulz:
You'd want a flag on the loadstate method which skips tool updates and the API callback if you do it that way.

dacin21:
> This feels like a misuse of tools. If you need to go back a large number of frames, use the piano roll. If you need to go back to a specific instant in time because that instant is particularly significant, take a state of it and load that specific state. @vadosnaprimer could you shed some more light on these use cases?

I'm assuming piano roll = TAStudio? TAStudio is an amazing tool for making highly optimized stuff, getting through very precise sections or re-syncing lag frames, but for doing glitch puzzles or less optimized runs, I've found it to slow me down more than it helps. 

I will give you that going back 4000 frames is easer in TAStudio and it comes with added benifits if you overshoot your target by accident. However, if I wanted to go back 24 frames, then I had to move my hand from the keyboard to the mouse to click on the piano roll and the back to the keybord. This is even worse that hitting the rewind key 12 times.

Ironically, the feature I would need to make TAStudio more usable for me is two rewind hotkeys (rewind + fast-forward rewind) that go back 2 and 20 frames respectively (configurable), so that I can nagivate the piano roll with just the keyboard. As of now, you can only go back in single frame steps, which is painfully slow at times. (Holding down the key briefly was too inconsistent to build muscle memory.) Also keep in mind that TAStudio will always have a more difficult case to make due long saving times, larger movie files and slower playback time (+ IIRC some quirks if you hit / hold rewind while it's rebuilding the rewind buffer).

Making manual save states is an option, but sooner or later you'll forget to make one and and hate yourself. A big appeal of rewinds is that you don't have to constantly remind yourself to make states. Loading a state also clears the rewind buffer, so you better hope the state is at just the right place.

Issue #3066, opened 2022-01-09
Loading recent rom while in TAStudio breaks the GUI


Morilli:
### Summary
Pressing the "load last rom" hotkey or otherwise reloading the currently loaded rom while in TAStudio seems to put TAStudio in some weird kind of read-only mode, for example it prevents the tasproj from getting longer, aka advancing past the last frame of the tasproj puts the cursor in the void.
Additionally, doing it while on a frame that has inputs results in those inputs appearing "frozen", aka those inputs will always be considered pressed until TAStudio is closed.

Got really confused by that one until I noticed that my "load last rom" hotkey is ctrl+s, which I naturally also use to save a tasproj...

### Repro
1. Open any rom.
2. Open TAStudio.
3. Advance any number of frames. Either save the tasproj now or when asked in 4.
4. Load the first "recent rom" (the currently loaded rom).

### Host env.
- BizHawk 2.7.1 dev, commit 9b88ae48e; Win11 Pro 22526.1000

PR #3069, opened 2022-01-10
Fix weird edge case putting TAStudio in an unusual state

YoshiRulz:
#3066


src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs
```diff
@@ -124,7 +124,8 @@ public override void Restart()
 				}
 				else
 				{
-					RefreshDialog();
+					TastudioStopMovie();
```
Morilli, line 127: As this is the same first call as in the other if branch, might as well merge it outside of the if statement.

Morilli (approved):
Have tried to fix this myself and was very unsure about what function calls did what and which were required and which weren't (so I'm not really sure why exactly this method works), but I've tested it and it seems fine 👍

PR #3089, opened 2022-01-23
Lua double update fix

SuuperW:
Is it safe to move this code?

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -1736,8 +1736,15 @@ namespace BizHawk.Client.EmuHawk
 
 		private void UpdateToolsAfter()
 		{
-			Tools.UpdateToolsAfter();
-			HandleToggleLightAndLink();
+			if (IsTurboing)
+			{
+				Tools.FastUpdateAfter();
+			}
+			else
+			{
+				Tools.UpdateToolsAfter();
+				HandleToggleLightAndLink();
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


YoshiRulz:
First commit could be done simpler:
```diff
InputManager.AutofireStickyXorAdapter.IncrementLoops(Emulator.CanPollInput() && Emulator.AsInputPollable().IsLagFrame);

PressFrameAdvance = false;

+if (!(IsTurboSeeking && Emulator.Frame == PauseOnFrame.Value))
+{
	if (IsTurboing)
	{
		Tools.FastUpdateAfter();
	}
	else
	{
		UpdateToolsAfter();
	}
+}

if (!PauseAvi && newFrame && !InvisibleEmulation)
{
	AvFrameAdvance();
}
```
...though I haven't looked into the bug you're trying to fix, so the best solution might be something completely different. I don't understand the purpose of the second commit.

SuuperW:
The bug is that tools get updated twice in certain situations. This is just a waste of time in many cases, but there are three cases that I know of where it results in bugs:
1) RAM search, with automatic search every frame turned on. If you search for values that have changed, then updating twice results in all values being removed.
2) RAM watch. In some cases (I don't know the details), it will update the changes counter on the second update when there was no change.
3) Lua scripts that rely on emu.frameadvnace or event.onframeend. These can exhibit bugs when they update twice in one frame.

The second commit really isn't necessary, and looking more at other code it may be a bad idea. I'll not use that commit.

I will take your code suggestion, and commit it to master since there's no longer any worry about moving the code around.

PR #3093, opened 2022-01-24
Nymashock

CasualPokePlayer:
Nymashock! Nyma version of Octoshock. Not sure about everything in the settings, some buttons might need some overrides added (thanks shape buttons). This has less debugging features than Octoshock, so might be best to keep the old core around and have this as a default for TASers.


Spikestuff:
I'm going to start off by saying this is a real good improvement to the already existing PlayStation core as it provides the missing controllers and actually has a ~120% improvement of performance at least on my system.

I personally believe that the Sync Settings should be adjusted ahead of release, like it was done in Octoshock (at a latter version) and that being:
* Setting Controller 1 to Dual Analog: This will basically match the other core, as well as for first timers who don't really change the settings to actually be able to use the analog input if they have to, or just use the d-pad.
* Controller 2 to 8 - None: Now multitap is false at a later step, so in reality it's just needed for controller 2 to be set to None but for the sake of reexplaining (and this is also in the theoretically true of the memory card step), some games do memory card and controller checks to which then the game can take a bit of a longer time to load and can result in a change of RNG as well.
* Multitap - False: This step is already done 🙂
* Memory Card 1 - True
* Memory Card 2 to 8 - False: Same explanation as the Controller section, and the allowance is what should be taken priority.

The second issue is resolutions and the overscans, at the current time any title that is at the 240/288p resolution has about six times the horizontal overscan for the quick rundowns on what I checked:
* Crash Bandicoot 2: Cortex Strieks Back - 512x240 (No Overscan) -> 560x240 (Octoshock) -> 800x240 (Nymashock)
* Crash Bandicoot 2: Cortex Strieks Back (PAL) - 512x256 (No Overscan) -> 560x288 (Octoshock) -> 800x288 (Nymashock)
* Castlevania: Symphony of the Night - 512x240 (No Overscan) -> 280x240 (Octoshock) -> 400x240 (Nymashock)
* Bishi Bashi Special 3 - 320x240 (No Overscan) -> 350x240 (Octoshock) -> 400x240 (Nymashock)

The good news is that the wrong horizontal overscan doesn't make its way to 480/576i titles (however, it should be noted that I also only checked Tekken 3).

The extended bad news, which is what I found when quickly checking the PAL version of Crash Bandicoot 2: Cortex Strikes Back is that the PlayStation logo appears on the veritcal overscan when it shouldn't be there at all:
Crash2PALNyma

Basically really good waterboxing, just that there's overscan issues that exists in the dumping/screenshot layer where something needs to be tweaked in order to fix that issue.

nattthebear:
Please adjust this file's contents:  https://github.com/CasualPokePlayer/BizHawk/blob/3e39b722957d2007668a109db0e944a374374423/waterbox/nyma/make-all-released-cores.sh

nattthebear:
> Basically really good waterboxing, just that there's overscan issues that exists in the dumping/screenshot layer where something needs to be tweaked in order to fix that issue.

Sounds like we may need some adjustments to `ECL_EXPORT void FrameAdvance(MyFrameInfo& frame)`, as it's probably not handling all of the Mednafen frame metadata correctly.  I've seen this before -- "simple" cores don't fully utilize all of the fields that Mednafen fills out in a frame struct, so there were some things not handled.

alyosha-tas:
Awesome! @TizianoMattei please test for desyncs if you are able. Hopefully this clears up a bunch of such issues.

adelikat (approved):

TizianoMattei:
> Awesome! @TizianoMattei please test for desyncs if you are able. Hopefully this clears up a bunch of such issues.

I already am! Im speechless, cant believe this is happening!
A small problem that is slowing a bit my workflow to test out if desyncs and such are solved, is the input roll on tastudio having weird names: 
Immagine 2022-01-24 180022


The directional keys are just (!) on the input roll, thats the main """issue""" atm, but im besides this small detail im not having any problem so far!

CasualPokePlayer:
I believe the above commit should fix these weird overscan issues. Turns out the "0" default I copy pasted from Saturn code just ends up being "false" internally but the settings display it as "true" (perhaps a bug?).

Spikestuff:
First thing's first, the overscan issues have been dealt with, thank you for correcting that.
Still have that issue that can be found in Crash Bandicoot 2: Cortex Strikes Back (Europe) but it's getting there.
Crash2PALNymaUpdate

First issue I noticed is with the other ports are still actually avaiable to not only use (Virtual Pad), but also to edit in Config > Controllers.
The alternative way of verifying is to use Point Blank, where despite being set to "none" the Blue Crosshair of player 2 was visiable when it shouldn't be.
image

The good news to this in comparison is that there's only one Memory Card inserted so it doesn't extend to here.
This was mainly for sanity checking.

Another issue with the player 1 controller being default set to DualShock is that you cannot config the controller to have Analog, meanwhile either setting it to Analog Controller or having Player 2 (or higher) to DualShock reveals the option for that controller port.

Last thing is that doing the Symbols for Triangle, Cross, Square, Circle results to this:
image

CasualPokePlayer:
> First thing's first, the overscan issues have been dealt with, thank you for correcting that. Still have that issue that can be found in Crash Bandicoot 2: Cortex Strikes Back (Europe) but it's getting there.
> 
> First issue I noticed is with the other ports are still actually avaiable to not only use (Virtual Pad), but also to edit in Config > Controllers. The alternative way of verifying is to use Point Blank, where despite being set to "none" the Blue Crosshair of player 2 was visiable when it shouldn't be.
> 
> The good news to this in comparison is that there's only one Memory Card inserted so it doesn't extend to here. This was mainly for sanity checking.
> 
> Another issue with the player 1 controller being default set to DualShock is that you cannot config the controller to have Analog, meanwhile either setting it to Analog Controller or having Player 2 (or higher) to DualShock reveals the option for that controller port.
> 
> Last thing is that the Symbols for Triangle, Cross, Square, Circle results to this: 

Issues 1 and 2 was due to the same issue, simply being there was a mismatch between what the nyma settings showed and what the core (and virtualpad/controller config) saw, the latter seeing all gamepads. The issue should be fixed in the latest commit.

Spikestuff:
> The issue should be fixed in the latest commit.

Yup those two issues are fixed.
The timing was honestly funny since I let you know privately beforehand and just want to let a public document of what's up.

nattthebear (approved):

PR #3098, opened 2022-01-27
TAStudio update self

SuuperW:
Commit 319455c reverted one of my previous commits, commenting that it caused a regression.
These commits redo my commit, along with fixing the regression and a very similar bug.

Morilli, you said that you would work on "a different fix". Have you begun this yet?

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs
@@ -19,7 +19,6 @@ namespace BizHawk.Client.EmuHawk
 		[OptionalService]
 		public ISaveRam SaveRamEmulator { get; private set; }
 
-		private bool _hackyDontUpdate;
 		private bool _initializing; // If true, will bypass restart logic, this is necessary since loading projects causes a movie to load which causes a rom to reload causing dialogs to restart
 
 		private int _lastRefresh;
@@ -63,11 +62,6 @@ namespace BizHawk.Client.EmuHawk
 		protected override void UpdateAfter()
 		{
 			if (!IsHandleCreated || IsDisposed || CurrentTasMovie == null)
-			{
-				return;
-			}
-
-			if (_hackyDontUpdate)
 			{
 				return;
 			}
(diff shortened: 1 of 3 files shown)


SuuperW:
4 commits? Oops, I forgot to push those to master before creating this PR.
I have pushed the first two commits to master, they should not be a part of this PR. I'm not familiar enough with git to know if there is a way to get them to not show up here anymore without closing this PR and creating a new one.

Morilli:
> Morilli, you said that you would work on "a different fix". Have you begun this yet?

When testing I could not see your commits actually fixing the bugs that you mentioned (specifically, clicking on the first frame would not update the emulator window, which is what I assume you meant), so in light of that, I didn't know how to proceed and just left as-is.

If you *can* measure any use from dropping the `_hackyDontUpdate`, I would be interesting in being able to verify them for myself (I will be testing this PR later).

SuuperW:
I was talking about the TAStudio window. Here are steps to reproduce the two bugs on the master branch. On this branch both expected behaviors are observed.

1) Open TAStudio.
2) Advance to frame 1.
3) Click on the cursor column for frame 0.
Expected behavior A: The current frame cursor and current frame highlighting show on frame 0.
Actual behavior A: The current frame cursor and current frame highlighting remain on frame 1.
4) With the current frame still at 0, save the tasproj.
5) Open the tasproj that was just saved.
Expected behavior B: The current frame should be 0.
Actual behavior B: The current frame is 1.

The emulator window not updating is a separate bug, which I made no attempt to fix.

I also just noticed that I did not remove a comment in the code or rename the method. Both of those should be done, I will do them now.

vadosnaprimer:
If you mean analog controls, you enter analog editing mode by double-clicking a cell and then enter numbers by:
- inputting digits
- dragging mouse pointer up and down
- using arrow keys

Morilli:
Okay I've checked analog controls now, and I'm actually kinda confused what was broken before that needed fixing.

Current master behavior:
Clicking on an analog control input in TAStudio will not scroll the view or seek to that frame. When double clicking it, the value can be set correctly by dragging the mouse or entering values, and a seek will only start when the mouse button is released (or immediately upon entering any value with the keyboard).

Behavior on this PR: 
Clicking on an analog control will cause a seek (if needed) to immediately start, which makes double clicking basically impossible as the view will immediately scroll upon the first click. 

So how is this behavior better than the previous? I would argue that this feels worse than before.

SuuperW:
Painting analog controls works the same as painting boolean values, except that you paint the same value that was clicked rather than toggling the value before painting. So, if you set an analog control to 9, you can then click+drag to paint input. Or, you can click a blank analog cell and drag to paint over the 9. This painting had the same scrolling before mouse up issue as boolean values did with the regression.

However you are correct that double-clicking to edit analog values that are earlier than current frame is messed up by that commit. (I just tested it.) I will push a fix for that.

Morilli:
Have chcked myself and can confirm both scenarios work correctly and as expected now.

PR #3100, opened 2022-01-28
Run testroms against SameBoy

YoshiRulz:
I have most of the tech is working locally, just need to tweak timing and update lists.


YoshiRulz:
Pushed https://github.com/TASEmulators/BizHawk/commit/5d92ca3b4fe132a5fe923537c33c25a4c0b05611 and 2 prior commits to do this.

Issue #3102, opened 2022-01-28
BSNESv115+: Savestate size is too large


Morilli:
When attempting to TAS on the new bsnes core, I noticed greenzone is much less frequently created than on the old core, most likely due to the savestate size being much higher. In the rewind settings, an average state size is displayed, which is ~350KB for old bsnes and 3-3.5MB for the new bsnes core.
This is not a huge issue for casual play when delta compression is enabled, as that shrinks the average state size down to about 20KB, but for TAStudio, where delta compression is not (and cannot be?) used, this quickly causes there to be large gaps between savestates (there is not even one savestate per frame on the current location, at least not with default settings).
Besides that, the savestate size is clearly unnecessarily large.

I see two potential ways to resolve this: 
1. Use bsnes's builtin `serialize` function, which *does* seem to work deterministically (unlike old bsnes, it does not need to synchronize co_threads to create a savestate, as long as the current libco version is used), which would require a libco update. That itself is an issue currently, as just slamming in the new libco version crashes the waterboxhost unless it is built in `no-dirty-detection` mode, which tells me something is pretty broken there (see also my proof-of-concept commit [here](https://github.com/Morilli/BizHawk/commit/8c152686ede3571bdaa8fd9ef6f59968fbc25729)).
2. Figure out what is causing the savestate size to be this large and use `alloc_invisible` or the like calls inside the core to mark those memory regions as non-savestatable. I have attempted this with the obvious big `new[]` calls, but did not see a reduction in savestate size. Would need an investigation to figure out the problematic spot.

CasualPokePlayer:
This really ends up affecting a lot more waterbox cores than just bsnes w/e (e.g. melonDS), which I've just always doubled/tripled/quadrupled savestate size with all the tastudio settings. Dynamically changing the default savestate settings could be better here? That or just get the damn delta thingie to work here (zlib compression seemingly works but is slow, why not get delta compression working?)

Dirty detection should catch anything allocated for giant ROMs and what not, and generally should be catching things which do not need to be savestated. Usually there isn't need for core modifications to reduce state besides dumb cases like memory buffers being deallocated then reallocated on hard reset.

Morilli:
> Perhaps old bsnes commits could shed some light on this, considering that itself predates the dirty detection model and likely have/once had modifications done to reduce state size.

Old bsnes did heavily use `alloc_invisible` calls and even communicates with the frontend to let it allocate some thing for it iirc, which is why I even considered that.

[...]

PR #3105, opened 2022-01-29
Reduce bsnes115 state size

CasualPokePlayer:
resolves https://github.com/TASEmulators/BizHawk/issues/3102


nattthebear:
This seems reasonable.  For each thing that you're moving to invisible, could you give a short explanation of what it is, where it's used, and why it's safe to not save?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesApi.cs
```diff
@@ -116,9 +116,9 @@ public BsnesApi(string dllPath, CoreComm comm, IEnumerable<Delegate> allCallback
 				Filename = "bsnes.wbx",
 				Path = dllPath,
 				SbrkHeapSizeKB = 14 * 1024,
-				InvisibleHeapSizeKB = 4,
+				InvisibleHeapSizeKB = 200 * 1024,
 				MmapHeapSizeKB = 110 * 1024, // TODO: check whether this needs to be larger; it depends on the rom size
-				PlainHeapSizeKB = 0,
+				PlainHeapSizeKB = 110 * 1024,
```
Morilli, line 121: These should ideally be adjusted to the lowest size that consistently works (and you can reduce the `MmapHeapSize` and `SbrkHeapSize` in the process`)

CasualPokePlayer:
Most of the things here are just used as cache for rendering with fast PPU, output is just ppu rendering output (invisible on old bsnes), lookup/target are mapped on bootup and only unmapped on unload (not supported, if you want to unload you throw the whole core out and get a new one). Overkill perhaps but other changes there to prevent deallocation/reallocation on reset. Things allocated with the ReadableMemory struct are just read only and do not need to be stated (changing to invisible here is probably overkill, although changes are there so it won't be deallocated/re-allocated on hard reset, ditto for WritableMemory/ProtectableMemory, which I shove into the plain pool for some kind of consistency).

A weird thing here is that I tried to move the output buffer of the slower PPU to the invisible pool, which seemed to crash games on bootup? Not even with any states loaded, just loading up the core with a game and black screen forever. It works fine if the buffer is allocated with new, but not with alloc_invisible.

waterbox/bsnescore/Makefile
```diff
@@ -1,7 +1,7 @@
 NEED_LIBCO := 1
 
 CXXFLAGS := -std=c++17 \
-	-I../libco -I./bsnes \
+	-I../emulibc -I../libco -I./bsnes \
```
Morilli, line 4: This should not be required, as it gets included by `common.mak` below

waterbox/bsnescore/bsnes/sfc/ppu-fast/ppu.cpp
```diff
@@ -52,13 +54,21 @@ PPU::PPU() {
   }
 
   for(uint y : range(240)) {
+    lines[y].cgram = (uint16_t*)alloc_invisible(256 * sizeof (uint16_t));
+    lines[y].items = (ObjectItem*)alloc_invisible(128 * sizeof (ObjectItem));
+    lines[y].tiles = (ObjectTile*)alloc_invisible(128 * sizeof (ObjectTile));
+    lines[y].above = (Pixel*)alloc_invisible(256 * 9 * 9 * sizeof (Pixel));
+    lines[y].below = (Pixel*)alloc_invisible(256 * 9 * 9 * sizeof (Pixel));
+    lines[y].windowAbove = (bool*)alloc_invisible(256 * sizeof (bool));
+    lines[y].windowBelow = (bool*)alloc_invisible(256 * sizeof (bool));
```
Morilli, line 63: Consider using the C++ version `alloc_invisible<T>(size)` instead of `(T*) alloc_invisible(size * sizeof(T))` (here and in all places where it makes sense)

waterbox/bsnescore/bsnes/sfc/ppu-fast/ppu.cpp
```diff
@@ -52,13 +54,21 @@ PPU::PPU() {
   }
```
Morilli, line 55: How about allocating `lines` itself here as well? Considering `sizeof(Line) == 312`, a `lines = alloc_invisible<Line>(240);` should give another ~70KB boost.

nattthebear:
For the things that are written once but never changed afterwards, we could use alloc_sealed -- so long as it's written during the initial boot/init process and not afterwards, savestates won't contain it.  The benefit is that when you use alloc_sealed on something that you think never changes, if it does happen to change, you crash, whereas with invisible, if it does happen to change, you silently nondeterminize savestates.

waterbox/bsnescore/bsnes/sfc/ppu-fast/ppu.cpp
nattthebear: I think the bsnes 87 core predates the C++ typed overloads on the alloc functions, which is why you wouldn't have seen them there.

CasualPokePlayer:
> For the things that are written once but never changed afterwards, we could use alloc_sealed -- so long as it's written during the initial boot/init process and not afterwards, savestates won't contain it. The benefit is that when you use alloc_sealed on something that you think never changes, if it does happen to change, you crash, whereas with invisible, if it does happen to change, you silently nondeterminize savestates.

ROMs and shit are made writable I believe with memory domains? Don't want to crash that at least. A lot of these things anyways do end up re-written on hard reset the exact same way as initial core boot (which I've made to make them not deallocate, although they might still be writing a copy of itself on hard reset still). Not sure if going through the logic here is very worth it here.

waterbox/bsnescore/bsnes/sfc/ppu-fast/ppu.cpp
CasualPokePlayer: Ah, I was largely looking at old bsnes core for clues so didn't know c++ got some nice options for this.

waterbox/bsnescore/bsnes/sfc/ppu-fast/ppu.cpp
Morilli: 312 is the size after, and in my testing I did see a decent reduction in state size.

CasualPokePlayer: Not sure of the consequences of putting a struct with functions into the invisible pool, paging @nattthebear here.

nattthebear: The pointers are just as stable as any others, so the same downside as everything else:  If you're wrong about it, the core becomes nondeterministic, or maybe crashes if it expects something to be restored there that's not.

nattthebear:
> ROMs and shit are made writable I believe with memory domains? Don't want to crash that at least. A lot of these things anyways do end up re-written on hard reset the exact same way as initial core boot (which I've made to make them not deallocate, although they might still be writing a copy of itself on hard reset still). Not sure if going through the logic here is very worth it here.

The hard reset case means that the core will balloon in savestate size after a hard reset, then.  (It wouldn't be the first waterbox core to do so...)

CasualPokePlayer:
Hence why I'm slapping them in invisible so the ROMs/etc don't get shoved into states (which before these changes a hard reset throws state size to 12 MB, with these changes there aren't any changes).

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesApi.cs
```diff
@@ -115,10 +115,10 @@ public BsnesApi(string dllPath, CoreComm comm, IEnumerable<Delegate> allCallback
 			{
 				Filename = "bsnes.wbx",
 				Path = dllPath,
-				SbrkHeapSizeKB = 14 * 1024,
-				InvisibleHeapSizeKB = 4,
-				MmapHeapSizeKB = 110 * 1024, // TODO: check whether this needs to be larger; it depends on the rom size
-				PlainHeapSizeKB = 0,
+				SbrkHeapSizeKB = 12 * 1024,
+				InvisibleHeapSizeKB = 136 * 1024, // TODO: check whether this needs to be larger; it depends on the rom size
```
Morilli, line 119: This comment has changed positions now, which I don't think is right, considering roms are still saved as vector, which uses `new` internally (see for examples bsnescore.cpp line 204).

waterbox/bsnescore/bsnes/sfc/ppu-fast/ppu.hpp
```diff
@@ -360,13 +361,23 @@ struct PPU : PPUcounter {
...
+
+  typedef struct {
+    int count = -1;
+    int startLine[32];
+    int endLine[32];
+    int startLerpLine[32];
+    int endLerpLine[32];
+  } Mode7LineGroups;
+  
+  Mode7LineGroups* mode7LineGroups;
```
Morilli, line 380: Am I seeing it right that this is saving exactly 516 bytes? If so, I'm against having to do such a big change/diff like in `mode7hd.cpp` just to save this little space (it's like 1 or 2 promille of the entire savestate size, not worth it).

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesApi.cs
CasualPokePlayer: Ah, right. I guess it's more technically being double loaded here, as copies would be being placed in invisible heap. Not ideal but w/e.

nattthebear:
> Hence why I'm slapping them in invisible 

If the writes on hard reset doesn't matter, and can be safely skipped, why not remove them entirely?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesApi.cs
Morilli: Why would copies be placed in the invisible heap? And where do they come from?

CasualPokePlayer:
> > Hence why I'm slapping them in invisible
> 
> If the writes on hard reset doesn't matter, and can be safely skipped, why not remove them entirely?

Can't be removed entirely, power (which reset calls) is called on core bootup and is used a general initializer. Example could be the lookup and target LUTs that get init on bootup. Code for mapping is scattered around, they all happen on load or power functions in many different classes, you'd have to do some somewhat complicated logic to make it not just map again on hard reset. Same logic applies for ROMs and shit since there's several ReadableMemory structs floating all around (thanks co-processors). It'd just blow up diff size really.

nattthebear:
I'm not convinced that this is bulletproof :/  What's our plan to make sure it causes 0 determinism issues?

CasualPokePlayer:
Could just not merge and actually go make the delta compressor work lol

nattthebear:
Can the bsnes core be savestated in the middle of a video frame, and if so will that cause a problem here?  It's fine if the visual output from a single frame is bad after such a loadstate, but crashes or determinism failures are no good.  Alternatively, if the core always stops in vblank, that's fine too.

CasualPokePlayer:
Maybe it could cause a problem if we ever decide to make a sub-frame core out of this. Could maybe need to remove the cache invisibility if that ever happens. But that's neither here or there for now.

Morilli (approved):
All fine from my side now.

nattthebear (approved):
Have fun debugging this when it breaks!

PR #3110, opened 2022-01-30
MSXHawk Linux

CasualPokePlayer:
Noting there were some uninitialized warnings with Ztemp2 when going through with this, no idea what's up with that. Potentially bad?


alyosha-tas:
Neat! Looks good. For Ztemp2 it just doesn't like that I put in the instruction builder without initializing it. It gets overwritten on instruction call so it doesn't matter. You can just initialize it to zero if you want to gt rid of the warnings.

alyosha-tas (approved):

Issue #3060, opened 2022-01-05
BSNESv115+ hex editor: viewing bytes at location hex:xx2130 causes graphical bugs


getCursorsExe:
xx is the bank (set of 0xFFFF bytes)

Morilli:
Judging from a comment in the disassembler part of the code
```
    //$00-3f,80-bf:2000-5fff: do not attempt to read I/O registers from the disassembler:
    //this is because such reads are much more likely to have side effects to emulation.
```
, it seem that certain system bus reads may have emulation side effects.
I'm not entirely sure why those exist or how to prevent them.

CasualPokePlayer:
> Judging from a comment in the disassembler part of the code
> 
> ```
>     //$00-3f,80-bf:2000-5fff: do not attempt to read I/O registers from the disassembler:
>     //this is because such reads are much more likely to have side effects to emulation.
> ```
> 
> , it seem that certain system bus reads may have emulation side effects. I'm not entirely sure why those exist or how to prevent them.

In general having something like "peek at system bus the same way as the CPU" can have side-effects if it causes the state of various components to be updated (common for lazy evaluation emulation etc). You need to either have them not be updated and try to return whatever is present (which granted may be "wrong") or just block the reads.

getCursorsExe:
I think the culprit is the address xx213C and xx213D, which is the location of the scanning beam.

PR #3118, opened 2022-02-04
hack a peek into bsnesv115

CasualPokePlayer:
Resolves https://github.com/TASEmulators/BizHawk/issues/3060 I guess


Morilli:
Any specifics on which regions were made peekable and which not? Like, is there any reason e.g. the entire `APU` region is not peekable?

CasualPokePlayer:
Anything region I saw that updated emulator state I marked as not peekable (APU did a synchronize which would update emulator state).

In the future it could be worth it to make separate methods for peeks which avoid updating emulator state, but that would require a lot of work I don't care for doing right now.

Morilli (approved):
Given the above reasoning, I'll approve this one for now.

Not returning "correct" system bus values is preferrable over potentially altering emulation state.

PR #3130, opened 2022-02-07
Fix Deadlock in DiscoHawk

### LINKED ISSUE #3128: DiscoHawk chokes on Silverload
1. Load Silverload (PS1) in DiscoHawk as a .ccd file. Use "Bizhawk" mode.
2. Discohawk stops responding.
3. It loads about 928 MB, and seems to get stuck.
4. Try again, but with "Mednafen" mode.
5. It loads about 440 MB into RAM, and seems to get stuck.

The full .img is only 650 MB. Track 2 is about 435 MB.

If you kill the ffmpeg subprocess, then DiscoHawk continues and finishes track 03 and 04 with no problems.

It's kind of weird that this file seems to really not work well with DiscoHawk. Track 02 is 15+ minutes long, so there might be some kind of issue when processing long tracks?

Edit: I let it run for 12+ hours, but it made no more progress than originally.

Meerkov:
This code deadlocks if the Standard Error stream fills up before the standard output stream does.

Code based on https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-6.0

Fixes #3128


Meerkov:
It looks like I was working on a bit of an older source. This fix from 4 months ago looks bad. Can we use my 2 line fix instead?
https://github.com/TASEmulators/BizHawk/commit/253d532a63c7073068bec246db288dc6a84cc180

zeromus:
1. mine interleaves stdout and stderr (maybe.. or maybe not). at least you should fix yours to do that
2. no. don't mess with it unless you have reason to believe it's failing. if anyone runs into any trouble with mine
3. your eOut string begins null... how does this even work at all? it should be initialized to empty string instead
4. if we're revising this we should at least use a stringbuilder. I have no idea why i didn't do that originally

Meerkov:
I'm bad at git, so I made a new pull request. #3133

PR #3133, opened 2022-02-08
Interleave output and error async in Discohawk

Meerkov:
Fixes deadlocks, and removes confusing custom parser.


Meerkov:
Thanks for reviewing @zeromus ,
1. This new one interleaves the output asynchronously, using mutexes for extra safety.
2. I disagree, the code is confusing, and confusing card is hard to maintain. I don't think this code should be doing any parsing of the contents of the output of ffmpeg.
3. Good catch, I don't know why my first version worked either. It's undefined behavior, I guess.
4. StringBuilder is now used, thanks.

Issue #3073, opened 2022-01-13
Genesis Game Demons of Asteborg Not Emulated Properly (SRAM Fails to Load)


ShinobiWannabe:
Demons of Asteborg (https://www.demonsofasteborg.com/) does not load the SRAM properly.  This is fixed in the genesis plus gx according to the link below.  I have tested the current build of Bizhawk alongside 2.6.2 and 2.7.0. Older Bizhawks like 2.3.2 just displays graphical garbage on loading the game.

https://github.com/ekeeke/Genesis-Plus-GX/issues/381

I think what is needed is to update using Waterbox for the genesis plus gx core?  

The way to tell the game is not loading properly is to run the game and when you start a new game no difficulty or save slot is selectable and no files can be saved. 

https://youtu.be/3lPBlLZvgXk?t=335

I have not tested Retroarch, but Blastem does load the rom and SRAM as long as you do a hex edit where you replace "SEGA DOA" with "SEGA SSF" in the header. https://www.reddit.com/r/SEGAGENESIS/comments/p1x3hh/demons_of_asteborg_review/

Rom Info
DemonsOfAsterborg v1.1
SHA1:3274194684FE28DA04F6E57981CA7B00E10835D2
MD5:7EF77BBB609124BCFC18FA400DE99BF3

PM me CoolHandMike#0352 on Discord if necessary.
![image](https://user-images.githubusercontent.com/34147391/149338481-9eab5e5e-9ddb-4ac5-bf8a-d99c4c365113.png)

![image](https://user-images.githubusercontent.com/34147391/149338396-4e99b860-42bd-4c05-aeb0-41736fcc1b17.png)

CasualPokePlayer:
Fixed by https://github.com/TASEmulators/BizHawk/pull/3139

PR #3139, opened 2022-02-14
Genesis GPGX SRAM support for games larger than 8MB

ShinobiWannabe:
Cherry picked changes in order to run games with SRAM larger than 8MB based on the existing commit by ekeeke in Genesis-Plus-GX.

https://github.com/ekeeke/Genesis-Plus-GX/commit/e6ab7563e7cfb4e18c29e557d36e59cc51e83815

This should resolve "Genesis Game Demons of Asteborg Not Emulated Properly (SRAM Fails to Load)".
 https://github.com/TASEmulators/BizHawk/issues/3073

__
Tested These to Run to End. Last change to GPGX core in Bizhawk was 2.5 apparently.
[Genesis Barney's Hide & Seek Game by Samsara & Spikestuff in 04:54.51](https://tasvideos.org/4580M)
[Genesis Beavis and Butt-Head "2 players" by adelikat, EZGames69 in 09:29.37](https://tasvideos.org/4426M)
[Genesis Flicky by ParadaxeTH in 14:40.25](https://tasvideos.org/4375M)

Thanks to RiYAN for showing me that the needed change on cd_cart.h worked fine.


waterbox/gpgx/core/cd_hw/cd_cart.h
```diff
@@ -40,7 +40,7 @@
  /* CD compatible ROM/RAM cartridge */
 typedef struct
 {
-  uint8 *area;    // always 16K in bizhawk
+  uint8 area[0x830000]; /* cartridge ROM/RAM area (max. 8MB ROM + Pro Action Replay 128KB ROM / 64KB RAM) */   // always 16K in bizhawk
```
nattthebear, line 43: Why are we static allocating an extra 8MB of junk at all times now?  Dirty states reduces the perf impact of this, but there's still some perf impact.

ShinobiWannabe: You should ask ekeeke.  I just tried to get the changes as close as possible to his commit that fixed the issue.  If you think it would be best to change it back, then by all means go ahead.  I want that Demons of Asteborg game to work without introducing breaking changes.

nattthebear: ekeeke's answer won't be relevant here, as they had no requirements to adhere to the Waterbox memory model and they were changing code that was already diverged between the two branches; in the upstream commit it's changed from a static 0x840000 to 0x830000, but here it was dynamically allocated.

There's clearly some other mismatch going on between the code, as the original comment mentions that this is holding both cartridge ROM and some other things like Action replay, where as we had an "always 16k" comment.

This PR looks quite sloppy and poorly-researched with that juxtaposition:

```c
uint8 area[0x830000]; // always 16K in bizhawk
```

If I recall correctly, a lot of this was work I did to make things more amenable to the waterbox memory model; the old code punned a single ROM area for a lot of things, but for waterbox it made sense to have the ROM be actually ROM (in the sealed area) to keep state size down.

With lazystates, this may not be as relevant anymore, but if that's our path forward, we'll want to completely and holistically change the memory mapping back to match upstream, not half-change things like here.

nattthebear: SRAM change should be fine.  Judging from the number of comments related to sanic and kunkle, someone should test that game on the new code as well.

ShinobiWannabe: Thank you.  I will remove those changes to cd_cart.h and cd_cart.c and rebuild gpgx.  Test those tases again as well as S&K.

ShinobiWannabe: Made those changes as above and committed. Should not show the cd_cart.h and cdcart.c, and the gpgx.wbx.gz file changing in size slightly.

Tested those tases as done previously and played some of s&k, sonic 3 w/ knuckles, sonic2 w/ knuckles, sonic 3, sonic 2.

nattthebear (approved):

PR #3169, opened 2022-02-28
BSNES Region Override

CasualPokePlayer:
Someone wanted this and Faust had it so eh


waterbox/bsnescore/bsnes/target-bsnescore/bsnescore.cpp
```diff
@@ -122,32 +122,41 @@ EXPORT void snes_set_callbacks(SnesCallbacks* callbacks)
     snesCallbacks = SnesCallbacks(*callbacks);
 }
 
-EXPORT void snes_init(int entropy, uint left_port, uint right_port, uint16_t merged_bools)// bool hotfixes, bool fast_ppu)
+struct SnesInitData {
+    int entropy;
+    uint left_port;
+    uint right_port;
+    int hotfixes;
+    int fast_ppu;
```
Morilli, line 130: Any reason for making these `int` instead of `bool`?

CasualPokePlayer: /shrug

CasualPokePlayer: Well apparently it breaks as bool, fun

Morilli:
I have pushed a commit that should make the struct definition a bit cleaner on the C# side (and should work), please check that those changes are fine with you.

The rest of the changes look good, just a couple thoughts for later: `ENTROPY` and `REGION_OVERRIDE` should probably be *enum*'d on the c++ side as well for clarity on what the individual values mean, and as per the comment in [IVideoProvider](https://github.com/TASEmulators/BizHawk/blob/master/src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.IVideoProvider.cs#L8), the selected (or determined) region should most likely affect the aspect ratio as well.

CasualPokePlayer:
Seems good to me.

Issue #3082, opened 2022-01-20
add nes rom to data base? NRS 2.0 headered


Eggo-dotcom:
```
Inferred from iNES header; potentially wrong

------
BEGIN NES rom analysis:
Found iNES header:
pr=64,ch=16,wr=8,vr=0,ba=0,pa=1|0,brd=MAPPER066,sys=
Found iNES V2 header:
pr=64,ch=16,wr=0,vr=0,ba=0,pa=1|0,brd=MAPPER0066-00,sys=NES-NTSC
Since this is iNES we can (somewhat) confidently parse PRG/CHR banks to hash.
headerless rom hash: SHA1:5F3091D337063BFC1031666E7114446E2C130560
headerless rom hash:  MD5:D148B0E2674CC2BAC57446F635D11AE2
Could not locate game in bizhawk gamedb
Could not locate game in nescartdb
Attempting inference from iNES header
Failed to load as iNES V2
Chose board from iNES heuristics:
pr=64,ch=16,wr=8,vr=0,ba=0,pa=1|0,brd=MAPPER066,sys=
Final game detection results:
pr=64,ch=16,wr=8,vr=0,ba=0,pa=1|0,brd=MAPPER066,sys=
""
Implemented by: class GxROM
END NES rom analysis
------
```
```
Inferred from iNES header; potentially wrong

------
BEGIN NES rom analysis:
Found iNES header:
pr=64,ch=16,wr=8,vr=0,ba=0,pa=1|0,brd=MAPPER066,sys=
Found iNES V2 header:
pr=64,ch=16,wr=0,vr=0,ba=0,pa=1|0,brd=MAPPER0066-00,sys=NES-NTSC
Since this is iNES we can (somewhat) confidently parse PRG/CHR banks to hash.
headerless rom hash: SHA1:5F3091D337063BFC1031666E7114446E2C130560
headerless rom hash:  MD5:D148B0E2674CC2BAC57446F635D11AE2
Could not locate game in bizhawk gamedb
Could not locate game in nescartdb
Attempting inference from iNES header
Failed to load as iNES V2
Chose board from iNES heuristics:
pr=64,ch=16,wr=8,vr=0,ba=0,pa=1|0,brd=MAPPER066,sys=
Final game detection results:
pr=64,ch=16,wr=8,vr=0,ba=0,pa=1|0,brd=MAPPER066,sys=
""
Implemented by: class GxROM
END NES rom analysis
------
```

[...]

PR #3170, opened 2022-03-01
Fix SXROM Detection

### LINKED ISSUE #3168: MMC1 SXROM Support
Hi, BizHawk doesn't properly support the SXROM variant of MMC1. The Z1M1 (Zelda 1 / Metroid 1) crossover randomizer requires SXROM and tests for 32 KB of bankable PRG-RAM on boot. This test fails on BizHawk.

[This nesdev link](https://wiki.nesdev.com/w/index.php/MMC1#SOROM.2C_SUROM_and_SXROM) contains the relevant information on SXROM. This is the error screen I see in BizHawk:
screencap

SXROM is also used by the official Final Fantasy I+II multicart, so support for it is useful for the commercial library. I have not seen any other bug reports related to this issue.

Thanks for your time!

CasualPokePlayer:
fixes #3168. 

Skeptical about these changes as I also discovered iNES 2.0 detection ended up producing 4 digits for the board number instead of 3 like iNES 1.0 does. I assume 3 digits is the intended width (otherwise iNES 2.0 seems to just be useless?). Not sure if this is correct however. EDIT: Might be related to https://github.com/TASEmulators/BizHawk/issues/3082#issuecomment-1017690152

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.iNES.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.iNES.cs
@@ -52,11 +52,11 @@ namespace BizHawk.Emulation.Cores.Nintendo.NES
 				// only add submapper if it is non-zero
 				if (submapper != 0)
 				{
-					CartV2.BoardType = $"MAPPER{mapper:d4}-{submapper:d2}";
+					CartV2.BoardType = $"MAPPER{mapper:d3}-{submapper:d2}";
 				}
 				else
 				{
-					CartV2.BoardType = $"MAPPER{mapper:d4}";
+					CartV2.BoardType = $"MAPPER{mapper:d3}";
 				}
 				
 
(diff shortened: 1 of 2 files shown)


nattthebear:
The four digit numbers are intentional.  If you want a board to respond specifically to the iNES 2.0 signature, you add that string to the `Configure` method as an option.  There are very few of them because most of our detection is iNES 1.0, and when we added iNES 2.0 we didn't want any backwards compatibility issues; we didn't want any already existing correct detections to function differently or fail.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.iNES.cs
```diff
@@ -52,11 +52,11 @@ public static bool DetectFromINES(byte[] data, out CartInfo Cart, out CartInfo C
 				// only add submapper if it is non-zero
 				if (submapper != 0)
 				{
-					CartV2.BoardType = $"MAPPER{mapper:d4}-{submapper:d2}";
+					CartV2.BoardType = $"MAPPER{mapper:d3}-{submapper:d2}";
```
nattthebear, line 55: Don't change this or the lines below, you will break existing working identifications

nattthebear:
Thanks for pointing out https://github.com/TASEmulators/BizHawk/commit/4ff77456e49c662612fe7a57200ff935d490a545; that was incorrect and should be reverted.

The point of this is that all iNES detection logic is a crapshoot, but it only makes sense to shoot craps with whatever data you actually have.  For instance,  MAPPER0001 should use the wram size to choose between SUROM and SXROM so people can play their pirate garbage.  But for MAPPER001, that makes no sense, since there is no such thing as wram size in the iNES 1.0 header.

nattthebear (approved):

PR #3174, opened 2022-03-05
Melon Cleanups & Updates & Prep For Multithread Rendering Support & Prep For DSi NAND bs

CasualPokePlayer:
Tries to implement the waterbox multithreading for Melon's threaded renderer. Seems to deadlock after a random amount of time, not sure what's wrong :/


nattthebear:
Ahh right, `_frameThreadStart`.  What a fucking mess.

1. The code run by _frameThread is forbidden to make any stdlib calls at all.  _This includes heap allocations._  __This includes threading synchronization primitives.__
2. The code in the main thread is forbidden to use stdlib treading synchronization primitives to speak to the frame thread.  All synchronization must be done by atomics only.
3. It's a wonky pile of shit.

CasualPokePlayer:
Right, I read the big warning about that waterbox unsafe shit. I kinda ended up just bullshitting a Semaphore using atomic operations and spin looping the thread. It seems to generally work, although I'm randomly getting deadlocks after a random amount of use, so I'm not sure if I did that right.

I also had to do some bullshit in order to make the render thread stop once we're done with the frame. Upstream logic in melon is to forever spin the render thread until changing renderer/shutdown. So I stop the render thread after a RunFrame.

Which on that note shutdown by firmware hardlocks, so fun.

nattthebear:
Yeah, having the render thread return after every frame is critical; we can't save meaningful savestates if any actual thread of execution is running in the core when we do (because we won't be able to restore it).

CasualPokePlayer:
And of course "can't call stdlib functions" means "can't printf abuse for debugging." And I've learned the hard way that callbacks to c# code are definitely not supported.

So figuring out random deadlocks is not going to be easy. I can't see anything wrong with the Semaphore hackery I put in. The best I can even get out from gdb is there's some crash in or after SetupPolygonRightEdge...

The speed benefit from this is rather minor anyways, might just ifdef this out and come back to it later.

PR #3176, opened 2022-03-08
Fix #3173 by only calling bus.map() on initial power, not subsequent calls

### LINKED ISSUE #3173: Loading savestate after reset is broken on bsnesv115+
### Emulator
Bizhawk 2.8

### Repro
1. Load any snes games with said core.
2. Hard reset at any point.
3. Try to load savestate.

That process leads crashing emu without warning, making emu stuck loading state for good or open state normally. That didn't occur in 2.7.

Morilli:
I'm honestly not 100% happy with this change, but it logically more or less matches what was done in the old core as well, and it also makes sense to not do those calls again anyway.

I have also tried to `alloc_invisible` the reader, writer and counter in `sfc/memory.hpp` instead, which *SHOULD* have worked imo, but somehow it always crashed badly and displays a `0x... in mal ()` in gdb, so it must be corrupting the stack somehow...

As a bit more background what is even broken: `reset` and `power` calls cause the core to re-map memory regions in the system bus, and because of the way it is implemented internally in `sfc/memory.cpp`, the `id` for those regions change and therefor, when a savestate is loaded and `lookup` is not restored (`target` doesn't matter actually), the id is mismatched between the reader, writer and counter arrays and the `lookup` array and... well, things go badly.


nattthebear:
What I'm not understanding is why the remapping on repower doesn't produce the exact same ids as the first time around.

nattthebear:
If `auto Bus::reset() -> void {` is called then that would reset all of the `counter` values right?  So when it went to remap everything, it should start at the first slot again and presumably map them all in the same order

Morilli:
> If auto Bus::reset() -> void { is called then that would reset all of the counter values right?

It should, yes. However, `bus.reset()` is not called on savestate load, so that doesn't happen. I guess it could potentially be put in the frontend to the loadSavestate or whatever call, although we'd probably have to call `system.load()` in its entirety as well then?

nattthebear:
> It should, yes. However, bus.reset() is not called on savestate load,

No, I wouldn't expect that; but it is called on hard reset, right?  The only time we ever call `Bus::map` is right after calling `Bus::reset` on initial power or hard reset, I think?  That's all that should be needed; then the lookup should contain the same IDs every time.

Morilli:
> [ bus.reset()] is called on hard reset, right?

Nope, both `reset` and `power` are both internally the same command (power), just with the argument of `reset` set to true/false.

One issue is that the cartridge does some mapping on `load()`, and components like cpu or ppu do it on `power()`, so to catch everything you would need to start from the `system.load()` (which calls `bus.reset()`) and then do `emulator->power()` I guess.

Also I'm still not sure if having the same ids would be enough; as I've stated above I've tried to not let the counter, reader and writer arrays get savestated, and it still broke shit after loading a savestate, so it could be possible that the function pointers become broken after loading the waterbox state as well...

nattthebear (approved):

CasualPokePlayer (commented):
May want to remove the peekable[id] = true; in Bus::reset() so that everything isn't just marked as peekable.

Morilli:
> May want to remove the peekable[id] = true; in Bus::reset() so that everything isn't just marked as peekable.

I guess I can do that for safety (will just change to false I think), but it shouldn't generally be causing anything bad right now due to the explicit set in every bus.map() call (any bus.reset() is only called once anyway, so all those overwrites are effectively no-ops anyway).

CasualPokePlayer:
Ah I see, nevermind then, I was thinking bus.reset() can get called on soft/hard reset, but it's only called on load (lol)

CasualPokePlayer (approved):

PR #3181, opened 2022-03-13
Allow negative values to be entered in ram watch for fixed-point watches

### LINKED ISSUE #3175: Cannot poke negative numbers into fixed point RAM Watch
### Summary
Negative numbers cannot be entered into fixed point RAM Watches, despite fixed point numbers being treated as signed.

### Repro
1. Create a new RAM watch
2. Set "Size" to 4 bytes and "Display type" to "Fixed point 20.16" (or any other fixed point setting)
3. Right-click the entry and select "Poke"
4. Attempt to enter "-1" into the "Value" field
5. The negative sign will not be registered

### Host env.
Bizhawk 2.8; Windows 10

Morilli:
closes #3175

float and fixed-point numbers are basically handled the same right now, whereas previously float was the only type that was allowed a `-` sign (should be correct now).

In general, this code that is supposed to do text validation and value checking seems to be massively overengineered, otherwise such issues wouldn't even come up (a simple float.TryParse or whatever should catch everything that isn't valid and return sane values in all cases).
Anyway, I was too scared to do any major logic changes so this is what we get now.

### DIFF
--- a/src/BizHawk.Client.Common/tools/Watch/ByteWatch.cs
+++ b/src/BizHawk.Client.Common/tools/Watch/ByteWatch.cs
@@ -1,8 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Globalization;
-
-using BizHawk.Common.StringExtensions;
 using BizHawk.Emulation.Common;
 
 namespace BizHawk.Client.Common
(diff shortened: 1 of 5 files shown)


YoshiRulz:
If you're going to change the `IsFixedPoint` garbage, just remove it entirely and use `double.TryParse`.

Morilli:
Just the string overload, I take it? Because the other thing is also used for handling `OnKeyPress`. Or should all the logic there be removed too and all inputs allowed, only to be validated later on?

YoshiRulz:
Oh, nevermind, I didn't see it was in `OnKeyPress`. This LGTM. (Unless you want to burninate `WatchValueBox` too, because *why* is there a 700 LOC `TextBox` subclass which only affects text validation.)

Morilli:
Well, I have started doing some changes and I'm also noticing other issues, so I guess something will have to change.

As an example: floats can be e.g. `1E+09`, while the textbox does not support the characters `E` or `+` at all, not even minding the validation issues with that format.

Morilli:
Okay have just removed all the textbox input validation now, so now you can write everything and it will only get checked at the very end, when the entered value is getting converted to whatever format it's supposed to be in.

PR #3190, opened 2022-03-25
Implement msu1 handling for bsnes115+

### LINKED ISSUE #1693: Can't get MSU-1 support to work
From my understanding MSU-1 support should be working but I can't find any documentation on what the XML format should be to load it.  When I get what I think is the correct format (based on research) the emulator just crashes and eventually closes itself when I tell it to continue (albeit without audio).

Am I doing something wrong with the format or is this a bug?  Linked is my xml file.

This is for the Link to the Past randomizer.

https://gist.github.com/shinji257/3be89054b6a906c798f1d0a2bede7a71

Morilli:
closes #1693

The `RomAsset`/`IRomAsset` change here is my biggest concern, but I didn't see a way around it.

msu1 files are loaded on demand, by taking the game name and loading the `[game name].msu`, `[game name].bml` and `[game name]-N.pcm` files. While the msu and bml files are loaded on init, the pcm files can be loaded immediately, later in the game or potentially never.

This current implementation should be safe in regards to waterbox state integrity.


src/BizHawk.Client.Common/RomLoader.cs
```diff
@@ -481,6 +482,7 @@ private void LoadOther(CoreComm nextComm, HawkFile file, string forcedCoreName,
 						RomData = rom.RomData,
 						FileData = rom.FileData,
 						Extension = rom.Extension,
+						RomPath = file.Directory,
```
YoshiRulz, line 485: Don't call it `RomPath` if it contains the parent directory

Morilli: What else would you suggest? `RomDirectory`?

YoshiRulz: That's fine. Or use `RomPath` but include the filename, it seems the callsite is reconstructing it anyway.

Issue #1861, opened 2020-02-23
[REQUEST] Auto resume old play session


dart193:
When I moved to BizHawk from FCEUX for NES emulation, the option that I missed immediately was "Auto resume old play session". What it did was to automatically save state of the game in a separate ".resume" save slot when the emulator is closed, then when the next time the emulator was opened, it first automatically opened the last game (like Autoload ROM in Biz) and then loaded aforementioned save state (something that Biz can do too). Right now, Biz can only autosave SRAM every now and then, but no option to autosavestate on exit and load that save.

So I would like to see this feature done, like in FCEUX. As adelikat suggests, "savestate on close" is what is missing, and if added, can be paired with already present autoload to allow "autosave and restore" kind of functionality.

PR #3218, opened 2022-04-16
Add auto save state on close

peteyus:
- Add configuration for auto-saving state on exit
- Update MainForm to auto save on close game if configured
- Fix config serialization test.

#1861


YoshiRulz (changes requested):
The new setting(s) should be beside the autoload setting in `File` > `Load State`. That means the config properties are in the wrong class too.

This autosave uses a specific, configurable slot, but the autoload counterpart uses the currently selected slot:
https://github.com/TASEmulators/BizHawk/blob/8ea09eebf591a88b679d01f70784b84fd8fcbe5b/src/BizHawk.Client.EmuHawk/MainForm.cs#L597-L600
Not sure which we should go with.

I'm also not sure if the autosave call belongs in `CloseGame`. I believe that runs every time a rom is reloaded or unloaded, including when EmuHawk is closed. Autoload only happens on startup.

YoshiRulz:
Bump?

peteyus:
> 

Saw this, more or less agreed with it... and promptly lost it all in the shuffle of summer craziness with kids. I'll get back to this in the next week or two and apply the changes requested.

peteyus:
> I'm also not sure if the autosave call belongs in CloseGame. I believe that runs every time a rom is reloaded or unloaded, including when EmuHawk is closed. Autoload only happens on startup.

My thinking is that for autosave, it should happen any time the ROM changes. I'm not sure how it worked in FCEUX, although for my use case only doing it on shutdown works. Open to whichever here, calling it on `CloseGame` felt the safest (save early and often) but we could definitely make it happen on shutdown instead to reduce the number of times this happens.

peteyus:
> > I'm also not sure if the autosave call belongs in CloseGame. I believe that runs every time a rom is reloaded or unloaded, including when EmuHawk is closed. Autoload only happens on startup.
> 
> My thinking is that for autosave, it should happen any time the ROM changes. I'm not sure how it worked in FCEUX, although for my use case only doing it on shutdown works. Open to whichever here, calling it on `CloseGame` felt the safest (save early and often) but we could definitely make it happen on shutdown instead to reduce the number of times this happens.

I'm going to do it on CloseGame, it looks like the autoload code is executed during LoadRomInternal as well, so it runs any time the ROM is loaded (not just if it's passed as a command line argument). I think Autosave should have the same logic.

YoshiRulz (approved):
re: call location, agreed.

LGTM. Will merge tomorrow if no objections.

PR #3233, opened 2022-05-02
[WIP] Make input callbacks actually useful by passing gamepad index

YoshiRulz:
A check in the Impl. column indicates that the core's `IInputPollable` implementation has been updated i.e. to call `InputCallbacks.Call(int)` with a meaningful parameter. Cores which aren't `IInputPollable` or which implemented `IInputPollable.InputCallbacks` as a throw are not included in the table.

Core | Impl. | Field-tested
--:|:-:|:-:
A7800Hawk | ☑ | ☐
Atari2600Hawk | ☑ | ☐
BSNESv115+ | ☐ | ☐
C64Hawk | ☑ | ☐
CPCHawk | ☑ | ☐
ChannelFHawk | ☑ | ☐
Cygne | ☑ | ☐
Emu83 | ☑ | ☐
GBHawk | ☑ | ☐
GBHawkLink | ☑ | ☐
GBHawkLink3x | ☑ | ☐
GBHawkLink4x | ☑ | ☐
GGHawkLink | ☑ | ☐
Gambatte | ☐ | ☐
Genplus-gx | ☐ | ☐
IntelliHawk | ☑ | ☐
mGBA | ☑ | ☐
Mupen64Plus | ☑ | ☐
NesHawk | ☐ | ☐
O2Hawk | ☑ | ☐
PCEHawk | ☑ | ☐
SMSHawk | ☑ | ☐
SameBoy | ☑ | ☐
SubGBHawk | ☑ | ☐
TI83Hawk | ☑ | ☐
Virtu | ☑ | ☐
All Waterbox cores | ☐ | ☐
ZXHawk | ☑ | ☐

Notes:
- (Old) BSNES instantiates a `InputCallbackSystem` but never triggers it?
- Not sure if GambatteLink re-uses Gambatte implementation.
- Callback with param n should be immediately before player n is processed. If there is conceptually a player 0, either system/disc controls or a keyboard, those may be processed before or after the callback for player 1.
- Markdown checkboxes don't work in tables >:(

### DIFF
--- a/src/BizHawk.Client.Common/lua/INamedLuaFunction.cs
+++ b/src/BizHawk.Client.Common/lua/INamedLuaFunction.cs
@@ -6,7 +6,7 @@ namespace BizHawk.Client.Common
 {
 	public interface INamedLuaFunction
 	{
-		Action InputCallback { get; }
+		Action<int> InputCallback { get; }
 
 		Guid Guid { get; }
 
--- a/src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
+++ b/src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
@@ -43,7 +43,7 @@ namespace BizHawk.Client.Common
 					logCallback($"error running function attached by the event {Event}\nError message: {ex.Message}");
 				}
 			};
-			InputCallback = () => Callback(Array.Empty<object>());
+			InputCallback = gamepadIndex => Callback(new object[] { gamepadIndex });
 			MemCallback = (addr, val, flags) => Callback(new object[] { addr, val, flags });
 		}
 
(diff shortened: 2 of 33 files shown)


CasualPokePlayer:
https://github.com/TASEmulators/BizHawk/blob/8112be31e1ea9c2ab40e0ff4e8b8d86a51f2c5da/src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.cs#L50
https://github.com/TASEmulators/BizHawk/blob/d4bb5e047e07dbfc078d171c99c80eafd01b5eb0/src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs#L330-L333
re: GambatteLink, it just replaces the internal core's input callback system with the link core's so all of them internal cores use the same input callback system instance.

Probably simple enough to make something like `_whichPlayer`, set to 1 on internal core constructor, only changed in a ConnectInputCallbackSystem call, and send that _whichPlayer var through the callback (unless in SGB mode then the value returned by gambatte_getjoypadindex should be used, may need revisiting if SGB support is added to link core).

PR #3234, opened 2022-05-02
fix absolute dimensions of shaders

notwa:
previously, a height specified absolutely would affect the output width instead of the height.


zeromus:
what's there to review? this seems obviously a bug

PR #3259, opened 2022-05-29
depend on versioned ffmpeg release instead of nightly

vadosnaprimer:
It may be hard to find the exact nightly hawk wants for linux, and building it yourself with all the right libs is also quite a challenge sometimes, and even then hawk will depend on what it says in version.

I tested all our current ffmpeg commands against 4.0 and 4.4.1 and they all work great.

https://ffmpeg.org/doxygen/3.2/index.html says that API remains compatible within the same major version, so that's what I'm checking now.

`FFmpegDownloaderForm` may need to be reworded, because we can host working versioned builds for both windows and linux (ubuntu), taken from https://github.com/BtbN/FFmpeg-Builds/releases/tag/autobuild-2022-05-28-12-32 and for those who can't install pre-built stuff we'd at least provide the right version info.

I couldn't make our dearchiver extract a file from a nested path, so we may repackage them to only contain ffmpeg executable, or we can tweak `HawkFile` to support nested paths.


zeromus:
If you want to support any 4.x instead of a single version, that's up to you

nattthebear:
From memory, the main issues we've seen are:

1. For some releases, the nutdemuxer was broken and looked for a seek block even on a non-seekable stream.  This was catastrophic because nut is how we get all of the A/V in, so everything stopped working.
2. Sometimes, codecs are obsoleted and replaced, and some of the commands need to change.  This isn't as bad because you can get around it with a custom command, but it sure is annoying and it'd be nice for our "out of the box" commands to actually work out of the box.

If you want to take on the risks associated with these issues, go ahead.  There's some reward in simplifying our process here, and the likelihood of breakage isn't big.

YoshiRulz (commented):
I think using 4.4.1 rather than a custom build is a good idea. Unfortunately, while the build from https://johnvansickle.com/ffmpeg works if manually installed, I can't get the downloader to work. It refuses to extract the main executable.
```csharp
// (abridged from FFmpegDownloaderForm.Download)
using var hf = new HawkFile(fn);
var ai = hf.ArchiveItems.First(ai => ai.Name == "ffmpeg");
Console.WriteLine(ai.Size); // 77173696
using var exe = hf.BindArchiveMember(ai);
var data = exe!.ReadAllBytes();
Console.WriteLine(data.Length); // 4715!?
```

src/BizHawk.Client.EmuHawk/AVOut/FFmpegDownloaderForm.cs
```diff
@@ -20,7 +20,7 @@ public FFmpegDownloaderForm()
 			txtLocation.Text = FFmpegService.FFmpegPath;
 			txtUrl.Text = FFmpegService.Url;
 
-			if (OSTailoredCode.IsUnixHost) textBox1.Text = string.Join("\n", textBox1.Text.Split('\n').Take(3)) + $"\n\n(Linux user: Create a symlink with the below filename pointing to the ffmpeg binary with version {FFmpegService.Version}.)";
+			if (OSTailoredCode.IsUnixHost) textBox1.Text = string.Join("\n", textBox1.Text.Split('\n').Take(3)) + "\n\n(Linux user: If installing manually, you can use a symlink.)";
```
vadosnaprimer, line 23: I think we should also tell the user right away what version we consider compatible, so they know what to look for if they can't use our build.

PR #3263, opened 2022-06-05
Support editing nyma settings without the core loaded

CasualPokePlayer:
The details are icky here. Info on the settings is stored in the unmanaged core. So no matter what a core instance is needed at some point.

Now settings info here should be constant for a given core, so we can cache settings info after the core has been loaded and use that.

If in the case the user wants to edit the settings but the core has never been loaded, a minimal load will be done to obtain the settings (and cache them) and quickly dispose thereafter.

This code could probably be cleaner, but not sure how to do so.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
@@ -62,6 +62,7 @@ using BizHawk.Emulation.Cores.Sega.GGHawkLink;
 using BizHawk.Emulation.Cores.Sega.MasterSystem;
 using BizHawk.Emulation.Cores.Sony.PS2;
 using BizHawk.Emulation.Cores.Sony.PSX;
+using BizHawk.Emulation.Cores.Waterbox;
 using BizHawk.Emulation.Cores.WonderSwan;
 using BizHawk.WinForms.Controls;
 
@@ -1898,6 +1899,10 @@ namespace BizHawk.Client.EmuHawk
 			OpenGenericCoreConfig($"{coreName} Settings");
 		}
 
+		private DialogResult OpenNymaConfigFor<T>(NymaCore.NymaSettingsInfo settingsInfo)
+			where T : IEmulator
+			=> GenericCoreConfig.DoNymaDialogFor(this, GetSettingsAdapterFor<T>(), settingsInfo, isMovieActive: MovieSession.Movie.IsActive());
+
 		private DialogResult OpenVirtuSettingsDialog()
 			=> OpenGenericCoreConfigFor<AppleII>("Apple II Settings");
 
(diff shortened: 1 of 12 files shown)


src/BizHawk.Emulation.Cores/Waterbox/NymaCore.cs
```diff
@@ -185,6 +185,31 @@ public abstract unsafe partial class NymaCore : WaterboxCore
...
 		}
 
+		// inits only to get settings info
+		// should only ever be called if no SettingsInfo cache exists statically within the core
+		protected void InitForSettingsInfo(string wbxFilename)
+		{
+			_nyma = PreInit<LibNymaCore>(new WaterboxOptions
+			{
+				Filename = wbxFilename,
+				// todo: maybe make these smaller? not much is going to occur within this core instance...
```
nattthebear, line 195: I like this the way it is now:  The exact same memory settings should be used to guarantee a reliable core bootstrap.  All these megabytes don't cost much since they're eliminated soon after, and you'd need that much ram to load the core anyway.

To make this more explicit, could we abstract away this `new WaterboxOptions` call and make it exactly the same one that is used in `DoInit`?

CasualPokePlayer: Do you mean making a separate function that returns a new WaterboxOptions returning the same options for each? Not sure if that's exactly a good idea given a dummy firmware delegate would probably need to be made. Perhaps best to keep it as is (although todo comment removed).

nattthebear: The `WaterboxOptions` class doesn't contain any of the delegate info.

CasualPokePlayer: Ah, misinterpreted the delegate array as part of WaterboxOptions when that's just the next arg in PreInit, I see.

nattthebear (approved):

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
```diff
@@ -2821,9 +2826,8 @@ ToolStripMenuItemEx CreateCoreSubmenu(VSystemCategory cat, string coreName, para
 			items.Add(CreateCoreSubmenu(VSystemCategory.Other, CoreNames.Emu83, CreateSettingsItem("Palette...", (_, _) => OpenTI83PaletteSettingsDialog(GetSettingsAdapterFor<Emu83>()))));
 
 			// Faust
-			var faustSettingsItem = CreateSettingsItem("Settings...", (_, _) => OpenGenericCoreConfig($"{CoreNames.Faust} Settings"));
+			var faustSettingsItem = CreateSettingsItem("Settings...", (_, _) => OpenNymaConfigFor<Faust>(Faust.CachedSettingsInfo(CreateCoreComm())));
```
YoshiRulz, line 2829: Does this work when the core is loaded?

Issue #2843, opened 2021-07-11
Log Key changes when movie is saved without considering whether saved input reflects it


CasualPokePlayer:
Noticed this mostly when I went insane over N64 timing bullshit, since resaving the movie at the end seemed to desync on a subsequent replay, and it was due to the log key changing.

YoshiRulz:
see also #2567

PR #3274, opened 2022-06-15
Fix KeyLog not being respected in TAStudio

Morilli:
See: #2843, #2567.

Before, loading a bk2 with a specific `KeyLog` set in TAStudio would just discard it, effectively mis-matching all inputs.
It was possible to craft a .tasproj with a different `KeyLog` by manually editing the `InputLog` file, but on saving the same issue would occur.

I have fixed the `KeyLog` getting discarded in the cases of bk2 -> tasproj, tasproj -> bk2 and saving a tasproj in general.

Here are two files (one .bk2, one .tasproj) that should show the difference in the `InputLog` and what is displayed in TAStudio when loading them:
[Movies.zip](https://github.com/TASEmulators/BizHawk/files/8904678/Movies.zip)

I have tested this and I believe this to work fine, however I would like to get some feedback on whether I may have missed something (or there is a better way of handling the `KeyLog`).

### DIFF
--- a/src/BizHawk.Client.Common/movie/MovieConversionExtensions.cs
+++ b/src/BizHawk.Client.Common/movie/MovieConversionExtensions.cs
@@ -21,6 +21,7 @@ namespace BizHawk.Client.Common
 			string newFilename = ConvertFileNameToTasMovie(old.Filename);
 			var tas = (ITasMovie)old.Session.Get(newFilename);
 			tas.CopyLog(old.GetLogEntries());
+			tas.LogKey = old.LogKey;
 
 			old.Truncate(0); // Trying to minimize ram usage
 
@@ -56,6 +57,7 @@ namespace BizHawk.Client.Common
 		{
 			var bk2 = old.Session.Get(old.Filename.Replace(old.PreferredExtension, Bk2Movie.Extension));
 			bk2.CopyLog(old.GetLogEntries());
+			bk2.LogKey = old.LogKey;
 
 			bk2.HeaderEntries.Clear();
 			foreach (var (k, v) in old.HeaderEntries) bk2.HeaderEntries[k] = v;
@@ -95,6 +97,7 @@ namespace BizHawk.Client.Common
 			var entries = old.GetLogEntries();
 
 			tas.CopyLog(entries.Skip(frame));
+			tas.LogKey = old.LogKey;
 			tas.CopyVerificationLog(old.VerificationLog);
 			tas.CopyVerificationLog(entries.Take(frame));
 
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.InputLog.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.InputLog.cs
@@ -1,6 +1,5 @@
 ﻿using System.Collections.Generic;
 using System.IO;
-
 using BizHawk.Common;
 
 namespace BizHawk.Client.Common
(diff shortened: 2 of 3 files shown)


YoshiRulz (approved):
I don't completely understand this but it makes sense that it needs to be set when copying mem-->mem as well as disk-->mem.

Morilli:
> Does this fix either of the linked issues?

Should definitely fix https://github.com/TASEmulators/BizHawk/issues/2843. For #2567, I actually haven't tried the movie importer (didn't even know such a feature existed lol), but I did see empty logkeys in my testing (due to them not being copied), so this is probably resolved as well. Will check again though to confirm, only I get some test movie file.

Morilli:
Okay correction: The movie importer does still import a null LogKey. Will see what the best way to address that is and push a commit after.

Morilli:
Okay, so I think this is not actually trivial to do, at least in a proper way. The following diff shows the minimum changes to get *some* logkey into the resulting .bk2 file:
```diff
diff --git a/src/BizHawk.Client.Common/movie/bk2/Bk2Controller.cs b/src/BizHawk.Client.Common/movie/bk2/Bk2Controller.cs
index b549dd1f0..dc673d0ff 100755
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Controller.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Controller.cs
@@ -7,7 +7,7 @@ using BizHawk.Emulation.Common;

 namespace BizHawk.Client.Common
 {
-       internal class Bk2Controller : IMovieController
+       public class Bk2Controller : IMovieController
        {
                private readonly WorkingDictionary<string, bool> _myBoolButtons = new WorkingDictionary<string, bool>();
                private readonly WorkingDictionary<string, int> _myAxisControls = new WorkingDictionary<string, int>();
diff --git a/src/BizHawk.Client.Common/movie/interfaces/IMovieSession.cs b/src/BizHawk.Client.Common/movie/interfaces/IMovieSession.cs
index 7b375733f..7b9d76e89 100644
--- a/src/BizHawk.Client.Common/movie/interfaces/IMovieSession.cs
+++ b/src/BizHawk.Client.Common/movie/interfaces/IMovieSession.cs
@@ -25,7 +25,7 @@ namespace BizHawk.Client.Common

                IDictionary<string, object> UserBag { get; set; }

-               IMovieController MovieController { get; }
+               IMovieController MovieController { get; set; }

                /// <summary>
                /// Provides a source for sticky controls ot use when recording
diff --git a/src/BizHawk.Client.EmuHawk/MainForm.cs b/src/BizHawk.Client.EmuHawk/MainForm.cs
index b9add8567..ee7db798f 100644
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -4128,6 +4128,7 @@ namespace BizHawk.Client.EmuHawk

                private void ProcessMovieImport(string fn, bool start)
                {
+                       MovieSession.MovieController = new Bk2Controller(Emulator.ControllerDefinition);
                        var result = MovieImport.ImportFile(this, MovieSession, Emulator, fn, Config);

                        if (result.Errors.Any())
```
Although this technically works, during my testing I immediately noticed that the resulting LogKey may not be correct at all, as for example the `LsmvImport` hardcodes the controllers that should be should, and completely disregards the current instantiated emulator's controller choice (which makes sense, except for when it also ignores the lsmv's controller specification... but that's another issue).

So, while the above may be "more correct", in the end all 14 importers will have to manually set the LogKey to match with what they are actually exporting, and I'm not willing to implement that in this PR.

YoshiRulz:
That looks a lot like [my earlier suggestion](https://github.com/TASEmulators/BizHawk/pull/2568/files) and [its failings](https://github.com/TASEmulators/BizHawk/pull/2568#issuecomment-761195035), so I feel validated.

Morilli:
For reference, how it could probably work if done for every `Import` class:
```diff
diff --git a/src/BizHawk.Client.Common/movie/import/LsmvImport.cs b/src/BizHawk.Client.Common/movie/import/LsmvImport.cs
index 1da398c44..6dfc48850 100644
--- a/src/BizHawk.Client.Common/movie/import/LsmvImport.cs
+++ b/src/BizHawk.Client.Common/movie/import/LsmvImport.cs
@@ -42,6 +42,7 @@ namespace BizHawk.Client.Common.movie.import
                                RightPort = LibsnesControllerDeck.ControllerType.Gamepad
                        };
                        _deck = new LibsnesControllerDeck(ss);
+                       Result.Movie.LogKey = new Bk2LogEntryGenerator("SNES", new Bk2Controller(_deck.Definition)).GenerateLogKey();

                        string platform = VSystemID.Raw.SNES;

```

Issue #2754, opened 2021-05-26
"New" BSNES (v115) megathread


YoshiRulz:
BizHawk 2.6.2 includes a fourth option for SNES core, namely `BSNESv115+`. Its purpose is to bring EmuHawk in line with standalone BSNES/lsnes in terms of console-accuracy. Once it's proven itself it can replace the existing core, but that means it needs field-testing, which is where YOU come in!

**Give an unverifiable run a new life by resyncing it with this new core, and post your triumphs and failures below.**

And thanks again to @Morilli for swooping in and contributing a whole core!

anewuser:
Is there any way to use the run-ahead feature with this core?

YoshiRulz:
Nope.

Eggo-dotcom:
The official 115 can run special chip games in hle and lle mode. It is able to indicate what is happening at the moment as well. On bizhawk how can i tell if it is running lle or hle?

in 115 this is normally exposed as selectable even

YoshiRulz:
That doesn't seem to be exposed. There is a `snes_get_mapper` but it's unused.

Morilli:
There are currently a lot of configuration options that are not exposed and left at their default value, mostly because I didn't found them necessary to be exposed (see [configuration.hpp](https://github.com/TASVideos/BizHawk/blob/master/waterbox/bsnescore/bsnes/sfc/interface/configuration.hpp) for a list). That is subject to change though, in case some of those turn out to be worthy enough.

Eggo-dotcom:
But the main question is if it is running in hle or lle mode for special chip games?

ozderya:
If TAStudio open the time on the right up corner doesn't work for BS Super Mario USA Power Challenge.

Issue #973, opened 2017-08-23
Request: Subframe inputs where games poll for input multiple time in a frame


meshuggahtas:
List of platforms I'm specifically looking for:
-NES
-SMS
-GBA

edit: my threads on various forums so far:
SMS - http://www.smspower.org/forums/17362-SMSTechnicalQuestionsAndLookingForOddities#105134
VBoy - https://www.planetvb.com/modules/newbb/viewtopic.php?topic_id=7279&forum=1
Lynx - http://atariage.com/forums/topic/288180-lynx-technical-questions-and-looking-for-oddities/
A7800 - http://atariage.com/forums/topic/288181-technical-questions-and-looking-for-oddities/

zeromus:
we know.

YoshiRulz:
SubNESHawk was fully merged into master at 9696a7583.

meshuggahtas:
Totally forgot about this issue. Thanks again for working on it!

My page with relevant informations for sub frame inputs: http://tasvideos.org/MESHUGGAH/ForbiddenTechniques.html#NotPresentInBizhawk231 --> **Subframe inputs** tab.

In the OP I listed those 3 platforms because of known titles with potential techniques based on subframe level. The GBA one is only mentioned for the ability to subframe reset, so I'm not sure about which GitHub labels you would like to hold/throw away as only SubNESHawk came to materalization.

YoshiRulz:
If adding a toggle for sub-frame inputs to a core allows a really cool TAS to be made, then that core is missing features. Like [Chrono Trigger](http://tasvideos.org/2047M.html), which you linked to.

meshuggahtas:
The SMS one points to the discovery thread. The example is SMS Batman Returns where it might be possible to:

> with subframe inputs there would be enough time to decrement the counter in only a couple of frames. 

--> skipping the loading screen very fast. SG-1000 Bank Panic might have a similar trick.

(Just noting: this doesn't makes a "really cool TAS")

We (me and Alyosha) don't have knowledge about does it works in real life, at least I don't have consoles but it's on my TODO list to start gathering informations about these techniques from large forums dedicated to those platforms.

alyosha-tas:
Yeah I would like to have some confirmation that pause button in SMS actually does have the theorized effect on real console before attempting a subframe SMS mode.

As for SNES and GBA, I don't see how subframe mode would be even remotely possible for them.

vadosnaprimer:
[...]

PR #3281, opened 2022-06-18
BSNESv115: allow subframe inputs

Morilli:
Not many cores seem to implement such functionality so I wasn't entirely sure on how to do this properly, but the seemingly common way of adding a cycle count "input" seemed too complex and annoying to work with, so I've went with this relatively minimal solution.

Was very unsure about the frame and lag count, so I'd appreciate feedback on whether this implementation can be improved, ideally in a non-intrusive way.

Here's an example tasproj that shows this feature in action: [SuperMarioWorld_agdq2014TotalControl_BSNESNEW.zip](https://github.com/TASEmulators/BizHawk/files/8933083/SuperMarioWorld_agdq2014TotalControl_BSNESNEW.zip). This is a slightly modified version of [this TASVideos submission](https://tasvideos.org/4156S) (and ported to BizHawk's .bk2/.tasproj format).

see #973, #2754


CasualPokePlayer:
For BSNES, it might be best to look at SubNESHawk for inspiration. It has "frames" defined as either input latch or vblank. It also has a "cycle count input" in a limited fashion: it only applies for resets. This is actually fairly important here as save corruption fun (e.g. https://tasvideos.org/4540M) isn't going to align to input latches, and of course powering off and on the console wouldn't care for input latches.

Also for proper timing ICycleTiming would need to be implemented. Which granted, that will cause a nag to the user to save cycle count no matter what (even if no subframe inputs are done, as a core implementing ICycleTiming is assumed to always need cycle count timing). May be best to split it off as a new "core" (which is fairly simple anyways since it's mostly a wrapper).

Morilli:
> "cycle count input" in a limited fashion: it only applies for resets.

That seems like a decent solution, as that doesn't require manually setting the cycle time for each frame. Will look into getting something like that working.

CasualPokePlayer:
> > "cycle count input" in a limited fashion: it only applies for resets.
> 
> That seems like a decent solution, as that doesn't require manually setting the cycle time for each frame. Will look into getting something like that working.

In reality not really, the input wouldn't do anything if power isn't pressed, and the neutral input would probably be 0 (i.e. instant reset). Only frames which you actually want to reset the console have to be manually set.

Morilli:
> Only frames which you actually want to reset the console have to be manually set.

Yeah, I was just thinking about the case where you don't have latch detection logic and need to specify the cycle duration for each frame in which you want to perform subframe inputs manually.

CasualPokePlayer:
> > Only frames which you actually want to reset the console have to be manually set.
> 
> Yeah, I was just thinking about the case where you don't have latch detection logic and need to specify the cycle duration for each frame in which you want to perform subframe inputs manually.

I'm realizing I misread your sentence, oops.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs
```diff
@@ -10,11 +10,13 @@ namespace BizHawk.Emulation.Cores.Nintendo.BSNES
 {
 	[PortedCore(CoreNames.Bsnes115, "bsnes team", "v115+", "https://github.com/bsnes-emu/bsnes")]
 	[ServiceNotApplicable(new[] { typeof(IDriveLight) })]
-	public unsafe partial class BsnesCore : IEmulator, IDebuggable, IVideoProvider, ISaveRam, IStatable, IInputPollable, IRegionable, ISettable<BsnesCore.SnesSettings, BsnesCore.SnesSyncSettings>
+	public unsafe partial class BsnesCore : IEmulator, IDebuggable, IVideoProvider, ISaveRam, IStatable, IInputPollable,
+		IRegionable, ISettable<BsnesCore.SnesSettings, BsnesCore.SnesSyncSettings>, ICycleTiming
```
CasualPokePlayer, line 14: I suggest not adding an ICycleTiming here, as otherwise the movie recorded will complain if cycle count is not present in movie (not needed as subframe is a different core).

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/SubBsnesCore.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Collections.Generic;
+using System.IO;
+using BizHawk.Emulation.Common;
+
+namespace BizHawk.Emulation.Cores.Nintendo.BSNES
+{
+	[PortedCore(CoreNames.SubBsnes115, "")]
+	[ServiceNotApplicable(new[] { typeof(IDriveLight) })]
+	public class SubBsnesCore
+		: IEmulator, IDebuggable, IVideoProvider, ISaveRam, IStatable, IInputPollable, IRegionable,
```
CasualPokePlayer, line 11: You're registering most of these services from the BsnesCore instance anyways, don't re-implement them here (except ICycleTiming, which you don't want "implemented" in the non-subframe core)

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/SubBsnesCore.cs
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		private readonly BsnesCore _bsnesCore;
+		
+		public IEmulatorServiceProvider ServiceProvider { get; }
+
+		public ControllerDefinition ControllerDefinition => _bsnesCore.ControllerDefinition;
+		public bool FrameAdvance(IController controller, bool render, bool renderSound = true)
+		{
+			return _bsnesCore.FrameAdvance(controller, render, renderSound);
```
CasualPokePlayer, line 37: It would be best I think to just reimplement FrameAdvance here instead of calling the _bsnesCore one (and keeping subframe logic in here).

Morilli:
Should have resolved all suggestions. New tasproj:
[SuperMarioWorld_agdq2014TotalControl_SubBSNESv115.zip](https://github.com/TASEmulators/BizHawk/files/8936265/SuperMarioWorld_agdq2014TotalControl_SubBSNESv115.zip)

Morilli:
So, it seems like checking the current cpu cycle in c# code only (after each cpu_step returns) may not be cycle-accurate enough to support all possible reset cycles, so this would require core changes.

I have been trying to get https://tasvideos.org/2466M to work to verify correctness, but couldn't. If I'm understanding correctly lsnes does not handle all cycles, but instead just increments a cycle counter for each cpu instruction executed, no matter how many cycles they take.
I have also tried to replicate this behavior, even though I probably wouldn't want this to be in bizhawk at the end, but that also failed to sync the movie.
Will need to figure out why this is failing to make sure the reset cycle input actually works as intended.

Morilli:
Okay, I've changed the `Reset Cycle` axis value to `Reset Instruction`, which actually matches with what is checked in code now. I don't believe having higher reset granularity than between individual cpu instructions is necessary, however technically speaking that could be realized as well by letting the core do the counting work in the cpu itself.

I have also managed to get the subframe reset of the mentioned submission (https://tasvideos.org/2466M) to work; it required ~700 additional instructions, which are probably all wait / stop instructions which are not counted by lsnes, but are in this core currently. Here's a tasproj that should perform the subframe resets correctly and create a "glitched" save file: [mother2_glitchedNEW.zip](https://github.com/TASEmulators/BizHawk/files/8979019/mother2_glitchedNEW.zip)

Part of what makes the rest not work is the value of $7E9979 not matching the expected one. I did make some progress by creating an extended gamepad controller (instead of the extended super multitap controller), which did seem to alter the behavior in the last frames of the tasproj, but I still couldn't manage to recreate the behavior of the original submission. Perhaps some additional lag frames break a frame counter that was read or something, idk.

In any case, this subframe reset thing should work properly now 👍

CasualPokePlayer (approved):
Seems to all work, rebase then LGTM

PR #3283, opened 2022-06-19
Simplify SGB related UIs (and make it possible to use SGB under Gambatte link)

Fortranm:
https://en.wikipedia.org/wiki/Super_Game_Boy#Super_Game_Boy_2

> The Super Game Boy 2 was released exclusively in Japan in 1998. Additions included a link port to allow a user to access two-player mode via the link cable...

Currently, the SGB mode of Gambatte has to be turned on via Config > Cores > GB in SGB instead of in the sync setting and cannot be used in GB Link as a result.

This PR does three things related to SGB:
1. SGB is added (or rather, merged in) as a Console Mode option for Gambatte.
image

2. "GB in SGB" check box and SGB core picker are removed and the BSNES cores are added under the GB core picket. GB/GBC games will load under SGB mode of the corresponding BSNES core if one of those options are chosen.
image

3. As a result of the first change (with addition adjustments), it is now possible to use SGB mode in Gambatte Link too.
image
image
Each SGB core can have up to 4 controllers linked to it. Note that the tab parsing process is currently hardcoded to categorize every item that starts with "Px " under the tab "Player x" and everything else under "Console". There have been suggestions to use "Console x" for a core in link mode but that requires enough refactoring to be considered a separate task IMO.


CasualPokePlayer:
There's some issues here:

1. I'm not sure if this UI is any better than before. The old UI seems more intuitive imo (still not the best of course)
2. This change was mainly to get GBL in for SGB2 (which yes, the SGB2 did have a link port). However, getting it working requires extra effort here. The 4 players in SGB2 per Gambatte core makes controller handling a mess (I'm not sure if this is PR is exactly the best solution here). SGB Borders are also an issue as they are absent in the current implementation, not to mention SGB sound, which is missing here.

If we really want GBL for SGB2 maybe it would maybe be better to do some sort of hack that makes the GB in SGB work with GBL for GambatteLink.

Fortranm:
SGB borders can now be displayed in GBL mode. Note that _all_ borders will be displayed when _any_ of the SGB cores have the Display Border option enabled.
image
(the second and fourth cores are in GBA mode here)

src/BizHawk.Client.Common/config/Config.cs
```diff
@@ -325,7 +322,6 @@ public void ResolveDefaults()
 			[VSystemID.Raw.GB] = CoreNames.Gambatte,
 			[VSystemID.Raw.GBC] = CoreNames.Gambatte,
 			[VSystemID.Raw.GBL] = CoreNames.GambatteLink,
-			[VSystemID.Raw.SGB] = CoreNames.Gambatte,
```
YoshiRulz, line 328: This is incorrect, all `AppliesTo` sysIDs are used here (note GB and GBC). Actually I don't think the above change to `CorePickerUIData` is correct either.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.IEmulator.cs
```diff
@@ -23,7 +23,11 @@ public bool FrameAdvance(IController controller, bool render, bool rendersound =
 				{
 					for (int i = 0; i < _numCores; i++)
 					{
-						if (s.Contains($"P{i + 1} "))
+						if (s.Contains($"P{i + 1} -"))
+						{
+							_linkedConts[i].Set(s.Replace($"P{i + 1} - ", ""));
+						}
+						else if (s.Contains($"P{i + 1} "))
```
YoshiRulz, line 30: This forced you to add a hack to `Bk2MnemonicLookup.cs`. What does it actually do? Can you implement it another way?

Fortranm: This deals with the fact that an SGB core has 4 controllers, as shown in the last image in the original description. I don't have a better implementation in mind for now.

YoshiRulz: Number them 1-4 and make the next console start from 5

Fortranm: That wouldn't be intuitive on which core a controller corresponds to, especially when console modes are mixed.

Fortranm: image

getCursorsExe:
Unable to reproduce this, since it is impossible to load GambatteLink for me.

src/BizHawk.Client.Common/config/Config.cs
Fortranm: This has been restored in a commit. As for the other changes in the file, I have repeatedly switched between cores and it seems to work fine aside from the minor issue mentioned in the comment.

PR #3288, opened 2022-06-24
A simple boolean check.

### LINKED ISSUE #3287: [Debugger] It should be checking to see if a breakpoint set at the given address exists, to prevent it from crashing when attempting to remove said breakpoint

### Summary
Bizhawk Debugger cannot remove a single breakpoint where the address is pointing to the same address of another breakpoint. This also prevents the user from closing Bizhawk windows because the Debugger attempts to remove all of the breakpoints and fails.

Below included the error exception. However, the error exception stack trace is misleading, because the actual cause is a code logic error from this source file here:  
https://github.com/TASEmulators/BizHawk/blob/79994070899581981005bc385170384167772c33/src/BizHawk.Client.Common/BreakpointList.cs#L12-L15

The issue is the lack of breakpoint validations before adding the new breakpoint object into the Breakpoint List. It should be checking to see if a breakpoint with the same address exists in the `BreakpointList` class, before adding the new breakpoint. If an existing breakpoint whose address matches in the list, it should either tell the user the breakpoint already exists, or skip adding the breakpoint.

That way, when the time comes for the user to attempt to remove the breakpoint, it will not crash, because the delegate LINQ function assumes there exists only a breakpoint whose address is unique among the list of breakpoints. (The `_callbacks.SingleOrDefault()` on line 136.)

https://github.com/TASEmulators/BizHawk/blob/79994070899581981005bc385170384167772c33/src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBA/MGBAMemoryCallbackSystem.cs#L134-L142

### Repro
1. Download the open-source GBA rom here:  ( https://github.com/Hpmason/gba-dvd/releases/tag/v0.1.1 )
2. In Bizhawk, open the gba_dvd.rom, making sure it is running on mGBA v0.9.1 Core.
3. Open TAStudio.
4. Open Debugger.
5. At the bottom of the Debugger tool, click on the green [+] icon to add a new breakpoint.
6. In the pop-up, in the Address field, enter 0x08000286.
7. In the Mask field, leave 0xFFFFFFFF as the default.
8. In the Scope field, select the radio button next to Execute.
9. Click on the Add button.
10. Repeat Steps 5 through 9, then go to Step 11.
11. Attempt to remove one of the breakpoints.
12. Encounter error exception.

### Output

```
System.InvalidOperationException: Sequence contains more than one matching element
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at BizHawk.Emulation.Cores.Nintendo.GBA.MGBAMemoryCallbackSystem.Remove(MemoryCallbackDelegate action)
   at BizHawk.Client.Common.Breakpoint.set_Active(Boolean value)
   at BizHawk.Client.Common.BreakpointList.Remove(Breakpoint b)
   at BizHawk.Client.EmuHawk.BreakpointControl.RemoveBreakpointButton_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

### Host env.
- BizHawk 2.8.0; Win10 Pro 21H1 Build 19043; Intel/NVIDIA

tommai78101:
Added a simple boolean check to see if an existing breakpoint contains the same address as the one we wanted to add. If none, we go ahead and add the new breakpoint.

Fixes #3287.

### DIFF
--- a/src/BizHawk.Client.Common/BreakpointList.cs
+++ b/src/BizHawk.Client.Common/BreakpointList.cs
@@ -11,7 +11,10 @@ namespace BizHawk.Client.Common
 
 		public void Add(IDebuggable core, string scope, uint address, uint mask, MemoryCallbackType type)
 		{
-			Add(new Breakpoint(core, scope, Callback, address, mask, type));
+			if (!Exists(x => x.Address == address))
+			{
+				Add(new Breakpoint(core, scope, Callback, address, mask, type));
+			}
 		}
 
 		public new void Clear()


CasualPokePlayer:
This doesn't seem to actually fix the issue, as the address equaling each other doesn't matter, it's the the same callback is being sent through here.

CasualPokePlayer:
Issue fixed in d3d90eb70d31e494069b8ba781b9819dc32fe3ca

tommai78101:
Thank you for looking into the issue.

PR #3299, opened 2022-07-01
Fixed the "To PC" button from not updating the disassembler view in the Debugger window.

tommai78101:


src/BizHawk.Client.EmuHawk/tools/Debugger/GenericDebugger.cs
```diff
@@ -294,6 +294,8 @@ private void CancelSeekBtn_Click(object sender, EventArgs e)
 		private void ToPCBtn_Click(object sender, EventArgs e)
 		{
 			UpdateDisassembler();
+			UpdatePC();
```
YoshiRulz, line 297: Call order is reversed at other callsites. `UpdatePC` sets a field which `UpdateDisassembler` needs to read.

tommai78101: Oh I see. Will reverse it. Do I need to reverse the order at other call sites, because it would feel it's out of scope for this PR?

PR #3307, opened 2022-07-07
Added "Clear Output" button to Lua Console tool.

tommai78101:
image

Minor UI improvement. It's not that apparent for new users to right-click on the Output window, and click "Clear" to clear away the messages.


YoshiRulz:
Is the icon your own work?

PR #3310, opened 2022-07-10
Change wording of `RewindConfig.TargetFrameLength` edit UI

### LINKED ISSUE #3309: Confusing "Desired frame length" option name in Rewind configuration
image

The name of the option "Desired frame length" is confusing. First off, what does it even mean to have a frame length of 300, and what about it is desirable? You have to somehow figure out that "frame length" doesn't talk about the length of a frame, but about a range of frames? Not ideal.
Additionally, it says nothing about the case when I have more storage than is needed. Does it cut off at the range anyway? I looked at the code but I think it doesn't cut off the range and just continues filling up the storage. This also isn't explained in the setting name.

So I have a new suggestion:
"**Minimum rewind range in frames**"

It explains that it's only a minimum, so you don't have to worry about it cutting off your rewind. And it also makes it apparent that it's a range.

(BizHawk 2.8; Win10)

tommai78101:
fixes #3309


nattthebear:
I'm pretty neutral on this one.  Old label was fine, but if you want to change it, have fun.

Masterjun3:
In that case, why not make it an actual Minimum and use Math.Ceiling here?
https://github.com/TASEmulators/BizHawk/blob/d8fc32f1a820ee5a044ad9422c0fbf967101642b/src/BizHawk.Client.Common/rewind/ZwinderBuffer.cs#L152
Can't do anything about state size differences, but a minimum setting like that is not only more useful but also a less confusing feature than whatever there is right now.

Masterjun3:
If all else fails though, in my opinion, the most accurate and easiest to understand label in this case is
"Aim for a rewind range of at least `[ ]` frames".

tommai78101:
I'm not keen on changing anything in the `ZwinderBuffer.cs`, as I'm not familiar with how Rewind Frequency is calculated, nor know how to test this properly if I had to tweak the calculations. 

The method name, however, hinted the value should be called "Ideal Rewind Interval" or "Ideal Rewind Buffer Size". Would these label name be sufficient enough to make it more understandable than the current label? 

Either that, or I might need to seek advice from others.

YoshiRulz:
"desired" indicates user preference, which isn't necessarily "ideal"

Masterjun3:
I disagree with "Ideal".
The point I was trying to make with the issue is the important explanation to the user of "what happens if I have waay much more space than I specify here?".
Say I have space for 4 hours of rewind, but specify 10 seconds here. All the "desired" or "ideal" terms imply that the emulator may just cut off the range to 10 seconds. This is precisely why I opened the issue ticket in the first place.

tommai78101:
Good to know the word "Ideal" is not suitable. 

> Say I have space for 4 hours of rewind, but specify 10 seconds here. All the "desired" or "ideal" terms imply that the emulator may just cut off the range to 10 seconds. 

The first thing that popped into my mind is "Rewind buffer chunk size". In short, I'm coming up with label names including something to do with the word "Size" or "Capacity" in it. I am starting to feel the word, "Length", is not a good word to use because "Length" implies there is a cut-off point.

tommai78101:
Closing pull request, because I no longer feel confident in my abilities to give this label a proper renaming.

Issue #2959, opened 2021-10-07
Lowering emulation speed wih "-" key can occassionally hang BizHawk for some time


getCursorsExe:
Whenever you use a "-" key, the emulation speed decreases, but because of how frames are processed, it can occassionally hang BizHawk for some time.

zeromus:
The emulator is unusably unresponsive also.

CasualPokePlayer:
This I think is more a side effect with how emulation speed is tied with the speed of main loop, which does both message pumping and emulation. Speedup the emulator and you end up spamming message pumping faster, slow it down and you slow down message pumping, the slower that is, less responsive the emulator becomes.

Issue #1483, opened 2019-02-10
Drag and drop with missing firmware blocks everything until firmware window is closed


Masterjun3:
Drag and drop a ROM that triggers the `No BIOS found` window into BizHawk.
After this, the "blocking" already starts and is only fixed by clicking No, or clicking Yes and closing the firmware window.

What I mean by "blocking" is being unable to bring the explorer window (source of the drag and drop) to the front, or even doing anything with it.
Other side effects are, for example, being unable to initiate a drag and drop from a 7-Zip window (it errors with "Unspecified error").

YoshiRulz:
Reproducible with 2.3.1 on Win10, note only the mouse is unable to give focus to Explorer, Alt+TAB works.

Asnivor:
Sounds like this should be easy enough to fix.

Although drag-n-drop from a 7zip window - i'd be interested to see a windows application that doesn't complain about this...

zeromus:
this is how windows works. Everything blocks til the drag is done. If this issue was only about 'no' then I would close this issue right now as "will never, ever fix".
With the firmware dialog opening, the expectations are a little different. I could see this getting fixed.

I don't see how the 7z drag is related at all. It should get blocked same as windows. If it errors, that's a separate bug.

Masterjun3:
The 7-Zip window was just an example to show that blocking is indeed going on. I'm not dragging anything out of the 7-Zip window into anything like how @Asnivor assumed. I'm just holding down on an item, and as soon as I move my mouse one pixel, it errors (obviously, because drag and drop is blocked).

The problem here is how the whole program flow is linked to that one event that was called
`DragDrop -> FormDragDrop -> FormDragDrop_Internal -> _FormDragDrop_internal -> _LoadRom -> LoadRom -> ... -> DoLoadErrorCallback -> OnLoadError -> ShowLoadError -> MessageBox.Show -> FirmwaresMenuItem_Click`

zeromus:
that's not a problem, that's normal. If youve got drag and drop blocked and meanwhile 7zip malfunctions, I don't care. Blocked drag and drop, and whatever consequences, are not our problem. I still say the only consideration here that's important is the firmware manager--if the user chooses that, it needs to be launched by posting a message to the mainform and canceling the drag

Masterjun3:
This issue isn't about 7-Zip anyways. I never said it was. I never said 7-Zip shouldn't error.

[...]

PR #3314, opened 2022-07-15
Opt/run loop optimizations

aineros85:
This changeset includes some (somewhat experimental) performance optimizations for the main run loop.  Optimization was guided with instrumentation and measured with displayed FPS while unthrottled and with Rewind disabled.  I also optimized a few passes to reduce heap allocations and converted some of the most frequently allocated types to structs (e.g. ReturnData).  I attempted to isolate changes to code on the hot path.  For example, ReturnDataStruc could and probably should replace ReturnData universally, but wouldn't have a significant impact on performance.

Each commit was evaluated separately, and for the most part, each commit may be cherry-picked and applied independently.

I verified the changes don't break gameplay on MelonDS, but did not test with other cores.  The most significant change (and most significant gain) was alling wbx_activate_host and wbx_deactivate_host only once, which alone represents about a 30% gain.  FilterProgram is cached, based on the parameters used during creation, although this likely needs to be refined to include additional global configuration settings.

Check if completed:


Morilli:
While some of these changes (like converting heap allocations to struct allocation and thereby reducing GC load) seem like senseful changes we probably want, I cannot seem to reproduce your mentioned 30% gain in fps. Were you testing in release or debug config?

When checking on a normal release build, unthrottled speed, rewind disabled, in pokemon platin (after pressing start on the main menu), I get around 400fps on both and master and your branch (it does seem to be a bit faster on yours though, maybe 5%).

aineros85:
The ~33% gain was with the Release build, running without debugging or instrumentation.  Here are the actual numbers I was seeing with HeartGold.  I haven't seen anything in the neighborhood of 400 fps.  The framerate that I experience seems to match the streamer I follow.

Debug fps (before): 27 fps
Debug fps (after): 109 fps

Release fps (before): 90 fps
Release fps (after): 127 fps

Specifically, I loaded the same savestate each time, which is in the lab before selecting a mon.  On Release build, framerate jumped as high as 170fps but averaged around 127fps over about 10 seconds.  With instrumentation enabled, the wbx_activate_host and wbx_deactivate_host accounted for about 37% of the CPU time within the run loop, with the other enhancements contributing about another 14% reduction.

Morilli:
It would probably be interesting to get data from other people, or alternatively figure out why you're seeing such a high cost of `wbx_activate_host`, because this is how it looks for me:
image

nattthebear:
The ReturnData change should be fine, and I see no reason not to use it both here and for every waterbox call that uses ReturnData.



Please remove the waterbox host enter/exit changes and put them in a separate PR.  There's a lot of complexity there that I'll need some time to think about and discuss it.

aineros85:
Reverted and reapplied all changes except to wbx_activate_host.  Made changes to ReturnData for all function calls.

InfamousKnight:
Can someone send me a link with this pr implemented? As in, one that's compiled already? I wanted to test ares

YoshiRulz:
[automated build](https://gitlab.com/TASVideos/BizHawk/-/pipelines/611503219/builds)

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
```diff
@@ -2707,7 +2707,21 @@ private static void FormDragEnter(object sender, DragEventArgs e)
 		}
 
 		private void FormDragDrop(object sender, DragEventArgs e)
-			=> PathsFromDragDrop = (string[]) e.Data.GetData(DataFormats.FileDrop);
+			=> this.BeginInvoke(((Action<DragEventArgs>)(args =>
+			{
+				var pathsFromDragDrop = (string[])args.Data.GetData(DataFormats.FileDrop);
+				if (pathsFromDragDrop is not null) this.DoWithTempMute(() =>
```
YoshiRulz, line 2713: This was moved out of the drop event handler because of #1483—does this commit reintroduce that bug?

src/BizHawk.Client.Common/Controller.cs
```diff
@@ -34,11 +34,11 @@ public IReadOnlyCollection<(string Name, int Strength)> GetHapticsSnapshot()
...
 
 		private readonly WorkingDictionary<string, List<string>> _bindings = new WorkingDictionary<string, List<string>>();
-		private readonly WorkingDictionary<string, bool> _buttons = new WorkingDictionary<string, bool>();
-		private readonly WorkingDictionary<string, int> _axes = new WorkingDictionary<string, int>();
+		private readonly WorkingValueDictionary<string, bool> _buttons = new WorkingValueDictionary<string, bool>();
+		private readonly WorkingValueDictionary<string, int> _axes = new WorkingValueDictionary<string, int>();
 		private readonly Dictionary<string, AxisSpec> _axisRanges = new WorkingDictionary<string, AxisSpec>();
 		private readonly Dictionary<string, AnalogBind> _axisBindings = new Dictionary<string, AnalogBind>();
-		private readonly Dictionary<string, int> _haptics = new WorkingDictionary<string, int>();
+		private readonly WorkingValueDictionary<string, int> _haptics = new WorkingValueDictionary<string, int>();
```
YoshiRulz, line 41: I'm interested in just how expensive this boxing was. If it's a lot I'm thinking we could make a new class specifically for this use-case rather than use the same thin wrapper over `Dictionary<K, V>`.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -784,7 +787,9 @@ public int ProgramRunLoop()
 
 				Render();
 
-				CheckMessages();
+				++frameCount;
+				if (0 == (frameCount % MessageCheckFrame))
+					CheckMessages();
```
Morilli, line 792: I very much disagree with this change. This will cause the UI to become even slower to respond to user interaction than it already is, given it waits for a full emulator frame to complete (and now 5!).

CasualPokePlayer: This would make https://github.com/TASEmulators/BizHawk/issues/2959 even worse than it is

YoshiRulz:
[Backup on GitLab mirror](https://gitlab.com/TASVideos/BizHawk/-/commits/aineros85-optim-202207)

PR #3326, opened 2022-07-23
Prevent the BasicBot from loading invalid addresses into the Tiebreaker text boxes.

tommai78101:

Before this change, when the Basic Bot is saving the bot parameters the user made, if the user did not set any address values to the address boxes (including Main Value address box), by default, it will store `0` as the value in the produced .BOT file. When Basic Bot is loading a .BOT file (a file that's produced by Basic Bot itself when saving the bot parameters), if the address boxes' values are `0` in the .BOT file, it will populate it with `0000`. Because it's assuming the user entered in `0000` when the user really didn't do this.

With this change, what I did is to prevent the Basic Bot from saving an empty address box as `0`, so that when it's loading the .BOT file, it won't produce the address value `0000`. I used `int.MinValue` to indicate this, because I'm pretty sure any emulators won't be using some address value located around 2.1 billion places away.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/BasicBot/BasicBot.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/BasicBot/BasicBot.cs
@@ -643,10 +643,18 @@ namespace BizHawk.Client.EmuHawk
 				control.Probability = p;
 			}
 
-			MaximizeAddress = botData.Maximize;
-			TieBreaker1Address = botData.TieBreaker1;
-			TieBreaker2Address = botData.TieBreaker2;
-			TieBreaker3Address = botData.TieBreaker3;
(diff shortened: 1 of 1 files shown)


YoshiRulz:
Not sure about this. Addresses will be `ulong` eventually so using negative numbers won't go well. Maybe `null` would be better here?

tommai78101:
@YoshiRulz I don't really know how to put a `null` in a type that's `int` in the first place, and I was afraid that if I change the type from `int` to something like `int?` in `BotData` class, I may break existing .BOT files. 

On second thought, if the addresses will eventually become `ulong`, shall we make this change before 2.9?

YoshiRulz:
^ seems to work, closing

PR #3343, opened 2022-07-29
Source Generator for null-safe `[RequiredService]` boilerplate

YoshiRulz:
Applying the new attribute one or more times, such as:
https://github.com/TASEmulators/BizHawk/blob/e34e199e551303b1b1d28435396a0f582e440c17/src/BizHawk.Client.EmuHawk/tools/Watch/RamSearch.cs#L24-L27
results in:
```csharp
public partial class RamSearch
{
	[RequiredService]
	public IEmulator? _maybeEmu
	{
		[Obsolete("use Emu")] get;
		[Obsolete("only service provider should be writing to this")] private set;
	} = null;
	private IEmulator Emu => _maybeEmu!;

	[OptionalService]
	public IInputPollable? InputPollableCore
	{
		get;
		[Obsolete("only service provider should be writing to this")] private set;
	} = null;

	[RequiredService]
	public IMemoryDomains? _maybeMemoryDomains
	{
		[Obsolete("use MemoryDomains")] get;
		[Obsolete("only service provider should be writing to this")] private set;
	} = null;
	private IMemoryDomains MemoryDomains => _maybeMemoryDomains!;
}
```
This boilerplate uses nullity annotations and gives build warnings if the wrong prop is used. But rather than copy it across the whole codebase, I thought codegen would be more suitable. What do people think? If it's well-recieved I can do the same for `[RequiredApi]`.


YoshiRulz:
I've been rebasing this changeset this week. 5f77a310c5 and 59562d893 were to address 16755d345, and now I only have a few classes left to double-check. One of the problematic ones is `SNESGraphicsDebugger`. @Morilli Could you check it's unhooking itself from the core instance properly?

Morilli:
> Could you check it's unhooking itself from the core instance properly?

Uhh I don't actually know how it should be doing that, but the way the code works I can almost guarantee you it's not doing stuff correctly.

Issue #3344, opened 2022-07-31
Use better compression for savestates and project files


Morilli:
As discussed on discord, the zip compression currently used is far inferior to more modern methods, and would do good being replaced with a better and faster method, for example zstd.

This could benefit both savestates and .bk2 and .tasproj files, resulting in smaller project files and faster savestates.

getCursorsExe:
Seems to be a good idea, but this does NOT MEAN dolphin and citra will appear.

nattthebear:
What, zip file with zstd compressed entries?  Not until all major zip software supports it.

Morilli:
> Not until all major zip software supports it.

Well in that case, let's wait another 20 years, or potentially till the end of time. If you consider this path problematic, perhaps just compressing specific binary files (inside of the zip) additionally could work as well.

nattthebear:
No, we don't need every last random zip program, but certainly the major ones on Windows, which would include the Windows file explorer.  One of the major design goals of the current line of formats is for end users to be able to open them and manipulate them without programming knowledge or special programs.

Morilli:
> One of the major design goals of the current line of formats is for end users to be able to open them and manipulate them without programming knowledge or special programs.

Right, which is why I think this "double-compression" could work fine, aka keep zips as they are, and just compress and decompress specific files an additional time internally. Especially the greenzone file can become quite large (500MB for me), which takes a long time to compress using deflate and is not really editable for an end-user anyway.

nattthebear:
I'm interested to see the real world gains that we'll get with 2c56da5

vadosnaprimer:
Greenzone indeed takes way too long to save to file.

The previous implementation was able to only save every 16th state (user adjustable iirc), because you don't need all the immediate states from the previous session, only some basic anchors. And if saving is quick, it feels transparent and doesn't interrupt the workflow.

Since to work comfortably you need to have a decent number of immediate states, now they take several seconds to save every time. So autosaving is now a problem, unless one forces saving to a backup file, but then autosaving as regular tasproj is kinda pointless...

CasualPokePlayer:
Started a separate go here: https://github.com/CasualPokePlayer/BizHawk/commit/aae12f3c1f847cc58fa15bd3217950a16092f847

[...]

PR #3345, opened 2022-08-03
Zstd Compression

CasualPokePlayer:
Resolves https://github.com/TASEmulators/BizHawk/issues/3344

A bit of an alternative to https://github.com/TASEmulators/BizHawk/commit/2c56da52a05e914b903a4ac439d5ac14b9342931, opting to create a slimmed down wrapper around libzstd. Only use is compression and decompression streams currently (wrapped around libzstd's streaming API).

Zip files created by Hawk after this change will now always any binary blobs in them zstd compressed first (before the zip itself is deflated). Anything text is not zstd compressed (as that would be rather annoying for users and of course very annoying for TASVideos' movie parser). The BizState 1.0 is now included in all zips and bumped up to 1.0.2 indicate this change, with code in place to ensure backwards compatibility where possible.

With this greenzone files are compressed MUCH better than with just deflate, I'm seeing compression ratios of 798 vs 9 for similar greenzone files, and tasproj's are saved FASTER to boot, never more than a second.

~~Kept as draft currently, while this all works fine on Windows, Linux just shits the bed for some reason, somehow compression grows a savestate to the point it's too big for the rewind buffer??? And loading in a state just crashes with some src size is wrong error. I suspect there's an interop mistake somewhere...~~

Check if completed:


nattthebear:
>I suspect there's an interop mistake somewhere...

I took a look at the zstd header file to compare, and the interop definitions seem fine so long as we stick to 64 bit only.  For debugging purposes, you could try the compatibility mode on BizImport to see if it changes anything.

CasualPokePlayer:
Seems like the issue is my libzstd.so I provided. Or rather, its conflict with my system libzstd.so. It seems like both get loaded... somehow. It's probably due to the weird symlinking that library normally does with its so's for some reason.

CasualPokePlayer:
Seems like trying to open libzstd.so.1 is the "solution." 

Probably best to not provide a libzstd.so in the first place, libgdiplus has a dependency on libzstd so a Linux user must have it present and loaded anyways.

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
```diff
@@ -9,7 +9,8 @@ namespace BizHawk.Client.Common
 	public class ZipStateLoader : IDisposable
 	{
 		private ZipArchive _zip;
-		private Version _ver;
+		private Version _zipVer;
+		private Version _stateVer;
```
Morilli, line 13: Someone used `Version` here initially, which is definitely not the right class to use. Doesn't an `int` suffice?
That would probably make the entire logic simpler and easier to understand as well.

src/BizHawk.Client.Common/zstd/Zstd.cs
```diff
@@ -0,0 +1,342 @@
...
+			private readonly byte[] _outputBuffer;
+			private readonly GCHandle _outputHandle;
+			private LibZstd.StreamBuffer _output;
+
+			// TODO: tweak these sizes
+
+			// 4 MB input buffer
+			private const int INPUT_BUFFER_SIZE = 1024 * 1024 * 4;
+			// 1 MB output buffer
+			private const int OUTPUT_BUFFER_SIZE = 1024 * 1024 * 1;
```
Morilli, line 30: Would recommend using `ZSTD_CStreamInSize` and `ZSTD_CStreamOutSize` here instead of these hardcoded (and very large) sizes.

src/BizHawk.Client.Common/savestates/ZipStateSaver.cs
```diff
@@ -8,9 +8,16 @@ namespace BizHawk.Client.Common
...
 	{
 		private readonly IZipWriter _zip;
-		private bool _isDisposed;
+		private bool _isDisposed;
+
+		private static void WriteZipVersion(Stream s)
+		{
+			var sw = new StreamWriter(s);
+			sw.WriteLine("0"); // version 1.0.0
+			sw.Flush();
```
Morilli, line 17: Probably reads better with a `using var sw = new StreamWriter(s);`; which should also nullify the need for flushing.

Morilli (commented):
Have tested tasproj loading, saving and general states with the rewinder using this, and it all seems to work just fine.

What exactly happens with old compressed savestate files or tasproj with compressed states though?

src/BizHawk.Client.Common/zstd/Zstd.cs
CasualPokePlayer: I used very large sizes due to the .h file stated those functions are only good in a C environment, in an environment like C# where there is a cost to interop it's recommended to just use very big buffers to reduce interop calls.

CasualPokePlayer:
Old compressed files are still backwards compatible, as they will be missing the BizZip 1.0 file completely, in which case I just initialize the _zipVersion with version 0.0.0, and only decompress if major version > 0.

src/BizHawk.Client.Common/zstd/Zstd.cs
Morilli: Do you have some link for that? I can't find it right now, and when testing myself I'm also not really seeing any performance loss from using the recommended buffer sizes.

Morilli:
> Old compressed files are still backwards compatible, as they will be missing the BizZip 1.0 file completely

Ah, didn't realize savestate files are also just zip. In that case, looks good.

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -276,10 +276,9 @@ private void AddToReserved(ZwinderBuffer.StateInformation state)
 				return;
 			}
 
-			var bb = new byte[state.Size];
-			var ms = new MemoryStream(bb);
+			var ms = new MemoryStream();
 			state.GetReadStream().CopyTo(ms);
-			_reserved.Add(state.Frame, bb);
+			_reserved.Add(state.Frame, ms.ToArray());
```
nattthebear, line 281: This is regrettable, although I understand why you did it.

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
```diff
@@ -9,7 +9,8 @@ namespace BizHawk.Client.Common
 	public class ZipStateLoader : IDisposable
 	{
 		private ZipArchive _zip;
-		private Version _ver;
+		private Version _zipVer;
```
nattthebear, line 12: Could we just bump the state version to something new instead of introducing a second version?

nattthebear (commented):
How fast is the rewinder with compression on compared to the old one?

src/BizHawk.Client.Common/zstd/Zstd.cs
CasualPokePlayer: https://github.com/facebook/zstd/blob/f5c4ec465817f77ea04346659e284e2c1c5c6f27/lib/zstd.h#L724-L731

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
CasualPokePlayer: Movies do not have this included it seems (seems like this was meant only for savestates?)

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
CasualPokePlayer: The previous logic crashes on 2.8 too fwiw (I guess showing people don't use this compression in tastudio).

CasualPokePlayer: The size used for creating the array is the compressed size, and reading the stream like CopyTo does will read out the decompressed size, and memory streams init with an array cannot be resized.

nattthebear: Right, I get that.  Does the current live code actually allow you to select compression for the state manager?

CasualPokePlayer: The current code allows for selecting compression or no compression. In rewind menu it's "zlib compression" (reworded in this PR as "zstd compression"). In TAStudio it's in the "State History Settings." Delta compression is a completely separate rewinder.

nattthebear: So it's busted, hmm hmm.

The delta rewinder doesn't allow the kind of access pattern that tarsestudio needs, so yeah, that's a nonstarter.

CasualPokePlayer:
Comparing Pokemon SoulSilver in Oak's intro dialogue:

Rewind Disabled: ~185 FPS
No compression: ~175 FPS, Avg. State Size 13.38 MB
zlib (Deflate) compression: ~14 FPS, Avg. State Size 2.04 MB
zstd compression: ~50 FPS, Avg. State Size 2.04 MB
delta compression: ~69 FPS, Avg. State Size 3.47 KB

So yes zstd performs much faster than zlib compression (seems to be around the same ratio in this case). Delta beats it out in all fronts, although it can't be used in TAStudio anyways so zstd has that going for it.

nattthebear:
Nice improvement.  Yeah, I think we could merge this.

src/BizHawk.Client.Common/zstd/Zstd.cs
Morilli: Thanks. I guess what I just don't like is the fact that 5MB of buffer are allocated every time a zstd stream is initialized, which could happen as often as once per frame (300mb of managed allocations per second).

CasualPokePlayer: Perhaps this could be reworked to have a separate "context" class, dictionary of which is kept around statically and may be reused for different streams (only if said dictionary doesn't have an unused context instead wil a new context is created and added).

nattthebear: Ideally we'd get rid of the extra buffers+copies entirely, but at least it's faster than what we have now

CasualPokePlayer: See e5514e25e5611a1adc1896ed558186c29e754801

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
CasualPokePlayer: Ended up just making all movies put in the BizState 1.0 file and bumped it up to 1.0.2.

On related https://github.com/TASEmulators/BizHawk/pull/3345#discussion_r936672311

src/BizHawk.Emulation.Common/zstd/Zstd.cs
```diff
@@ -0,0 +1,505 @@
...
+			public ZstdCompressionStream(Stream baseStream, ZstdCompressionStreamContext ctx)
+			{
+				_baseStream = baseStream;
+				_ctx = ctx;
+			}
+
+			~ZstdCompressionStream()
+			{
+				Dispose();
+			}
```
Morilli, line 106: Due to the unconditional access of managed fields in the dispose function, this isn't actually safe to do and could crash the entire program.
This should probably either be a `Dispose(false)` or not implement the finalizer at all and hope this always gets disposed correctly.

You should be able to reproduce this failure by slamming
```csharp
for (int i = 0; i < 100; i++)
	new Zstd().CreateZstdCompressionStream(new MemoryStream());
```
somewhere, like in main.

Issue #3347, opened 2022-08-06
Implement HW Accelerated CRC32/SHA1


CasualPokePlayer:
More QoL for a user (especially for big ROMs), and especially nice for PSX full disc hashing. CRC32 hw acceleration needs `CLMUL` and SHA1 needs `SHA` instructions. While it's fairly unlikely we'd see anyone with CPUs without support for CRC32, SHA instructions are a bit recent and probably not supported on a random user's CPU. In any case, these instruction sets should be detected and used only if possible, and otherwise fallback on the currently used CRC32/SHA1 implementations.

CRC32 is really only used for disc hashing, which considering it's very likely HW acceleration would be supported by a user's CPU (and if it isn't have fun running any disc based system in the first place lol), implementing this would offer a substantial boost for a full disc hash. So much so it might be a good idea to just do full disc hashing with CRC32 always (and don't bother with the "quick" hashing).

There's also a related problem of hashing functions being called multiple times when GameInfo could maybe just contain every hash (except for discs obviously, those could just put in N/A for MD5/SHA1 fields). But that's for another time.

PR #3348, opened 2022-08-07
Implement hardware accelerated CRC32 and SHA1, using them if possible

CasualPokePlayer:
Resolves https://github.com/TASEmulators/BizHawk/issues/3347

CRC32's generic function is also replaced with zlib's as it is much more performant than our implementation.
Full hash of a ~731MB disc took only ~369 ms with this, and the generic CRC32 isn't so far behind at ~659 ms.
We probably should consider just always doing a full disc hash with this change, as it no longer takes around 4.3 seconds to do a full disc hash.

Also if someone's computer supports SHA instructions I'd like them to test this as I'm unable to do so with my PC.


ExternalProjects/LibBizHash/README.md
```diff
@@ -0,0 +1,21 @@
+LibBizHash is the unmanaged side for BizHawk's hashing.
+
+CRC32 code is taken from [zlib-ng](https://github.com/zlib-ng/zlib-ng) with massive slashing of code and various tweaks. This code is licensed under the zlib license.
+SHA1 is code is taken from [SHA-Intrinsics](https://github.com/noloader/SHA-Intrinsics) with some tweaks. This code is under the public domain.
+All modifications by the BizHawk team are licensed under MIT.
```
YoshiRulz, line 5: nitpick: "modifications" are not licensed, the modified copy of the source is. More importantly this should include build instructions.

src/BizHawk.Common/checksums/CRC32.cs
```diff
@@ -66,69 +83,41 @@ private static uint gf2_matrix_times(ReadOnlySpan<uint> mat, uint vec)
...
-				// if no more bits set, then done
-			} while (len != 0U);
-
-			// finally, combine and return
-			_current ^= crc;
+				len >>= 1;
+			} while (len != 0U); // if no more bits set, then done
+
+			// finally, combine and return
+			_current ^= crc;
```
YoshiRulz, line 120: no whitespace-only changes please

CasualPokePlayer: I think most of that was Visual Studio being a butt here with newlines, not sure how I fix this.

Morilli (approved):
Changes look good from my side.
According to the SHA and CRC tests, everything works fine.
I can't really test the performance improvement of the fast SHA implementation given I don't have a capable system, but I'm certain there's a ~4x speedup here for people that do.

Feel free to check my commit again; I tested it, but another look can't hurt.

PR #3349, opened 2022-08-08
Add .git-blame-ignore-revs

CasualPokePlayer:
This could be used for ignoring giant whitespace changes in git blames.
With this we could commit to doing giant tab/space + newline normalization in .cs files without blame noise.
The commit used here is used more for POC, although still nice for getting rid of the blame noise from it.


YoshiRulz:
GitHub doesn't respect this unfortunately. I still like this idea.

Morilli:
Well, this feature seems to be intended for `git blame` only; it works here: https://github.com/YoshiRulz/git-blame-ignore-revs-poc/blame/master/src/BizHawk.Client.EmuHawk/EmuHawkUtil.cs

YoshiRulz (approved):

YoshiRulz:
> it works here

...or would, if the commit changed any lines in that file.

I've actually changed my mind and am now skeptical of this feature. How exactly does it "see beyond" the filtered commit(s)? Does it only try if the filter commit(s) are whitespace-only? In any case, I'd like to see it in action.

Morilli:
https://github.com/Morilli/BizHawk/blame/master/.editorconfig ... I think that's working?

YoshiRulz (approved):

adelikat (approved):

YoshiRulz:
Care to revisit this? As per my edits above, I think we should use this feature only for whitespace changes (and file renames).

Issue #2261, opened 2020-07-31
Problematic dependencies


YoshiRulz:
package | targets | purpose / notes
--:|:--|:--
`FlatBuffers.Core` | ? | Unknown (`Cores`). IDK, natt hacked this in for Nyma. Presumably serialisation.
`GongShell` | Framework (Win, Linux?) | UI (`EmuHawk`). Used in the code that generates "recents" submenus.
`JunitXml.TestLogger` v`2.1.32` | Core 3.1 (Win, Linux) | Tests. Writes test results in a format readable by GitLab CI. As GitLab can read several formats, this may be replaceable.
`Microsoft.VisualBasic` v`10.3.0` | Standard 2.0 | Startup (`EmuHawk`). Type `WindowsFormsApplicationBase` inherited by `EmuHawk.Program.SingleInstanceController`. No compatibility issues, but adds a runtime dep that's annoying to install on Linux.
`Newtonsoft.Json` v`12.0.3` | Standard 2.0 | Serialisation. (There's actually nothing wrong with this, I just think we should migrate to `System.Text.Json`.)
`NLua` | Framework (Win) | Lua (`Client.Common`, `EmuHawk`). Needs to be loaded at runtime for either Lua engine to function. See #2260.
`OpenTK` v`3.0.1` | Framework (Win, Linux) | Host graphics (`BizwareGL`, `EmuHawk`), host input/sound (`EmuHawk`). The rewrite targeting Standard, OpenToolkit 4, is nearing release.
`OpenTK.GLControl` v`3.0.1` | Framework (Win, Linux) | Host graphics (`EmuHawk`). IIRC natt said this isn't in OTK4.
`SlimDX` | Framework (Win) | Host graphics (`EmuHawk`).
`System.Data.SQLite` | Framework (Win, Linux?) | DB manip (`Client.Common`). Probably very easy to replace w/ NuGet package.
`System.Web` | Framework (Win, Linux?) | Startup (`DiscoHawk`). Single usage in static method `DiscoHawk.GetExeDirectoryAbsolute`.
`System.Windows.Forms` | Framework (Win, Linux), .NET 5 (Win) | UI (`WinForms.Controls`, `EmuHawk`). I don't see us changing UI framework before .NET 5.

`GongShell`, `Microsoft.VisualBasic`, `System.Data.SQLite`, and `System.Web` seem to be easy targets.

nattthebear:
Could we leave the net standard libs out of this?  They're already net core compatible

[...]

Issue #2260, opened 2020-07-31
MoonSharp / the future of Lua


YoshiRulz:
(gonna flesh this out more later, see existing discussion in https://github.com/TASVideos/BizHawk/issues/971#issuecomment-549660627, #1415, and #1422)

YoshiRulz:
(reserved because I can imagine this thread getting real long)

CasualPokePlayer:
> edit 2022-05-07: [KeraLua](https://github.com/NLua/KeraLua) is another .NET Standard 2.0 library, but supports Lua 5.4 features, unlike MoonSharp.

To clarify here, KeraLua is the new Lua "backend" for NLua. Our NLua is still when they used KopiLua, and we only are able to use native Lua with a hacky switcharoo to give it our LuaInterface instead of KopiLua.

It would probably be easiest (and best in terms of lua functionality) overall to just update our NLua to upstream and delete our LuaInterface. I tried a bit here: https://github.com/CasualPokePlayer/BizHawk/tree/moonshot and it seems to execute commands within the console without completely crashing, so it doesn't seem so bad to do.

YoshiRulz:
https://github.com/CasualPokePlayer/BizHawk/commit/c41096ac01bc88040b286ceb483e2860dc9fdb7b seems to work on Linux

CasualPokePlayer:
> * [Pluto](https://pluto-lang.org) (fork of Lua 5.4, probably no C# bindings)

It would appear the API (as far as bindings are concerned) is identical to Lua 5.4, so simply dropping in a different .dll/.so would suffice to "support" this (hell you could make it literally a switch in the config whether you want Pluto or pure Lua 5.4)
edit: Even better, it has a [Lua-compatible parsing mode](https://pluto-lang.org/docs/Compatibility) which scripts can bypass with an opt-in comment! --yoshi

PR #3361, opened 2022-08-18
Move to NLua/KeraLua

CasualPokePlayer:
This is a moonshot update to our NLua, which now uses KeraLua / native Lua 5.4. Some things work here, a lot of things just crash horribly. A bit of refactoring might be needed here? See #2260 and #2261


CasualPokePlayer:
Most things seem to work fine now? I had to remove a yield that was placed in stop that resolved some earlier issue, as it now just causes NREs (something NativeWindow callback was garbage collected? not sure what's going on). May not be needed anymore (need to check).

There seems to be an interesting quirk with the console window. If you put in print("foo") (or some other printing command), it prints foo out then nil. Internally this is actually console.log(print("foo")). In old Lua print returns nothing, and outputs is null, and thus that console.log doesn't output anything. But now with newer lua versions outputs contains 1 element with simply null, and thus nil is outputted. Not sure what's the proper way to deal with this, should we treat outputs with one null the same as no return?

YoshiRulz:
> should we treat outputs with one null the same as no return

Lua does not have void returns, so yes.

YoshiRulz:
> How does it directly print tables now?

before:
```
> memory.read_bytes_as_array(0xABCD, 10)
"1": "33"
"10": "253"
"2": "159"
"3": "53"
"4": "47"
"5": "184"
"6": "61"
"7": "193"
"8": "127"
"9": "38"

> memory.read_bytes_as_dict(0xABCD, 10)
"43981": "33"
"43982": "159"
"43983": "53"
"43984": "47"
"43985": "184"
"43986": "61"
"43987": "193"
"43988": "127"
"43989": "38"
"43990": "253"

```
(both kinds have an extra blank line)
after:
```
> memory.read_bytes_as_array(0xABCD, 10)
33	159	53	47	184	61	193	127	38	253
> memory.read_bytes_as_dict(0xABCD, 10)
47	184	61	193	127	38	253	33	159	53
```
(those are tabs)

vadosnaprimer:
Was the table dumper our custom thing? Regular lua can't unwrap tables when printing, and something similar seems to be the case with that new thing.

CasualPokePlayer:
The table dumper was custom in hawk. The outputs of the Log function would contain a raw LuaTable, which we have custom code to turn that into a string.

It seems however that outputs just contains only the keys of a LuaTable rather than the raw LuaTable itself, so Lua might be unpacking it now?

CasualPokePlayer:
Seems the unpacking stuff is just "normal" upstream behavior, not unpacking seems to be something tacked on by @zeromus. The source of the hacked NLua doesn't seem available, so not sure what exactly we could even submit to upstream to "fix" this.

zeromus:
should be here
https://github.com/TASEmulators/nlua

CasualPokePlayer:
Hmm, suppose might be a good idea to just keep forking off NLua and just update our fork. I'd like to PR some of these changes anyways to upstream and hopefully have them accepted to we can just NuGet them and not worry about maintaining a fork, although no telling when they do their next release.

@zeromus Could I be given write access to our NLua fork?

zeromus:
done

CasualPokePlayer:
Turns out most of our changes in NLua are already implemented upstream/not needed in NLua/irrelevant, only the Lua table unpacking differs in our fork. Probably easy enough to swap over to NuGet after upstream has that fixed (I plan on making a PR to add a switch for disabling that unpacking) and a new release is out.

Still, what's here should be the same, although might want to do some checking before merging this here.

YoshiRulz:
I'd like to see regression tests implemented on master before merging this.

PR #3362, opened 2022-08-19
Use NuGet central package management

YoshiRulz:
[docs](https://docs.microsoft.com/en-gb/nuget/consume-packages/Central-Package-Management)

This feature is supported in VS2022 from 17.2 and in the SDK from 6.0.300. ~~@zeromus plz check~~ This PR is finished, just waiting for everyone to have a compatible dev environment before merging.

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -7,12 +7,14 @@
 		<AnalysisModePerformance>Recommended</AnalysisModePerformance>
 		<AnalysisModeUsage>Recommended</AnalysisModeUsage>
 		<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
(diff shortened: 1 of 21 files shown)


zeromus:
1. Are you asking me if I'll sanction dropping support for vs2019? No. I won't require any bleeding edge "released this year" visual studio.
2. Again with the commits with no given rationale. WHY?

YoshiRulz:
1. Was hoping you'd updated since we had to revert C# 10 in March. There's really no rush, you've still got more than 2 years until .NET 6 goes EOL.
2. This NuGet feature is mostly for convenience. If it's not obvious why only having to touch one file is advantageous, well, nothing I say is going to convince you. But unlike some of the other shiny new things I advocate for, this isn't something you'd spend a lot of time looking at.

PR #3367, opened 2022-08-21
Implement snes graphics debugger for the new bsnes core

### LINKED ISSUE #3045: Implement SNES Graphics Debugger to BSNESv115+ core

Morilli:
- closes #3045.

This PR aims to implement the functionality of the existing SNES graphics debugger for the new bsnes core (BSNESv115+).
Note that most of the implementation was copied and just adjusted for better code style or readability.

I have cleaned up the code a bit in the process, hopefully without breaking anything.

Noteworthy things:
- changing palette is currently not implemented; I decided that the effort required to support it is just not worth it, especially when considering the fact that bsnes and bizhawk palettes appear undifferentiable and the snes9x palette is probably wrong anyway
- the ScanlineHookManager is not implemented currently, but that could come in the future when someone wants to put in the effort required for it
- a new memory domain `OBJECTS` is exposed which is basically a parsed/objectified version of the `OAM` memory domain's content. This may be weird for people trying to use this as the `OAM` memory domain that the old core provided, but changing this would require core changes that I'm not willing to invest into.


src/BizHawk.Client.EmuHawk/tools/SNES/SNESGraphicsDebugger.cs
```diff
@@ -1263,15 +1268,16 @@ private void MessageTimer_Tick(object sender, EventArgs e)
 		private void comboPalette_SelectedIndexChanged(object sender, EventArgs e)
 		{
 			if (suppression) return;
-			var pal = (SnesColors.ColorType)comboPalette.SelectedValue;
+			var pal = comboPalette.SelectedValue.ToString();
 			Console.WriteLine("set {0}", pal);
-			var s = Emulator.GetSettings();
-			s.Palette = pal.ToString();
-			currentSnesCore?.PutSettings(s);
+			currentSnesCore?.SetPalette((SnesColors.ColorType)comboPalette.SelectedIndex);
```
YoshiRulz, line 1273: `pal` is now unused, what did you intend?

Morilli: Well, it's now only used in the `Console.WriteLine` call, which I just kept from before.

nattthebear (approved):

PR #3373, opened 2022-08-27
Lua tests

CasualPokePlayer:
re: https://github.com/TASEmulators/BizHawk/pull/3361#issuecomment-1221564853

### DIFF
--- a/src/BizHawk.Tests/Client.Common/lua/LuaTests.cs
+++ b/src/BizHawk.Tests/Client.Common/lua/LuaTests.cs
@@ -149,7 +149,7 @@ namespace BizHawk.Tests.Client.Common.Lua
 						o!.Call(d);
 						break;
 					case string s:
-						o!.Call(s);
+						o!.Call(UnFixString(s));
 						break;
 					case NLua.LuaTable t:
 						o!.Call(t);
@@ -423,7 +423,7 @@ namespace BizHawk.Tests.Client.Common.Lua
 		public void Lua_Argument_String_Utf8()
 		{
 			ExpectedValue = true;
-			CallackArg = UnFixString("こんにちは"); // seriously, wtf?
+			CallackArg = "こんにちは";
 			LuaInstance.DoString("pass_callback(function(foo) pass_bool(foo == \"こんにちは\") end)");
 		}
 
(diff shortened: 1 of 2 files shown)


src/BizHawk.Tests/Client.Common/lua/LuaTests.cs
```diff
@@ -0,0 +1,427 @@
...
+		public static void PassF32(float? o)
+			=> Assert.IsTrue(o == (float?)ExpectedValue);
+
+		[LuaMethod("pass_f64", "")]
+		public static void PassF64(double? o)
+			=> Assert.IsTrue(o == (double?)ExpectedValue);
+
+		[LuaMethod("pass_f128", "")]
+		public static void PassF128(decimal? o)
+			=> Assert.IsTrue(o == (decimal?)ExpectedValue);
```
YoshiRulz, line 86: `decimal` is not [binary128](https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format)

CasualPokePlayer: https://docs.microsoft.com/en-us/dotnet/api/system.decimal?view=netstandard-2.0
"The binary representation of a Decimal value is 128-bits consisting of a 96-bit integer number, and a 32-bit set of flags representing things such as the sign and scaling factor used to specify what portion of it is a decimal fraction."

Different significand/exponent width but it adds up to 128 bits in the end.

src/BizHawk.Tests/Client.Common/lua/LuaTests.cs
```diff
@@ -0,0 +1,874 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace BizHawk.Tests.Client.Common.Lua
+{
+	[TestClass]
+	public class LuaTests
+	{
+		private static readonly NLua.Lua LuaInstance = new();
+		private static readonly NLuaTableHelper _th = new(LuaInstance, Console.WriteLine);
```
YoshiRulz, line 19: Can you subclass `LuaLibraryBase` to get this stuff?

CasualPokePlayer: Doesn't work. LuaLibraryBase doesn't have an NLua.Lua in the first place, and the NLuaTableHelper is just obtained using the IPlatformLuaLibEnv passed in the ctor.

PR #3402, opened 2022-10-02
Don't send key release events with modifiers

### LINKED ISSUE #3327: Input Bug
Bizhawk is skipping input when I hold down "CTRL" key as input.
On the video below you can notice that Lara stops shooting for a while and then starts shooting again. She should shoot all the time, because I hold down "CTRL" key all the time.

https://youtu.be/xO-29MNUT_U

Morilli:
This would close #3327. As I've written in that issue thread, the main problem seems to be the fact that key release events were sent with their respective modifier, aka when holding e.g. `Ctrl+C` and releasing `C`, a `Release:Ctrl+C` event would be generated which also happned to release the `Ctrl` key, even though that was still pressed.

The comments and code which I've removed seem to imply that this functionality was (more or less) intended and had a use case, although in my testing I couldn't figure out an issue with removing it; @zeromus do you recall the specific reason this code was needed?

### DIFF
--- a/src/BizHawk.Client.Common/input/InputCoalescerControllers.cs
+++ b/src/BizHawk.Client.Common/input/InputCoalescerControllers.cs
@@ -1,7 +1,7 @@
 #nullable enable
 
 using System.Linq;
-
+using BizHawk.Common.StringExtensions;
 using BizHawk.Emulation.Common;
 
 namespace BizHawk.Client.Common
@@ -20,8 +20,11 @@ namespace BizHawk.Client.Common
 			Buttons[button] = state;
 			ProcessSubsets(button, state);
 			if (state) return;
-			// when a button is released, all modified variants of it are released as well
-			foreach (var k in Buttons.Keys.Where(k => k.EndsWith($"+{ie.LogicalButton.Button}")).ToList()) Buttons[k] = false;
+			// when a button or modifier key is released, all modified key variants with it are released as well
+			foreach (var k in Buttons.Keys.Where(k =>
+						k.EndsWith($"+{ie.LogicalButton.Button}") || k.StartsWith($"{ie.LogicalButton.Button}+") || k.Contains($"+{ie.LogicalButton.Button}+"))
+						.ToArray())
+				Buttons[k] = false;
 		}
 	}
 
@@ -29,6 +32,13 @@ namespace BizHawk.Client.Common
 	{
 		protected override void ProcessSubsets(string button, bool state)
 		{
+			// we don't want a release of e.g. "Ctrl+E" to release Ctrl as well
+			if (!state)
+			{
+				Buttons[button.SubstringAfterLast('+')] = state;
+				return;
+			}
+
 			// For controller input, we want Shift+X to register as both Shift and X (for Keyboard controllers)
 			foreach (var s in button.Split('+')) Buttons[s] = state;
 		}
--- a/src/BizHawk.Client.EmuHawk/Input/Input.cs
+++ b/src/BizHawk.Client.EmuHawk/Input/Input.cs
@@ -64,7 +64,6 @@ namespace BizHawk.Client.EmuHawk
 			_updateThread.Start();
 		}
 
-		private readonly Dictionary<string, LogicalButton> _modifierState = new Dictionary<string, LogicalButton>();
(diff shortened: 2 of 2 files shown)


zeromus:
Without even thinking about it or checking the code, it seems obvious to me that inputs and bindings connected to e.g. Ctrl+C need to receive a PRESS and RELEASE both or else they'll malfunction. "Release:Ctrl+C event would be generated which also happened to release the Ctrl key" sounds like a bug to be solved directly.

Morilli:
Hm, you're probably right about that. I couldn't find the exact place where that input event was handled though.

CasualPokePlayer:
I believe https://github.com/TASEmulators/BizHawk/blob/5c59e6d9c4d56583f5818fbacc24f01d579f78ac/src/BizHawk.Client.Common/input/InputCoalescerControllers.cs#L30-L34 would be the cause of this issue.

zeromus:
commit 726f1f3f112e9b8947f1b35ea659717604bfdd31 seems considerably different, where it had this before; unclear why a MOVE CODE commit had to be a CHANGE ALGORITHM commit also at the same time
```
var releases = Buttons.Where((kvp) => kvp.Key.Contains("+") && kvp.Key.EndsWith(ie.LogicalButton.Button)).ToArray();
foreach (var kvp in releases)
	Buttons[kvp.Key] = false;
```

YoshiRulz:
You're comparing different code blocks; the matching block before the move is:
https://github.com/TASEmulators/BizHawk/blob/25da1153b69a25ba9bac21ccf394a8f439640536/src/BizHawk.Client.EmuHawk/Input/Input.cs#L43-L51

CasualPokePlayer:
Another thing I guess to note (logging events enqueued):
Doing `hold Ctrl` `hold M` `spam unpress/press M`
```
Press:Ctrl
Press:Ctrl+M
Release:Ctrl+M
Press:Ctrl+M
Release:Ctrl+M
```
Doing `hold Ctrl` `hold M` `spam unpress/press Ctrl`
```
Press:Ctrl
Press:Ctrl+M
Release:Ctrl
Press:Ctrl
Release:Ctrl
```
Something seems wrong here in any case

zeromus:
The first log doesn't cause Ctrl to be released. So the original problem isn't happening. Thats strange...

The second log seems wrong.. ctrl+M should be released... but that's the only problem I see.

CasualPokePlayer:
The first log would have the "original problem" (or well #3327) since the ProcessSubsets will eventually interpret that as unpressing both Ctrl and M. The second would probably not have any issues per se. But it does seem to indicate the logic this PR touches doesn't seem to function correctly entirely anyways.

Morilli:
Okay, how about this current logic? It should behave correctly in all scenarios, even ones which weren't handled correctly (I hope) before.

For example, for `Hold Ctrl`, `Hold E`, `Release Ctrl`, `E` will still count as pressed, but `Ctrl+E` won't, whereas it did before.
Similarly, for `Hold Ctrl`, `Hold E`, `Release E`, `Ctrl` will still count as pressed while `Ctrl+E` will not.

CasualPokePlayer:
Testing this, Hold Ctrl+Z -> Release Ctrl -> Hold Ctrl results in only Z recognized as pressed.

Morilli:
> Testing this, Hold Ctrl+Z -> Release Ctrl -> Hold Ctrl results in only Z recognized as pressed.

That is intentional; after all, holding Z and pressing Ctrl does currently not consider `Ctrl+Z` as pressed either.

CasualPokePlayer:
Ah, testing it further and it seems Ctrl and Z are at least consider pressed, but not Ctrl+Z. I guess practically it doesn't matter so much.

CasualPokePlayer (approved):

YoshiRulz:
Holding Shift then holding Z still results in `Shift` keybinds being released.

Morilli:
> Holding Shift then holding Z still results in keybinds for `Shift` alone being released.

I can't replicate this in my testing, where are you seeing this?

YoshiRulz:
https://youtu.be/4j6s4_dHiQo
I expect
```
keydown Shift
#=> U
keydown Z
#=> UDL
```
but see
```
keydown Shift
#=> U
keydown Z
#=> DL
```

YoshiRulz (approved):

PR #3407, opened 2022-10-08
RetroAchievements Support

CasualPokePlayer:
Implementation with rcheevos (works for both Windows and Linux) will by default be used (with a Windows user optionally allowed to slap RA_Integration's dll in).

Some of this is a bit ugly, but most of it is contained within the RetroAchievements folder. BizInvokeUtilities was modified to support [return: MarshalAs()] and BOOL return types.

rcheevo's source can be found here https://github.com/RetroAchievements/rcheevos (MIT licenced, also submoduled with custom Makefile next to it for our purposes), and RAIntegration can be found here https://github.com/RetroAchievements/RAIntegration (also MIT licenced).


YoshiRulz:
blocked by #2235

No Linux :(

CasualPokePlayer:
Should I just have a disclaimer about data being sent.

CasualPokePlayer:
> No Linux :(

Linux support is now in!

PR #3423, opened 2022-10-21
BSNES Memory Domain Fixes

### LINKED ISSUE #3405: BSNESv115+ and original BSNES have different memory domain names, causing Lua scripts to fail
Example: https://github.com/brunovalads/yoshis-island/blob/master/BizHawk/Yoshi's-Island-BizHawk.lua

CasualPokePlayer:
Fixes #3405, along with some other things:
Symlink to our SameBoy submodule
Add more SGB domains, according to SameBoy's GB_get_direct_access.
Add SGB System Bus (note: needed to update SameBoy for GB_safe_read_memory which symlink does).
Correct SaveRAM usage with SGB (need to call SameBoy's functions for this due to potential extra data like RTC)
Set proper Main Memory and System Bus accordingly (WRAM / System Bus or the SGB variants depending on SGB flag)
Various cleanups, avoiding extraneous unsafe use.


waterbox/common.mak
```diff
@@ -44,7 +44,7 @@ ifneq ($(filter %.cpp,$(SRCS)),)
 EXTRA_LIBS := -lc++ -lc++abi -lunwind $(EXTRA_LIBS)
 endif
 
-_OBJS := $(addsuffix .o,$(realpath $(SRCS)))
```
Morilli, line 47: All changes look sane to me, but I'd just like to confirm with @nattthebear that this change does not break anything unexpectedly.

nattthebear: Makefiles, huh.  Uhhhhh....

Make clean all and make all and see if it still works.  So long as an alligator doesn't come out, I'm inclined to say it's fine.

Morilli: Okay this breaks the `libco` makefile, possibly among others. This can be resolved by changing the `realpath` to an `abspath` call instead of removing it altogether.

PR #3437, opened 2022-11-02
MAME Waterbox

CasualPokePlayer:


.gitmodules
```diff
@@ -51,3 +51,6 @@
 	path = libmupen64plus/mupen64plus-video-angrylion-rdp
 	url = https://github.com/TASEmulators/angrylion-rdp.git
 	branch = biz_mupen64plus
+[submodule "waterbox/mame-arcade/mame"]
+	path = waterbox/mame-arcade/mame
+	url = https://github.com/CasualPokePlayer/mame
```
vadosnaprimer, line 56: https://github.com/TASEmulators/mame/tree/mamehawk

src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
```diff
@@ -1,170 +1,88 @@
...
-When we want to get a C string, we Marshal.PtrToStringAnsi().
-With buffers, we Marshal.Copy() to our new buffer.
-MameGetString() only covers the former because it's the same steps
-every time, while buffers use to need aditional logic.
-
-In both cases MAME wants us to manually free the string buffer. It's
-made that way to make the buffer persist actoss C API calls.
-
-*/
-
```
vadosnaprimer, line 74: Please add some notes on what crazy unconventional stuff we have to do to make it all work. Also I think explanation of mame lua strings is nice to have.

src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
```diff
@@ -174,21 +92,21 @@ private void ExecuteMAMEThread()
 				, "-skip_gameinfo"                      // forbid this blocking screen that requires user input
 				, "-nothrottle"                         // forbid throttling to "real" speed of the device
 				, "-update_in_pause"                    // ^ including frame-advancing
-				, "-rompath",            _gameDirectory // mame doesn't load roms from full paths, only from dirs to scan
+				, "-rompath",                      ""   // mame doesn't load roms from full paths, only from dirs to scan
```
vadosnaprimer, line 95: Legend?

src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
```diff
@@ -174,21 +92,21 @@ private void ExecuteMAMEThread()
...
 				, "-nonvram_save"                       // prevent dumping non-volatile ram to disk
-				, "-artpath",          "mame\\artwork"  // path to load artowrk from
-				, "-diff_directory",      "mame\\diff"  // hdd diffs, whenever stuff is written back to an image
+				, "-artpath",           "mame/artwork"  // path to load artowrk from
+				, "-diff_directory",       "mame/diff"  // hdd diffs, whenever stuff is written back to an image
 				, "-cfg_directory",                "?"  // send invalid path to prevent cfg handling
 				, "-volume",                     "-32"  // lowest attenuation means mame osd remains silent
 				, "-output",                 "console"  // print everything to hawk console
 				, "-samplerate", _sampleRate.ToString() // match hawk samplerate
+				, "-sound",                     "none"  // forbid osd sound driver
```
vadosnaprimer, line 104: Are we sure samplerate is not butchered?
https://github.com/mamedev/mame/pull/10483 isn't pulled yet.

CasualPokePlayer: I removed the problematic code in the last commit I did.

src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
CasualPokePlayer: The absolute path within wbx's virtual filesystem for the ROM is whatever I set in the AddReadonlyFile call. Which would be the rom filename itself. So rompath needs to be "".

vadosnaprimer: Does it launch without `-rompath`?

CasualPokePlayer: nope

vadosnaprimer (approved):

Morilli:
fwiw the build still fails for me when not having `sdl2` installed due to this:
```
Compiling mame/src/lib/formats/all.cpp...
Archiving libformats.a...
Compiling mame/src/osd/modules/input/input_common.cpp...
In file included from ../../../../generated/includes/SDL2/SDL_config.h:33,
                 from ../../../../generated/includes/SDL2/SDL_stdinc.h:31,
                 from ../../../../generated/includes/SDL2/SDL_main.h:25,
                 from ../../../../generated/includes/SDL2/SDL.h:32,
                 from ../../../../../../mame/src/osd/modules/input/input_common.cpp:34:
../../../../../../mame/3rdparty/SDL2-override/waterbox/SDL_config_premake.h:33:10: fatal error: SDL_platform.h: No such file or directory
   33 | #include "SDL_platform.h"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [osd_sdl.make:975: ../../../../linux_gcc/obj/x64/Release/osd_sdl/mame/src/osd/modules/input/input_common.o] Error 1
```
Should probably be fixable by editing the include line in the failing header.

CasualPokePlayer:
> Should probably be fixable by editing the include line in the failing header.

Hopefully should be fixed with this? https://github.com/TASEmulators/BizHawk/pull/3437/commits/a196fee63f105bb606019866958bf40efed71f6e

src/BizHawk.Client.EmuHawk/tools/MultiDiskBundler/MultiDiskBundler.cs
```diff
@@ -31,6 +31,23 @@ public MultiDiskBundler()
...
+			SystemDropDown.Items.AddRange(new[]
+			{
+				VSystemID.Raw.AmstradCPC,
+				VSystemID.Raw.AppleII,
+				VSystemID.Raw.C64,
+				VSystemID.Raw.GBL,
+				VSystemID.Raw.GEN,
+				VSystemID.Raw.GGL,
+				VSystemID.Raw.Jaguar,
+				VSystemID.Raw.MAME,
```
vadosnaprimer, line 43: Arcade.

src/BizHawk.Emulation.Cores/Arcades/MAME/LibMAME.cs
```diff
@@ -28,64 +24,85 @@ public enum SaveError
...
-		[DllImport(dll, CallingConvention = cc)]
-		public static extern int mame_lua_get_int(string code);
+		[BizImport(cc)]
+		public abstract int mame_lua_get_int(string code);
 
-		// get double
-		[DllImport(dll, CallingConvention = cc)]
-		public static extern double mame_lua_get_double(string code);
+		// get long
+		// nb: this is actually a double cast to long internally
```
vadosnaprimer, line 80: Comment expired?

InfamousKnight:
Could someone send a precompiled link so I can try this out please?

ThunderAxe31:
So, does this replace the previous MAME core? Do I simply use Open Advanced to try this?

CasualPokePlayer:
It does replace it. Open Advanced shouldn't be needed anymore in most cases (it's still there to force something to be shoved into MAME).

ThunderAxe31:
Can you please add the various MAME firmware files for BizHawk firmware file detection? The only workaround I found so far is to put the necessary firmware inside the game .zip, but that's unintuitive and hacky.

CasualPokePlayer:
You should normally be placing them as additional files in the multidisk bundler. We can't really go ahead with detecting firmware (as they need to be in an unknown amount of zip files, and we really just get opaque blobs as far as the c# side of the core is concerned).

CasualPokePlayer:
> Edit: also, I've just done a quick test, and it seems that the core discards all persistent in-game settings data upon starting recording from power-on, or upon playing back an existing movie. However, if in-game settings are changed by a movie, they will remain even after rebooting the core. The game I've tried is Puzzle Bobble, Japan version (pbobble.zip).

This sounds like normal SaveRAM behavior? You start a movie, it will start with clean SaveRAM assuming you didn't ask to start with preexisting SaveRAM. If the existing movie didn't have SaveRAM in it, it won't load any up and start with clean SaveRAM. Rebooting a core saves current SaveRAM and that will also load up the core without a movie loaded, so that SaveRAM will be used.

ThunderAxe31:
Uhm, ok then.

Moving on, another thing. I'm using Metal Slug X (mslugx.zip) and I get this error every time I reboot the core after changing the BIOS in the Sync Settings.

image

Weirdly enough, if I reboot the core again it doesn't give any error.

CasualPokePlayer:
BIOS is put in internal info since that's what MAME wants, and it throws if it tries to load a mismatch. Maybe it shouldn't do this? Might be a bad idea if BIOS changes end up changing more than just the name here per se...

ThunderAxe31:
Can you please set to 0% the default deadzone setting for analog inputs? I've tried Time Crisis (timecris.zip) and the initial value of 10% was bad.

CasualPokePlayer:
Deadzone being bad for lightguns is known, but it is still useful for actual analog sticks and mice.

Can you please make separate issues for these anyways instead of clogging this PR thread.

PR #3442, opened 2022-11-06
Use NS cursor for TAStudio axis double-click-and-drag

YoshiRulz:
~~Set is in the right place, unset is in the wrong place (may need to be in multiple places). Anyone is welcome to finish this.~~

### DIFF
--- a/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/InputRoll.cs
+++ b/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/InputRoll.cs
@@ -1076,7 +1076,7 @@ namespace BizHawk.Client.EmuHawk
 
 			Cursor = IsHoveringOnColumnEdge || _columnResizing != null
 				? Cursors.VSplit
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
What is NS cursor?

YoshiRulz:
North&ndash;South (C [`IDC_SIZENS`](https://learn.microsoft.com/en-us/windows/win32/menurc/about-cursors), C# [`Cursors.SizeNS`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.cursors.sizens?view=netframework-4.8))

vadosnaprimer:
InputRoll's OnMouseMove overrides it to default.

Otherwise I'd just add cursor logic to AxisEditRow setter.

SuuperW:
The solution is to set `Cursor` to `null`, which allows parent controls to decide the cursor, instead of setting it to `Default`.

YoshiRulz (commented):
Doesn't seem to work under Mono :( but that's not what this was about

PR #3451, opened 2022-11-16
Use a small dll for handling the msabi<->sysv adapter

### LINKED ISSUE #2735: Waterbox breaks in debug mode with too many args
### Repro
1. Rebuild waterboxhost (the rust code) in debug mode.
  a. We always ship the release mode binary.
2. Create a function to call in a waterbox core that takes 5 or 6 args.
  a. At present, all actual waterbox cores have 4 or fewer args on all of these functions.
3. Call the function

### Output
CRASH.  `rax` was stomped over by the rust code in `depart5` / `depart6` before it could call it.

### Host env.
Windows only.

Want to fix this, can't think of an easy way right now so I'm dumping it here.

CasualPokePlayer:
This dll only uses assembly (taken from generated optimized rustc output) and handcrafted unwind information (c# exceptions in a callback seem to work fine in testing).

Additionally, allow floating point arguments. This really only needs to occur on the c# side. Msabi and sysv agree on the first 4 floating point args and for returns, so no work actually has to be done adapting these. With assembly being used, we can guarantee rax will not be stomped by compiler whims (and avoid potential floating point args from being trashed).

Resolves #2735


nattthebear (approved):
This looks good, and this approach is probably the cleanest possible overall approach.

Issue #3435, opened 2022-10-31
Stop using file extensions to determine `GameInfo.System`


YoshiRulz:
see also #969

CasualPokePlayer:
Note that these issues only apply if a match isn't found in the gamedb anyways (so bad dumps, hacks, homebrew, or systems where the gamedb is not available).

YoshiRulz:
gamedb should be optional #2939

Issue #2939, opened 2021-09-24
Refactor gamedb


YoshiRulz:
(I'm referring here to the way it's read from disk, not necessarily to the file formats.)

- 7z is annoying for CI.
- EmuHawk should not require any part of the gamedb to be usable.

PR #3457, opened 2022-11-21
Implement bsx + bs memory loading for the BSNESv115+ core

Morilli:
The core already supports everything, so all that's required is wiring it up with the frontend.
I've structured this similarly to SGB loading; where the BS-X rom is a required firmware for all bs memory cartridges.

I'm not aware of a way to detect that the BS-X cartridge should be used, so I've created a new BSX system with file extension `.bs` that allows loading roms as bs memory cartridge. Open to feedback on whether there is a better way to do this.


YoshiRulz:
I think this is more like FDS than SGB. (Neither BSNES core handles sysIDs correctly, which may have given you the wrong idea.) Give `.bs` files the `SNES` sysID, then check the rom header ([docs](https://wiki.superfamicom.org/bs-x-satellaview-header), [alt.](https://satellaview.fandom.com/wiki/Satellaview_ROM_header)) to differentiate Satellaview roms from SNES roms, assuming there isn't already a function for that in BSNES. I'm guessing there isn't one because it loads the firmware "game" when given 1 MiB of zeroes.

If you were to stick with the separate sysID, `gamedb_snes.txt` needs to be updated or many games will get a sysID override and fail to load.

Morilli:
> Neither BSNES core handles sysIDs correctly, which may have given you the wrong idea

What's wrong about it and how should it be done instead? Are you referring to this line? https://github.com/TASEmulators/BizHawk/blob/5a4dc9fd88e3aa7d20afa79758fe69529ed813bc/src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.IEmulator.cs#L97

In any case I guess the path you're recommending is just adding `.bs` to the SNES system's file extensions and differentiating between normal SNES rom and BS-X rom based on the rom header. That's also what I would prefer to do; I'll have to check the docs to see how to make that differentiation.

Morilli:
> differentiating between normal SNES rom and BS-X rom based on the rom header

Actually, this may not be possible or desired; judging from https://github.com/bsnes-emu/bsnes/issues/262 (and testing it myself) it is totally possible to load BS memory cartridges as normal rom files without the BS-X rom, even if it may not be a senseful thing to do. I don't actually know enough about the hardware to saw whether that was also possible on a real console so I'm not sure it makes sense to allow, but it may be worth considering.

YoshiRulz:
> What's wrong about it and how should it be done instead? Are you referring to this line? 
> https://github.com/TASEmulators/BizHawk/blob/5a4dc9fd88e3aa7d20afa79758fe69529ed813bc/src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.IEmulator.cs#L97

Yes. There's a separate flag for no reason:
https://github.com/TASEmulators/BizHawk/blob/5a4dc9fd88e3aa7d20afa79758fe69529ed813bc/src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs#L30
Which, in your defence, is a holdout from old BSNES:
https://github.com/TASEmulators/BizHawk/blob/5a4dc9fd88e3aa7d20afa79758fe69529ed813bc/src/BizHawk.Emulation.Cores/Consoles/Nintendo/SNES/LibsnesCore.cs#L113-L116
Resulting in this spaghetti:
https://github.com/TASEmulators/BizHawk/blob/5a4dc9fd88e3aa7d20afa79758fe69529ed813bc/src/BizHawk.Client.EmuHawk/MainForm.cs#L2051-L2066

> [...] it is totally possible to load BS memory cartridges as normal rom files without the BS-X rom, even if it may not be a senseful thing to do. I don't actually know enough about the hardware to saw whether that was also possible on a real console [...]

I've also never seen one of these things IRL.

Morilli:
> If you were to stick with the separate sysID, gamedb_snes.txt needs to be updated or many games will get a sysID override and fail to load.

Right, there should probably be a separate one for BS memory cartridges, unless they can be recognized as BS cartridges before getting loaded, which doesn't seem possible without using hacks, at least judging from my understanding of the header format and the fact bsnes does also not seem to identify anything special about them.

 I'm also currently struggling with the fact that BS memory cartridges can work with multiple normal sfc cartridges, not just the ``BS-X - Sore wa Namae o Nusumareta Machi no Monogatari`` one... there seem to be a couple, and I'm unsure if marking them all as firmware is a proper solution, see here for a list (no idea how exhaustive it is): https://satellaview.fandom.com/wiki/Category:Slotted_cartridges

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -2048,20 +2048,9 @@ private void HandlePlatformMenus()
...
-					SNESSubMenu.Visible = true;
-					break;
-				case VSystemID.Raw.SNES when Emulator is BsnesCore bsnesCore:
-					SNESSubMenu.Text = bsnesCore.IsSGB ? "&SGB" : "&SNES";
-					SNESSubMenu.Visible = true;
-					break;
-				case VSystemID.Raw.SNES when Emulator is SubBsnesCore subBsnesCore:
-					SNESSubMenu.Text = subBsnesCore.IsSGB ? "&SGB" : "&SNES";
+				case VSystemID.Raw.SGB when Emulator is BsnesCore or SubBsnesCore:
+				case VSystemID.Raw.SNES when Emulator is LibsnesCore or BsnesCore or SubBsnesCore:
```
YoshiRulz, line 2052: Don't need type checks anymore, and you can keep the label assignment as `SNESSubMenu.Text = Emulator.SystemId is VSystemID.Raw.SGB ? "&SGB" : "&SNES";`

Morilli: > you can keep the label assignment as SNESSubMenu.Text = Emulator.SystemId is VSystemID.Raw.SGB ? "&SGB" : "&SNES";

No other core is doing that and when thinking about it I also found it kind of counterintuitive that the label text changes even though the menu is exactly the same. Not really opposed to changing the text dynamically, but it doesn't seem useful to me right now.

CasualPokePlayer:
Based on the wikipedia article's images, it looks like the BS-X cartridge, which acts like a "BIOS", is the actual SNES cartridge. BS games are stored onto smaller "memory packs" which themselves are not SNES cartridges and would not fit in the SNES cartridge slot due to their size: they can only fit onto the BS-X cartridge's own slot. You can shove the memory packs as normal roms on an emulator as they do contain SNES code and would work in some capacity, but that wouldn't be possible with the actual carts on real hardware.

Morilli:
> Putting BS-X "BIOS" as a firmware file is actually probably correct here given this, there's no need to do the MDB bs here.

As I've linked earlier, the BS-X cartridge is just one of many that have a bs memory slot, so logically I would have to add all of them to the list of firmware files. And even then, it's not clear which of those cartridges the BS game wants; that would have to be somehow configured by the user still.

That being said, there is currently the issue of `.bs` files being loadable by any SNES core alone, which probably should be avoided if possible as as you've correctly stated BS cartridges are not real cartridges themselves. However they are in the snes gamedb and seem to commonly use the `.sfc` extension anyways, so...

CasualPokePlayer:
If they're in the db, surely you could just add in some special key in the db which the core can refer to to understand if it's a bs memory pack?

CasualPokePlayer:
Also, perhaps just a sync setting to know which slotted cart to use for BS-X could suffice? Could just a dropdown menu in the settings dialog. There's only 5 options anyways, so might be better?

YoshiRulz:
Looking at https://satellaview.fandom.com/wiki/Category:Slotted_cartridges, there's at least 8 others. But I think they're intended to be [used with specific roms](https://satellaview.fandom.com/wiki/Game_data), not as generic replacement "firmwares", so maybe a gamedb flag to pick the base cart from firmware is the way to go.

Morilli:
> so maybe a gamedb flag to pick the base cart from firmware is the way to go.

That could work (if I knew how to set and retrieve such a flag), but I think the gamedb is missing entries for some of the downloadable content. Generally I think it's easier (and less invasive) to require loading both the base cartridge and its corresponding bsmemory pack instead of modifying the gamedb and requiring people to use perfect dumps or otherwise their game won't work,
You also mentioned in #2939 that the gamedb should be optional, which this idea goes against.

Maybe BizHawk needs a "Load ROM**S**" button...

YoshiRulz:
Ah, I was trying to say that I agreed with CPP's suggestion
> Also, perhaps just a sync setting to know which slotted cart to use for BS-X could suffice?

And that this could be done automagically with a gamedb flag. Which would be `((GameInfo) loadParameters.Game)["flag"]`.

CasualPokePlayer:
Yeah, we could have it so it's a dropdown of items for each "firmware", with one option actually just be "Autodetect" which relies on gamedb.

Morilli:
> which falls back to the base BS-X cart if gamedb isn't present or no entry can be found

That still requires a way to clearly identify bs memory pack data (data which may not necessarily be a loadable game at all), and I still don't know how that would be done reliably.

Morilli:
> Need to check if that game works properly in Satellaview mode once it's finished.

The entirety of RTC in SNES does not currently work correctly in bsnes. The code tries to use host `time()` functions to access real time, which isn't supported in waterbox (and wouldn't be deterministic anyways). Lots of changes need to be done to support RTC correctly, both in satellaview and other games that have an RTC chip, like Tengai Makyō Zero.

CasualPokePlayer:
> I'm not aware of a way to detect that the BS-X cartridge should be used, so I've created a new BSX system with file extension `.bs` that allows loading roms as bs memory cartridge. Open to feedback on whether there is a better way to do this.

Not sure if this is why you choose .bs, but it seems be the standard way no-intro is labeling bs memory cartridges.

Morilli:
I mainly relied on what bsnes uses internally, which is `.bs` for bsmemory cartridge files: https://github.com/TASEmulators/BizHawk/blob/7b4bd52150943b91524fb64aa378d10f639e1a26/waterbox/bsnescore/bsnes/sfc/cartridge/load.cpp#L211
(also, the corresponding thing for satellaview would apparently be `.st`)

CasualPokePlayer:
> > Need to check if that game works properly in Satellaview mode once it's finished.
> 
> The entirety of RTC in SNES does not currently work correctly in bsnes. The code tries to use host `time()` functions to access real time, which isn't supported in waterbox (and wouldn't be deterministic anyways). Lots of changes need to be done to support RTC correctly, both in satellaview and other games that have an RTC chip, like Tengai Makyō Zero.

It's actually relatively trivial to support time, you just need the C# side to provide time (as its the only one able to query host time, and can provide a deterministic time easily itself). [The generic WaterboxCore provides functionality to easily support this](https://github.com/TASEmulators/BizHawk/blob/7b4bd52150943b91524fb64aa378d10f639e1a26/src/BizHawk.Emulation.Cores/Waterbox/WaterboxCore.cs#L91-L115), which can be easily copy pasted over to BSNES. Probably easy to just hook up a callback to query the C# side whenever the core wants time.

YoshiRulz:
So are we all agreed on this?
- The rom you need to load is the memory pack, not the base cart, nor a multi-disk bundle.
	- Because of this, Satellaview should be a separate sysID.
- The base cart rom is pulled from firmware. As there are only a handful, they're separate records in the FWDB.
	- This lets you change the base cart without needing to manually set FW customisations, only a sync setting.
	- Either an "automatic" option for that setting, or a separate bool, enables a gamedb flag to inform which base cart is chosen.
	- Movie headers still have the base cart rom hash in firmware.
	- Homebrew base cart roms, if anyone's crazy enough to make one, would be possible by simply setting a FW customisation to the custom rom.
- File extension nonsense should keep with the status quo for this PR: use the most common extension(s) found in the wild.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs
```diff
@@ -85,11 +86,18 @@ public BsnesCore(CoreLoadParameters<SnesSettings, SnesSyncSettings> loadParamete
...
+			if (SystemId == VSystemID.Raw.Satellaview)
 			{
-				Api.core.snes_load_cartridge_bsmemory(loadParameters.Roms[0].RomData, loadParameters.Roms[1].RomData,
-					loadParameters.Roms[0].RomData.Length, loadParameters.Roms[1].RomData.Length);
+				SATELLAVIEW_CARTRIDGE slottedCartridge = _syncSettings.SatellaviewCartridge;
+				if (slottedCartridge == SATELLAVIEW_CARTRIDGE.Autodetect)
+				{
+					slottedCartridge = (SATELLAVIEW_CARTRIDGE)loadParameters.Game.GetInt("baseCartridge", -1);
+					// TODO: this exception is kinda ugly... we could just (silently?) fallback to the BS-X cartridge as 99% of games use that one
+					if ((int)slottedCartridge == -1) throw new Exception("Unable to autodetect satellaview base cartridge for unknown game!");
```
CasualPokePlayer, line 96: Using the BS-X cartridge as a fallback would be best here (that's what I was suggesting), the user can set sync settings to manually override if somehow this cartridge isn't right.

Morilli: I guess it's probably fine to do that. I just generally dislike silently failing and potentially doing unexpected things.

CasualPokePlayer: You can give a warning to the user perhaps (CoreComm ShowMessage for a popup or Notify for an OSD message)

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.ISettable.cs
```diff
@@ -4,17 +4,32 @@ namespace BizHawk.Emulation.Cores.Nintendo.BSNES
 {
 	public partial class BsnesCore : ISettable<BsnesCore.SnesSettings, BsnesCore.SnesSyncSettings>
 	{
+		// names need to exactly match the firmware db names
```
YoshiRulz, line 7: Oof. If you can't fix this, make sure to add a note in the FWDB too.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.ISettable.cs
```diff
@@ -4,17 +4,32 @@ namespace BizHawk.Emulation.Cores.Nintendo.BSNES
...
+			Rom_DS96,
+			Rom_Ongaku_Tsukuru,
+			Rom_SoundNovel_Tsukuru,
+			Rom_Tsuri
+		}
+
 		public SnesSettings GetSettings()
 		{
-			return _settings.Clone();
+			return _settings with {};
```
YoshiRulz, line 24: IIRC this is a compile-time thing, but I'll leave this reminder for myself to check it works under Mono just in case.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.ISettable.cs
```diff
@@ -29,25 +44,15 @@ void SNES.IBSNESForGfxDebugger.PutSettings(SNES.IBSNESForGfxDebugger.SettingsObj
...
-			bool ret = o.LeftPort != _syncSettings.LeftPort
-				|| o.RightPort != _syncSettings.RightPort
-				|| o.LimitAnalogChangeSensitivity != _syncSettings.LimitAnalogChangeSensitivity
-				|| o.Entropy != _syncSettings.Entropy
-				|| o.RegionOverride != _syncSettings.RegionOverride
-				|| o.Hotfixes != _syncSettings.Hotfixes
-				|| o.FastPPU != _syncSettings.FastPPU
-				|| o.FastDSP != _syncSettings.FastDSP
-				|| o.FastCoprocessors != _syncSettings.FastCoprocessors
-				|| o.UseSGB2 != _syncSettings.UseSGB2;
```
YoshiRulz, line 41: Oh yeah, and like the settings dialog refactor, can you do this separately? Actually, maybe we should do this for all cores?

waterbox/bsnescore/bsnes/target-bsnescore/bsnescore.hpp
```diff
@@ -10,9 +10,9 @@ enum SNES_MEMORY {
     CARTRIDGE_RAM,
     CARTRIDGE_ROM,
 
-    // bsx and sufamiturbo unused cause unsupported by frontend
-    BSX_RAM,
-    BSX_PRAM,
+    BSMEMORY_ROM,
+    BSMEMORY_PSRAM,
+    // sufamiturbo unused cause unsupported by frontend
```
YoshiRulz, line 15: I just looked up the SuFami Turbo and oh god not again

YoshiRulz (changes requested):

src/BizHawk.Client.EmuHawk/config/FirmwaresConfig.cs
```diff
@@ -64,6 +64,7 @@ public partial class FirmwaresConfig : Form, IDialogParent
 		{
 			["NES"] = "NES",
 			["SNES"] = "SNES",
+			["Satellaview"] = "SNES+Satellaview",
```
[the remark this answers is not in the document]
Morilli: Have renamed the system id. About the displayed name: I was following the code comment above that read `// friendlier names than the system Ids`, and I wanted to somehow make it clear that the required roms are actually SNES roms, just to be used in conjunction with the satellaview.

src/BizHawk.Emulation.Common/filetype_detectors/SatellaviewFileTypeDetector.cs
```diff
@@ -0,0 +1,162 @@
...
+			else
+			{
+//				Util.DebugWriteLine($"didn't match {(checkHiROM ? "HiROM" : "LoROM")}, -{corruption} pts.: {header.ToString()}");
+				warnings.Clear();
+			}
+			return detected;
+		}
+
+		/// <remarks>not to be confused with a "slotted cart" i.e. base game, which we treat as either firmware or a normal SNES rom</remarks>
+		public static bool IsSatellaviewRom(ReadOnlySpan<byte> rom, out IReadOnlyList<string> warnings)
```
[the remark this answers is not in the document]
Morilli: If I'm reading it right the logic here aligns with the proposal in #3435, and given the fact that bsmemory rom data is commonly given the `.sfc` extension anyways I feel like the ways it's currently done is good here.

src/BizHawk.Client.EmuHawk/config/FirmwaresConfig.cs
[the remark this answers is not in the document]
YoshiRulz: Sounds good, though I'd be wary of the comments in this class :P

src/BizHawk.Emulation.Common/VSystemID.cs
```diff
@@ -55,6 +55,7 @@ public static class Raw
 			public const string PSP = "PSP";
 			public const string PSX = "PSX";
 			public const string SAT = "SAT";
+			public const string Satellaview = "Satellaview";
```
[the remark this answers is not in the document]
YoshiRulz: Staff were 3:0 in favour of "BSX".

YoshiRulz (approved):
I'm done with caring about Designer file diffs. No further questions from me.

CasualPokePlayer (approved):
LGTM

PR #3471, opened 2022-12-03
Add `--userdata` CLI flag

YoshiRulz:
As an example, `./EmuHawkMono.sh '--userdata=a:1;b:.5;c:1b;d:00;e:true;f:TRUE'` (quoted because ';' is a special char in BASH) sets these userdata entries:
key | type | value
--:|:--|:--
`a` | int | `1`
`b` | double | `0.5`
`c` | string | `"1b"`
`d` | int | `0`
`e` | bool | `true`
`f` | string | `"TRUE"`


YoshiRulz:
Will merge in a few days if no-one objects.

PR #3472, opened 2022-12-04
`Span`-based bulk peeks (the boring kind)

YoshiRulz:
This isn't the switching to `ReadOnlySpan` to let the frontend and APIs read memory without copying. This still copies, but into a `Span` instead of an array, which probably makes no difference for internal use, and is slightly more convenient for ext. tools.
I could have exposed the overload which takes an array reference and writes into it, but I figured this way would mitigate user error. The use of an unsigned integer for the start offset, especially.

I'd appreciate if someone could double-check my pointer arithmetic is correct, i.e. it matches the arithmetic in the other overload's implementation.
edit: Rebased, would still like that second opinion, and I think this should wait for #3965.

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
@@ -260,6 +260,26 @@ namespace BizHawk.Client.Common
 			return newBytes;
 		}
 
+		public bool ReadByteRange(ulong srcStartOffset, Span<byte> dstBuffer, string domain = null)
+		{
+			var d = NamedDomainOrCurrent(domain);
+			if (srcStartOffset >= (ulong) d.Size)
+			{
+				LogCallback($"Warning: attempted read of {srcStartOffset} outside the memory size of {d.Size}");
+				return false;
+			}
+			try
+			{
+				d.BulkPeekByte(srcStartOffset, dstBuffer);
+				return true;
+			}
+			catch (Exception e)
+			{
+				LogCallback($"bulk-peek threw {e.GetType().Name}: {e.Message}");
+				return false;
+			}
+		}
+
 		public void WriteByteRange(long addr, IReadOnlyList<byte> memoryblock, string domain = null)
 		{
 			var d = NamedDomainOrCurrent(domain);
--- a/src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
@@ -252,7 +252,7 @@ namespace BizHawk.Client.Common
 			var iSrc = Math.Min(Math.Max(0L, addr), d.Size);
 			var iDst = iSrc - addr;
 			var bytes = new byte[indexAfterLast - iSrc];
-			if (iSrc < indexAfterLast) using (d.EnterExit()) d.BulkPeekByte(iSrc.RangeToExclusive(indexAfterLast), bytes);
+			if (iSrc < indexAfterLast) using (d.EnterExit()) d.BulkPeekByte(unchecked((ulong) iSrc), bytes);
 			if (lastReqAddr >= d.Size) LogCallback($"Warning: Attempted reads on addresses {d.Size}..{lastReqAddr} outside range of domain {d.Name} in {nameof(ReadByteRange)}()");
 			if (bytes.Length == length) return bytes;
 			var newBytes = new byte[length];
--- a/src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
+++ b/src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
@@ -17,6 +17,15 @@ namespace BizHawk.Client.Common
 
 		uint ReadByte(long addr, string domain = null);
 		IReadOnlyList<byte> ReadByteRange(long addr, int length, string domain = null);
+
+		/// <summary>copies a region of memory into <paramref name="dstBuffer"/>, starting at <paramref name="srcStartOffset"/> and taking as many bytes as will fit in the buffer</summary>
+		/// <returns>
+		/// <see langword="false"/> iff the internal API call threw an exception (for example, if you request bytes beyond the end of the domain).<br/>
+		/// If this happens, the state of <paramref name="dstBuffer"/> is undefined. Anywhere from <c>0</c> bytes to all of them could have been copied.
+		/// The exception is when <paramref name="srcStartOffset"/> falls outside the domain, as that is checked first and results in a no-op.
+		/// </returns>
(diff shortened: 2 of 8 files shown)


src/BizHawk.Emulation.Cores/Waterbox/WaterboxMemoryDomain.cs
```diff
@@ -99,6 +99,26 @@ public override void PokeByte(long addr, byte val)
...
+			if ((ulong) Size < srcStartOffset + (ulong) dstBuffer.Length) throw new ArgumentException(paramName: nameof(dstBuffer), message: ERR_MSG_TOO_MANY_BYTES_REQ);
+			try
+			{
+				_monitor.Enter();
+				new Span<byte>((void*) Z.US((ulong) _data + srcStartOffset), dstBuffer.Length).CopyTo(dstBuffer);
+			}
+			finally
+			{
+				_monitor.Exit();
+			}
```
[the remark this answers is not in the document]
YoshiRulz: It was copy-pasted from an older copy of another overload, will fix

src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
```diff
@@ -252,14 +252,34 @@ public IReadOnlyList<byte> ReadByteRange(long addr, int length, string domain =
...
 
+		public bool ReadByteRange(ulong srcStartOffset, Span<byte> dstBuffer, string domain = null)
+		{
+			var d = NamedDomainOrCurrent(domain);
+			if (srcStartOffset >= (ulong) d.Size)
+			{
+				LogCallback($"Warning: attempted read of {srcStartOffset} outside the memory size of {d.Size}");
+				return false;
+			}
+			try
```
SuuperW, line 271: Why does the new method have a `try` block?
Lua already puts everything inside a `try` block, and I think that is better: an exception should stop the Lua script. For C# users, I think it better to let them decide what to do.

(IMO attempting to read/write outside the range of the memory domain should also throw but that's a separate change)

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
```diff
@@ -17,6 +17,15 @@ public interface IMemoryApi : IExternalApi
...
 		uint ReadByte(long addr, string domain = null);
 		IReadOnlyList<byte> ReadByteRange(long addr, int length, string domain = null);
+
+		/// <summary>copies a region of memory into <paramref name="dstBuffer"/>, starting at <paramref name="srcStartOffset"/> and taking as many bytes as will fit in the buffer</summary>
+		/// <returns>
+		/// <see langword="false"/> iff the internal API call threw an exception (for example, if you request bytes beyond the end of the domain).<br/>
+		/// If this happens, the state of <paramref name="dstBuffer"/> is undefined. Anywhere from <c>0</c> bytes to all of them could have been copied.
+		/// The exception is when <paramref name="srcStartOffset"/> falls outside the domain, as that is checked first and results in a no-op.
+		/// </returns>
+		bool ReadByteRange(ulong srcStartOffset, Span<byte> dstBuffer, string domain = null);
```
SuuperW, line 27: The doc comment here is wrong. It returns false if the internal API call threw an exception, but not iff: the method itself contains a check for when the start falls outside the domain.
This is also different behavior from the other overload, which will still return some bytes just starting at the beginning of the memory domain. I think they should have the same behavior. (which again IMO should be throwing)

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
```diff
@@ -119,13 +141,12 @@ public virtual void BulkPeekUshort(Range<long> addresses, bool bigEndian, ushort
 			var start = addresses.Start;
 			var end = addresses.EndInclusive + 1;
 
-			if ((start & 1) != 0 || (end & 1) != 0)
-				throw new InvalidOperationException("The API contract doesn't define what to do for unaligned reads and writes!");
+			if (start % 2 is not 0 || end % 2 is not 0) throw new InvalidOperationException(ERR_MSG_NOT_ALIGNED);
```
SuuperW, line 144: Why are you changing this? `(start & 1) != 0` is easier to understand, and far nicer than `start % 2 is not 0`.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
```diff
@@ -119,13 +141,12 @@ public virtual void BulkPeekUshort(Range<long> addresses, bool bigEndian, ushort
 			var start = addresses.Start;
 			var end = addresses.EndInclusive + 1;
 
-			if ((start & 1) != 0 || (end & 1) != 0)
-				throw new InvalidOperationException("The API contract doesn't define what to do for unaligned reads and writes!");
+			if (start % 2 is not 0 || end % 2 is not 0) throw new InvalidOperationException(ERR_MSG_NOT_ALIGNED);
 
-			if (values.LongLength * 2 != end - start)
+			if (checked((ulong) values.Length * 2UL) != addresses.Count())
```
SuuperW, line 146: The `checked` here does nothing. Unless `Length` can somehow be negative, this expression can never overflow.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
```diff
@@ -92,14 +98,30 @@ public virtual void PokeUint(long addr, uint val, bool bigEndian)
 			}
 		}
 
+		public virtual byte[] BulkPeekByte(Range<long> addresses)
```
SuuperW, line 101: Would it be a good idea to make `void BulkPeekByte(Range<long> addresses, byte[] values)` protected, and change any call sites to use this (or the span version)?

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
vadosnaprimer: Yeah `&` is a standard way to check individual bits.

YoshiRulz: This isn't querying a bitfield though, it's checking address alignment. Later in the file is a `% 4`. Let the compiler optimise it to a mask.

SuuperW: Address alignment is inherently tied to bits, which makes the bitwise operator more appropriate.
I also much prefer the standard equality operator `==` over `is` and especially prefer `!=` over `is not`. (`is` would make sense with pattern matching, maybe even checking for `null`, but not for plain equality checks. And having whitespace inside the `is not` "operator" is just terrible imo.)
But even disregarding all that, this is a style change that isn't relevant to this PR.

PR #3476, opened 2022-12-06
LuaConsole fixes

kalimag:
This PR fixes a number of issues in `LuaConsole`, mostly related to script loading/unloading

# Changes

1. Change references to `Config.DisableLuaScriptsOnLoad` to `Settings.DisableLuaScriptsOnLoad`
This was inconsistently done in an old commit, leading to some functions obeying this setting and others ignoring it.
This will effectively reset the setting

2. `Close()` existing `Win32LuaLibraries` during `Restart()`.
Before, only closing the `LuaConsole` would call close on the current `LuaImp`, causing possible memory and file handle leaks, and leaving behind 'zombie' forms still referencing the old libraries after loading a rom/restarting the core

3. Dispose `FileSystemWatcher` when no longer needed, avoid relying on path string comparisons, prevent crash in case of missing directory, and prevent some cross-thread risks

4. Remove buggy `RunLuaScripts` and replace it with calls to ` EnableLuaFile`/`ToggleLuaScript` in the two places it was used
This resolves scripts not starting correctly when loading a session, and all scripts being stopped when loading a script that already exists in the list.

5. Remove existing scripts before loading a session or creating a new session, to run cleanup code
Previously, registered functions would hang around

6. Refactor removing and disabling scripts into methods to reduce some code duplication

7. Update registered functions dialog after instantiating a new `LuaImp`

8. Fix luases files always being saved in the exe directory regardless of selected save location
I don't really understand what the intent of the existing code was, seems like a refactor gone wrong? 

9. Stop converting script paths between relative paths and absolute paths all over the place, stick with absolute paths
Again, I'm not sure why it worked this way, but the existing code has a number of issues:
   1. `PathEntries.TryMakeRelative` doesn't handle paths containing dot notation or inconsistent forward/backward slashes on Windows
   2. Paths were made relative to the configured base path, but converted back to absolute to the exe directory, which can be different
   3. Relative paths in a .luases were interpreted as relative to the .luases, which can be different again
   4. FileSystemWatcher code didn't recognize differently formatted paths (resolved by 3.)

# Open questions

1. Should removing a script call `LuaFile.Stop()` and call the exit event like disabling a script does? I would think so, but it didn't before. Also goes for scripts cleared when loading a session/creating a new session.

2. Should "Stop all scripts" behave the same as disabling scripts? Currently it only changes the `State` flag.

3. Should the relative path conversion when loading lua scripts be fixed instead of removed? If so, what is the intended behavior?
No other file loading code behaves like this from what I've seen (roms, lua session files, cheats, watches, etc.) 

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
@@ -373,7 +373,7 @@ namespace BizHawk.Client.EmuHawk
 		{
 			if (!item.IsSeparator)
 			{
-				LuaImp.RegisteredFunctions.RemoveForFile(item, Emulator);
+				DisableLuaScript(item);
 				RemoveFileWatcher(item);
 			}
 			LuaImp.ScriptList.Remove(item);
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
@@ -341,36 +341,6 @@ namespace BizHawk.Client.EmuHawk
 			UpdateRegisteredFunctionsDialog();
 		}
 
-		private void RunLuaScripts()
-		{
-			foreach (var file in LuaImp.ScriptList.Where(s => !s.IsSeparator))
-			{
-				if (!file.Enabled && file.Thread is null)
-				{
-					try
-					{
-						LuaSandbox.Sandbox(null, () =>
-						{
-							string pathToLoad = ProcessPath(file.Path);
-							LuaImp.SpawnAndSetFileThread(file.Path, file);
-							LuaSandbox.CreateSandbox(file.Thread, Path.GetDirectoryName(pathToLoad));
-						}, () =>
-						{
-							file.State = LuaFile.RunState.Disabled;
-						});
-					}
-					catch (Exception e)
-					{
-						DialogController.ShowMessageBox(e.ToString());
-					}
-				}
-				else
-				{
-					file.Stop();
-				}
-			}
-		}
-
 		private void SessionChangedCallback()
 		{
 			OutputMessages.Text =
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
@@ -203,9 +203,7 @@ namespace BizHawk.Client.EmuHawk
 				// we don't use runningScripts here as the other scripts need to be stopped too
 				foreach (var file in luaLibsImpl.ScriptList)
 				{
-					if (file.Enabled) luaLibsImpl.CallExitEvent(file);
-					luaLibsImpl.RegisteredFunctions.RemoveForFile(file, Emulator);
-					file.Stop();
+					DisableLuaScript(file);
 				}
 			}
 
@@ -1509,11 +1507,24 @@ namespace BizHawk.Client.EmuHawk
 			}
 			else if (!file.Enabled && file.Thread is not null)
 			{
+				DisableLuaScript(file);
+				// there used to be a call here which did a redraw of the Gui/OSD, which included a call to `Tools.UpdateToolsAfter` --yoshi
+			}
+		}
+
+		private void DisableLuaScript(LuaFile file)
+		{
+			if (LuaImp is not Win32LuaLibraries luaLibsImpl) return;
+
+			if (file.IsSeparator) return;
+
+			file.State = LuaFile.RunState.Disabled;
+
+			if (file.Thread is not null)
+			{
 				luaLibsImpl.CallExitEvent(file);
 				luaLibsImpl.RegisteredFunctions.RemoveForFile(file, Emulator);
-				luaLibsImpl.CallExitEvent(file);
 				file.Stop();
-				// there used to be a call here which did a redraw of the Gui/OSD, which included a call to `Tools.UpdateToolsAfter` --yoshi
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -687,13 +687,9 @@ private void SaveSessionAs()
...
-				var path = Config.PathEntries
-					.AbsolutePathFor(file.FullName, "")
-					.MakeRelativeTo(Path.GetDirectoryName(file.FullName));
-				LuaImp.ScriptList.Save(path);
-
-				Config.RecentLuaSession.Add(file.FullName); // TODO: should path be used here?
-				OutputMessages.Text = $"{Path.GetFileName(LuaImp.ScriptList.Filename)} saved.";
+				LuaImp.ScriptList.Save(file.FullName);
+				Config.RecentLuaSession.Add(file.FullName);
+				OutputMessages.Text = $"{file.Name} saved.";
```
YoshiRulz, line 691: Do you believe https://github.com/TASEmulators/BizHawk/commit/99dc0e03df4c4cd20420507a06dd9987cbdf7140 broke this? I'm guessing you can reproduce this easily, so please check 2.4.2.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -209,6 +209,7 @@ public override void Restart()
 
 			LuaFileList newScripts = new(LuaImp?.ScriptList, onChanged: SessionChangedCallback);
 			LuaFunctionList registeredFuncList = new(onChanged: UpdateRegisteredFunctionsDialog);
+			(LuaImp as Win32LuaLibraries)?.Close();
```
YoshiRulz, line 212: > Causes open forms to be closed on core restart

Not sure I like this. Ext. tool forms remain open, though they get `Restart` called...

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -933,21 +927,14 @@ private void PauseScriptMenuItem_Click(object sender, EventArgs e)
 			UpdateDialog();
 		}
 
-		private string ProcessPath(string path)
-		{
-			return Path.IsPathRooted(path)
-				? path
-				: path.MakeProgramRelativePath();
-		}
-
```
YoshiRulz, line 919: I'd like to test this further on Linux

YoshiRulz (commented):
I've looked through all of this and nothing else jumped out at me.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
kalimag: Yes, what happened here is that the original code was supposed to make the script paths inside the .luases relative to the path of the .luases
Then 99dc0e03df4c4cd20420507a06dd9987cbdf7140 accidentally made *the path of the .luases* relative to itself, so they always end up as ".\foo.luases" and are saved in the current directory

CasualPokePlayer (commented):
Normal lua functionality seems OK, nothing seems particularly wrong with these changes.

YoshiRulz:
> Should removing a script call LuaFile.Stop() and call the exit event like disabling a script does? I would think so, but it didn't before. Also goes for scripts cleared when loading a session/creating a new session.

If you're referring to `event.onexit`, yes obviously removing should trigger those.

> Should "Stop all scripts" behave the same as disabling scripts? Currently it only changes the State flag.

It should be equivalent to clicking on each running script individually and toggling them.

> Should the relative path conversion when loading lua scripts be fixed instead of removed? If so, what is the intended behavior? No other file loading code behaves like this from what I've seen (roms, lua session files, cheats, watches, etc.)

Maybe it's different because Lua scripts can be loaded from the command-line? Not sure. But I think there should be no relative paths sitting around in memory, they should be made absolute ASAP after they're entered by the user, and only made relative again for display. That problem permeates the whole codebase though.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
kalimag: I can see why keeping lua forms open may be desirable, but as it is it's not functional.

Any event handlers of the forms will still be attached to the old `LuaImp` and even the services of the old core (memory etc.). And there's no way for the restarted script in the new `LuaImp` to even interact with any forms/controls that remained open, the handles will be lost and the new `FormsLuaLibrary` isn't able to map them anyway.

YoshiRulz: For event handlers, could we add a level of indirection?

What do you mean by "the handles will be lost"? If you mean `FormsLuaLibrary._luaForms` is garbage collected, that can be changed.

kalimag: Variables, including global ones, won't persist through the `Restart`. So scripts would have to be specifically written to persist the form handles somewhere and retrieve them after restart.

`Restart` will instantiate a new `FormsLuaLibrary` that doesn't know about the previous `_luaForms`, so currently even if the restarted script knows the handle of the open form or its controls, functions like `forms.settext` won't work. This would have to be changed somehow.

Also, I don't think there's any non-hacky way for a script to check if an old handle is still valid

kalimag: Oh, sorry, I didn't even think of that.  `client.reboot_core` already works fine with keeping forms open, none of the issues I mentioned apply to that.

This change doesn't affect that case, `Restart()` bails out early if  `luaLibsImpl.IsRebootingCore` is set.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
YoshiRulz: So this restores the old behaviour?

YoshiRulz:
Pushed https://github.com/TASEmulators/BizHawk/commit/01ab9416b5dede3b1e374bfbc905d5ec119fa7ee post-merge.

PR #3486, opened 2022-12-08
Clean up and enable SNES Game Genie cheatcode converter

### LINKED ISSUE #3597: Game Genie support for SNES

YoshiRulz:
Implementation looked finished to me besides needing to remove the '-', but apparently it doesn't work.

resolves #3597

### DIFF
--- a/src/BizHawk.Client.Common/cheats/GameSharkDecoder.cs
+++ b/src/BizHawk.Client.Common/cheats/GameSharkDecoder.cs
@@ -129,12 +129,7 @@ namespace BizHawk.Client.Common.cheats
 
 		private static IDecodeResult Snes(string code)
 		{
-			if (code.Contains("-") && code.Length == 9)
-			{
-				return new InvalidCheatCode("Game genie codes are not currently supported for SNES");
-				////return SnesGameGenieDecoder.Decode(code);
-			}
-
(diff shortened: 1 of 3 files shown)


YoshiRulz:
Rebased with no changes besides adding to the unit tests. Algorithm seems sane (compare [Mesen](https://github.com/SourMesen/Mesen2/blob/7b84a75fac5a1eb44486b8c302fe55abd6dfdbc0/Core/Shared/CheatManager.cpp#L321-L354)), but I still have yet to find a code which works—the one in the unit test should be "infinite time" for SMW but it doesn't work.
Some notes:
- https://www.almarsguides.com/retro/walkthroughs/snes/games/SuperMarioWorld/GameGenie/
- https://etherealgames.com/snes/s/super-mario-world/game-genie-codes/
- https://dlevere1.proboards.com/thread/4571/snes-game-genie-handbook
- https://gamehacking.org/system/snes

Morilli:
I've looked at the code and it looks to be correct and matches other implementations. Understanding the logic is quite hard but I guess that's unavoidable with these codes.

As to why codes don't seem to work: I can only reasonably assume that most codes are garbage that may have done something at some point (or not), but aren't generally usable. For example there are multiple different "infinite lives" codes for smw like `14B4-6F07` or `C222-D4DD` that write to completely unrelated memory areas used for level or graphics data and have nothing to do with lives.

Issue #1823, opened 2020-01-29
foreach loop in MemoryCallbackSystem.cs can cause enumeration exceptions


Trivial-Man:
In commit 2b7cb1bb7d01c0acd29c955e80ef9e06c075a1e9 the Call function inside BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs was changed to use a foreach loop instead of a for loop. The problem is that the collection of memory callbacks can change itself when it is called, and thus crash Bizhawk with a System.InvalidOperationException. Can't modify a collection while iterating through it.

Reverting this specific change should remove the crash. If need be I can create a small sample script to demonstrate this behavior.

Zinfidel:
Just as a data point for this one: I had a collection of scripts with callbacks for SNES Harvest Moon that would do exactly this kind of thing. Some of the callbacks would enable or disable other callbacks that were extremely expensive and only needed to be run on a couple frames.

zeromus:
There were bugs in the original logic too. Please don't close this bug without addressing it more seriously, by actually accumulating edits to the callbacks list in a separated "added" and "removed" list and attempting to skip ones that have been removed. Don't do it with a dictionary unless you're absolutely sure you won't make the order the callbacks are run become chaotic.

adelikat:
Well, I'm not going to do all that, I'll put it back to a for loop and unassign myself.  I at least fixed what I broke

PR #3488, opened 2022-12-09
Improve performance and correctness of memory callbacks

kalimag:
## Changes

- Implement a specialized collection for `IMemoryCallback` used in `MemoryCallbackSystem` with copy-on-write behavior to safely handle adding/removing callbacks from within a callback.
- Add some basic tests for `MemoryCallbackSystem`

635fff6c5d574b161129d52a61ccead2e8897b0d addressed #1823, but the logic is still not quite correct: when a callback is removed, it will skip whatever callback is last in the list rather than the one that was removed. It also causes a significant performance hit on something that can run countless times per frame, due to the allocations and (un)subscribing `CollectionChanged` (which can be surprisingly expensive on multicast delegates).

This PR replaces the use of `ObservableCollection<T>` with a custom, pared down collection that only handles the needs of  `MemoryCallbackSystem`. It uses copy-on-write while the collection is being enumerated to allow modifying callbacks but keep the enumeration allocation free.

Some very rough performance numbers for Gargoyles (Genesis):
| Script                                                     | FPS before PR | FPS after PR |
|------------------------------------------------------------|---------------|--------------|
| None                                                       | 1000          | 1000         |
| Dummy script (only frameadvance loop)                      | 700           | 700          |
| Dummy script plus `event.on_bus_exec(function() end, 0x0)` | 215           | 500          |
| Gargoyles.lua (bundled)                                    | 115           | 175          |

## Issues/limitations

- Adding/removing callbacks from within a memory callback can be slightly more expensive
- Adding/removing a callback from within a callback will only come into effect on the next read/write/execute
In previous implementations this behavior has been inconsistent

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
+++ b/src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
@@ -4,9 +4,8 @@ using System;
 using System.Linq;
 using System.Collections;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
-using System.Collections.Specialized;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
 namespace BizHawk.Emulation.Common
 {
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
Oh wow thanks for using my script here :D

kalimag:
1. Any item is removed from `cbs`
2. The `CollectionChanged` handler does `e--`
3. The `while` loop will exit earlier
5. The last item in `cbsCopy` is not processed, regardless of which item was removed from `cbs`.

ItEndsWithTens:
I don't have the skillset to speak to the correctness of this PR, but in terms of performance I can confirm it's a night and day difference!

I'm working on [an external tool for the original Silent Hill](https://github.com/ItEndsWithTens/SilentHillMapExaminer) to help with reverse engineering, namely picking apart the map format. To help me visualize things I'm drawing a debug overlay with either the GUI API or, when available (i.e. in Octoshock), with a memory callback that's called when the game swaps its double buffers, letting me draw directly into the PSX's VRAM. Drawing right on the framebuffer keeps everything rock solid synced with the game content, and is less jarring than the swimmy disconnect otherwise seen when waiting for a PostFrame ToolUpdate call and rendering on top of the tearing. Not strictly necessary, but nice to have.

Preparing for BizHawk 2.9 I'm seeing a substantial perf drop for the same code, from 60fps down to the mid 30s or so, between 2.8 and the latest HEAD (currently b1e8d93f5f4d26198ee7be2d9af7f718b332a22f); git bisect suggests it's the fault of the aforementioned 635fff6. Checking out this pull request instead I'm right back up to full speed without breaking a sweat.

What my anecdotal evidence is worth I can't say, but I thought I should share. If you'd rather I open a proper bug report for the performance drop, I can do that too, but since this PR seems to have addressed the issue it seemed redundant to bring it up elsewhere.

PR #3491, opened 2022-12-12
Update virtualpads immediate when the movie status changes, huge impr…

adelikat:
…ovement when using tastudio and virtualpad together as the record checkbox toggle now changes the readonly status of the pad


YoshiRulz:
Will use #3490 instead.

PR #3492, opened 2022-12-13
Add missing `UpdateDialog` call in `LuaConsole`

kalimag:
Call `UpdateDialog` after the script is refreshed, in case the script was disabled by a new syntax error etc.


YoshiRulz:
For anyone else who was confused by this: `OnLuaFileChanged` can only be called when "Reload When Script File Changes" is enabled (defaults to disabled).

YoshiRulz (commented):
I feel like this is in the wrong spot. `RefreshLuaScript` should check `item.State` after the first call to `ToggleLuaScript` (and before the first call for that matter), then this can go just before `RefreshLuaScript` returns.

PR #3517, opened 2023-01-07
Improve `DisplaySurface` performance

kalimag:
- Replace `DisplaySurface.Clear()` with a significantly faster implementation
- Remove unused `int[]` pixel buffer and superfluous conversion between the array and `Bitmap`

My aim was to improve the performance of drawing on the "client" surface from Lua, which I've always avoided using simply because it causes huge FPS drops compared to drawing on "emucore" regardless of how simple the drawing is, especially at larger window sizes.

Much of the performance drain comes from `DisplaySurface.Clear()`, which uses a fairly slow byte-for-byte loop to clear the `_pixels` array. And then the cleared `_pixels` data is later copied back to the `Bitmap`.

With this PR, `DisplaySurface.Clear()` now clears the `Bitmap` directly (`LockBits`), and uses `Span<byte>.Clear()` which is a lot faster than the custom `Utils.Memset` method. (I also tested `Graphics.Clear(Color)` which surprisingly was also comparatively slow.)

As far as I can tell nothing other than the `Clear()` method actually references the `_pixels` array or the pointer to that array. Drawing functions draw to the bitmap, and rendering reads from the bitmap. So I ended up removing that aspect completely. `DrawingSurface` now doesn't really have functionality other than the `Clear()` method, and there was already a very old comment at the top saying it should be phased out, but the code around it looked fairly daunting and I didn't want to get even deeper into the weeds.

### DIFF
--- a/src/BizHawk.Client.Common/DisplayManager/DisplaySurface.cs
+++ b/src/BizHawk.Client.Common/DisplayManager/DisplaySurface.cs
@@ -1,99 +1,34 @@
+using System;
 using System.Drawing;
 using System.Drawing.Imaging;
-using System.Runtime.InteropServices;
-
-using BizHawk.Common;
 
 namespace BizHawk.Client.Common
 {
 	/// <summary>
-	/// This is a wrapper for a Bitmap, basically, which can also be a int[].
+	/// This is a wrapper for a Bitmap.
 	/// It should be phased out, in favor of BitmapBuffer and Texture2d's
 	/// </summary>
 	public unsafe class DisplaySurface : IDisplaySurface
 	{
+		private const PixelFormat Format = PixelFormat.Format32bppArgb;
+
 		private Bitmap _bmp;
-		private BitmapData _bmpData;
-		private readonly int[] _pixels;
 
 		public void Clear()
 		{
-			FromBitmap(false);
-			Util.Memset(PixelPtr, 0, Stride * Height);
+			var bmpData = _bmp.LockBits(new(0, 0, Width, Height), ImageLockMode.WriteOnly, Format);
+			new Span<byte>((void*)bmpData.Scan0, bmpData.Stride * bmpData.Height).Clear();
+			_bmp.UnlockBits(bmpData);
 		}
 
 		public Bitmap PeekBitmap()
 		{
-			ToBitmap();
 			return _bmp;
 		}
 
 		public Graphics GetGraphics()
 		{
-			ToBitmap();
 			return Graphics.FromImage(_bmp);
-		}
-
-		public void ToBitmap(bool copy = true)
-		{
-			if (_isBitmap) return;
-			_isBitmap = true;
-
-			if (_bmp == null)
-			{
-				_bmp = new Bitmap(Width, Height, PixelFormat.Format32bppArgb);
-			}
-
-			if (copy)
-			{
-				_bmpData = _bmp.LockBits(new Rectangle(0, 0, Width, Height), ImageLockMode.WriteOnly, PixelFormat.Format32bppArgb);
-
-				int w = Width;
-				int h = Height;
-				int stride = _bmpData.Stride / 4;
-				int* bmpBuf = (int*)_bmpData.Scan0.ToPointer();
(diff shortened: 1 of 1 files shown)


zeromus:
I read the detailed description from kalimag of what was done here and it sounds fine to me. It seems like some low hanging fruit was found without disturbing the hornets nests.

YoshiRulz:
Can you provide a sample script? Maybe it's because I was using Debug builds, but I didn't see any difference between this PR and `bc823f479` (or between the two surfaces for that matter).

kalimag:
```Lua
print(string.format('emu: %dx%d', client.bufferwidth(), client.bufferheight()))
print(string.format('client: %dx%d', client.screenwidth(), client.screenheight()))
print(string.format('window: %dx', client.getwindowsize()))

while true do
	gui.drawLine(0, 0, 100, 100, 0xFFFF00FF, 'client')
	--gui.drawLine(0, 0, 100, 100, 0xFFFF00FF, 'emucore')
	emu.frameadvance()
end
```

```
emu:    256x224
client: 1465x1120
window: 5x
```

With a release build, SMW Snes9x, unthrottled, no rewind, OpenGL, Windows 10, the rough numbers I'm seeing:

No Lua/drawing: ~600 FPS

Executing `gui.drawLine(0, 0, 100, 100, 0xFFFF00FF, 'emucore')` once in the Lua console has no significant effect on FPS.

After executing `gui.drawLine(0, 0, 100, 100, 0xFFFF00FF, 'client')` once in the Lua console, drops to ~230 FPS until I close the Lua console or erase the drawings.

Enabling the test script before this PR further drops me to ~85 FPS. `DisplaySurface.Clear()` and `DisplaySurface.ToBitmap()` are among the top functions according to the VS CPU profiler.

Enabling the test script after this PR only drops me to ~190 FPS.

Of course this is a fairly artificial test. The smaller the client surface, and the lower the overall performance (heavy Lua script etc.), the lower the difference will be.

YoshiRulz:
While I don't doubt you saw some improvement at large window sizes, that's hardly a fair test.

Release builds, loading a savestate in the middle of the *Gargoyles* demo in GPGX, clock throttle + unthrottled, no rewind, window size 1x, and a more realistic number of draw calls:
```lua
--gui.use_surface("client");
while true do
	for i = 0, 200 do
		gui.drawLine(200, i, 200 - i, 200, 0xFFFF00FF);
	end
	emu.frameadvance();
end
```

rev \\ surface | `emucore` | `client` | no script
--:|:-:|:-:|:-:
bc823f479 | 155 fps | 155 fps | 550 fps
abee97b44 | 165 fps | 165 fps | 640 fps

With feos' script for this game, which makes so many draw calls it falls below 100%:

rev \\ surface | `emucore` | `client`
--:|:-:|:-:
bc823f479 | 40 fps | 40 fps
abee97b44 | 35 fps | 35 fps

I have no idea why my framerates improved after rebuilding, but it seems it didn't translate to the real script.

kalimag:
At 1x window size there shouldn't be any difference between `emucore` and `client`, but I'm not playing anything at 1x SNES resolution when that's like 2% of my screen. And the only reason why I even want to draw on the `client` surface is to have a larger resolution to work with to do more detailed drawings and fit more than a couple of lines of readable text on the screen.

In this old script for Misadventures of Flink (Gen) that does a more modest number of draw calls than the Gargoyles script, at a (for me) reasonable 3x window size, I'm seeing a notable improvement from ~160 FPS to ~230 FPS. https://gist.github.com/kalimag/a2061d45d384f7c1ad194b01952d0aee

I don't know, maybe there are other factors (config? OS?) that make this change have a much more significant effect on my system.

CasualPokePlayer:
> I don't know, maybe there are other factors (config? OS?) that make this change have a much more significant effect on my system.

From what I can tell the Span Clear() here is more or less the same on Windows and Linux, ultimately just PInvoking for the native libc memset to be used (I'm assuming libc on both Windows and Linux would be around the same speed). It's probable this just doesn't have a significant performance effect when clearing out a buffer with 1x window size, but with larger resolutions it does. It should be an improvement in any case given the native memset use.

PR #3525, opened 2023-01-18
Fix hex editor domain issues with MAME

kalimag:
- Replace `OpenAdvanced_MAME` check with `_romDomain` `null` check
Fix `NullReferenceException` when changing memory domains on MAME, now that MAME roms can be opened without using the Open Advanced dialog (related to bae71326bf21a392cb360db807214d4132ecb288) 

- Clear `_romDomain` reference when loading MAME rom
Previously, when loading a MAME rom after a non-MAME rom, the "File on Disk" domain would still appear with the contents of the previous rom

For context: `_romDomain` is the "File on Disk" pseudo memory domain in the hex editor, which is not initialized for MAME roms.

Check if completed:


YoshiRulz:
LGTM. @vadosnaprimer?

PR #3529, opened 2023-01-20
Revamp Waterbox Setup

### LINKED ISSUE #3593: Expose MAME's machine/game compatibility status in frontend
_Insert me not paying attention to the repo and not realizing that MAME is now a release for 2.9 instead of experimental here._

As we know with MAME upstream it flashes two types of warnings when a title that can be ran, but has issues.
An Orange one, and a Red one.

For the over-explain something that we know but just doing it for same page sake:
The Orange one can get into game with some minor issues, and is a setting that can be toggled to stop being intrusive "Skip Imperfect Emulation Warnings".
The Red one can not and wouldn't get into game, or have insanely terrible compatibility with a "This System Doesn't Work". Some titles are done like this to hide that they're actually functional to hide from certain companies.

In a mockup provided below would having something like this be an acceptable way to make it not only non-intrusive, but also more acceptable of a core to at least let the user know.
image

This is something that the MAME development team would like to have towards an end user when MAME is a ported core, so the user would at least know that the title has an issue.
And rather than always pointing to the upstream. Having something in Hawk's Core would be at least helpful to our end users.

CasualPokePlayer:
This PR revamps the waterbox setup. Previously gcc would be used. Now clang will be used, along with other llvm tooling (lld, llvm-ar, llvm-ranlib). In addition, the compiler_rt / libunwind / libcxxabi / libcxx setup has been updated to the latest stable llvm, 15.0.7 (previously it was some 10.x commit). The old version notably caused an issue with ares as it needed some function implemented in a newer compiler_rt. This new llvm should also generally help prepare for any hot new C++ toys emulators may use.

Now why clang over gcc?

Speed: clang in recent times generally beats out gcc in optimization. This can cause noticable speed boosts in cores, ares is noted as a 15% (!?) speed boost. Other times it doesn't cause any measurable difference in speed.

Size: clang generally produces smaller binaries, which is a small plus.

Warnings / less stupid code: clang also has an advantage (or disadvantage, depending on how you look at it) when it comes to warnings, having way more warnings than gcc. These could be just noise, but sometimes can be very useful. A few functions in bsnes/ares' nall were caught as being infinitely recursive as clang does have a warning against that apparently (they are fixed in this commit). Less warning but more not allowing dumb code: clang evaluates an entire template struct/class, which ended up causing a compilation error in libsnes as it had uncompilable code in an unused function. gcc would only error on this function if it was actually used due to its more lazy approach. This overall hopefully should make for less stupid code by us and maybe catch some dumb upstream bugs.

clang also has the misc advantage of better suited for cross-compiling Windows code, especially given clang-cl (so a Linux system could be made to compile waterbox and native cores with clang as a "universal" compiler). This is fairly minor of course, but a nice note.

The main disadvantage of this new setup would simply be it requires redoing any developers current setups, but that will be needed anyways due to the updated libcxx part, so changing it to clang at the same time seems reasonable enough to do.

Another thing this PR does (not particular to clang but became apparent due to it): it appears with the previous setup builtin compiler headers would be included. This was more or less fine, as they in practice only got pulled in when intrinistics were used by the core. But since we probably don't want waterbox cores exposed to the host headers (who knows what gcc/clang may decide to put in there), I included a script which would copy all the intrinsic files from our cloned llvm-project into a special folder in sysroot, which may be included for a core with NEED_INTRINSICS (similar to NEED_LIBCO).

This PR also updates MAME.

resolves #3593


YoshiRulz:
Does this (and the musl PR) prevent GCC from being used?

Will try this branch in Nix later and edit results in here.

CasualPokePlayer:
It orients the setup towards clang, although GCC should still be able to be used by tweaking build scripts. If desired, I probably could tweak all the scripts to consider CC, and use clang when it is not defined.

nattthebear (approved):

Morilli:
Note: this will mostly likely require a `musl` submodule commit update once <https://github.com/nattthebear/musl/pull/4> is merged (which needs to happen before this PR is merged also!).

CasualPokePlayer:
Both gcc and clang should work now with 0bc6d72a5ee3b3832003d2bae137c36835236400

Morilli:
llvm 16 has been released: https://github.com/llvm/llvm-project/releases/tag/llvmorg-16.0.0

.gitmodules
```diff
@@ -31,6 +31,7 @@
 [submodule "waterbox/llvm-project"]
 	path = waterbox/llvm-project
 	url = https://github.com/llvm/llvm-project.git
+	branch = llvmorg-16.0.0
```
Morilli, line 34: Does this actually do anything? Because the branch name is actually `release/16.x` while the tag is called `llvmorg-16.0.0`.

CasualPokePlayer: The release/16.x isn't for 16.0.0, it's for 16.x (it'll get updates eventually for 16.0.1 and beyond most likely). A tag is a branch itself (kinda). I'm not sure it really does anything here other than give us documentation that was at in fact at exactly 16.0.0.

YoshiRulz: I believe it dictates what `git pull` does when it's ambiguous.

CasualPokePlayer:
With llvm-16, both clang and gcc now compile all cores.

There is a catch however, it seems gcc-11 does not work here. Only gcc-12 appears to work (gcc-12/g++-12 packages are available for Ubuntu 22.04, so this isn't that bad). This would drop (simple) support for gcc for 20.04 and older. Also, this appears to drop (simple) gcc support entirely for Debian, at least until bookworm is out. Regardless, they can use PPAs or just use clang (and on that, llvm has an [apt packages](https://apt.llvm.org/) themselves that offers the latest clang for any not EOL Debian/Ubuntu version, so even as far back as Ubuntu 18.04 can setup a recent enough clang suitable for the waterbox environment).

YoshiRulz (commented):
After a few snags, I got the toolchain building on my Manjaro machine. GPGX, old BSNES, PicoDrive, Snes9x, TIC-80, Uzem, and all the Nyma cores sans Saturnus also built.

~~Ares64 fails because `__builtin_ia32_*` doesn't exist. (Note: I don't think I've built Ares64 before.)~~ fixed

~~New BSNES fails with:~~ fixed
```
cxx bsnes/processor/spc700/spc700.cpp
cxx bsnes/processor/wdc65816/wdc65816.cpp
cxx bsnes/processor/arm7tdmi/arm7tdmi.cpp
cxx bsnes/emulator/emulator.cpp
make: *** No rule to make target '/home/yoshi/Documents/.repos/BizHawk/waterbox/bsnescore/obj/release/bsnes/gb/Core/apu.c.o', needed by '/home/yoshi/Documents/.repos/BizHawk/waterbox/bsnescore/obj/release/bsnes.wbx'.  Stop.
```

MAME fails with:
```
make[1]: Entering directory '/home/yoshi/Documents/.repos/BizHawk/waterbox/mame-arcade/mame'
makefile:514: *** PLATFORM is not x86.  Stop.
make[1]: Leaving directory '/home/yoshi/Documents/.repos/BizHawk/waterbox/mame-arcade/mame'
make: *** [Makefile:24: /home/yoshi/Documents/.repos/BizHawk/waterbox/mame-arcade/obj/release/libmamearcade.wbx] Error 2
```
(Note: I haven't built MAME before.)

~~I wasn't able to build melonDS because of submodule shenanigans.~~ fixed

`make` hung whenever I tried to build Saturnus. The last things it prints are:
```
cxx mednafen/src/ss/vdp2_render.cpp
mednafen/src/ss/vdp2_render.cpp:1327:13: warning: unknown pragma ignored [-Wunknown-pragmas]
#pragma GCC push_options
            ^
mednafen/src/ss/vdp2_render.cpp:1328:13: warning: unknown pragma ignored [-Wunknown-pragmas]
#pragma GCC optimize("no-unroll-loops,no-peel-loops,no-crossjumping")
            ^
mednafen/src/ss/vdp2_render.cpp:1362:13: warning: unknown pragma ignored [-Wunknown-pragmas]
#pragma GCC pop_options
            ^
mednafen/src/ss/vdp2_render.cpp:3445:12: warning: unused variable 'wdcq' [-Wunused-variable]
  uint32_t wdcq = __atomic_fetch_add(&DrawCounter, 1, __ATOMIC_RELEASE);
           ^
mednafen/src/ss/vdp2_render.cpp:3436:17: warning: unused variable 'bwthresh' [-Wunused-variable]
 const unsigned bwthresh = VisibleLines - 48;
                ^
5 warnings generated.
cxx mednafen/src/hw_cpu/m68k/m68k.cpp
mednafen/src/hw_cpu/m68k/m68k.cpp:52:13: warning: unknown pragma ignored [-Wunknown-pragmas]
#pragma GCC optimize ("no-crossjumping,no-gcse")
```

CasualPokePlayer:
Not sure what the ares fail is about, might be unintended consequences from me directly using clang Headers from out cloned llvm-project? (this was intended to prevent host system headers from needing to be used, I would assume anything "too new" wouldn't be an issue unless they were specificially used. Even then, ia32 functions should be present regardless? Those functions would be old mmx/sse intrinsics (26/24 years old!) which by default should be available on any x86-64 target.

New BSNES fail is you didn't init the SameBoy submodule in submodules/sameboy/libsameboy, so it can't find SameBoy's source files. Should be a trivial fix on your side.

MAME's issue would be `uname -p` didn't give x86_64. I had this issue too with Debian. GNU docs on uname note that -p is non-portable, even across Linux/GNU, so I assume that is the reason here. I'll add in a fix later.

melonDS seems to be I had changes (well, more just pulling upstream) not yet actually committed to the main repo, oops, that's fixed now.

When did make hang on Saturnus? It does take a long while when linking, you might just need to wait longer?

CasualPokePlayer:
> Not sure what the ares fail is about, might be unintended consequences from me directly using clang headers from out cloned llvm-project?

This seems to be the issue, fails with clang-15 but not with clang-16. Annoying, but I guess that makes sense. Don't particularly like letting host headers be used (especially as they sometimes bring in other unintended consequences there aren't only intrinsic headers here), I'll just revert it back to using host headers.

YoshiRulz (approved):
Edited my earlier review; everything now seems to be working. I was going to say I'm looking forward to the download size reduction, but it seems to be about the same :P Still waiting on MAME though

src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
```diff
@@ -48,15 +49,15 @@ public MAME(CoreLoadParameters<object, MAMESyncSettings> lp)
...
+				_adapter = CallingConventionAdapters.MakeWaterbox(new Delegate[] { _logCallback, _baseTimeCallback, _inputPollCallback, _filenameCallback, _infoCallback }, _exe);
 				_core = BizInvoker.GetInvoker<LibMAME>(_exe, _exe, _adapter);
 				StartMAME(lp.Roms);
 			}
 
 			if (_loadFailure != string.Empty)
 			{
 				Dispose();
-				throw new Exception("\n\n" + _loadFailure);
+				throw new("\n\n" + _loadFailure);
```
YoshiRulz, line 60: Uhh???

CasualPokePlayer: You can do that btw (thank Rider)

YoshiRulz: Thanks Rider, I hate it

waterbox/readme.txt
```diff
@@ -9,25 +9,27 @@ It consists of a modified musl libc, and build scripts to tie it all together.
 	PREPARE A WIN10 WORKSTATION:
 	Using the guidance at (https://docs.microsoft.com/en-us/windows/wsl/wsl2-kernel & https://docs.microsoft.com/en-us/windows/wsl/install-win10):
 	1. Install WSL2 
-	2. Install Ubuntu 20.04 LTS (https://www.microsoft.com/en-us/p/ubuntu-2004-lts/9n6svws3rx71)
-	3. Clone the bizhawk repository. You can use it through /mnt or /home if you really like 
-	4. Install build tools: sudo apt-get update && sudo apt-get install gcc g++ make cmake llvm zstd
-	4b. (Note for future work: ideally the llvm installed above would not be required)
+	2. Install Ubuntu 22.04.1 LTS (https://apps.microsoft.com/store/detail/ubuntu-22041-lts/9PN20MSR04DW)
```
Morilli, line 12: Small nitpick: the LTS is now on `22.04.2` instead of `22.04.1`. The link already redirects, but might as well change the text + link accordingly.

waterbox/readme.txt
```diff
@@ -41,14 +43,19 @@ It consists of a modified musl libc, and build scripts to tie it all together.
...
+5. Some additional preparation is required before all the cores can be built:
 	cd nyma && ./build-and-install-zlib.sh
 
-7. You are now ready to start building cores. Each supports `make` and `make install`, as well as `make debug` and `make install-debug` for local development.  From the root directory, the following should all be valid:
+6. You are now ready to start building cores. Each supports `make` and `make install`, as well as `make debug` and `make install-debug` for local development.  From the root directory, the following should all be valid:
+	cd ares64 && make install
+	cd bsnescore && make install
 	cd gpgx && make install
 	cd libsnes && make install
+	cd mame-arcade && make install
```
Morilli, line 58: Might as well add `virtualjaguar` here as well while we're at it.

waterbox/readme.txt
```diff
@@ -9,25 +9,27 @@ It consists of a modified musl libc, and build scripts to tie it all together.
...
 
 	PREPARE A LINUX WORKSTATION:
-	1. TODO. This should work, but no one has tested it yet
+	1. Debian/Ubuntu based distros work fine with the WIN10 Workstation instructions (confirmed working with Debian 11). Other distros should be able provide the needed programs in their package managers. make, cmake, gcc/g++ or clang/clang++, ld or lld (lld is required for clang), gcc-ar/gcc-ranlib or llvm-ar/llv-ranlib (which one you choose doesn't matter), llvm-config, zstd.
 
 2. Clone bizhawk sources
 	* Make sure git's core.autocrlf is set to false, as the alternatives cause git to modify the line endings in .sh-looking files which WILL break the build process, from the very first step.
 	* This is NOT git's default. You will need to change it!! Go ahead and set it false globally permanently, since do you really want git modifying files?
 	* Make sure you have initialized and updated the needed submodules in the waterbox directory (for example, /waterbox/llvm-project and /waterbox/musl, etc.)
```
Morilli, line 30: There should be a note here that the `llvm-project` submodule can be checked out faster by using the setup-llvm script.
 Similarly the `libcxx/readme.txt` should be updated.

waterbox/libcxx/configure-for-waterbox-phase--
```diff
@@ -2,25 +2,34 @@
...
-export CXXFLAGS="-mcmodel=large -mstack-protector-guard=global -no-pie -fno-pic -fno-pie"
+export ASMFLAGS="-w -mcmodel=large -mstack-protector-guard=global -fno-pic -fno-pie -fcf-protection=none"
+export CFLAGS="-w -mcmodel=large -mstack-protector-guard=global -fno-pic -fno-pie -fcf-protection=none"
+export CXXFLAGS="-w -mcmodel=large -mstack-protector-guard=global -fno-pic -fno-pie -fcf-protection=none -fno-use-cxa-atexit"
+export LDFLAGS="-no-pie"
+
 cmake \
-	-DCMAKE_C_COMPILER="$SYSROOT/bin/musl-gcc" \
-	-DCMAKE_CXX_COMPILER="$SYSROOT/bin/musl-gcc" \
+	-DCMAKE_ASM_COMPILER="$CC" \
```
Morilli, line 17: cmake warns that `CMAKE_ASM_COMPILER` is unused here (and in the 1 and 2 scripts as well), so it's probably best to remove the `ASMFLAGS` and `CMAKE_ASM_COMPILER` specifications.

waterbox/libsnes/bsnes/target-libsnes/libsnes_pwrap.cpp
```diff
@@ -597,9 +597,8 @@ void new_emuthread()
 //DLL INTERFACE
 
 #include <emulibc.h>
-#define EXPORT extern "C" ECL_EXPORT
 
-EXPORT void* DllInit()
+ECL_EXPORT void* DllInit()
```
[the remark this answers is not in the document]
CasualPokePlayer: ECL_EXPORT already has extern "C", I assume this was done at a time before that (and subsequently copy pasted everywhere...) clang gives out a warning due to the essentially double extern "C" usage.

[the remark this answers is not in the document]
Morilli: Ah I see, other cores just define EXPORT to ECL_EXPORT. That's fine then.

waterbox/libcxx/configure-for-waterbox-phase--
CasualPokePlayer: I'm assuming you meant to review another file since compiler_rt does have assembly bits (and thus cmake gives no warnings here), unless something weird with your setup is causing it to not use assembly bits here? Only 1 and 2 have asm compiler unused.

Morilli: I didn't actually run through the entire libcxx setup and was instead just going by memory, my bad 😅

Issue #3172, opened 2022-03-02
Hostile Bug Report Responses are Hurting the Project


alyosha-tas:
Natt's responses to the recent NES mapper issue have led to me having to field a very frustrating complaint about his behaviour. I shouldn't have to do this. Over the years I personally have become de-sensitized to it, as I'm sure may other contributors have, but you should know that ordinary users and other general observers have not. In particular, fiskbit's work has directly benefited NESHawk in ways that I am incapable of doing on my own, so you are even alienating people who you don't even know are important contributors.

To at least hopefully bring attention to this issue, I am withdrawing from this project and will no longer be contributing.

TiKevin83:
In my opinion there are core problems with the contribution methodology of this project that run contrary to standard open source development practices and this has contributed to the friction experienced between devs. There are experiences with individual devs that could be named but it's not relevant when the issue is exacerbated by the project's design philosophy.

- Nobody, not even the most seasoned devs should be committing directly to the main branch. This is a mature project with a regular release cadence, not an unused experiment. Every commit should be PR'ed and have at least one approval from another developer before merge.
- The main branch should be renamed to "main" from "master" as this has become a common open source convention and hurts nothing to change.
- A Contributing guideline should be implemented similar to the one created for TASVideos which links out to a Code of Conduct, local development setup (the readme), coding conventions used in the project, and frontend design conventions. This could then be used as an authoritative reference to resolve disputes cleanly within issues and PR reviews.

Without structured contribution guidelines and code reviews this project will continue to alienate seasoned developers who are used to a welcoming and team-building atmosphere in an open source community.

CasualPokePlayer:
Nobody should be committing to the main branch? I don't see how well that would work out, given how segmented the project is itself, and how many submodules there are. Primarily it could encourage moving/creating cores to/in submodules like ported cores often are. Which at that point the only "review" will be "reviewing the submodule update." Remember too there aren't a ton of developers, and we have different skills which can make a difference to however meaningful a review actually is.

[...]

PR #3530, opened 2023-01-21
Add a Code of Conduct

CasualPokePlayer:
This is taken largely from https://github.com/TASVideos/tasvideos/blob/main/CODE_OF_CONDUCT.md

Who to connect for these kinds of issues is still up for discussion. I wasn't sure what to put in here so I only kept in adelikat's email.

Before this is merged, I would like for all currently "active" (i.e. anyone with push access and has committed within the last year) developers to agree to this CoC before merging, and affirming they understand consquences of not following it:

- [ ] @adelikat 
- [ ] @vadosnaprimer 
- [ ] @zeromus 
- [ ] @YoshiRulz 
- [ ] @Morilli 
- [ ] @nattthebear 
- [ ] @SuuperW
- [ ] @CasualPokePlayer 

A reply like this to this PR will be sufficient:
```
I, <username>, agree to follow the code of conduct defined in the BizHawk project. I understand violating this agreement will result in disciplinary action and may result in temporary or permanent removal from the BizHawk team.
```

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,51 @@
+# Contributor Code of Conduct
+
+BizHawk is a highly segmented and interconnected project, which ultimately relies the work of many different developers, whether it is developers of emulation cores we port, or the developers of our emulation frontend. To that end, we expect contributors of this project to follow a basic code of conduct both inside and outside project spaces.
+
+## Attribution
+
+This Code of Conduct is adapted from the [Contributor Covenant version 1.4](https://www.contributor-covenant.org/version/1/4/code-of-conduct.html)
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+- Using welcoming and inclusive language
+- Being respectful of differing viewpoints and experiences
+- Gracefully accepting constructive criticism
+- Attempting collaboration before conflict
+- Focusing on what is best for the community
+- Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+- Violence, threats of violence, or inciting others to commit self-harm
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
> using an official project e-mail address, posting via an official social media account

Do we have those?

> If you are unsure whether an incident is a violation, or whether the space where the incident took place is covered by our Code of Conduct, we encourage you to still report it.

Does that include events before acceptance of that code?

CasualPokePlayer:
> > using an official project e-mail address, posting via an official social media account
> 
> Do we have those?

I suppose adelikat's email would be the closest to an official project e-mail address, but overall not really. We don't have any official social media accounts either. Although I suppose these are more examples and are things that could in theory become reality.

> > If you are unsure whether an incident is a violation, or whether the space where the incident took place is covered by our Code of Conduct, we encourage you to still report it.
> 
> Does that include events before acceptance of that code?

I wouldn't be so sure, which I would hope discussion on this PR would have that be decided. Personally, I would think it would be best to have it somewhat "grandfathered" or so any events beforehand should be given more as a warning / thinner ice for those people.

zeromus:
Hahaha I appreciate your courtesy in including me in the list of vetoes. Count me as an actual veto.

adelikat:
For the record, I agree with the assessment of #3172 and think behavior needs to be addressed.  I also think a code of conduct is a good idea, if everyone would be willing to agree with this and treat it seriously.

An important issue though is that my email that's mention no longer exists and likely won't ever exist (I'll have to figure something out for the tasvideos repo, but that's not a relevant problem here)

vadosnaprimer:
Regarding personal approval/agreement to the code, I don't like to speak in copypaste but I' ready to follow the suggested thing.

YoshiRulz:
> An important issue though is that my email that's mention no longer exists and likely won't ever exist

Could you update [your homepage](https://tasvideos.org/HomePages/Adelikat#Contact) and/or [the security page](https://github.com/TASEmulators/BizHawk/security/policy) then?

CasualPokePlayer:
These kinds of policies would *mostly* just apply to regular developers, who currently are ones who have pushed to the repo "recently" (perhaps this access should be more restricted?). Optimally the governance structure would ultimately be the org "owners" (adelikat / zeromus / feos), but given zeromus' reply I suppose that wouldn't work out :(

zeromus:
Run through some scenarios of how this would play out in your head. It's almost comical. The lords of the place aren't administering justice, so the contributors tweet about it, crash the palace, run them out of town on a rail, and take up administrative duties?

PR #3619, opened 2023-04-09
SharpAudio

CasualPokePlayer:
Knocks out audio stuff from SlimDX/OpenTK to one package (see https://github.com/TASEmulators/BizHawk/issues/2261). See here for the modifications done: https://github.com/CasualPokePlayer/SharpAudio/tree/netstandard20. Dunno what to do with the modified fork in terms of the BizHawk repo (should it just be slapped in a submodule? a link just placed somewhere? should it go to TASEmulators org?). Leaving this draft for now (as far as I can tell though it just works)

EDIT: Also CI fails for whatever reason lol

EDIT2: It might actually not work, after a rebuild it is failing for me. I suspect actually it's related to the latest commit which I am yet checking, I would guess things were not getting rebuilt properly so it was still using old code?


vadosnaprimer:
TASEmulators fork + submodule sounds the sanest.

CasualPokePlayer:
After further use of this library, I don't think this is a good choice for us. In debugging I found multiple issues throughout the library, some of which seem to just indicate the lack of actual use (e.g. OpenAL disposing will always crash due to faulty dispose code). Using Vortice directly seems to be a better option for our needs (for XAudio2 at least).

This isn't a complete loss at least, the device enumeration issue has been figured out, and was using Vortice's bindings so it would be a copy paste to port it over.

PR #3622, opened 2023-04-11
Switch XAudio2SoundOutput to use Vortice instead of SlimDX

CasualPokePlayer:
A lot less painful than SharpAudio was, also just works!


CasualPokePlayer:
Seems that Vortice is no longer supports standard 2.0 in its 3.x releases (it only supports net 6+), so we'll have to stay in 2.4.2 for the time being (not really an issue, outside of this [slightly annoying hack](https://github.com/TASEmulators/BizHawk/blob/120e42ddca64cdb54bbe119b8198e1c2d2753cd4/src/BizHawk.Bizware.DirectX/GamePad.cs#L37-L59))

PR #3625, opened 2023-04-11
Use source generation for VersionInfo

Morilli:
Doing this instead of what was being done previously has a couple of advantages:
1. It gets rid of a legacy build tool that seemingly only ran 3 git commands
2. It allows building the solution on the first attempt (wow) instead of needing the `VersionInfo.gen.cs` to get generated first
3. Changes are reflected immediately in the IDE without requiring a solution rebuild
4. It simplifies the process which allows us to remove some obscure hacks


src/BizHawk.Common/BizHawk.Common.csproj
```diff
@@ -14,8 +14,6 @@
 		<PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="5.0.0" />
 	</ItemGroup>
 	<ItemGroup>
-		<!-- target in Version project generates VersionInfo.gen.cs; note that the Version project MUST BE FIRST in solution file -->
-		<ProjectReference Include="$(ProjectDir)../BizHawk.Version/BizHawk.Version.csproj" ReferenceOutputAssembly="false" />
-		<Compile Update="VersionInfo.gen.cs" DependentUpon="VersionInfo.cs" />
+		<Analyzer Include="$(MSBuildProjectDirectory)/../../References/BizHawk.SrcGen.VersionInfo.dll" />
```
YoshiRulz, line 17: Can be private

ExternalProjects/BizHawk.SrcGen.VersionInfo/VersionInfoGenerator.cs
```diff
@@ -0,0 +1,61 @@
...
+		// This is written in a way to (hopefully) work both for build and IDE analyzers
+		// FIXME: This should probably be done in a better way, but I haven't found any
+		string projectDir = Path.GetDirectoryName(context.Compilation.SyntaxTrees.First(x => x.HasCompilationUnitRoot && x.FilePath.Contains("BizHawk.Common")).FilePath)!;
+
+		string? rev = ExecuteGitWithArguments($"-C {projectDir} rev-list HEAD --count");
+		string? branch = ExecuteGitWithArguments($"-C {projectDir} rev-parse --abbrev-ref HEAD");
+		string? shortHash = ExecuteGitWithArguments($"-C {projectDir} log -1 --format=\"%h\"");
+
+		// Generated source code
+		string source = $@"// This file (VersionInfo.g.cs) is auto-generated. Any edits will be overwritten before compilation.
```
YoshiRulz, line 46: Don't need this comment anymore

ExternalProjects/BizHawk.SrcGen.VersionInfo/VersionInfoGenerator.cs
```diff
@@ -0,0 +1,61 @@
...
+			return $"{e.GetType()}: {e.Message}";
+		}
+	}
+
+	public void Execute(GeneratorExecutionContext context)
+	{
+		// Finds the current project directory in order to pass to git commands.
+		// This is written in a way to (hopefully) work both for build and IDE analyzers
+		// FIXME: This should probably be done in a better way, but I haven't found any
+		string projectDir = Path.GetDirectoryName(context.Compilation.SyntaxTrees.First(x => x.HasCompilationUnitRoot && x.FilePath.Contains("BizHawk.Common")).FilePath)!;
```
YoshiRulz, line 39: How does this not fail on the rest of the projects? `PrivateAssets="all"` from my other suggestion would fix it in any case.

Dist/QuickTestBuildAndPackage.bat
```diff
@@ -27,7 +22,7 @@ rem slimdx has a way of not making it into the output directory, so this is a go
 copy ..\..\SlimDx.dll
 
 rem Now, we're about to zip and then unzip. Why, you ask? Because that's just the way this evolved.
-..\dist\zip.exe -X -r ..\Dist\%NAME% EmuHawk.exe EmuHawk.exe.config DiscoHawk.exe DiscoHawk.exe.config defctrl.json EmuHawkMono.sh dll Shaders gamedb Tools NES\Palettes Lua Gameboy\Palettes overlay -x *.pdb -x *.lib -x *.pgd -x *.ipdb -x *.iobj -x *.exp -x dll\libsneshawk-64*.exe -x *.ilk -x dll\gpgx.elf -x dll\miniclient.* 
+..\dist\zip.exe -X -r ..\Dist\%NAME% EmuHawk.exe EmuHawk.exe.config DiscoHawk.exe DiscoHawk.exe.config defctrl.json EmuHawkMono.sh dll Shaders gamedb Tools NES\Palettes Lua Gameboy\Palettes overlay -x *.pdb -x *.lib -x *.pgd -x *.ipdb -x *.iobj -x *.exp -x dll\libsneshawk-64*.exe -x *.ilk -x dll\gpgx.elf -x dll\miniclient.*
```
YoshiRulz, line 25: ws change

Morilli: Yes, but does this need to be reverted?

src/BizHawk.Common/BizHawk.Common.csproj
Morilli: This has the disadvantage of not actually using the source generator in the IDE, just on build.
Also, how would this fix the issue about project directory resolving inside of the generator?

YoshiRulz: `<ProjectReference/>` Analyzers don't run in VS? That sounds like a bug.

Git doesn't need to know where `.git` is, the CWD just needs to be set to any subdir.

ExternalProjects/BizHawk.SrcGen.VersionInfo/VersionInfoGenerator.cs
Morilli: I can actually not find any documentation on this, but to me it appears like `PrivateAssets="all"` is doing the same to a `ProjectReference` as `Analyzer` is already doing implicitly? In any case I could not find any behavioral difference between the two in my limited testing.

Morilli:
> It should work without Git.

Have added a fallback to empty strings when git fails.

ExternalProjects/BizHawk.SrcGen.VersionInfo/VersionInfoGenerator.cs
YoshiRulz: Go ahead and check-in the assembly then.

Morilli: > Go ahead and check-in the assembly then.

Not sure what you mean, the assembly is committed and should be working fine.

YoshiRulz (approved):

zeromus:
Since ISourceGenerator is about to be deprecated, it's old enough for me to approve using. I'm sure this is better and works fine with vs2019 so I can't see any reason not to use it. As long as it works for you guys using rider, I guess

YoshiRulz:
Should probably record this somewhere: Besides reading Git (and Subversion) revision info, `BizHawk.Build.Tool` had 3 additional features that were disused at the time of this PR.
https://github.com/TASEmulators/BizHawk/blob/03aa420bd5f8ad652203410944689d9eb237f747/Build/BizHawk.Build.Tool/Tool.cs#L72-L163

PR #3633, opened 2023-04-15
Nes.Core now updates the DataBus in WriteMemory()

100thCoin:
This will correct open bus execution following write instructions.

Here's an example: 

Address $0056 has a value of $CC. Before this fix, an LSR instruction would leave $CC on the open bus:

image

However, when the LSR instruction writes the result to memory, the new result will be on the bus. Here's the result with the changes I made:

image

This corrected behavior is also demonstrated on Visual6502:

image

This issue was relevant for the verification of a recent TAS involving open bus execution. In the most recent version of Bizhawk, the tracelog looks like this:

image

However, the actual result should be this:

image

This run was console verified with help from Alyosha, despite failing in the emulator: [https://www.youtube.com/watch?v=_Qs0G_gLEyk](url)

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.Core.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.Core.cs
@@ -1149,8 +1149,8 @@ namespace BizHawk.Emulation.Cores.Nintendo.NES
 				uint flags = (uint)(MemoryCallbackFlags.CPUZero | MemoryCallbackFlags.AccessWrite | MemoryCallbackFlags.SizeByte);
 				MemoryCallbacks.CallMemoryCallbacks(addr, value, flags, "System Bus");
 			}
-			DB = value;
 
+			DB = value;
 		}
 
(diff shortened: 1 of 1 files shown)


nattthebear (approved):

src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.Core.cs
```diff
@@ -1149,6 +1149,8 @@ public void WriteMemory(ushort addr, byte value)
 				uint flags = (uint)(MemoryCallbackFlags.CPUZero | MemoryCallbackFlags.AccessWrite | MemoryCallbackFlags.SizeByte);
 				MemoryCallbacks.CallMemoryCallbacks(addr, value, flags, "System Bus");
 			}
+			DB = value;
+
```
CasualPokePlayer, line 1153: Spacing, otherwise LGTM
```suggestion

			DB = value;
```

CasualPokePlayer (changes requested):

CasualPokePlayer (approved):

PR #3634, opened 2023-04-16
Implement IBasicMovieInfo interface for IMovie

Morilli:
This is an RFC.

This change allows parsing basic movie fields into an own class that is independent from Bk2Movie/TasMovie.
The current Hierarchy is `IBasicMovieInfo` <- `IMovie` (<- `ITasMovie`).
The current separation of fields between `IMovie` and `IBasicMovieInfo` is mainly driven by what fields seem useful outside of a movie context and which don't, but I'm open for change requests.

This is mainly useful for the PlayMovie dialog which can now load movie information from files on disk without having to go through the entire Bk2Movie/TasMovie loading process.

The main purpose of this change is to allow the `PlayMovie` dialog to operate quickly and without churning memory. This should also resolve all reported issues about VirtualAlloc returning NULL.

Some additional changes I've done:
- `IBasicMovieInfo.Load` no longer takes a `preload` parameter, as that one failed to do its purpose anyways and should no longer be necessary in the current logic
- `PlayMovie` no longer potentially performs `ScanFiles();` 3 times
- the `Comments` and `Subtitles` edit buttons in the `PlayMovie` dialog now actually work and have been annotated with TODOs

### DIFF
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
@@ -1,17 +1,10 @@
-﻿using System.Collections.Generic;
-using System.Text;
+﻿using System.Text;
 
 namespace BizHawk.Client.Common
 {
 	public partial class Bk2Movie
 	{
-		protected readonly Bk2Header Header = new Bk2Header();
 		private string _syncSettingsJson = "";
-
-		public IDictionary<string, string> HeaderEntries => Header;
-
-		public SubtitleList Subtitles { get; } = new SubtitleList();
-		public IList<string> Comments { get; } = new List<string>();
 
 		public string SyncSettingsJson
 		{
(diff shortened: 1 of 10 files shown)


src/BizHawk.Client.Common/movie/interfaces/IBasicMovieInfo.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace BizHawk.Client.Common
+{
+	public interface IBasicMovieInfo
```
YoshiRulz, line 6: Is this necessary if the base class is in the same assembly?

src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
```diff
@@ -80,9 +70,8 @@ public bool StartsFromSaveRam
 			}
 		}
 
-		public string GameName
+		public override string GameName
 		{
-			get => Header.TryGetValue(HeaderKeys.GameName, out var s) ? s : string.Empty;
```
YoshiRulz, line 85: This isn't the same implementation as the base. Is it wise to only override one?

Morilli: It should be, if you look at the definition of `Bk2Header`s `this[string key]` implementation: 
```csharp
	public class Bk2Header : Dictionary<string, string>
	{
		public new string this[string key]
		{
			get => TryGetValue(key, out var s) ? s : string.Empty;
			set => base[key] = value;
		}
```

PR #3635, opened 2023-04-16
Simplify SGB related UIs and make it possible to use SGB under Gambatte link

Fortranm:
This is a revision of https://github.com/TASEmulators/BizHawk/pull/3283 made to work on BizHawk 2.9.

https://en.wikipedia.org/wiki/Super_Game_Boy#Super_Game_Boy_2

> The Super Game Boy 2 was released exclusively in Japan in 1998. Additions included a link port to allow a user to access two-player mode via the link cable...

Currently, the SGB mode of Gambatte has to be turned on via Config > Cores > GB in SGB instead of in the sync setting and cannot be used in GB Link as a result.

This PR does three things related to SGB:
1. SGB is added (or rather, merged in) as a Console Mode option for Gambatte.
image

2. "GB in SGB" check box and SGB core picker are removed and the BSNES cores are added under the GB core picket (legacy BSNES core not included). GB/GBC games will load under SGB mode of the corresponding BSNES core if one of those options are chosen.
image

3. As a result of the first change (with addition adjustments), it is now possible to use SGB mode in Gambatte Link too.
image
image
Each SGB core can have up to 4 controllers linked to it. Note that the tab parsing process is currently hardcoded to categorize every item that starts with "Px " under the tab "Player x" and everything else under "Console". There have been suggestions to use "Console x" for a core in link mode but that requires enough refactoring to be considered a separate task IMO.


src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs
```diff
@@ -15,8 +15,9 @@ namespace BizHawk.Emulation.Cores.Nintendo.BSNES
 	[ServiceNotApplicable(new[] { typeof(IDriveLight) })]
 	public partial class BsnesCore : IEmulator, IDebuggable, IVideoProvider, ISaveRam, IStatable, IInputPollable, IRegionable, ISettable<BsnesCore.SnesSettings, BsnesCore.SnesSyncSettings>, IBSNESForGfxDebugger, IBoardInfo
 	{
+		[CoreConstructor(VSystemID.Raw.GB)]
+		[CoreConstructor(VSystemID.Raw.GBC)]
```
YoshiRulz, line 19: This is just incorrect. BSNES doesn't magically emulate a new CPU variant by just adding an attribute.

This is my main concern with this PR, a needless weakening of the "system" semantics that are already too weak.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.IEmulator.cs
```diff
@@ -23,7 +23,11 @@ public bool FrameAdvance(IController controller, bool render, bool rendersound =
 				{
 					for (int i = 0; i < _numCores; i++)
 					{
-						if (s.Contains($"P{i + 1} "))
+						if (s.Contains($"P{i + 1} - C"))
+						{
+							_linkedConts[i].Set(s.Replace($"P{i + 1} - C", "P"));
+						}
+						else if (s.Contains($"P{i + 1} "))
```
YoshiRulz, line 30: I've thought this over more since reviewing the old PR and I'm now firmly against it, as I believe it will break a lot of scripts and possibly parts of EmuHawk.

YoshiRulz (changes requested):

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs
Fortranm: I have changed it so that the default system id is still SGB when the core is BSNES and now the BSNES classes don't need any changes.

YoshiRulz: That's not what I meant, but it could work.

src/BizHawk.Client.Common/movie/bk2/Bk2MnemonicLookup.cs
```diff
@@ -718,12 +718,12 @@ public static string LookupAxis(string button, string systemId)
...
+			},
 			[VSystemID.Raw.TIC80] = new()
-			{
-				["Mouse Position X"] = "mpX",
-				["Mouse Position Y"] = "mpY",
-				["Mouse Scroll X"] = "msX",
+			{
+				["Mouse Position X"] = "mpX",
+				["Mouse Position Y"] = "mpY",
+				["Mouse Scroll X"] = "msX",
```
TiKevin83, line 726: I don't actually see any changes in this file, is there a spaces vs tabs change or something?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.IVideoProvider.cs
```diff
@@ -19,19 +22,35 @@ public partial class GambatteLink : IVideoProvider
...
 
+		private int[] CreateSGBVideoBuffer()
+		{
+			if (isAnySgb)
+			{
+				return new int[256 * _numCores * 224];
+			}
+			else
+			{
+				return new int[1]; // create a dummy so that the fixed statement can work regardless
```
TiKevin83, line 52: Something seems off here, like you might want this to be nullable or throw some sort of error if an SGBVideoBuffer was asked for when isAnySgb is false and then only call SGBVideoBuffer if you know it's true

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.cs
```diff
@@ -160,6 +174,31 @@ private ControllerDefinition CreateControllerDefinition()
...
+		public bool isAnySgb { get; set; }
+
+		private bool showBorder(int i)
+		{
+			return isSgb(i) && _settings._linkedSettings[i].ShowBorder;
+		}
+
+		public bool showAnyBorder()
+		{
+			for (int i = 0; i < _numCores; i++)
```
TiKevin83, line 192: A more idiomatic way to do this in C#/LINQ:
```c#
int[] coreIndices = Enumerable.Range(0, _numCores).ToArray()
return coreIndices.Any(core => showBorder(core))
```
There may be other places in this code where you could use more idiomatic C# like this but this is just an example

TiKevin83 (changes requested):

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.cs
YoshiRulz: You don't need all the allocations that `ToArray` brings: `public bool showAnyBorder() => Enumerable.Range(0, _numCores).Any(showBorder);`

TiKevin83 (approved):
approving from a code perspective, haven't done a QA of the functionality yet

src/BizHawk.Client.Common/RomLoader.cs
```diff
@@ -483,7 +483,7 @@ private IEmulator MakeCoreFromCoreInventory(CoreInventoryParameters cip, string
 						return;
 					}
 
-					if (_config.GbAsSgb)
+					if (_config.PreferredCores[game.System].Contains(CoreNames.Bsnes115))
```
Morilli, line 486: This is a bit of an unfortunate hack that also happens to exclude legacy BSNES (even though that is currently not supported anyway). I would prefer to do this explicitly like this:
```csharp
if (_config.PreferredCores[game.System] is CoreNames.Bsnes115 or CoreNames.SubBsnes115)
```

src/BizHawk.Client.Common/config/Config.cs
```diff
@@ -30,10 +30,8 @@ public class Config
 				new[] { CoreNames.Faust, CoreNames.Snes9X, CoreNames.Bsnes, CoreNames.Bsnes115, CoreNames.SubBsnes115 }),
 			(new[] { VSystemID.Raw.N64 },
 				new[] { CoreNames.Mupen64Plus, CoreNames.Ares64 }),
-			(new[] { VSystemID.Raw.SGB },
-				new[] { CoreNames.Gambatte, CoreNames.Bsnes, CoreNames.Bsnes115, CoreNames.SubBsnes115 }),
-			(new[] { VSystemID.Raw.GB, VSystemID.Raw.GBC },
-				new[] { CoreNames.Gambatte, CoreNames.Sameboy, CoreNames.GbHawk, CoreNames.SubGbHawk }),
+			(new[] { VSystemID.Raw.GB, VSystemID.Raw.GBC, VSystemID.Raw.SGB },
+				new[] { CoreNames.Gambatte, CoreNames.Sameboy, CoreNames.GbHawk, CoreNames.SubGbHawk, CoreNames.Bsnes115, CoreNames.SubBsnes115 }),
```
Morilli, line 34: You say `the BSNES cores are added under the GB core picket (legacy BSNES core not included)`; is there a reason for the exclusion of the legacy BSNES core here?

src/BizHawk.Client.Common/movie/import/LsmvImport.cs
```diff
@@ -150,7 +150,6 @@ protected override void RunImport()
 						case "sgb_ntsc":
 						case "sgb_pal":
 							platform = VSystemID.Raw.SNES;
```
Morilli, line 152: This isn't your fault, but this statement should be changed to `platform = VSystemID.Raw.SGB`.

Morilli (dismissed):
Haven't tested the gambatte changes extensively but I think this is solid enough to get in. The "gb as sgb" UX flow was pretty bad imo so this is a definite improvement.

YoshiRulz (changes requested):
I'm curious of what @CasualPokePlayer has to say, specifically about the video buffer stuff. But this now LGTM in terms of code. (I think you deserve some kind of trophy for hacking together the button mapping by doing it in the wrong direction and with 3 nested loops.)
~~As for functionality, opening a 4x bundle and immediately closing it results in:~~
<details>
<summary>stacktrace</summary>

```
System.NullReferenceException: Object reference not set to an instance of an object
  at BizHawk.Emulation.Common.LinkedSaveRam.CloneSaveRam () [0x0002f] in <f3a4da15026e4f05ab7072e9a135677c>:0
  at BizHawk.Client.EmuHawk.MainForm.FlushSaveRAM (System.Boolean autosave) [0x000f0] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at (wrapper remoting-invoke-with-check) BizHawk.Client.EmuHawk.MainForm.FlushSaveRAM(bool)
  at BizHawk.Client.EmuHawk.MainForm.CloseGame (System.Boolean clearSram) [0x0006f] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at BizHawk.Client.EmuHawk.MainForm.CloseRom (System.Boolean clearSram) [0x00010] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at (wrapper remoting-invoke-with-check) BizHawk.Client.EmuHawk.MainForm.CloseRom(bool)
  at BizHawk.Client.EmuHawk.MainForm.CheckHotkey (System.String trigger) [0x01588] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at BizHawk.Client.EmuHawk.MainForm.<ProcessInput>b__143_0 (System.Boolean current, System.String trigger) [0x00000] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at System.Linq.Enumerable.Aggregate[TSource,TAccumulate] (System.Collections.Generic.IEnumerable`1[T] source, TAccumulate seed, System.Func`3[T1,T2,TResult] func) [0x0002e] in <89a8dc5bc33447a4a98f23a2d52e8aee>:0
  at BizHawk.Client.EmuHawk.MainForm.ProcessInput (BizHawk.Client.Common.InputCoalescer hotkeyCoalescer, BizHawk.Client.Common.ControllerInputCoalescer finalHostController, System.Func`2[T,TResult] searchHotkeyBindings, System.Func`2[T,TResult] activeControllerHasBinding) [0x00135] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop () [0x0009c] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
  at (wrapper remoting-invoke-with-check) BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop()
  at BizHawk.Client.EmuHawk.Program.SubMain (System.String[] args) [0x0042b] in <5e7f3991f9b246bb80e8d9725eda76e1>:0
```
</details>

Morilli:
> As for functionality, opening a 4x bundle and immediately closing it results in: [System.NullReferenceException]

fwiw I cannot reproduce that exception when testing with either GambatteLink or GBHawk4x as everything seems to work and play and close just fine. Do you have a specific bundle that shows this issue?

YoshiRulz:
4x *Chikyū Kaihō Gun ZAS* `SHA1:2FF6D0242D845669445E3D97FFFAE5C5F7943CFD` in a bundle and the crash is immediate, not on close. The rom on its own is fine in Gambatte. This is on 9ee38301a under Mono. I suspect it has something to do with the mapper, because 4x *Pokemon Yellow* is fine.

YoshiRulz:
I still don't have a nice repro, but I found the culprit with print-debugging, and it wasn't related to this PR. I'll push a fix to `master` later.
In working on that, I've seen that SGB in GambatteLink is working, so I won't block this any further. @CasualPokePlayer?

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -22,8 +22,7 @@ public partial class Gameboy : IInputPollable, IRomInfo, IGameboyCommon, ICycleT
 
 		[CoreConstructor(VSystemID.Raw.GB)]
 		[CoreConstructor(VSystemID.Raw.GBC)]
-		[CoreConstructor(VSystemID.Raw.SGB)]
```
CasualPokePlayer, line 25: Why is SGB being removed here? BSNES/BSNESv115 still have it.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.IEmulator.cs
```diff
@@ -19,16 +21,26 @@ public bool FrameAdvance(IController controller, bool render, bool rendersound =
 
 			foreach (var s in GBLinkController.BoolButtons)
 			{
-				if (controller.IsPressed(s))
+				if (!controller.IsPressed(s)) continue;
```
CasualPokePlayer, line 24: How much has this code been tested? Does it work as expected even with mix matching SGB vs non-SGB with multiple different combinations?

CasualPokePlayer (changes requested):

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
Morilli: Because for core choosing, the system is only set to `SGB` when the core is bsnes: https://github.com/TASEmulators/BizHawk/blob/9ee38301aff5b0322c9e10dfc3ecce75a5304c1e/src/BizHawk.Client.Common/RomLoader.cs#L485-L490

CasualPokePlayer: How does this exactly end up working for say, loading a GBC only ROM into BSNES. Does it proceed to fail completely due to not finding a fallback with the SGB system ID ctor, or does it end up still picking Gambatte with the GBC system ID?

On that, the code tries to handle old BSNES, yet it's not added into the config selection (for whatever reason).

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/GambatteLink.IEmulator.cs
YoshiRulz: I should say I didn't do much testing as I was rewriting it, only really checking for crashes.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
Morilli: That's a good point. It does NOT work currently for that exact reason and instead runs into this code whichthen no longer makes sense: https://github.com/TASEmulators/BizHawk/blob/9ee38301aff5b0322c9e10dfc3ecce75a5304c1e/src/BizHawk.Client.Common/RomLoader.cs#L859

YoshiRulz: Unrelated to the above bug, I think these `[CoreConstructor]`s should be kept, so that `emu.SystemId` always matches a `[CoreConstructor]`.

Morilli: There's more to this, and there is already broken behavior currently. I'll make an issue later.

quicknes/core
Morilli, line 1: image

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs
```diff
@@ -17,8 +17,9 @@ namespace BizHawk.Emulation.Cores.Nintendo.BSNES
...
 		[CoreConstructor(VSystemID.Raw.Satellaview)]
-		[CoreConstructor(VSystemID.Raw.SGB)]
```
YoshiRulz, line 21: Keep this attribute...

src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/SubBsnesCore.cs
```diff
@@ -8,8 +8,9 @@ namespace BizHawk.Emulation.Cores.Nintendo.BSNES
 	[ServiceNotApplicable(new[] { typeof(IDriveLight) })]
 	public class SubBsnesCore : IEmulator, ICycleTiming
 	{
+		[CoreConstructor(VSystemID.Raw.GB)]
+		[CoreConstructor(VSystemID.Raw.GBC)]
 		[CoreConstructor(VSystemID.Raw.Satellaview)]
-		[CoreConstructor(VSystemID.Raw.SGB)]
```
YoshiRulz, line 12: ...and here...

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -21,8 +21,7 @@ public partial class Gameboy : IInputPollable, IRomInfo, IGameboyCommon, ICycleT
...
```
YoshiRulz, line 24: ...and here...

src/BizHawk.Emulation.Cores/Consoles/Nintendo/SNES/LibsnesCore.cs
```diff
@@ -22,7 +22,8 @@ namespace BizHawk.Emulation.Cores.Nintendo.SNES
 	public unsafe partial class LibsnesCore : IEmulator, IVideoProvider, ISaveRam, IStatable, IInputPollable, IRegionable, ICodeDataLogger,
 		IDebuggable, ISettable<LibsnesCore.SnesSettings, LibsnesCore.SnesSyncSettings>, IBSNESForGfxDebugger
 	{
-		[CoreConstructor(VSystemID.Raw.SGB)]
```
YoshiRulz, line 25: ...and here.

YoshiRulz (changes requested):

src/BizHawk.Emulation.Cores/Consoles/Nintendo/SNES/LibsnesCore.cs
Morilli: At the very least this one is questionable as the core doesn't ever actually sets its `SystemID` to SGB. I don't really mind this too much, but I do think we should be able to agree on what this attribute is supposed to mean.

YoshiRulz: Semantically, it indicates which constructor to use for a given `RomGame`.
Also, the set across all of the core's constructors doubles as the list of sysIDs which that core can emulate. We could split that out to the `[Core]` attr instead, but I think a single-source-of-truth is better.
https://github.com/TASEmulators/BizHawk/blob/dbc6acc8fb6b98ad3ce3b9a415f6a5566471eb66/src/BizHawk.Emulation.Cores/Consoles/Nintendo/SNES/LibsnesCore.cs#L112-L115
^ This is clearly incorrect. I imagine it was a hack to get the frontend to show the right menus/tools.

Morilli: > the set across all of the core's constructors doubles as the list of sysIDs which that core can emulate.

This is precisely the issue; I don't think this invariant necessarily needs to hold true (it doesn't even right now with that old bsnes core), and even with your suggested change bsnes* core cannot provide the GB/C SystemID, even though they have a corresponding CoreConstructor attribute.

That attribute is currently entirely internal and only used for the internal selection process of which cores to consider for a given rom, so putting semantic stuff here that is supposed to communicate meaning to the reader while doing nothing for the core or romloader that's using it is misguided imho.
It's fine to communicate this SGB stuff with a comment, but the attribute itself doesn't need to be here.

The conceptual question here is whether it is acceptable for a core to load a rom of type `X` and declare a SystemID of type `Y != X` using that loaded rom, and I say yes.

YoshiRulz: How about this:
- `RomLoader` never sets `SGB` on any rom (since any GB rom could be used with SGB, they should all be detected as the same, and `GB`/`GBC` is the better choice)
- There is no core preference for `SGB`, only `GB` and maybe `GBC`
- Cores are perfectly honest with setting `IEmulator.SystemId` (not necessarily matching the one passed in, since a sync setting can choose between `GB`/`GBC`/`SGB`)
- `SystemId` always matches a `[CoreConstructor]`—if it doesn't make sense to have that ctor, like with `SGB` which `RomLoader` will never use, it can simply be `private` and call something like `this(default)`

I think that would give the consistency I'm after while being simpler for the end-user.

Morilli: The first three should already be present here.

If you feel like adding dummy constructors with the SGB attribute just to signal intent, fine.

YoshiRulz: The first 3 are not already implemented. This PR addresses some of them, but I'm saying we should clean it up on master first.

YoshiRulz (approved):
New approach LGTM, just need to clean up the diff.

vadosnaprimer:
Any news?

vadosnaprimer:
@Fortranm

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.cs
```diff
@@ -47,6 +47,11 @@ public Gameboy(CoreComm comm, IGameInfo game, byte[] file, GambatteSettings sett
 
 				var flags = LibGambatte.LoadFlags.READONLY_SAV;
 
+				if (game.System == VSystemID.Raw.SGB)
+				{
+					_syncSettings.ConsoleMode = GambatteSyncSettings.ConsoleModeType.SGB2;
```
YoshiRulz: Not sure if this should be mutating the sync settings, IIRC we had some problem with that in mGBA

src/BizHawk.Emulation.Cores/Consoles/Nintendo/Gameboy/Gambatte.ISettable.cs
```diff
@@ -115,7 +115,7 @@ public enum ConsoleModeType
 			}
 
 			[DisplayName("Console Mode")]
-			[Description("Picks which console to emulate. Forced to be 'SGB2' when 'GB in SGB' is checked.")]
+			[Description("Picks which console to emulate. 'SGB2' always in effect when 'GB in SGB' is checked.")]
```
RetroEdit:
```suggestion
			[Description("Picks which console to emulate. 'SGB2' is always in effect when 'GB in SGB' is checked.")]
```

YoshiRulz:
@Fortranm Can you go to https://github.com/Fortranm/BizHawk/tree/0e32693d06bceb256260412c0796eecad94aec8f and create a new branch like so:
Screenshot_20251124_221528
...then open a new PR from there?

PR #3637, opened 2023-04-17
Fix the build?

nattthebear:
I've been unable to build since 65ffa3f.  The primary issue appeared to be code analysis tools no longer succeeding in loading:

```
CSC : warning CS8032: An instance of analyzer BizHawk.Analyzers.FeatureNotImplementedAnalyzer cannot be created from C:\Develop\BizHawk\References\BizHawk.Analyzer.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=4.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.. [C:\Develop\BizHawk\src\BizHawk.Common\BizHawk.Common.csproj]
```

This ultimately resulted in the git hash information not processing correctly.

This change fixes the build, but I'm not sure if we should do it, as the solution is a bit of a hack.

### DIFF
--- a/src/MainSlnCommon.props
+++ b/src/MainSlnCommon.props
@@ -22,4 +22,7 @@
 	<ItemGroup Condition=" '$(Configuration)' != 'Debug' ">
 		<Compile Remove="$(MSBuildProjectDirectory)/debug/**/*" />
 	</ItemGroup>
+	<ItemGroup>
+		<PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="4.3.0-3.final" />
+	</ItemGroup>
 </Project>


nattthebear:
Updating to SDK `7.0.203` fixed the build for me without needing this commit.  If that's expected and we want that to be the required thing, we could do that as well.

Morilli:
It seems like the current latest .NET 6 sdk only supports version `4.2` of the `Microsoft.CodeAnalysis.CSharp` dll, so downgrading from 4.3 to 4.2 would also be an option (or, dare I say it, require the .NET 7 sdk for build).

vadosnaprimer:
Builds fine without this commit **or** 7 sdk on vs2022.

Morilli:
This will build fine with sdk version 6.0.3XX and above (including any .net 7 sdk I believe) and not work with any older version, including anything that is supported with vs2019.

YoshiRulz:
In order of my preference:
1. Downgrade the package (I think new versions expand Roslyn plugins' capabilities, but I haven't changed any of ours)
2. This hack
3. Up SDK requirement

vadosnaprimer:
№1 sounds the sanest because I don't think we already have to depend on something brand new, until there's a conscious decision to upgrade the whole thing.

Morilli:
My personal preferred order would be `3 > 1 > 2`, just because I don't like introducing a hack when upgrading the required sdk version to a 1 year one old is enough.

That said, if we do want to continue supporting older .net sdk versions, the latest 6.0.1XX .net 6 sdk seems to support roslyn 4.0, so downgrading to that should™️ be enough.

YoshiRulz:
Redundant with https://github.com/TASEmulators/BizHawk/commit/131bc65e8f5127b05e93e7513198efe184ab1a6a.

Issue #3615, opened 2023-04-08
MAME rom loading regression


YoshiRulz:
Initially reported by Memory on Discord. She was running the TMNT romset in 2.9, and EmuHawk crashed while loading the MAME core. The same game loads correctly on other machines. Other games work on her machine. edit: Builds back to 2.9 RC1 behave the same. Current suspect is wbox re-port.

<details>
<summary>stdout</summary>

```
Using DirectInput+XInput for host input (keyboard + gamepads)
ROM was not a multiple of 1024 bytes, and not a recognized header size: 848. Assume it's purely ROM data.
Game was not in DB. CRC: 4A804626 MD5: 5375BCCB99C466E638E75DE1299426CE
MemoryBlock created for address 36f00000000:36f6b189000 with mirror 2858a490000:285f5619000
Mouting `libmamearcade.wbx` @36f00000000
  Sections:
    @0:0     `` 0 bytes
    @36f000002a8:36f000002c0 R   `.note.gnu.build-id` 24 bytes
    @36f00001000:36f00001003 R X `.init` 3 bytes
    @36f00001010:36f05a7b625 R X `.text` 94873109 bytes
    @36f05a7b625:36f05a7b628 R X `.fini` 3 bytes
    @36f05a7c000:36f08b75c20 R   `.rodata` 51354656 bytes
    @36f08b75c20:36f08e871cc R   `.eh_frame_hdr` 3216812 bytes
    @36f08e871d0:36f0a3c86b4 R   `.eh_frame` 22287588 bytes
    @36f0a3c86b4:36f0a6a8520 R   `.gcc_except_table` 3014252 bytes
    @36f0a6a9000:36f0a6a902c RW  `.tbss` 44 bytes
    @36f0a6a9000:36f0a6ae840 RW  `.init_array` 22592 bytes
    @36f0a6ae840:36f0a6ae840 RW  `.fini_array` 0 bytes
    @36f0a6ae840:36f0a6ae898 RW  `.data.rel.ro` 88 bytes
    @36f0a6ae898:36f0a6ae8b0 RW  `.got` 24 bytes
    @36f0a6af000:36f0a6af018 RW  `.got.plt` 24 bytes
    @36f0a6b0000:36f0a6b0080 RW  `.invis` 128 bytes
    @36f0a6b1000:36f0a6bd3f8 RW  `.data` 50168 bytes
    @36f0a6bd400:36f0af84ea8 RW  `.bss` 9206440 bytes
[...]

PR #3639, opened 2023-04-18
Workaround crashes in MAME's zlib decompression code

CasualPokePlayer:
"Fixes" #3615

### DIFF
--- a/src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
+++ b/src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
@@ -139,27 +139,29 @@ namespace BizHawk.Emulation.Cores.Arcades.MAME
 					// this is to get around some zlib bug?
 					using var ret = new MemoryStream();
 					ret.Write(rom.FileData, 0, rom.FileData.Length);
-					using var zip = new ZipArchive(ret, ZipArchiveMode.Update, leaveOpen: true);
-					foreach (var entryName in zip.Entries.Select(e => e.FullName).ToList())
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Do we need to put it back into a zip? MAME can load from a folder if it's named correctly. If it's possible to send it such folders it would be cool.

CasualPokePlayer:
If the zip contains mixed deflate and lzma compression, then this code would be needed (or bring in SharpCompress into the Emulation.Cores project), as ZipArchive doesn't support anything other than deflate.

src/BizHawk.Emulation.Cores/Arcades/MAME/MAME.cs
```diff
@@ -129,15 +131,50 @@ private void StartMAME(List<IRomAsset> roms)
...
+							using var newEntryStream = newEntry.Open();
+							newEntryStream.Write(contents, 0, contents.Length);
+						}
+						catch
+						{
+							// ignored
+						}
+					}
+
+					zip.Dispose(); // this actually modifies the MemoryStream
```
nattthebear, line 162: Style nit --  this would be better as a `using` block

CasualPokePlayer: Dispose needs to be called before ret.ToArray() is done. In hindsight, I would suppose another pair of parens would work for that.

nattthebear: This is a using block, as opposed to a using declaration:

```C#
using (var zip = new ZipArchive(ret, ZipArchiveMode.Update, leaveOpen: true))
{
  // ...
}
```

Issue #3565, opened 2023-02-21
kinst + kinst2 = crash


BrunoRaji:
I know it's an experimental core but the only CHD games that won't open are Killer Instincts
SF3 + Naomi, OK

5804ff04

Best regards,

nattthebear:
What error message are you getting?

BrunoRaji:
none, the emulator closes
yesterday I was doing some tests and I saw that other "Ultra/Nintendo 64" games don't open either, Cruis'n USA/World/Exotica, Off Road Challenge, but The Grid* opens.

* - https://system16.com/hardware.php?id=617

crusnusa
crusnwld
crusnexo
offroadc

CasualPokePlayer:
Keep in mind our MAME core is Arcade only, if you shove non-Arcade crap in it it will reject it.

Considering there's some silent emulator crash I suspect it's the same issue with other games I've noted which have some crash due to some getpid() call *somewhere* (and that crashing since the syscall for that is unimplemented, although the places where it possibly gets used would likely just crash later on so implementing pid syscalls probably won't fix anything)

nattthebear:
We can easily stub out syscalls like getpid to return some constant dummy value, but from what I've seen in the past it's not useful because there's nothing meaningful the guest application can do with it.  Usually syscalls like that are the first step in some high level call that we can't possibly support, such as `abort()`.  The real fix is to figure out what the guest is trying to do and fix it.

I could see improving our own abort situation here, though.  When the guest crashes internally we might be able to at least recover back to the host, although that guest would be permanently spoiled.  We do have recovery of this sort already when we throw inside a callback to managed that was passed to the guest.

YoshiRulz:
Would that be useful without a multi-instance-capable host?

nattthebear:
One goal is to simply give better error reporting and/or host error recovery.  Pop a reasonable error dialog, save the movie, etc.

An unimplemented syscall in the guest in principle shouldn't cause any direct problem for Bizhawk.  We can safely tear down the guest without knowing too much about its internal state.

CasualPokePlayer:
For MAME specifically it could potentially be more tricky, as emulation runs in a separate cothread. Other libco cores could have similar complexities.

CasualPokePlayer:
[...]

PR #3641, opened 2023-04-19
Waterbox host:  Bare minimum implementation of getpid/getppid

nattthebear:
This isn't that great because some things may assume pid 1 == init and pid 0 == sched, but you also want pid == tid (so long as you only have one thread), so I'd have to make some updates there and I haven't touched the threading code in a long time.

Hopefully it's good enough for anything that's just looking for pid as a """random value""".


nattthebear:
Does this fix https://github.com/TASEmulators/BizHawk/issues/3565?

CasualPokePlayer:
This does fix the NR_GETPID crash.

However, it seems that there is another crash after it, where NR_MKDIR is done. This isn't unknown anyways, some chd games were known to crash with this rather than NR_GETPID. The reason is MAME is attempting to make a dif directory to place `<chd game>.dif`. For chd games which don't do this, this PR allows them to boot successfully.

The NR_MKDIR crash can be avoided with C# changes by telling MAME to place the dif directory in `""` instead of `"?"` (the latter being done is a remnant of pre-waterbox MAME where that was an invalid directory intended to prevent the directory being made).

nattthebear:
Great, merged.  I'd prefer not to implement mkdir without taking a holistic look at virtual file system ideas, so yes it's best that we avoid that for now.

PR #3684, opened 2023-06-17
Sanitize text pasted into hex text boxes

### LINKED ISSUE #3682: [F.Rq.] Allow pasting `0xHHHH`-formatted numbers in a `HexTextBox` (e.g. in `WatchEditor`)
Went to add Ctrl+V check to keydown handler, but I have no idea what these magic numbers are. [Official docs](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.keypresseventargs.keychar?view=netframework-4.8#remarks) ([current](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.keypresseventargs.keychar?view=windowsdesktop-8.0#remarks)) were no help, nor was StackOverflow.
https://github.com/TASEmulators/BizHawk/blob/13dd04a026855f9ef087254114565eddfd6522b0/src/BizHawk.Client.EmuHawk/CustomControls/HexTextBox.cs#L70-L75

kalimag:
- Add `ClipboardEventTextBox` control with an `OnPaste` method and make  `HexTextBox` and `WatchValueBox` inherit from that

- Override normal paste behavior to trim `0x`/`$`/whitespace from pasted text

- Prevent pasting of non-hex text

Resolves  #3682

### DIFF
--- a/src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
+++ b/src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Windows.Forms;
+using BizHawk.Common;
 
 namespace BizHawk.Client.EmuHawk.CustomControls
 {
@@ -9,30 +10,46 @@ namespace BizHawk.Client.EmuHawk.CustomControls
 		{
 			const int WM_PASTE = 0x302;
 
-			if (m.Msg == WM_PASTE)
+			if (m.Msg == WM_PASTE && !OSTailoredCode.IsUnixHost)
 			{
-				bool containsText;
-				string text;
-
-				try
+				if (OnPasteInternal())
 				{
-					containsText = Clipboard.ContainsText();
-					text = containsText ? Clipboard.GetText() : "";
-				}
-				catch (Exception)
-				{
-					// Clipboard is busy? No idea if this ever happens in practice
 					return;
 				}
-
-				var args = new PasteEventArgs(containsText, text);
-				OnPaste(args);
-
-				if (args.Handled)
-					return;
 			}
 
 			base.WndProc(ref m);
+		}
+
+		protected override bool ProcessCmdKey(ref Message m, Keys keyData)
+		{
+			if (OSTailoredCode.IsUnixHost && keyData is (Keys.Control | Keys.V) or (Keys.Shift | Keys.Insert) && !ReadOnly)
+			{
+				return OnPasteInternal();
+			}
+
+			return base.ProcessCmdKey(ref m, keyData);
+		}
+
+		private bool OnPasteInternal()
+		{
+			bool containsText;
+			string text;
+
+			try
+			{
+				containsText = Clipboard.ContainsText();
+				text = containsText ? Clipboard.GetText() : "";
+			}
+			catch (Exception)
+			{
+				// Clipboard is busy? No idea if this ever happens in practice
+				return true;
+			}
+
+			var args = new PasteEventArgs(containsText, text);
+			OnPaste(args);
+			return args.Handled;
 		}
 
 		protected virtual void OnPaste(PasteEventArgs e)
--- a/src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
+++ b/src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Windows.Forms;
+
+namespace BizHawk.Client.EmuHawk.CustomControls
+{
+	public class ClipboardEventTextBox : TextBox
+	{
+		protected override void WndProc(ref Message m)
+		{
+			const int WM_PASTE = 0x302;
+
+			if (m.Msg == WM_PASTE)
+			{
+				bool containsText;
+				string text;
+
+				try
+				{
+					containsText = Clipboard.ContainsText();
+					text = containsText ? Clipboard.GetText() : "";
+				}
+				catch (Exception)
+				{
+					// Clipboard is busy? No idea if this ever happens in practice
+					return;
+				}
+
+				var args = new PasteEventArgs(containsText, text);
+				OnPaste(args);
+
+				if (args.Handled)
+					return;
+			}
+
+			base.WndProc(ref m);
+		}
+
+		protected virtual void OnPaste(PasteEventArgs e)
+		{ }
+
+		/// <summary>
+		/// Paste <paramref name="text"/> at selected position without exceeding the <see cref="TextBoxBase.MaxLength"/> limit.
+		/// The pasted string will be truncated if necessary.
+		/// </summary>
+		/// <remarks>
+		/// Does not raise <see cref="OnPaste"/>.
+		/// </remarks>
+		public void PasteWithMaxLength(string text)
+		{
+			if (MaxLength > 0)
+			{
+				int availableLength = MaxLength - TextLength + SelectionLength;
+				if (text.Length > availableLength)
+				{
+					text = text.Substring(0, availableLength);
+				}
+			}
+			Paste(text);
+		}
+
+		protected class PasteEventArgs : EventArgs
+		{
+			public bool ContainsText { get; }
+			public string Text { get; }
+			/// <summary>Prevents regular paste handling if set to <see langword="true"/>.</summary>
+			public bool Handled { get; set; }
+
+			public PasteEventArgs(bool containsText, string text)
+			{
+				ContainsText = containsText;
+				Text = text;
+			}
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace BizHawk.Client.EmuHawk.CustomControls
+{
+	public class ClipboardEventTextBox : TextBox
+	{
+		protected override void WndProc(ref Message m)
+		{
+			const int WM_PASTE = 0x302;
+
+			if (m.Msg == WM_PASTE)
```
kalimag, line 12: This still needs to be tested on Linux. I hope Mono emulates this window message?

src/BizHawk.Client.EmuHawk/CustomControls/HexTextBox.cs
```diff
@@ -135,6 +135,21 @@ protected override void OnTextChanged(EventArgs e)
 			base.OnTextChanged(e);
 		}
 
+		protected override void OnPaste(PasteEventArgs e)
+		{
+			if (e.ContainsText)
+			{
+				string text = e.Text.Trim().RemovePrefix("0x").RemovePrefix('$');
```
kalimag, line 142: This could be deduped and perhaps turned into a more robust utility method. Wasn't sure if it was worth it and where to put it, if so.

src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
YoshiRulz: Doesn't work :(

Printing `m.Msg` while doing focus, Ctrl+V, defocus gives: 0x00F, 0x085, 0x007, 0x100, 0x100, 0x102, 0x101, 0x101, 0x008

kalimag: Just focus/paint/key messages it seems.

Could fall back to trapping Ctrl+V on Linux, I suppose.

The benefit of `WM_PASTE` is that it also works when pasting through the native right click menu on Windows.

YoshiRulz: I still don't understand why this Win32 stuff was used when there's a `ValueChange` event. You can even detect copy-paste by checking if the Levenshtein distance between the current and pending values is > 1, though I'm not sure that's necessary.

kalimag: As it is, doing this in a change event is too late because the pasted text will have been truncated by the `MaxLength` already, so if a user pastes `0x1234` you just get `0x12` and can't recover the full text. Trying to indirectly infer copy-paste seems much more hacky and fragile to me.

Of course you could go with Morilli's suggestion and overhaul the controls completely. Get rid of all the input filtering and max length, let the user enter whatever they want, and then write new validation/sanitation logic in the `Validating` event or something.

src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
```diff
@@ -0,0 +1,92 @@
...
+
+			base.WndProc(ref m);
+		}
+
+		protected override bool ProcessCmdKey(ref Message m, Keys keyData)
+		{
+			if (OSTailoredCode.IsUnixHost && keyData is (Keys.Control | Keys.V) or (Keys.Shift | Keys.Insert) && !ReadOnly)
+			{
+				return OnPasteInternal();
+			}
```
kalimag, line 29: Alright, second try. Hopefully this works on Linux.

YoshiRulz: Yep. And as expected, right-click > `Paste` doesn't.

kalimag: Yeah I can't think of anything to do about that without finding a completely different approach. I'm afraid this is as good as my PR will get on Linux

src/BizHawk.Client.EmuHawk/CustomControls/HexTextBox.cs
```diff
@@ -135,6 +135,21 @@ protected override void OnTextChanged(EventArgs e)
...
+		protected override void OnPaste(PasteEventArgs e)
+		{
+			if (e.ContainsText)
+			{
+				string text = e.Text.Trim().RemovePrefix("0x").RemovePrefix('$');
+				if (text.IsHex())
+				{
+					PasteWithMaxLength(text);
+				}
+				e.Handled = true;
```
YoshiRulz, line 147: I don't think you're using `Handled` correctly...

src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
```diff
@@ -0,0 +1,92 @@
...
+		/// <summary>
+		/// Paste <paramref name="text"/> at selected position without exceeding the <see cref="TextBoxBase.MaxLength"/> limit.
+		/// The pasted string will be truncated if necessary.
+		/// </summary>
+		/// <remarks>
+		/// Does not raise <see cref="OnPaste"/>.
+		/// </remarks>
+		public void PasteWithMaxLength(string text)
+		{
+			if (MaxLength > 0)
```
YoshiRulz, line 67: Is this how the BCL treats `MaxLength == 0`?

src/BizHawk.Client.EmuHawk/tools/Watch/WatchValueBox.cs
```diff
@@ -435,6 +437,21 @@ protected override void OnTextChanged(EventArgs e)
 			base.OnTextChanged(e);
 		}
 
+		protected override void OnPaste(PasteEventArgs e)
```
YoshiRulz, line 440: ditto

src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
```diff
@@ -0,0 +1,92 @@
...
+			{
+				int availableLength = MaxLength - TextLength + SelectionLength;
+				if (text.Length > availableLength)
+				{
+					text = text.Substring(0, availableLength);
```
YoshiRulz, line 72: Style nitpicks: Add argument names here, use `var` everywhere, use `is` for enum/const comparison, use `string.Empty`, and make `PasteEventArgs` `sealed`.

src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
```diff
@@ -0,0 +1,92 @@
...
+					return;
+				}
+			}
+
+			base.WndProc(ref m);
+		}
+
+		protected override bool ProcessCmdKey(ref Message m, Keys keyData)
+		{
+			if (OSTailoredCode.IsUnixHost && keyData is (Keys.Control | Keys.V) or (Keys.Shift | Keys.Insert) && !ReadOnly)
```
YoshiRulz, line 26: Move `!ReadOnly` to front

YoshiRulz (changes requested):
adelikat is right, this is better than what we have. Please rebase.

src/BizHawk.Client.EmuHawk/CustomControls/ClipboardEventTextBox.cs
kalimag: WinForms just passes it through to the Windows API, Windows apparently treats it as `Int32.MaxValue - 1`

src/BizHawk.Client.EmuHawk/CustomControls/HexTextBox.cs
kalimag: In what way?

If the text is a valid hex string, manually paste the cleaned string. If it's not hex, it shouldn't be pasted at all. In either case, the default paste handling is suppressed by setting `Handled = true`.

YoshiRulz: On second thought I think that's fine. I'm just used to a `Handled` being used for `event`s, but this is simpler.

src/BizHawk.Client.EmuHawk/CustomControls/HexTextBox.cs
```diff
@@ -15,7 +15,7 @@ public interface INumberBox
 		void SetFromRawInt(int? rawInt);
 	}
 
-	public class HexTextBox : TextBox, INumberBox
+	public class HexTextBox : ClipboardEventTextBox, INumberBox
```
YoshiRulz, line 18: Can the `INumberBox` inheritance be moved to the base class?

kalimag: Should it be? The relevant implementations (`HexTextBox` and `WatchValueBox`) are all different, and while `ClipboardEventTextBox` might never actually be used for anything else, its functionality is not tied to numeric inputs.

kalimag:
`HexEditor` currently uses `NumericStringExtensions.OnlyHex` for its own custom paste handling. This could be unified with the new `NumericStringExtensions.CleanHex`, but it's more broad (e.g. `"0xwhat_ever".OnlyHex() == "0AEE"`).

YoshiRulz (approved):
LGTM. Will merge tomorrow if there are no objections.

PR #3719, opened 2023-08-04
System.Data.SQLite -> Microsoft.Data.Sqlite

CasualPokePlayer:
https://learn.microsoft.com/en-us/dotnet/standard/data/sqlite/compare

Seems to be a sort of successor / "modern variant" of System.Data.SQLite? netstandard2.0 so great for us I suppose. I assume it works for Linux? (System.Data.SQLite didn't lol). A quick testing of commands seems to say it works on Windows at least.

Also would eliminate another .NET Framework dep (see #2261).

Check if completed:


src/BizHawk.Client.Common/Api/Classes/SQLiteApi.cs
```diff
@@ -1,42 +1,45 @@
...
 	{
-		private SQLiteConnection _dbConnection;
+		private SqliteConnection _dbConnection;
 
 		public string CreateDatabase(string name)
 		{
 			try
 			{
-				SQLiteConnection.CreateFile(name);
+				File.Create(name).Dispose();
```
YoshiRulz, line 17: If the equivalent of this in the new library is a no-op, there's no point checking it. We assume paths are good and let them throw on first use.

YoshiRulz (approved):

YoshiRulz:
Oh I probably shouldn't approve without testing on Linux, huh.

src/BizHawk.Client.Common/Api/Classes/SQLiteApi.cs
CasualPokePlayer: This is the equivalent of what the old library did (well, old library did `File.Create(name).Close();` technically, but Dispose() is probably more proper)

YoshiRulz:
Works under Mono on NixOS, without host SQLite lib even.

PR #3723, opened 2023-08-06
Added support for Tools.GeneralUpdate() in MainForm

### LINKED ISSUE #3626: [Feature Request] APIHawk: Allow a way for an external tool to always run Update while the game is paused and without frame advancing.
### Summary
I have written an external tool that requires the use of TAStudio to function. There is a use case where I need to always have the external tool be invoking `IToolFormBase.UpdateAfter()` when the game is paused and when TAStudio is active. 

Given the following source code snippet here:

https://github.com/TASEmulators/BizHawk/blob/03aa420bd5f8ad652203410944689d9eb237f747/src/BizHawk.Client.EmuHawk/MainForm.cs#L3120-L3121

When the game is paused, `runFrame` flag will be `false`, and `force` is typically `false` by default. With `runFrame` and `force` both set to `false`, EmuHawk will not be able to invoke the methods:
* `Tools.FastUpdateBefore()`
* `Tools.UpdateToolsBefore()`
* `Tools.FastUpdateAfter()`
* `UpdateToolsAfter()`

When frame advancing from TAStudio, `force` flag will be set to `true`, and we can invoke any of the 4 methods listed above. But it wouldn't be useful as I have a need to invoke `Tools.FastUpdateBefore()`, `Tools.UpdateToolsBefore()`, and `Tools.FastUpdateAfter()` while the game is paused and without TAStudio frame-advancing the game. 

Thus, we currently don't have a solution. I would like to request for a small modification so that there can be an External Tools API method that will always be invoked even when the game is paused. Perhaps something like:

`Tools.AlwaysUpdate()` right outside of that `if` statement block. It is an empty method, but it will be useful for external tools to override it, and be able to do things even when the game is paused.

### Host env.
- BizHawk 2.9; Win11 64-bit Pro 22H2; Intel/NVIDIA
- BizHawk 2.9.1 dev; Win11 64-bit Pro 22H2; Intel/NVIDIA

tommai78101:
Currently, `ToolManager` does not have `GeneralUpdate()` exposed in the `MainForm` class. I added very minimal changes and added the `GeneralUpdate()` function, so that when BizHawk is running, the emulation is paused, TAStudio is opened, and an external tool is loaded and opened, the external tool can finally be able to update by overriding `ToolFormBase.GeneralUpdate()` and do things in the background without the requirements of having to frame advance or unpause the emulation.

Check if completed:
- [x] I have run any relevant test suites (The test suite is an integration test, with me verifying that the external tool can be invoked by `GeneralUpdate()` when correctly implemented).

Should be able to close #3626 .

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -3281,6 +3281,10 @@ namespace BizHawk.Client.EmuHawk
 				// Tools will want to be updated after rewind (load state), but we only need to manually do this if we did not frame advance.
 				UpdateToolsAfter();
 			}
+			else
+			{
+				Tools.GeneralUpdate();
+			}
 
 			Sound.UpdateSound(atten, DisableSecondaryThrottling);
 		}
--- a/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs
@@ -753,6 +753,17 @@ namespace BizHawk.Client.EmuHawk
 			}
 		}
 
+		public void GeneralUpdate()
+		{
+			foreach (var tool in _tools)
+			{
+				if (tool.IsActive)
+				{
+					tool.UpdateValues(ToolFormUpdateType.General);
+				}
+			}
+		}
+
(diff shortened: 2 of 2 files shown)


YoshiRulz:
Added in https://github.com/TASEmulators/BizHawk/commit/5eb2cd8cb128928fc4fd9af00010a5d74c99adcd.

Issue #3675, opened 2023-06-02
Deleting too many frames is massively inefficient


RetroEdit:
### Summary
I brought this up on the TASVideos Discord server, and it seems like I'm not the only one with this issue.

Basically, deleting a large number of frames takes a disproportionate amount of time. By comparison, even performing a delete that's identical to a truncate takes much longer, while truncate is basically instant.

### Repro
1. Open a game
2. Open TAStudio
3. Insert 99999 frames (I used `Ctrl+Shift+Insert`)
4. Select frames 1-100000
5. Delete those frames (I used `Ctrl+Delete`)

Doing these steps with 50000 frames took me 30 seconds, and 99999 frames took me several minutes. I also did variations where the selected frames were more near the middle of the movie.

### Host env.
- BizHawk 2.4.2; Win11 Home 22H2; Intel/Intel
  - (So probably not a regression.)
- BizHawk 2.9.1; Win11 Home 22H2; Intel/Intel
- BizHawk dev build at 9e5fb35; Win11 Home 22H2; Intel/Intel

Morilli:
As discussed on discord, this is due to a large `RemoveWhere` call on a `SortedSet<Cell>`, which performs terribly on .net framework while working fine on .net6+. We could consider local workarounds for now, but ultimately .net6 would be great as it just solves this itself.

nattthebear:
Bizhawk can never move to .net6, so we'll need to consider local workarounds.

nattthebear:
It's easier to leave an emoji reaction on a comment than to do a full rewrite of the Bizhawk frontend to separate it from winforms.

Morilli:
Of course doing a full UI rewrite is costly and not trivially done in a day. However, it *could* be done.

Also, targeting .NET6 on windows is already possible without issues, we just need to target .net framework for linux.

CasualPokePlayer:
At this moment, we can already just try to migrate immediately with Avalonia given it supports netstandard 2.0. Even then, the other deps can easily be replaced by alternatives (the hardest being Veldrid, but a quick OpenGL only variant would be easy)

Issue #2219, opened 2020-07-10
Waterbox - Don't die in Control Flow Guard on Windows


nattthebear:
If you try running Bizhawk under .net 5 preview (which has CFG enabled by default), or mung the .net 4.8 executable header to turn on flow control, waterbox cores crash.  See https://github.com/dotnet/runtime/issues/39019

This isn't urgent, but if I don't make a ticket I'll never get around to it.

I'm befuddled on what to do; the [documentation ](https://docs.microsoft.com/en-us/windows/win32/devnotes/guardchecklongjumptarget)seems to be lying here as it claims to check the destination of a jump (`rip` presumably?) but in fact is actually checking some validity constraint on `rsp` instead??  Because everything works if no cothreads are in action and stack marshaling is off...

PR #3745, opened 2023-08-17
Multi-target .NET Framework 4.8 and .NET 6

Morilli:
This is a proof-of-concept PR multi-targeting both .NET Framework and .NET 6.
.NET 6 has lots of performance improvements that could immediately benefit us (e.g. #3675).
.NET 6 does not support WinForms under linux, which means that the .NET Framework version can be used under linux and windows, while the .NET 6 build is windows-only.

This is still very much at POC level as there will most likely be unexpected issues like differing APIs, dialogs being visually broken etc. Also, no script and command is currently aware of the output folder change: `output/[everything]` -> `output/net4.8/[everything_net4.8]`+`output/net6.0-windows/[everything_net6.0]`.

Known issues:
[ ] Some waterbox cores (just bsnes?) hardcrash on the first frameadvance. No amount of debug compilation has helped so far, but compiling the waterboxhost with dirty detection disabled fixes this. I'm afraid this may be #3615 all over again.


src/BizHawk.WinForms.Controls/BizHawk.WinForms.Controls.csproj
```diff
@@ -1,10 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>net48</TargetFramework>
+		<TargetFrameworks>net48;net6.0-windows</TargetFrameworks>
 	</PropertyGroup>
 	<Import Project="../MainSlnCommon.props" />
+	<PropertyGroup>
+		<UseWindowsForms>true</UseWindowsForms>
+		<EnableWindowsTargeting>true</EnableWindowsTargeting>
+	</PropertyGroup>
```
YoshiRulz, line 9: This will break compilation on Linux, you'll need to do it conditionally

src/BizHawk.BizInvoke/BizHawk.BizInvoke.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>net48</TargetFramework>
+		<TargetFrameworks>net48;net6.0</TargetFrameworks>
```
YoshiRulz, line 3: This shouldn't be necessary?

Morilli: There are issues with certain hacks in the code and `#if NET_6_0` defines that can sometimes cause missing functions on runtime when not every project has a separate .net framework and .net 6 target, and so to make sure everything works I've multi-targeted every single project instead of just the output projects.

src/BizHawk.WinForms.Controls/BizHawk.WinForms.Controls.csproj
Morilli: I've specifically set `EnableWindowsTargeting` to *fix* compilation on linux. What compilation problems are you seeing? It compiles fine for me.

CasualPokePlayer:
https://github.com/TASEmulators/BizHawk/blob/c79e17cb3c452ac1d82cf6a8adccb27c8ccaed23/src/BizHawk.Bizware.Graphics.Controls/BizHawk.Bizware.Graphics.Controls.csproj#L3
Should be multi-targeted too.

src/BizHawk.WinForms.Controls/BizHawk.WinForms.Controls.csproj
YoshiRulz: I haven't tried. I recall from last time I messed with `hammer-and-duct-tape` that `<UseWindowsForms/>` didn't work w/ the `net48` TFM, and multi-targeting `net6.0-windows` prevented building on Linux—`<EnableWindowsTargeting/>` has obviously been added to fix that, but does it also fix the WinForms references?

Morilli: Well the net4.8 variant runs fine for me in WSL, so I assume everything is good on that front.
From the tests I've done it seemed that the `System.Windows.Forms` reference is functionally equivalent to setting `UseWindowsForms`.

CasualPokePlayer:
https://github.com/TASEmulators/BizHawk/blob/11bd5dac572b96c35bbe704635cbba4d83863e90/src/BizHawk.Client.EmuHawk/app.config#L4
Does this need changing at all here?

YoshiRulz:
No, it's been unused since at least https://github.com/TASEmulators/BizHawk/commit/e435fb807b99cfc9327c01cbfe628d0e4eb4ee21.

src/BizHawk.BizInvoke/BizHawk.BizInvoke.csproj
YoshiRulz: This might be because I've been mistakenly using `NET6_0` instead of `NET6_0_OR_GREATER`...

src/BizHawk.Common/BizHawk.Common.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>netstandard2.0</TargetFramework>
+		<TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
```
YoshiRulz, line 3: Just mentioning that this in only necessary because we have BCL backports which get included in the `netstandard2.0` assemblies. All the other libs can target `netstandard2.0` only and the build succeeds.

YoshiRulz: Relevant, not sure if useful: https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/overview

YoshiRulz:
Is the Waterbox crash mentioned in the OP the same as #2219?

Morilli:
> Is the Waterbox crash mentioned in the OP the same as #2219?

It looks like it, yeah. I didn't know of that issue until now, and it looks like following the workaround to disable control flow guard for the executable makes bsnes load without issues.

src/MainSlnExecutable.props
Morilli, line 1: Honestly I don't remember why we don't do this on master anymore... this structure looks simpler at least.

YoshiRulz:
Should drop redundant `<PackageReference/>`s like so:
```xml
<PackageReference Include="System.Memory" Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'netstandard2.1')) " />
```

The SDK will complain about such redundant references starting with .NET 10.

src/BizHawk.Common/BizHawk.Common.csproj
```diff
@@ -23,4 +23,7 @@
 		<Exec Command="pwsh $(MSBuildProjectDirectory)/../../Dist/install_git_hooks.ps1"
 			IgnoreExitCode="true" /> <!-- i.e. may fail... though it was actually using the exit code properly, but oh well -->
 	</Target>
+	<ItemGroup>
+		<Compile Remove="IsExternalInit.cs" Condition="$(TargetFramework) == 'net8.0'" />
```
YoshiRulz, line 27: Replaced with PolySharp I think; might need to rebase

Morilli:
btw, do we want to extract the `#if`-defs from this PR to master? Given many already exist in the codebase even though they aren't currently hit.

YoshiRulz:
Referring to the rewinder classes? Yeah those can go on master.

src/BizHawk.Common/BizHawk.Common.csproj
Morilli: > All the other libs can target netstandard2.0 only and the build succeeds.

Yes, *the build* succeeds. You will however get hit with `System.MissingMethodException`s immediately upon startup because of `#if`-directives which cause mismatched references between projects which have been compiled with .NET8+ and those that haven't. And removing those isn't possible/trivial because they are usually extension methods that backport existing modern extension functions from modern .NET.

In general I think it's a good idea to compile all projects for modern .NET even if they *could* work with netstandard2.0 only.

YoshiRulz: That's news to me. I thought for extension methods it emits method calls referencing our extension class, which yes would be slower, but functional. Are you sure `MissingMethodException` isn't due to something like a `Range` polyfill?

Morilli: The problem is that when `BizHawk.Common` is compiled using net8.0 but referenced from a project using netstandard2.0, the missing polyfills generate the exception as on runtime, only the net8.0 dll exists. This could be fixed by moving the polyfills out of `BizHawk.Common` or not compiling it with net8.0 target (which generates `IsExternalInit` errors and probably many more if you somehow fixed those), but at that point why are we even targeting net8.0 in the first place when no project is allowed to target it?

Issue #2292, opened 2020-08-13
TAStudio records extra frames at the end during playback


RetroEdit:
In BizHawk 2.4.2 and the latest dev build, there's a very odd difference between TAStudio's behavior and tradional movie recording.

If you create a new movie in TAStudio, and then export to bk2, it will be a 0 frame movie exported.

If you then advance one frame, and export to bk2, it will be a two frame movie. This shouldn't happen; it should be a one frame movie. For some reason, TAStudio will create two new frames.

(I do not think this occurs in Recording Mode for TAStudio, though I need to test that again, but this should not occur at all.)

----

This behavior is not intended, because it was a regression introduced in 2.2.2. There is a similar behavior, where TAStudio always has a blank line at the end. For instance, if you click a button on the last line of TAStudio, it will add a single line afterwards. However, this is a distinct behavior, and only adds *one* frame, not two.

Also comparable, as mentioned above, TAStudio always starts with a single line, which actually translates into a 0-frame movie. And this seems like a sensible behavior to me, since nothing has occurred in the movie.

Ideally, TAStudio would always shows one more input line than the size of the movie's input log. This bug breaks that, and limits the ability to control TAStudio's playback precisely. For instance, this bug is a blocker for TAStudio being able to (conveniently) calculate the cycle count in Game Boy movies. *Edit* (2024-06-24): Yes, you can seek to the desired "last" frame and truncate, and that could save cycle count if saving cycle count in TAStudio worked, but ideally that wouldn't even be necessary.

----

I am going to try and address this in the [moviesession-cleanup](https://github.com/TASVideos/BizHawk/compare/moviesession-cleanup) branch.

----

This bug was introduced between 2.2.1 and 2.2.2. Specifically, it was introduced in e99ea11cf7e18309a05d1c58d38ca13dea808120.

nattthebear:
Take a look around the backlog, this sounds hauntingly similar to something else I remember reading recently

vadosnaprimer:
> For instance, this bug is a blocker for TAStudio being able to calculate the cycle count in Game Boy movies.

The blocker is the fact that TAStudio can't know the count until the frame has been emulated, and the input being send is drawn on the previous frame. Changing which visible frame on the roll it should consider last frame of the bk2 won't be easy, but it may help resolving some confusion, for example if your last button is on frame 100, the final frame count of the movie will be 101, which is the frame when the input actually arrives to the core.

RetroEdit:
[...]

PR #3747, opened 2023-08-18
Set inputdisplay correctly on LoadState

### LINKED ISSUE #3741: Input display doesn't update when loading states in read-only mode
@adelikat pointed this out on Discord: when you load a state in read-only mode, it is supposed to update the input display with the last input pressed from the movie.

I can confirm this doesn't display anything in a recent dev build, and I believe the intuitive behavior would be to update the input display, just as advancing frames in movie playback during read-only mode will show inputs.

It also doesn't display anything in 2.4.2, 2.0, or 1.8.1, suggesting that this is possibly how BizHawk has always behaved. *Edit*: But FCEUX does appear to have the behavior of displaying inputs when loading savestates.

### Host env.
- BizHawk dev build at c10d2927 (post-2.9.1)

Morilli:
Resolves #3741.

Previously loading a state would either not update the input display (in case of TAStudio) or make the input display display the current frame's inputs, not the inputs of the previous frame (outside of TAStudio).

I'm honestly unsure if this can break anything so I'm making a PR instead of pushing to master directly.

### DIFF
--- a/src/BizHawk.Client.Common/movie/MovieSession.cs
+++ b/src/BizHawk.Client.Common/movie/MovieSession.cs
@@ -154,7 +154,8 @@ namespace BizHawk.Client.Common
 				}
 				else if (Movie.IsPlayingOrFinished())
 				{
-					LatchInputToLog();
+					int previousFrame = Movie.Emulator.Frame - 1;
+					Movie.Session.MovieController.SetFrom(Movie.GetInputState(previousFrame));
 				}
 				else if (Movie.IsFinished())
 				{
--- a/src/BizHawk.Client.Common/movie/MovieSession.cs
+++ b/src/BizHawk.Client.Common/movie/MovieSession.cs
@@ -154,6 +154,7 @@ namespace BizHawk.Client.Common
 				}
 				else if (Movie.IsPlayingOrFinished())
 				{
+					// set the controller state to the previous frame for input display purposes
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.Common/movie/MovieSession.cs
```diff
@@ -154,7 +154,8 @@ public bool HandleLoadState(TextReader reader)
 				}
 				else if (Movie.IsPlayingOrFinished())
 				{
-					LatchInputToLog();
+					int previousFrame = Movie.Emulator.Frame - 1;
+					Movie.Session.MovieController.SetFrom(Movie.GetInputState(previousFrame));
```
YoshiRulz, line 158: Only thing I can think of is saving and loading after the end of a movie, does that break?

Morilli: That doesn't seem to be possible in the first place, so I assume that's safe.

RetroEdit:
Oh no... this changes some code that has been reliably the same since 2020 that I was recently looking at again (I ideally want to fix #2292).

I do probably approve of replacing that `LatchInputToLog()` log call, though... but the code is so complicated that I don't necessarily understand all the implications currently.

Morilli:
Well, I'm just going to go ahead with this for now. We can always revert if something breaks.

PR #3754, opened 2023-08-20
Standardize and Optimize Code with dotnet format

TiKevin83:

Follow up on https://github.com/TASEmulators/BizHawk/pull/3681

Proposes merging the corrected output of `dotnet format style -v detailed --severity info`, optimizing and/or simplifying code per dotnet standards and removing unused code.

Implements or standardizes the following:

- destructuring assignment including in foreach initialization and in combination with tuples for concurrent assignment operations, removing the need for temp variables
- shorthand new() that doesn't repeat the variable type
- new switch syntax
- new using syntax
- new local function syntax
- `|=`, `&=`, `>>=`, `<<=`, `++`, `--`, `+=`, `-=`, `%=`, `??=` operator usage
- consistent readonly usage
- removes unused/commented out code
- lots of simplified boolean logic
- use of the `is` and `is not` keywords in favor of `==` and `!==` where relevant
- array access from end with [^]
- array range operator with [..] (also simplifies many substring uses)
- optional chaining with `?.`

potential outlying changes:
There's a change to RCheevos.cs that looks possibly intended to be the existing way
Some removals of unused code in the Amstrad CPC, Commodore 64, PC Engine, and Sinclair Spectrum CPU cores may not make sense (it looks like a bunch of cpu registers end up unused)

Check if completed:


YoshiRulz (changes requested):
This would be a good idea to enable—*after* the codebase has been given a manual once-over. This way is like seeing a dead canary and deciding to pump clean air into the birdcage.

TiKevin83:
> This would be a good idea to enable—_after_ the codebase has been given a manual once-over. This way is like seeing a dead canary and deciding to pump clean air into the birdcage.

I don't think the order is that crucial. I did look briefly before and found I can definitely improve the codebase with a manual review in ways that the auto checks can't find with things like LINQ expressions, so I could work on that first though if preferred. This PR also doesn't strictly enforce following dotnet format as a pre-commit or CI check which could be desired if the manual combing is also in place.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
```diff
@@ -1355,19 +1355,6 @@ private void SetUpToolStripColumns()
...
-		{
-			TasView.AllColumns.Clear();
-			SetUpColumns();
-			TasView.Refresh();
-			CurrentTasMovie.FlagChanges();
-
-			MainVertialSplit.SplitterDistance = _defaultMainSplitDistance;
-			BranchesMarkersSplit.SplitterDistance = _defaultBranchMarkerSplitDistance;
-		}
-
```
Morilli, line 1370: Removing this will 99% break stuff. In general I believe it's not good for the format command to remove unused code, as there are multiple valid reasons for why stuff may be "unused" but should still stay.
Is there a way to disable that specifically?

TiKevin83: definitely and I think I what I would do here if we go that way would be to use a `#pragma` directive to disable the rule for those lines in case other people want to have the analyzers on in their IDE or if it gets run in CI/pre commit hooks in the future

Morilli: There should be some way to mark functions as used with attributes like `RestoreDefaults` if I recall correctly, perhaps we should look into that.

TiKevin83: It looks like it's hitting IDE0051 removing unused private members and that warning still hits when marked with the attribute per other issues

https://github.com/dotnet/roslyn/issues/34986

I think you have to wrap it with the pragma directive to ignore IDE0051

TiKevin83: I've updated the changes with IDE0051 warnings pragma'ed

tommai78101:
Would it be helpful to add some info about the formatting style to the CONTRIBUTING file, and also the suggested / recommended way of checking for the formatting style based on the outputs of the `dotnet` style? https://github.com/TASEmulators/BizHawk/blob/master/contributing.md

TiKevin83:
> Would it be helpful to add some info about the formatting style to the CONTRIBUTING file, and also the suggested / recommended way of checking for the formatting style based on the outputs of the `dotnet` style? https://github.com/TASEmulators/BizHawk/blob/master/contributing.md

Definitely would be helpful to add that to Contributing however "format" is a bit misleading here, this PR is less intended to change formatting standards like indent style and more making use of new language features where the format command happens to know how to implement them automatically.

tommai78101:
Ok, I look forward to seeing the new info added to CONTRIBUTING.

src/BizHawk.Client.EmuHawk/RetroAchievements/RCheevos.cs
```diff
@@ -414,7 +414,7 @@ public override void Restart()
 
 			// validate addresses now that we have cheevos init
 			// ReSharper disable once ConvertToLocalFunction
-			LibRCheevos.rc_runtime_validate_address_t peekcb = address => _readMap.ContainsKey(address);
+			bool peekcb(int address) => _readMap.ContainsKey(address);
```
CasualPokePlayer, line 417: This should not be changed. Passing in `peekcb` as a local function will create a Delegate in _lib.rc_runtime_validate_addresses, which actually allows it to be GC'd. Doing it as a local variable prevents that.

src/BizHawk.Client.EmuHawk/RetroAchievements/LibRCheevos.cs
```diff
@@ -104,8 +104,8 @@ public struct rc_api_request_t
 			public IntPtr post_data;
 			public rc_api_buffer_t buffer;
 
-			public string URL => Mershul.PtrToStringUtf8(url);
-			public string PostData => Mershul.PtrToStringUtf8(post_data);
+			public readonly string URL => Mershul.PtrToStringUtf8(url);
```
CasualPokePlayer, line 107: These are properties (=> means getter only), so this should just be completely invalid and error on build? How is this even valid (and what would it even mean here?)

CasualPokePlayer (changes requested):

src/BizHawk.Client.EmuHawk/RetroAchievements/LibRCheevos.cs
TiKevin83: I believe that the => getter only syntax does not actually enforce that the getter does not mutate state, it is theoretically possible though terrible practice to mutate state inside a getter. The readonly keyword actually enforces that the property is not modified at the compiler level.

src/BizHawk.Client.EmuHawk/RetroAchievements/RCheevos.cs
TiKevin83: agree I'll revert and pragma this

src/BizHawk.Client.EmuHawk/RetroAchievements/LibRCheevos.cs
Morilli: How would the property ever get "modified" anyway? Isn't `=>` basically shorthand for `{ get { return ...} }`?

CasualPokePlayer: If I were to modify a property in a class, I would get "Property cannot be 'readonly'".
image
However, if it's in a struct, somehow it is allowed?
image

CasualPokePlayer: Seems like it may be due to this: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#readonly-instance-members?

Morilli: So I guess making struct properties readonly is like a guarantee to the caller that the struct itself won't get modified by the call... probably fine, although it shouldn't be required for our code.
Maybe making the entire struct readonly instead would be even better in this case?

CasualPokePlayer: The structs in this case shouldn't be readonly as they get passed by ref in interop.

src/BizHawk.BizInvoke/MemoryBlockLinuxPal.cs
```diff
@@ -43,19 +43,14 @@ public void Dispose()
...
-				case Protection.RX:
-					return MemoryProtection.Read | MemoryProtection.Execute;
-				default:
-					throw new ArgumentOutOfRangeException(nameof(prot));
-			}
+				Protection.None => MemoryProtection.None,
+				Protection.R => MemoryProtection.Read,
+				Protection.RW => MemoryProtection.Read | MemoryProtection.Write,
+				Protection.RX => MemoryProtection.Read | MemoryProtection.Execute,
+				_ => throw new ArgumentOutOfRangeException(nameof(prot)),
```
Morilli, line 52: This and a couple of similar changes are now giving an analyzer error `BHI1005`:
image
@YoshiRulz thoughts?

src/BizHawk.Client.EmuHawk/RetroAchievements/LibRCheevos.cs
YoshiRulz: Yes, this is a simple way of marking a function (prop getter) as pure. Why it's not allowed in classes I have no idea.

src/BizHawk.BizInvoke/MemoryBlockLinuxPal.cs
YoshiRulz: I think we were going to get rid of that rule since the compiler generates the `_ => throw IOE` branch. That doesn't mean remove all the manually-written ones, or change them all, but this change at least is reasonable.

TiKevin83: yeah it looks like this is erroring because of a custom analyzer rule that is no longer necessary/not accurate in this scenario

src/BizHawk.Client.Common/movie/bk2/Bk2Movie.HeaderApi.cs
```diff
@@ -52,7 +52,7 @@ public virtual bool StartsFromSavestate
 		public bool StartsFromSaveRam
 		{
 			// ReSharper disable SimplifyConditionalTernaryExpression
-			get => Header.TryGetValue(HeaderKeys.StartsFromSaveram, out var s) ? bool.Parse(s) : false;
```
Morilli, line 55: These should be reverted, as even though it's technically a "simplification" it makes the code less readable. The resharper disable is already evident of this.

TiKevin83: I disagree on this one, usually something is less readable/over complicated if you're returning true in the true condition of a ternary or likewise false in the false condition, but if someone is already intentionally ignoring the optimization then sure we can revert it

TiKevin83:
Rebased on Morilli's PR and reran against the analyzer setting changes.  I think something might be wrong with the IDE0007/IDE0008 var vs explicit settings still, will play with that as I have time.

Morilli:
I think the main issue here is that a lot of the inspections at `suggestion` level only really work as just that: A suggestion.

Blanket-applying those everywhere unconditionally may not be desired in all cases.

TiKevin83:
for sure I'll toy with it more, I'll probably rerun it at warning level and retune individual rule warning levels if I see that missing more than I think people would agree on

PR #3755, opened 2023-08-21
Add support for generating LuaCATS definitions for the Lua API

### LINKED ISSUE #3860: Output Lua annotations for editors
Because Bizhawk's Lua functions are created inside Bizhawk, they can't be seen by external editors like VSCode.

The Lua console already has functions for generating code completion tags for Sublime Text/Notepad++, but this code is unfinished and not part of the Lua language. However, there's a better solution.

What would be awesome is an [annotations file](https://luals.github.io/) that could be added/linked into our projects. That way we could use any editor we like and get language support outside the bizhawk Lua console.

kalimag:
Adds the ability to generate API definitions in the LuaCATS format to the Lua Console.

[LuaCATS](https://luals.github.io/wiki/annotations/) is primarily the annotation format of [Lua Language Server](https://luals.github.io/), which drives the most popular VS Code Lua extension and can be used in other editors with language server support.

~~LuaCATS is an extension/fork of the annotation format used by EmmyLua, but I'm not sure what the status of the latter is. These definitions might be compatible with EmmyLua, but the only spec of its format I could find is [sparse and partially in Chinese](https://emmylua.github.io/annotation.html).~~ EmmyLua extension now uses LuaCATS annotations, too.

The generated definitions could be further improved, e.g. enumerating valid values for `surfaceName` parameters, specifying the structure of certain `table` values, and callback function signatures. But that would require extending the API metadata and I wanted to get some feedback first.

Current output:
- [BizHawk Dev Builds](https://gist.github.com/kalimag/901d9e076e74668611f5872e9e48b87d)
- [BizHawk 2.11](https://gist.github.com/kalimag/a45b110239d75aa4a1ffa88c6a90cc8c)

Check if completed:

will resolve #3860


kalimag:
Just for example, some issues LuaLS can detect in the bundled scripts using these definitions:

image

image

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,174 @@
+using System;
```
vadosnaprimer:
```suggestion
```

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -1184,6 +1184,18 @@ private void RegisterNotePadMenuItem_Click(object sender, EventArgs e)
...
 
+		private void GenerateLuaCatsDefinitionMenuItem_Click(object sender, EventArgs e)
+		{
+			string initDir = !string.IsNullOrWhiteSpace(LuaImp.ScriptList.Filename)
+				? Path.GetDirectoryName(LuaImp.ScriptList.Filename)
+				: Config!.PathEntries.LuaAbsolutePath();
+
+			if (this.ShowFileSaveDialog(initDir, initFileName: "BizHawk.lua", fileExt: ".lua", filter: JustScriptsFSFilterSet) is string path)
+			{
+				File.WriteAllText(path, LuaImp.Docs.ToLuaLanguageServerDefinitions());
```
vadosnaprimer:
```suggestion
				File.WriteAllText(path, LuaImp.Docs.ToLuaLanguageServerDefinitions());
```

vadosnaprimer:
Let's move forward with this one!

I decided to move from notepad++ to vscode for lua coding after 15 years, and ended up making [exactly the same thing](https://github.com/vadosnaprimer/BizHawk/commit/8959b25e7ddc35fddad05da6b04be454c297a994) independently yesterday, I didn't even remember this PR! And here it's done in a much nicer way, so let's have it!

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,187 @@
...
+				if (func.Example != null)
+				{
+					sb.AppendLine("---");
+					sb.AppendLine("---Example:");
+					sb.AppendLine("---");
+					sb.AppendFormat("---\t{0}", func.Example
+							.Replace("\r\n", "\r\n---\t")
+							.Replace("{{", "`")
+							.Replace("}}", "`"));
+					sb.AppendLine();
```
YoshiRulz: Examples break out of the format. Try
```csharp
if (func.Example != null)
{
	sb.AppendFormat(
		"---\n---Example:\n---\t{0}\n",
		func.Example.Replace("{{", "`").Replace("}}", "`")
			.ReplaceLineEndings("\n---\t"));
}
```

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,187 @@
...
+
+	public string Generate(LuaDocumentation docs)
+	{
+		var sb = new StringBuilder();
+
+		sb.AppendLine($"--Generated with BizHawk {VersionInfo.MainVersion}");
+
+		sb.AppendLine(Preamble);
+
+		foreach (var libraryGroup in docs.GroupBy(func => func.Library).OrderBy(group => group.Key))
```
YoshiRulz, line 101: Split to multiple files? https://luals.github.io/wiki/definition-files/#using-definition-files suggests you can.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,187 @@
...
+--- - A CSS3/X11 color name e.g. `""blue""`, `""palegoldenrod""`
+--- - Color created with `forms.createcolor`
+---@alias luacolor integer | string | color
+";
+
+	public string Generate(LuaDocumentation docs)
+	{
+		var sb = new StringBuilder();
+
+		sb.AppendLine($"--Generated with BizHawk {VersionInfo.MainVersion}");
```
YoshiRulz:
```suggestion
		sb.AppendLine($"--Generated with EmuHawk {VersionInfo.MainVersion}");
```

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -1184,6 +1184,18 @@ private void RegisterNotePadMenuItem_Click(object sender, EventArgs e)
 			_luaAutoInstaller.InstallBizLua(LuaAutocompleteInstaller.TextEditors.NotePad, LuaImp.Docs);
 		}
 
+		private void GenerateLuaCatsDefinitionMenuItem_Click(object sender, EventArgs e)
+		{
+			string initDir = !string.IsNullOrWhiteSpace(LuaImp.ScriptList.Filename)
+				? Path.GetDirectoryName(LuaImp.ScriptList.Filename)
+				: Config!.PathEntries.LuaAbsolutePath();
+
+			if (this.ShowFileSaveDialog(initDir, initFileName: "BizHawk.lua", fileExt: ".lua", filter: JustScriptsFSFilterSet) is string path)
```
YoshiRulz:
```suggestion
			if (this.ShowFileSaveDialog(initDir, initFileName: "emuhawk.d.lua", fileExt: ".d.lua", filter: JustScriptsFSFilterSet) is string path)
```

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -1184,6 +1184,18 @@ private void RegisterNotePadMenuItem_Click(object sender, EventArgs e)
 			_luaAutoInstaller.InstallBizLua(LuaAutocompleteInstaller.TextEditors.NotePad, LuaImp.Docs);
 		}
 
+		private void GenerateLuaCatsDefinitionMenuItem_Click(object sender, EventArgs e)
+		{
+			string initDir = !string.IsNullOrWhiteSpace(LuaImp.ScriptList.Filename)
+				? Path.GetDirectoryName(LuaImp.ScriptList.Filename)
+				: Config!.PathEntries.LuaAbsolutePath();
```
YoshiRulz: The dir containing the most recent `.luases`? Why?
We should be generating these API docs to be included in releases anyway, so just pick something like `Path.Combine(Config!.PathEntries.LuaAbsolutePath(), "_doc_luacats")`.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,187 @@
...
+
+namespace BizHawk.Client.Common;
+
+/// <summary>
+/// Generates API definitions in the LuaCATS format.
+/// </summary>
+/// <remarks>
+/// See https://luals.github.io/wiki/annotations
+/// </remarks>
+internal class LuaCatsGenerator
```
YoshiRulz: Why is this separate from `LuaDocumentation`, and why is it not `static`?

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,187 @@
...
+		[typeof(LuaFunction)] = "function",
+		[typeof(LuaTable)] = "table",
+		[typeof(System.Drawing.Color)] = "color",
+	};
+
+	private const string Preamble = @"error(""This is a definition file for Lua Language Server and not a usable script"")
+
+---@meta _
+--- Lua functions available for the BizHawk emulator
+--- https://tasvideos.org/Bizhawk
```
YoshiRulz:
```suggestion
--- https://github.com/TASEmulators/BizHawk
```

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,187 @@
...
+
+		sb.AppendLine(Preamble);
+
+		foreach (var libraryGroup in docs.GroupBy(func => func.Library).OrderBy(group => group.Key))
+		{
+			string library = libraryGroup.Key;
+			string libraryDescription = libraryGroup.First().LibraryDescription;
+
+			if (!string.IsNullOrEmpty(libraryDescription))
+				sb.AppendLine(FormatDescription(libraryDescription));
```
YoshiRulz: Braces XOR single-line
edit: more remain

YoshiRulz (changes requested):
After fixing malformed examples, works w/ SumnekoLua plugin in IDEA.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: We want code in examples to look like code, not like this

изображение

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: Too many libs. With one file you can trivially put it where needed without clutter, it's not meant for manual reading anyway.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: Not seeing the need for this. The overarching thing is Bizhawk, it can't be more confusing than Emuhawk.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
vadosnaprimer: Why?

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
vadosnaprimer: It'd still have to be manually put where the extension can absorb it, initial path doesn't matter much. If we just ship it with releases people won't know it's there. Menu item makes it more obvious it's a thing now.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: Either is fine IMO.

vadosnaprimer:
Dang rebasing while I'm still changing those scripts was a silly idea.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: I'm not even looking at rendered (which would depend on your IDE anyway?), I'm looking at the generated `.d.lua`.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
YoshiRulz: [Per docs](https://luals.github.io/wiki/definition-files/#creating-definition-files)

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
YoshiRulz: Moving the menu up for discoverability sounds good, but I don't see why users should have to generate these for themselves.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: Where does too many files cause a problem? When I tried it, I didn't have to enter any paths, it was completely automatic.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaAutocompleteInstaller.cs
```diff
@@ -56,7 +56,7 @@ private bool IsSublimeInstalled()
 		private bool IsNotepadInstalled()
 		{
 			// The most likely location of the app, eventually we should consider looking through the registry or installed apps as a more robust way to detect it;
-			string exePath = @"C:\Program Files (x86)\Notepad++\notepad++.exe";
+			string exePath = @"C:\Program Files\Notepad++\notepad++.exe";
```
YoshiRulz: `%PROGRAMFILES%`.
The proper way would be searching in `%PATH%`, but I recently tried and failed to find what the "correct" way of doing that in .NET was, so maybe hold off on that.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: What's wrong with it?

изображение

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: How did you install it exactly then?

src/BizHawk.Client.EmuHawk/tools/Lua/LuaAutocompleteInstaller.cs
vadosnaprimer: AHAHAHAHA

изображение

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: I think having multiple different output formats in one class is confusing, and my plan was to split this into base class plus separate impls for LuaCATS and EmmyLua if necessary.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaAutocompleteInstaller.cs
kalimag: The Notepad++ stuff isn't implemented anyway.

I looked into it when I started this PR, but as far as I could tell you can't ADD to NPP's autocomplete, you'd have to generate a whole new language definition including Lua's builtin libs.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: I think their definitions would most likely be the same for our rather generic stuff. Maybe it'd be a good idea to rename the menu command to something more generic too...

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: I'm warming up to the idea of making it a folder with files, but I still like the idea of making it all a single file more. Feels easier on all fronts.

kalimag: In the LSS/sumneko case it'll search the working directory automatically and you can define additional import directories in the workspace settings. Don't know about EmmyLua.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaAutocompleteInstaller.cs
vadosnaprimer: Hmmmm https://npp-user-manual.org/docs/auto-completion/#create-auto-completion-definition-files

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
vadosnaprimer: It got loaded when I put it into the dir I opened in vscode (with emmylua) but that's too specific, and nobody opens our entire Lua dir in editors/IDEs I think.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: When I  started working on this, LuaCATS was much more capable than EmmyLua, but [this here](https://github.com/EmmyLuaLs/emmylua-analyzer-rust) lists LuaCATS support as well so apparently that's out of date and they're compatible now? That certainly makes things easier.

vadosnaprimer: Yeah it doesn't look abandoned.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: Figured out the discrepancy: `[LuaMethodExample]` converts to *host* newlines, but you were assuming CRLF here, so that replacement was a no-op for me.
The code has changed so this is no longer relevant.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: I did have "manual reading" in mind, since if you hit "Go to definition" in your IDE it will open that ~5 kLOC file, but discoverability is covered by IDE autocomplete.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: If the other format does differ and you want to add that, would it not be a simple RegEx replacement on top of this one?
At least make it `static`, there's no reason for it to be instantiable.

vadosnaprimer: Again for things we have in annotations it'd work in both (potentially more) LLSs. Our usage of the syntax is minimal.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
kalimag: What would be the preferred way to do this automatically? Manually updating the file(s) before release like the wiki is done, or some kind of build process?

In principle all of this could be done easily enough with a source generator or something, but would need some reworking first so it doesn't need a running EmuHawk.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: >Our usage of the syntax is minimal.

For now. But yeah, seems both implementations are generally compatible now so this is no longer a concern.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
vadosnaprimer: I'm getting some serious problems with finding any info on how to add a universal path for emmylua to look for custom definitions. luals has **several** ways to do that, it can't be that limited or hard...

kalimag: Inconvenient, but seems to work:

1. Create a [`.emmyrc.json`](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/main/docs/config/emmyrc_json_EN.md) (in your user dir or wherever) like this:
```json
{
    "workspace": {
        "library": [
            "D:\\Test\\Lua\\definitions"
        ]
    }
}
```

2. In the EmmyLua extension settings, enter the full path of the `.emmyrc.json` under "Global Config Path".
3. Restart VSC? Not sure if needed.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,234 @@
...
+	---@alias surface
+	---| "emucore" # Draw on the emulated screen. Resolution depends on emulated system and game. Drawing is scaled with the rest of the display.
+	---| "client" # Draw on the BizHawk window. Resolution depends on the window size. Drawing is not scaled.
+	""";
+
+	private static string? GetHardcodedType(ParameterInfo parameter)
+	{
+		// technically any string parameter can be passed a number, but let's just focus on the ones where it's commonly used
+		// like `gui.text` and `forms.settext` instead of polluting the entire API surface
+		if (parameter.Name is "message" or "caption" && parameter.ParameterType == typeof(string))
```
YoshiRulz, line 76: We don't want to encourage this. There are enough footguns in calling `gui.*` already with Lua not having named arguments.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,234 @@
...
+		{
+			return "string | number";
+		}
+
+		if (parameter.Member.DeclaringType == typeof(GuiLuaLibrary) && parameter.Name == "surfaceName" && parameter.ParameterType == typeof(string))
+		{
+			return "surface";
```
YoshiRulz, line 83: We should introduce new attributes if this grows to any more enums.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
```diff
@@ -0,0 +1,234 @@
...
+
+		if (TypeConversions.TryGetValue(type, out luaType))
+		{
+			return luaType;
+		}
+
+		throw new NotSupportedException($"Unknown type {type.FullName} used in API. Generator must be updated to handle this.");
+	}
+
+	private static bool IsNullable(Type type) => type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
```
YoshiRulz: Reference types (i.e. `string` and `byte[]`) are nullable too

src/BizHawk.Client.EmuHawk/tools/Lua/LuaAutocompleteInstaller.cs
```diff
@@ -46,17 +46,19 @@ public void InstallBizLua(TextEditors editor, LuaDocumentation docs)
...
-			string exePath = @"C:\Program Files\Sublime Text 2\sublime_text.exe";
+			string exePath = Path.Combine(ProgramFilesFolder, @"Sublime Text 2\sublime_text.exe");
 			return File.Exists(exePath);
 		}
 
 		private bool IsNotepadInstalled()
 		{
 			// The most likely location of the app, eventually we should consider looking through the registry or installed apps as a more robust way to detect it;
-			string exePath = @"C:\Program Files (x86)\Notepad++\notepad++.exe";
+			string exePath = Path.Combine(ProgramFilesFolder, @"Notepad++\notepad++.exe");
```
YoshiRulz, line 61: Will this be finished in this PR? Because there's no point touching it if it's going to stay broken, just set `Visible = false` on the menu items and it can continue gathering dust.

vadosnaprimer: It'd be silly to fix sublime path and leave this one broken in exactly the same way. No, there's very little value in finishing it :D

> just set `Visible = false` on the menu items

It's always been false

https://github.com/TASEmulators/BizHawk/blob/15d9a778086ae4a6be41f98573c3cae137029ab6/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs#L1199

YoshiRulz: I was referring to this whole file, since I was under the impression that the Sublime path was incomplete [like np++ is](https://github.com/TASEmulators/BizHawk/blob/15d9a778086ae4a6be41f98573c3cae137029ab6/src/BizHawk.Client.EmuHawk/tools/Lua/LuaAutocompleteInstaller.cs#L98), but no it's there. Though it hasn't been meaningfully touched in the last 10 years, so I wonder if it still works? And who uses Sublime Text anyway?

vadosnaprimer: adelikat was using it when he added this and it was working, and with my fix it works for other paths (I have OS on `D:`). These days it's probably not the best option, especially after we support LuaCATS, but it seemed like a proper path fix regardless.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: True, but this is about documenting when `null`/`nil` is valid, and in the Lua API it's generally not unless the parameter is optional.

If the Lua libraries used `#nullabe` annotations that information should be taken into account here, but currently that's not the case (and also I don't think .NET Framework has a reflection API to read that information).

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: Removing this will likely cause a lot of unnecessary warnings

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: All reference types in the Lua API are nullable because Lua is not a .NET language and cannot "see" the NRT metadata.

kalimag: I do not follow. Are you saying want *every*  `string` parameter to be marked `string?` in the definitions?

YoshiRulz: Passing `nil` for a parameter annotated as `string` results in a warning from the language server, so yes they all need to be `string?`. Unless you want to create another attribute and mark all the functions which throw on null arguments.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
YoshiRulz: I think type coercion warnings are good, but then I also consider dynamic languages like Lua to always be inferior to statically-typed languages. Is passing numbers for strings idiomatic in general, or only in concatenation?

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: That's the point, though? `gui.addmessage(nil)` throws an error, so the diagnostics should warn about it. `event.onexit(function() end, nil)` doesn't throw an error, so diagnostics shouldn't warn.

That both are a nullable reference type on the technical level is irrelevant. As far as Lua is concerned, anything goes and you can pass a dozen table arguments to `memory.writebyte` if you want, but the diagnostics should help you catch that before it causes a runtime error.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: Well, number-to-string conversion is specifically a feature in BizHawk's Lua bindings, not a feature of the Lua runtime. That's why LLS warns about it unless the parameter definition allows numbers.

But reading numbers from memory and passing them to gui functions is a pretty common pattern in BizHawk scripts, so I added this special case for those functions to avoid warnings in existing, working scripts. I have a test directory with the bundled Lua scripts and some from the [Lua scripts catalog](https://github.com/TASEmulators/BizHawk-ExternalTools/wiki/Lua-scripts-catalog), and removing this code causes 73 additional warnings, 16 of which are in BizHawk's own scripts ([for example](https://github.com/TASEmulators/BizHawk/blob/72034a2b3ca5f242e36f6209575019dc6220ab6b/Assets/Lua/Genesis/The%20Adventures%20of%20Batman%20and%20Robin.lua#L139-L146)).

I have no strong opinions on deprecating this behavior, but currently it's valid so the definitions say it's valid.

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: (Didn't see the edit)

Currently the assumption for reference types is: parameters cannot be `null` unless they are optional. That seems to work well with the existing API surface.

If there turn out to be exceptions where this doesn't hold true, I think it would be preferable to mark nullable parameters rather than not-nullable ones, and for that I'd prefer to just reuse C# `#nullable` annotations rather than creating a bespoke attribute, if reasonably possible.

YoshiRulz:
Please rebase.
Do you mind if I add `[LuaZeroIndexed]` myself? It could be used in the existing TASVideos wiki markup, and those functions' current documentation make no mention of it. edit: fdec9a3b844d75f2174038efd4a3d419f1b923f5

src/BizHawk.Client.Common/lua/Documentation/LuaCatsGenerator.cs
kalimag: Reference types are now annotated as nullable, unless they are compiled with `#nullable enable` (and not marked as nullable in C#).

Still need to do the legwork of adding `#nullable` to existing Lua libs.

YoshiRulz: Wait they are available with reflection? I was under the impression that you needed `NullabilityInfoContext` from .NET Core.
I'll see about removing the need for `dynamic`.

kalimag: Sure, it just gets rather complicated to correctly interpret the attribute data without `NullabilityInfoContext` once you get into generics and nested arrays and whatever. But for this purpose it's simple enough.

YoshiRulz: Not sure what you mean.
Pushed https://github.com/TASEmulators/BizHawk/commit/227bb47d7f1ef854c2ba99d0cac2c7693c845fd2. It's mostly the same as what you had except it returns `null` in the fallback (for `#nullable disable`), so you may need `?? true` at the callsites.
I'll be adding it into the TASVideos wiki generator too (edit: #4688).
Calling this resolved pending a rebase, and the work of adopting NRTs in the Lua libraries can come later.

YoshiRulz (approved):
Works great, code LGTM. If I don't return to finish #4688 soon enough then by all means merge both PRs. I think nullable `joypad.get` would be the most annoying false positive.

vadosnaprimer:
How soon is "soon enough"? I'd like to release within a week.

SuuperW:
When I hover over a gui.draw method in VS Code to see the documentation, `color` is replaced with `string|integer|dotnetcolor` which I don't think is good. If there's a way for the definition to hint to IDEs that expansion isn't desired I think that'd be good, but as far as I can tell there doesn't seem to be a way to do that.

I can change the settings in the VS Code extension to disable alias expansion at least.

vadosnaprimer:
Maybe making a derived class instead of alias would do what you want, but I personally don't see a problem in it showing the expanded thing.

kalimag:
>I can change the settings in the VS Code extension to disable alias expansion at least.

Yeah I don't think the annotations have any control over that. That's up to the IDE and individual users' settings. EmmyLua doesn't do this by default, FWIW.

> Maybe making a derived class instead of alias would do what you want

No, that wouldn't work because a class can't inherit a union of types.

YoshiRulz:
Can you put `dotnetcolor` first in the union?

PR #3759, opened 2023-08-25
Establish basic style analyzer rules

Morilli:
This establishes some basic style rules that should be displayed in the IDE and (some should) surface as build warnings when `RunAnalyzersDuringBuild` is enabled. Ideally all of these should be agreed upon and followed in future commits.

Should perhaps be applied before #3754 to establish some ground rules. I'd also appreciate feedback on the severity and rules themselves; I've tried to enable rules and choose their severity sensibly but I'm open for differing opinions.

Check if completed:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -57,7 +57,7 @@ dotnet_diagnostic.IDE0042.severity = suggestion
 # Remove unused private member
 dotnet_diagnostic.IDE0051.severity = suggestion
 # Remove unread private member
-dotnet_diagnostic.IDE0052.severity = warning
+dotnet_diagnostic.IDE0052.severity = silent # should be warning imo, but there's too much violation
 # Use compound assignment
 dotnet_diagnostic.IDE0054.severity = warning
 # Use index operator
@@ -106,8 +106,6 @@ dotnet_diagnostic.IDE0241.severity = warning
 dotnet_diagnostic.IDE0250.severity = suggestion
 # Use pattern matching
 dotnet_diagnostic.IDE0260.severity = suggestion
-# Null check can be simplified
-dotnet_diagnostic.IDE0270.severity = warning
 # Use nameof
 dotnet_diagnostic.IDE0280.severity = error
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,20 +11,3 @@ csharp_new_line_before_finally = true
 csharp_indent_switch_labels = true
 csharp_indent_case_contents = true
 csharp_indent_labels = one_less_than_current
-
-# Globalization rules
-dotnet_code_quality.CA1305.excluded_symbol_names = T:System.Byte|T:System.SByte|T:System.Int16|T:System.UInt16|T:System.Int32|T:System.UInt32|T:System.Int64|T:System.UInt64|T:System.String|T:System.Text.StringBuilder|T:System.Convert
-
-dotnet_diagnostic.CA1305.severity = error
-dotnet_diagnostic.CA2101.severity = suggestion
-
-# Performance rules
-dotnet_code_quality.CA1826.exclude_ordefault_methods = true
-
-dotnet_diagnostic.CA1805.severity = silent
-dotnet_diagnostic.CA1822.severity = silent
-dotnet_diagnostic.CA1838.severity = suggestion
-
-# Usage rules
-dotnet_diagnostic.CA1816.severity = none
-dotnet_diagnostic.CA2201.severity = suggestion
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,110 @@ csharp_new_line_before_finally = true
 csharp_indent_switch_labels = true
 csharp_indent_case_contents = true
 csharp_indent_labels = one_less_than_current
+
+# Style rules
+# Can't be in .globalconfig because dotnet format doesn't respect that https://github.com/dotnet/format/issues/1643
+
+# Remove unnecessary cast
+dotnet_diagnostic.IDE0004.severity = warning
+# Remove unnecessary import
+dotnet_diagnostic.IDE0005.severity = warning
+# Use var instead of explicit type
+dotnet_diagnostic.IDE0007.severity = suggestion
+# Use explicit type instead of var
+dotnet_diagnostic.IDE0008.severity = suggestion
+# Inline variable declaration
+dotnet_diagnostic.IDE0018.severity = warning
+# Use pattern matching to avoid as followed by a null check
+dotnet_diagnostic.IDE0019.severity = warning
+# Use pattern matching to avoid is check followed by a cast (with variable)
+dotnet_diagnostic.IDE0020.severity = warning
+# Use expression body for methods
+dotnet_diagnostic.IDE0022.severity = suggestion
+# Use expression body for properties
+dotnet_diagnostic.IDE0025.severity = suggestion
+# Use expression body for indexers
+dotnet_diagnostic.IDE0026.severity = suggestion
+# Use expression body for accessors
+dotnet_diagnostic.IDE0027.severity = suggestion
+# Null check can be simplified
+dotnet_diagnostic.IDE0029.severity = warning
+# Null check can be simplified
+dotnet_diagnostic.IDE0030.severity = warning
+# Use null propagation
+dotnet_diagnostic.IDE0031.severity = warning
+# Use auto property
+dotnet_diagnostic.IDE0032.severity = suggestion
+# Simplify default expression
+dotnet_diagnostic.IDE0034.severity = suggestion
+# Use pattern matching to avoid is check followed by a cast (without variable)
+dotnet_diagnostic.IDE0038.severity = warning
+# Use is null check
+dotnet_diagnostic.IDE0041.severity = warning
+# Deconstruct variable declaration
+dotnet_diagnostic.IDE0042.severity = suggestion
+# dotnet_diagnostic.IDE0049.severity = error # see SA1121
+# Remove unused private member
+dotnet_diagnostic.IDE0051.severity = suggestion
+# Remove unread private member
+dotnet_diagnostic.IDE0052.severity = warning
+# Use compound assignment
+dotnet_diagnostic.IDE0054.severity = warning
+# Use index operator
+dotnet_diagnostic.IDE0056.severity = warning
+# Use range operator
+dotnet_diagnostic.IDE0057.severity = warning
+# Use simple using statement
+dotnet_diagnostic.IDE0063.severity = suggestion
+# Make struct fields writable
+dotnet_diagnostic.IDE0064.severity = error
+# using directive placement
+dotnet_diagnostic.IDE0065.severity = error
+# Use switch expression
+dotnet_diagnostic.IDE0066.severity = suggestion
+# Use System.HashCode.Combine
+dotnet_diagnostic.IDE0070.severity = warning
+# Simplify interpolation
+dotnet_diagnostic.IDE0071.severity = suggestion
+# Use coalesce compound assignment
+dotnet_diagnostic.IDE0074.severity = suggestion
+# Use pattern matching
+dotnet_diagnostic.IDE0078.severity = suggestion
+# Convert typeof to nameof
+dotnet_diagnostic.IDE0082.severity = warning
+# Use pattern matching (not operator)
+dotnet_diagnostic.IDE0083.severity = warning
+# Simplify new expression
+dotnet_diagnostic.IDE0090.severity = suggestion
+# Remove unnecessary equality operator
+dotnet_diagnostic.IDE0100.severity = warning
+# Remove unnecessary discard
+dotnet_diagnostic.IDE0110.severity = warning
+# Simplify LINQ expression
+dotnet_diagnostic.IDE0120.severity = error
+# Namespace does not match folder structure
+dotnet_diagnostic.IDE0130.severity = silent # should be warning imo
+# Use tuple to swap values
+dotnet_diagnostic.IDE0180.severity = suggestion
+# Use UTF-8 string literal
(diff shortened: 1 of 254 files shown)


src/BizHawk.BizInvoke/BizExvoker.cs
```diff
@@ -67,8 +67,7 @@ public DelegateStorage(Type type)
 
 				foreach (var a in methods)
 				{
-					MethodBuilder unused;
-					var delegateType = BizInvokeUtilities.CreateDelegateType(a.Info, a.Attr!.CallingConvention, typeBuilder, out unused).CreateType()!;
+					var delegateType = BizInvokeUtilities.CreateDelegateType(a.Info, a.Attr!.CallingConvention, typeBuilder, out var unused).CreateType()!;
```
YoshiRulz, line 70:
```csharp
var delegateType = BizInvokeUtilities.CreateDelegateType(a.Info, a.Attr!.CallingConvention, typeBuilder, out _).CreateType()!;
```

Morilli: Can do, but that was just the dotnet format.

src/BizHawk.Emulation.Cores/CPUs/FairchildF8/F3850.Registers.cs
```diff
@@ -174,7 +174,7 @@ public bool FlagICB
 		public bool IRQRequest
 		{
 			get => Regs[IRQR] > 0;
-			set => Regs[IRQR] = value == true ? (byte)1 : (byte)0;
+			set => Regs[IRQR] = value ? (byte)1 : (byte)0;
```
TiKevin83, line 177: This probably doesn't matter here because IRQRequest appears to be unused, but I would expect this to either be on one extreme for performance marking the public bool as unsafe and doing a bit hack `set => Regs[IRQR] = *(byte*)&value;` or for readability and safety `set => Regs[IRQR] = Convert.ToByte(value);`

src/BizHawk.Emulation.Cores/Computers/AmstradCPC/Hardware/Display/CRCT_6845.cs
```diff
@@ -868,7 +868,7 @@ private bool ReadRegister(ref int data)
 				case 9:
 				case 10:
 				case 11:
-					if ((int)ChipType == 0 || (int)ChipType == 1)
+					if (ChipType == 0 || (int)ChipType == 1)
```
TiKevin83, line 871: I think I would prefer ChipType == CRCTType.HD6845S || ChipType == CRCTType.HD6845S || ChipType == CRCTType.UM6845R here and in the other cases, the whole point of the enum would be able to say if it's this chip or that chip by name

src/BizHawk.Emulation.Cores/Consoles/Coleco/SN76489col.cs
```diff
@@ -160,14 +160,14 @@ public void generate_sound(int cycles_to_do)
 						if (noise_type)
 						{
 							int bit = (noise & 1) ^ ((noise >> 1) & 1);
-							noise = noise >> 1;
```
TiKevin83, line 163: This if/else should be refactored further:
```
noise = (noise >> 1) | ((noise_type ? (noise & 1) ^ ((noise >> 1) & 1) : noise & 1) << 14);
```
It also kinda looks performance critical since it's doing bit shifting, you could probably optimize it further if you can do an unsafe byte cast on the noise_type because you could then raise noise & 1 to noise_type * true case + (1 - noise_type) * false case and turn the branching into pure math

YoshiRulz: Can we not hyperoptimise cores without profiling please

src/BizHawk.Emulation.Cores/Consoles/GCE/Vectrex/PPU.cs
```diff
@@ -40,8 +40,8 @@ public void tick()
 			{
 				if (skip == 0)
 				{
-					x_pos = x_pos + (x_vel - 128.0) / 256.0 * (vec_scale + 2);
-					y_pos = y_pos - (y_vel - 128.0) / 256.0 * (vec_scale + 2);
+					x_pos += (x_vel - 128.0) / 256.0 * (vec_scale + 2);
+					y_pos -= (y_vel - 128.0) / 256.0 * (vec_scale + 2);
```
TiKevin83, line 44: the operation on these positions could be DRYed by abstracting out to a function

src/BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/GBC_GB_PPU.cs
```diff
@@ -1842,7 +1842,7 @@ public void color_compute_OBJ()
 			uint retG = ((G * 3 + B) << 1) & 0xFF;
 			uint retB = ((R * 3 + G * 2 + B * 11) >> 1) & 0xFF;
 
-			OBJ_palette[OBJ_bytes_index >> 1] = (uint)(0xFF000000 | (retR << 16) | (retG << 8) | retB);
+			OBJ_palette[OBJ_bytes_index >> 1] = 0xFF000000 | (retR << 16) | (retG << 8) | retB;
```
TiKevin83, line 1845: something looks not very DRY here too with 1819 and the surrounding code

src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/Boards/Mapper227.cs
```diff
@@ -64,12 +64,12 @@ public override void WritePrg(int addr, byte value)
 			//maybe only the multicarts do it, to keep the game from clobbering vram on accident
 			//vram_protected = o;
 
-			if (o == true && S == false)	
+			if (o && S == false)
```
TiKevin83, line 67: I'm not sure exactly how I'd do it because it looks like it could get a bit indent pyramid of doom-like but it might be more readable/efficient to group these conditions instead of rechecking the whole condition every if line

src/BizHawk.Emulation.Cores/Consoles/Coleco/SN76489col.cs
TiKevin83: for sure I went a bit overboard on the comments but the initial one liner is probably better readable than the original if split back up over a few lines.

src/BizHawk.Emulation.Cores/Consoles/GCE/Vectrex/PPU.cs
Morilli: This and the things below may be valid, but I don't really wanna refactor the code here, just apply basic style rules.

src/BizHawk.Emulation.Cores/CPUs/FairchildF8/F3850.Registers.cs
Morilli: Yeah I'm gonna say no to `unsafe` optimizations on the microscopic level for an unused property here.

TiKevin83: yeah it looks much preferable to do Convert.ToByte() I just figured it worth mentioning when the code is already using (albeit safe) byte casts

src/BizHawk.Emulation.Cores/Computers/AmstradCPC/Hardware/Display/CRCT_6845.cs
TiKevin83: I think the last comparison may be typoed now, it was comparing for bits 0 or 2 and it's now effectively against 0 1 or 2

src/BizHawk.Emulation.Cores/CPUs/FairchildF8/F3850.Registers.cs
YoshiRulz: These casts are const, and the ternary will probably be optimised by the JITter. (I'd go with the BCL helper still.)

.globalconfig
```diff
@@ -0,0 +1,18 @@
...
+dotnet_code_quality.CA1305.excluded_symbol_names = T:System.Byte|T:System.SByte|T:System.Int16|T:System.UInt16|T:System.Int32|T:System.UInt32|T:System.Int64|T:System.UInt64|T:System.String|T:System.Text.StringBuilder|T:System.Convert
+
+dotnet_diagnostic.CA1305.severity = error
+dotnet_diagnostic.CA2101.severity = suggestion
+
+# Performance rules
+dotnet_code_quality.CA1826.exclude_ordefault_methods = true
+
+dotnet_diagnostic.CA1805.severity = silent
+dotnet_diagnostic.CA1822.severity = silent
```
TiKevin83, line 13: Personally I would use this one at Error level, marking members as static is a safe performance win in C# and it's also good at finding places where there's a bug because the method should actually be referencing something on the class but because it isn't the suggestion is there to make it static.

.globalconfig
```diff
@@ -0,0 +1,18 @@
...
+dotnet_diagnostic.CA1838.severity = suggestion
+
+# Usage rules
+dotnet_diagnostic.CA1816.severity = none
```
TiKevin83, line 17: I agree on leaving this off for now because it hits a ton of lines and looks possibly complex to refactor around, but would prefer a # TODO to turn it on at some point

.editorconfig
```diff
@@ -12,19 +12,107 @@ csharp_indent_switch_labels = true
...
+# Can't be in .globalconfig because dotnet format doesn't respect that https://github.com/dotnet/format/issues/1643
 
-dotnet_diagnostic.CA1305.severity = error
-dotnet_diagnostic.CA2101.severity = suggestion
+# Remove unnecessary cast
+dotnet_diagnostic.IDE0004.severity = warning
+# Remove unnecessary import
+dotnet_diagnostic.IDE0005.severity = warning
+# Use var instead of explicit type
+dotnet_diagnostic.IDE0007.severity = suggestion
```
TiKevin83, line 23: we should prefer one or the other of IDE0007 and IDE0008, IMO at error level to require consistency through the repo but at the least one of them should be preferred with none

.editorconfig
```diff
@@ -12,19 +12,107 @@ csharp_indent_switch_labels = true
...
+dotnet_diagnostic.IDE0038.severity = warning
+# Use is null check
+dotnet_diagnostic.IDE0041.severity = warning
+# Deconstruct variable declaration
+dotnet_diagnostic.IDE0042.severity = suggestion
+# dotnet_diagnostic.IDE0049.severity = error # see SA1121
+# Remove unused private member
+dotnet_diagnostic.IDE0051.severity = suggestion
+# Remove unread private member
+dotnet_diagnostic.IDE0052.severity = silent # should be warning imo, but there's too much violation
```
TiKevin83, line 60: agree but with TODO in the comment

.globalconfig
```diff
@@ -0,0 +1,18 @@
+is_global = true
+
+# Globalization rules
+dotnet_code_quality.CA1305.excluded_symbol_names = T:System.Byte|T:System.SByte|T:System.Int16|T:System.UInt16|T:System.Int32|T:System.UInt32|T:System.Int64|T:System.UInt64|T:System.String|T:System.Text.StringBuilder|T:System.Convert
+
+dotnet_diagnostic.CA1305.severity = error
+dotnet_diagnostic.CA2101.severity = suggestion
```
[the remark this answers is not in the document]
Morilli: Well it mostly mentions security and possible attack vectors, which I don't think we really care about in BizHawk for the most part.
Also, I don't really understand the rule and also don't understand possible implications of applying the suggestion fix, so I kept it at a minimum level for now.

.globalconfig
Morilli: I agree it could be higher, but the amount of current violations and fixes required made me ignore it for now.

Also note that `silent` implies that the rule *should* not be ignored but is currently.

.globalconfig
Morilli: CA1816 is pretty useless imo, it's a nudge to call a method on dispose that is ONLY relevant when:
1. The class has been/will be/could be derived by anything at any point
2. The class itself does not implement a finalizer, but the derived class does. It may also only be relevent when that derived class itself doesn't call `GC.SuppressFinalize`, however I'm not sure about that

In any case, this is relevant for 0% of BizHawk code imho.

.editorconfig
Morilli: IDE0007 and IDE0008 are controlled through the variables `csharp_style_var_for_built_in_types`, `csharp_style_var_when_type_is_apparent` and `csharp_style_var_elsewhere`.
They are all set to `false` by default and `csharp_style_var_when_type_is_apparent` and `csharp_style_var_elsewhere` are configured to `true` below.

So the current severity levels will *suggest* using the explicit type when it's a builtin type and *suggest* using var when the type is apparent or not built-in.

We could probably bump these levels *eventually*, but that will require consensus from all active developers to make sure everyone agrees on these rules 100%, otherwise we'll just get edit wars. As is, great parts of the code are not consistent at all, which makes sense because these style rules have never been established.

.globalconfig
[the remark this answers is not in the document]
TiKevin83: makes sense I'm fine leaving it alone since I don't really understand it either and it's off by default

.editorconfig
TiKevin83: ok this makes more sense now, thanks

.globalconfig
TiKevin83: AH that silent is good to know thanks

TiKevin83 (approved):

PR #3766, opened 2023-09-07
Better Firmware Names

hegyak:
Renaming Firmware Files based on System Name as BizHawk uses or at least renamed to be related to what it's for.

This changes the Auto-Renamed Firmware Files to be named based on the System ID.
Or to be more Descriptive.  So the users can see helpful file names.

The File has been formatted to use LF instead of CRLF.  Which is why it feels like a "Full Replacement"

Examples of Old Name:
cx4.rom
KoF95.bin

Examples of New Name:
SNES_cx4.rom
SAT_KoF95.bin

Also changed the MSX 1.0 BIOS and BASIC vs MSX BIOS and BASIC have their own filenames now instead of a Filename Collision.

Check if completed:
- [X] I have run any relevant test suites

### DIFF
--- a/src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs
+++ b/src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs
@@ -113,8 +113,8 @@ namespace BizHawk.Emulation.Common
 			FirmwareAndOption("A9B124948A19BB5A295EC208FCA8180476BDFEBB", 3145728, "BSX", "Rom_SoundNovel_Tsukuru", "Satellaview_SoundNovel_Tsukuru.sfc", "Sound-Novel Tsukūru Rom");
 			FirmwareAndOption("357AC4826297A6496035E3951CACDA55DCAE4B1B", 4194304, "BSX", "Rom_Tsuri", "Satellaview_Bass_Tsuri.sfc", "Itoi Shigesato no Bass Tsuri No. 1 Rom");
 
-			FirmwareAndOption("79F5FF55DD10187C7FD7B8DAAB0B3FFBD1F56A2C", 262144, "PCECD", "Bios", "PCECD_3.0-(J).pce", "Super CD Bios (J)");
-			FirmwareAndOption("014881A959E045E00F4DB8F52955200865D40280", 32768, "PCECD", "GE-Bios", "PCECD_gecard.pce", "Games Express CD Card (Japan)");
+			FirmwareAndOption("79F5FF55DD10187C7FD7B8DAAB0B3FFBD1F56A2C", 262144, "PCECD", "Bios", "NEC_PCECD_3.0-(J).pce", "Super CD Bios (J)");
+			FirmwareAndOption("014881A959E045E00F4DB8F52955200865D40280", 32768, "PCECD", "GE-Bios", "NEC_PCECD_gecard.pce", "Games Express CD Card (Japan)");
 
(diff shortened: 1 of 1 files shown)


YoshiRulz:
> The File has been formatted to use LF instead of CRLF.

IMO good, but don't include unrelated changes in your PR. [With `-w`](https://github.com/TASEmulators/BizHawk/pull/3766/files?diff=split&w=1), I can immediately see a typo, and it looks like you've reordered all the sections. If you're going to do that, can you make it a separate commit? No-one can review this as-is.

hegyak:
I do not get why the Commit I made is changing the Line Endings.  But I have redone the changes.
They are:
PC-Engine, Vectrex, Lynx, MSX And resolve Name Collision, Odyssey 2 G7400, Gameboy, GameBoy Color.

That unifies the names for all the Firmware.

src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs
```diff
@@ -185,12 +185,12 @@ void AddPatchAndMaybeReverse(FirmwarePatchOption fpo)
...
 			//FirmwareAndOption("E998F0C441F4F1800EF44E42CD1659150206CF79", 16384, "MSX", "bios_pal", "MSX_8020-20bios.rom", "MSX BIOS (Philips VG-8020)");
 			//FirmwareAndOption("DF48902F5F12AF8867AE1A87F255145F0E5E0774", 16384, "MSX", "bios_jp", "MSX_4000bios.rom", "MSX BIOS (FS-4000)");
-			FirmwareAndOption("409E82ADAC40F6BDD18EB6C84E8B2FBDC7FB5498", 32768, "MSX", "bios_basic_usa", "MSX.rom", "MSX BIOS and BASIC");
-			FirmwareAndOption("3656BB3BBC17D280D2016FE4F6FF3CDED3082A41", 32768, "MSX", "bios_basic_usa", "MSX.rom", "MSX 1.0 BIOS and BASIC");
-			FirmwareAndOption("302AFB5D8BE26C758309CA3DF611AE69CCED2821", 32768, "MSX", "bios_basic_jpn", "MSX_jpn.rom", "MSX 1.0 JPN BIOS and BASIC");
+			FirmwareAndOption("409E82ADAC40F6BDD18EB6C84E8B2FBDC7FB5498", 32768, "MSX", "bios_basic_usa", "Microsoft_MSX.rom", "MSX BIOS and BASIC");
+			FirmwareAndOption("3656BB3BBC17D280D2016FE4F6FF3CDED3082A41", 32768, "MSX", "bios_basic_usa", "Microsoft_MSX_1.0.rom", "MSX 1.0 BIOS and BASIC");
+			FirmwareAndOption("302AFB5D8BE26C758309CA3DF611AE69CCED2821", 32768, "MSX", "bios_basic_jpn", "Microsoft_MSX_jpn.rom", "MSX 1.0 JPN BIOS and BASIC");
```
vadosnaprimer, line 193: Microsoft is not a part of the device name.

YoshiRulz:
I'd forgotten about #2204, but this is a follow-up to that.

vadosnaprimer:
I don't have an opinion on cosmetic changes changes that don't change UX.

Morilli:
Okay, unless there's any further disagreement I'll just merge this later then; I've resolved the previous review comments myself.

YoshiRulz (approved):
I still think we should be using filenames found "in the wild", but I too don't really care. Diff LGTM.

Issue #2600, opened 2021-01-30
Unable to fully display two .lua scripts at the same time in 2.6


ghost:
![1](https://user-images.githubusercontent.com/47687575/106359425-08513400-6313-11eb-8590-0f99a1774535.png)
![2](https://user-images.githubusercontent.com/47687575/106359428-0a1af780-6313-11eb-9503-7c121c28c2bf.png)

As shown in the screenshots, in 2.6 it seems i cant run these .lua scripts togheter: whichever i open first will show on screen. So as it is now i have to enable and disable each

No errors on the console, and it looks like part of 1 script, at least some text, shows up: it kinda looks like one overrides the other one, following a layer logic. Same windows sizes, it works without issues till 2.5.2

I'll also leave the lua scripts linked, just in case, idk if it might be useful

Game: CTR NTSC-U (PSX)
Lua Scripts: https://mega.nz/folder/r4lihSzA#u3C3sxXbE8hFsMJOrwgLOQ

YoshiRulz:
I haven't tried, but looking at the code I think there's a simple fix for this. Is the old behaviour (drawing on the same surface, overwriting pixels) something we want to allow?

YoshiRulz:
I think I accidentally fixed this with 8c8d7eab3, please test

CasualPokePlayer:
This should be fixed in the latest dev build.

PR #3775, opened 2023-09-17
Multi script drawing

SuuperW:
This pull request fixes issue #2600.


YoshiRulz:
> [...] PolySharp [...]

We do not use PolySharp, it seems to be a transitive dependency accidentally added to SharpGen (which is a dep of Vortice). Opened SharpGenTools/SharpGenTools#236. We may be able to downgrade Vortice as a workaround.

> Add a reference to BizHawk.Client.EmuHawk [...]

You can't do this. The test project is .NET Core and EmuHawk is .NET Framework.

I'm also not seeing what part of your `MockDisplayManager` references something from EmuHawk. I suggest you move my [`SimpleGDIPDisplayManager`](https://github.com/TASEmulators/BizHawk/blob/0496d45e0903779370acf3fd15269152ee7f503b/src/BizHawk.Tests.Testroms.GB/DummyFrontend.cs#L128) (and `FakeGraphicsControl`) to `BizHawk.Client.Common` and use those.

> [Testing Lua drawing edge cases with a unit test that doesn't involve Lua]

If you can't see what's wrong with this, nothing I say can help you.

SuuperW:
I've force-pushed with new commits, this time Lua scripts are run in tests.

The first several commits are all refactoring so that the code needed for the test can be moved to BizHawk.Client.Common. There should not be any behavior changes except for commit [0c8ce93](https://github.com/TASEmulators/BizHawk/pull/3775/commits/0c8ce9356f54bfde316520f01888457c66060304).

Regarding the SimpleGDIPDisplayManager in BizHawk.Tests.Testroms.GB, that implementation is outdated and does not compile. So I decided not to use it.

SuuperW:
I realized that this broke drawing via external tools (which already wasn't working as semi-documented anyway, but another undocumented way). My new commits address this issue by allowing external tools to use the new BeginFrame and EndFrame methods without having to worry if Lua is also running. External tools + LuaConsole tool can now draw on the same frame.

It looks like none of the external tools in /ExternalToolProjects used the GuiApi. However one of the ones linked in https://github.com/TASEmulators/BizHawk-ExternalTools/wiki/Catalog does, but it uses the obsolete (and now deleted) DrawNew/DrawFinish methods so hasn't worked for a couple years now. It might be a good idea to update documentation to include a working example of drawing (which may be as simple as forking Ecco_BizHawkTool and replacing the obsolete methods with the new ones and linking to that instead).

YoshiRulz:
Here's one of my in-development projects which uses `IGuiApi`: [RetrOCR_wip202309.zip](https://github.com/TASEmulators/BizHawk/files/12843581/RetrOCR_wip202309.zip)

CasualPokePlayer:
https://github.com/CasualPokePlayer/PokemonGBTASTool
Another project which uses the IGuiApi

SuuperW:
Thanks for sharing those. So, I guess we want to keep the old WithSurface behavior to preserve compatibility with these tools and any others people may have made?

I just looked at the Lua library and it does not expose the WithSurface method, meaning it's _only_ available to external tools right now. So we should be able to preserve compatibility by putting calls to BeginFrame and EndFrame inside WithSurface, without breaking anything. We could even give it an optional parameter to control that functionality so external tools could still control the calls to BeginFrame and EndFrame (when LuaConsole isn't open at least).

@CasualPokePlayer Correct me if I missed something, but the only use of IGuiApi I see in that Pokemon tool is using Gui.Text. That is not relevant here since it goes to the OSD instead of rendering like other methods do.

YoshiRulz:
IMO, if you're going to change it, it should be to add a `with_surface` to Lua (or something equivalent, point being to replace the `gui` library, drawing only being available via the callback's parameter†). I would have done that but adelikat complained that old scripts which were 'just drawing' should be supported.

† Which I haven't implemented for ApiHawk yet, but plan to.

SuuperW:
Okay, new force push reverts the changes to IGuiApi. There are now no changes to how external tools draw.

To summarize, the behavior changes in this PR are:
1) Drawing with multiple Lua scripts on the same frame now works. This is done by replacing the LuaAutolockHack and LuaAutoUnlockHack of the GuiApi implementation with BeginFrame and EndFrame methods (which are not on the IGuiApi interface), since the auto hack was fundamentally incompatible with having multiple scripts drawing on the same frame.
2) Drawing with external tools and Lua at the same time now works. This is done by making the new GuiApi methods that LuaLibraries calls recognize when another GuiApi instance has locked/unlocked surfaces.

There is also a fair bit of moving code around and refactoring in order to support the new tests, but this should not have any impact on behavior.

Note that drawing from multiple external tools still does not work. All external tools share API instances and the GuiApi class doesn't know who's calling it, so we cannot support multiple external tools without either changing the API usage or giving each external tool its own GuiApi instance.

I tested a Lua script making thousands of calls to drawPixel per frame and got the same FPS on both master and this branch so performance seems unaffected.

Morilli:
Seeing as #2600 has been resolved, can this be closed?

Issue #3036, opened 2021-12-15
Lua: Replacing the implementation of a library function will affect other scripts (Global variables are shared across scripts)


gocha:
### Summary
Replacing the implementation of a library function will affect other scripts.

### Repro
1. Open the first script 'print_hijacking.lua'
   ```lua
   unaltered_print = print
   print = function(...)
       unaltered_print("Goodbye")
   end
   ```
2. Open the second script 'hello.lua'
   ```lua
   print("Hello world")
   ```

### Output
Expected output of 'hello.lua':
```
Hello world
```

Actual output:
```
Goodbye
```

### Host env.
- BizHawk 2.7; Win10 Pro 20H2; Intel/NVIDIA

gocha:
I was expecting the two scripts to be running in different Lua VMs, is that not correct?

CasualPokePlayer:
Does doing `local print` instead of `print` for the `print` redefinition also have the same result.

gocha:
a.lua:
```lua
name = "Alice_Rabbit"
```

b.lua:
```lua
print(name)
```

Output of b.lua:
```
Alice_Rabbit
```

It seems that scripts can manipulate each others global variables.

gocha:
> I think that's intended?

Hmm... Even so, I don't think this is the way to go.

In my opinion, a script should not change the behavior of another script. If scripts need to communicate with each other, I think they should use another method that is not a global variable.

As an example of other emulators, DeSmuME, for example, does not share global variables between scripts.

YoshiRulz:
Tempted to call this a "feature" because we won't be changing the sandboxing (or lack thereof) anytime soon, and fixing this within the current system would suck...

nattthebear:
I assume someone has written a collection of scripts that rely on this data passing

PR #3780, opened 2023-09-25
Lua no global sharing

SuuperW:
This PR builds on another PR (#3775). (So don't merge this until/unless that one is merged/approved.)

This PR fixes issue #3036. I'm not certain if this is actual desired behavior, since that issue was marked as wontfix but a stated reason for that was "fixing this within the current system would suck". So it's unclear if a fix is wanted, or wasn't done for other reasons.


YoshiRulz:
The fact that globals are shared isn't great, but we'd want to have a replacement merged and field-tested before disabling this.

SuuperW:
I did a force-push. The only change compared to previous commmit is in NamedLuaFunction.DetachFromScript.

> The fact that globals are shared isn't great, but we'd want to have a replacement merged and field-tested before disabling this.

Do you mean that sharing globals should be an option available to the user? so that we can also keep the old behavior until this has been field-tested

CasualPokePlayer:
I don't like this. Especially with running simple commands in the Lua Console, as now it can no longer access global variables in scripts (something I've used before as a way to toggle options on/off or just call manually call functions within the script). Scripts sharing globals as is is just a feature. At most we could just have an option to reset Lua state (without the user needing to close/re-open the console). This PR as-is I will revert if merged.

SuuperW:
> I don't like this. Especially with running simple commands in the Lua Console, as now it can no longer access global variables in scripts (something I've used before as a way to toggle options on/off or just call manually call functions within the script).

I pushed a new commit which may suit your use case. You can now toggle options again if the script is currently highlighted. Please let me know if this is a good solution.

> At most we could just have an option to reset Lua state (without the user needing to close/re-open the console)

I don't see how this would solve any issues. Two incompatible scripts would still be incompatible.

> perhaps a way to open a script with a new lua state (but this not being the default).

I think making it a setting for the user to choose would make sense. Defaulting to preserve old behavior also makes sense.

CasualPokePlayer:
> I don't see how this would solve any issues.
One of the issue of everything sharing globals means if a script has trashed globals you can't restore it by loading up a new script, you need to close the lua console and re-open it.

vadosnaprimer:
If it shouldn't be merged yet, you can convert it to a draft.

Issue #3779, opened 2023-09-24
fm2 movies for FDS games cause an IndexOutOfRangeException


TakuikaNinja:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"
[//]: # "One more thing: if you're on Linux, please comment on #1430 instead of opening an issue so we don't annoy the other devs."

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
Attempting to play an FDS movie converted from fm2 files (drag & drop/import movie) causes an IndexOutOfRangeException. From what I can tell in the source code, this functionality does not appear to be finished.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Start an FDS game. (games tried: SMB2J, Tonkachi Mario)
2. Drag and drop an fm2 file, or import it and then play the converted movie file.
3. "Fatal Error: IndexOutOfRangeException" window appears

### Output
[//]: # "Paste the contents of the error dialog if there is one (try Ctrl+C, it usually works), or paste the output from the Lua Console, or delete this section."
![image](https://github.com/TASEmulators/BizHawk/assets/43363447/1d177655-a71a-468f-b93c-2b2b4335540b)

### Host env.
[//]: # "List the computers you've found the bug with. If there's a version that doesn't have the bug, please put that in too. Here are some examples:"
- BizHawk 2.9, Linux (EndeavourOS)

YoshiRulz:
Full stacktrace when importing https://tasvideos.org/3348M:
```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
  at BizHawk.Client.Common.Bk2Controller.SetFromMnemonic (System.String mnemonic) [0x0005b] in <973fb7a12a68494182a2aa93dd31a9ec>:0 
  at BizHawk.Client.Common.Bk2Movie.GetInputState (System.Int32 frame) [0x0007a] in <973fb7a12a68494182a2aa93dd31a9ec>:0 
[...]

PR #3742, opened 2023-08-16
Refactor InputDisplayGenerator and LogEntryGenerator handling

Morilli:
This PR aims to simplify the `Bk2InputDisplayGenerator` and `Bk2LogEntryGenerator` handling, especially in regards to the many `IController` classes that exist, and improve caching.

Is this safe? It *should* be. Further simplifying the different controller types would help with the clarity here.

Basically, there is now one `InputDisplayGenerator` field on the `InputManager` class that always uses the `ActiveController`'s `ControllerDefinition` as basis. As far as I understand the code, that's all that's required but it was being handled much more complicated before.

Additionally I've added a `ILogEntryController : IController` interface that is now applied to `IMovieController` as well, and just contains one `Bk2LogEntryGenerator` cached as a field on the `ILogEntryController` to be used for log entry generation from that controller, without having to create a new one every time or caching it on a `Bk2Movie` (which was apparently problematic before: f6135abe652fddc5c28e4bc61a36d13a4c801cfd).

PR #3782, opened 2023-09-28
Refactor InputDisplayGenerator + LogEntryGenerator + ControllerDefinition

Morilli:
This PR supersedes / expands #3742.
This also partially (fixes) #3779 (the exception is fixed and the movie imports fine, but the logkey is missing FDS controls and so the movie may misbehave when editing it).

The main goals here are to improve performance (e.g. for the movie importers) by improving caching, simplify code and make the structure better and clearer.

Now, the Bk2LogEntryGenerator and Bk2InputDisplayGenerator classes are entirely static and just take an `IController` as input. They now rely on the `IController::Definition`s `MnemonicCache` to be performant, whereas such a cache was previously built by the classes themselves.
I'm not entirely sure about this cache being on `ControllerDefinition`, but I feel like it does make sense in the current code structure.
The fact that this cache needs to be manually built is kinda ugly. The best solution I can see for that would be to move the `Bk2MnemonicLookup` to a place accessable by `ControllerDefinition` and build the mnemonic cache on `MakeImmutable` (which then needs an additional `systemId` arg).

Oh, I also removed the hacky `SetMovieController` function that was used for initial `default.tasproj` saving, didn't seem to be necessary? ~~this will 100% cause bugs later on~~

Check if completed:


adelikat (approved):

vadosnaprimer:
Please fix the conflicts.

src/BizHawk.Emulation.Common/Base Implementations/ControllerDefinition.cs
```diff
@@ -21,38 +19,48 @@ public class ControllerDefinition
...
 		public readonly string Name;
 
-		public IReadOnlyList<string> OrderedControlsFlat => _orderedControlsFlat ??= ControlsOrdered.SelectMany(static s => s).ToList();
+		private Dictionary<string, char>? _mnemonicsCache;
+		public IReadOnlyDictionary<string, char>? MnemonicsCache => _mnemonicsCache;
+
+		/// <remarks>
+		/// TODO: this should probably be called in <see cref="MakeImmutable"/>,
+		/// but the needed Bk2MnemonicsLookup is in Client.Common
+		/// </remarks>
```
YoshiRulz, line 44: ???
Move the file?

Morilli: iirc I didn't because of semantics: bk2 is a movie thing and I wasn't sure if it was proper for the file to be in BizHawk.Emulation.Common instead. But looking at your commit it makes sense to move it.

PR #3784, opened 2023-10-01
Added soft and hard reset event hooks for apis.

roydmerkel:

Adds api hooks: HardReset and SoftReset inside of IEmuClientApi interface, which allows consumers of APIHawk to detect when Soft Resets and Hard Resets are triggered via EmuHawk UX. This is potentially useful for HUD interfaces counting resets without needing to infer based on specific points in game code and would allow for uniform handling of Hard Resets which won't go through the Reset handler code in the Game Rom for certain systems such as GB and NES.

Check if completed:
- [X] I have run any relevant test suites

### DIFF
--- a/src/BizHawk.Client.Common/EventTypes.cs
+++ b/src/BizHawk.Client.Common/EventTypes.cs
@@ -98,6 +98,32 @@ namespace BizHawk.Client.Common
 	}
 
 	/// <summary>
+	/// This class holds event data for HardReset event
+	/// </summary>
+	public sealed class HardResetEventArgs : EventArgs
+	{
+		/// <summary>
+		/// Initialize a new instance of <see cref="HardResetEventArgs"/>
+		/// </summary>
+		public HardResetEventArgs()
+		{
+		}
+	}
+
+	/// <summary>
+	/// This class holds event data for SoftReset event
+	/// </summary>
+	public sealed class SoftResetEventArgs : EventArgs
+	{
+		/// <summary>
+		/// Initialize a new instance of <see cref="SoftResetEventArgs"/>
+		/// </summary>
+		public SoftResetEventArgs()
+		{
+		}
+	}
+
+	/// <summary>
 	/// Represent a method that will handle the event raised before a quickload is done
 	/// </summary>
 	/// <param name="sender">Object that raised the event</param>
@@ -124,4 +150,18 @@ namespace BizHawk.Client.Common
 	/// <param name="sender">Object that raised the event</param>
 	/// <param name="e">Event arguments</param>
 	public delegate void StateSavedEventHandler(object sender, StateSavedEventArgs e);
+
+	/// <summary>
+	/// Represent a method that will handle the event raised when a hard reset is performed
+	/// </summary>
+	/// <param name="sender">Object that raised the event</param>
+	/// <param name="e">Event arguments</param>
+	public delegate void HardResetEventHandler(object sender, HardResetEventArgs e);
+
+	/// <summary>
+	/// Represent a method that will handle the event raised when a soft reset is performed
+	/// </summary>
+	/// <param name="sender">Object that raised the event</param>
+	/// <param name="e">Event arguments</param>
+	public delegate void SoftResetEventHandler(object sender, SoftResetEventArgs e);
 }
--- a/src/BizHawk.Client.Common/Api/Classes/EmuClientApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/EmuClientApi.cs
@@ -31,6 +31,10 @@ namespace BizHawk.Client.Common
 
 		public event StateSavedEventHandler StateSaved;
 
+		public event HardResetEventHandler HardReset;
+
+		public event SoftResetEventHandler SoftReset;
+
 		public EmuClientApi(Action<string> logCallback, IMainFormForApi mainForm, DisplayManagerBase displayManager, Config config, IEmulator emulator, IGameInfo game)
 		{
 			_config = config;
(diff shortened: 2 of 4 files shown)


src/BizHawk.Client.Common/EventTypes.cs
```diff
@@ -124,4 +150,18 @@ public StateSavedEventArgs(string stateName)
...
+	/// <param name="sender">Object that raised the event</param>
+	/// <param name="e">Event arguments</param>
+	public delegate void HardResetEventHandler(object sender, HardResetEventArgs e);
+
+	/// <summary>
+	/// Represent a method that will handle the event raised when a soft reset is performed
+	/// </summary>
+	/// <param name="sender">Object that raised the event</param>
+	/// <param name="e">Event arguments</param>
+	public delegate void SoftResetEventHandler(object sender, SoftResetEventArgs e);
```
YoshiRulz, line 166: Don't need custom delegate types, can use `EventHandler<>`, but also don't need custom arg struct with no props, so can use `Action`.

YoshiRulz (changes requested):

CasualPokePlayer:
This will fail for cases where the user just binds in a regular input to Reset/Power. Keep in mind all those UI buttons actually do here is press "Reset" or "Power" on the emulated controller, it's no different from pressing "A" or "B" on the emulated controller.

roydmerkel:
> This will fail for cases where the user just binds in a regular input to Reset/Power, or if they're pressed in a movie, or maybe the Virtual Pad. Keep in mind all those UI buttons actually do here is press "Reset" or "Power" on the emulated controller, it's no different from pressing "A" or "B" on the emulated controller.
> 
> In fact, couldn't the user of the API just check what buttons are being pressed? Surely that would be a better way to check this (also required for when GB Link enters the picture which has each _player_ (i.e. each linked console) have a Power button)

> In fact, couldn't the user of the API just check what buttons are being pressed? Surely that would be a better way to check this (also required for when GB Link enters the picture which has each _player_ (i.e. each linked console) have a Power button)

This was something I tried, but when I run InputManager.ClickyVirtualPadController.IsPressed("Power"); in Update, Update* or FastUpdate*, it doesn't fire. It appears it's being released in src/BizHawk.Client.EmuHawk/MainForm.cs as well before Update callback is triggered.

I don't mind closing this one as fixed, I'm just having trouble catching this event properly without hooking it myself.  I also tried using reflection to get MainForm.InputManager (which is private, but still accessible via reflection), and even then IsPressed is cleared before Update, Update*, and FastUpdate* events are triggered.

Do you have an example of a class or external tool successfully listening for a InputManager.ClickyVirtualPadController press that I can use for reference?

src/BizHawk.Client.Common/EventTypes.cs
roydmerkel: Ok, I know @CasualPokePlayer below had suggested trying to poll  InputManager.ClickyVirtualPadController.IsPressed("Power") which doesn't work for me, but I'm seeing if he can give an example of something polling InputManager.ClickyVirtualPadController.IsPressed("Power")  or equivlent I can use for reference to fix my code instead.

 If his suggestion doesn't work, then I'll want to replace the delegates with a generic EventHandler<> class, correct? (Also, I'll remove the second e arg in all the calls in that event, I added them assuming that they should be in parity with existing callbacks, which was an incorrect assumption on my end -- I'll fix it if we need a follow up).

YoshiRulz: `event Action HardReset;` in `IEmuClientApi` should suffice

kalimag:
> This was something I tried, but when I run InputManager.ClickyVirtualPadController.IsPressed("Power"); in Update, Update* or FastUpdate*, it's always false. It appears it's being released in src/BizHawk.Client.EmuHawk/MainForm.cs as well before Update callback is triggered.

I think `IJoypadApi.Get()` (or `GetWithMovie`?) is what you're looking for. That should give you the processed input as seen by the core on that frame.

CasualPokePlayer:
You shouldn't be trying to check the ClickyVirtualPadController directly, input gets resolved with multiple sources (e.g. if you had a movie in read-only mode then the ClickyVirtualPadController would just get ignored), and the end result can just be seen the IJoypadApi you're given in external tools (and the lua equivalent)

roydmerkel:
> You shouldn't be trying to check the ClickyVirtualPadController directly, input gets resolved with multiple sources, some of those other sources might also try to reset (or even prevent the reset, e.g. if you had a movie in read-only mode then the ClickyVirtualPadController would just get ignored), and the end result can just be seen the IJoypadApi you're given in external tools (and the lua equivalent)

That looks like it worked, I'll leave my rough cut code snippit here for the next guy that's looking for this and close the PR:

`        protected override void UpdateAfter() 
        {
            //if (labelsWriter != null)
            //{
            //    labelsWriter.WriteLine("UpdateAfter");
            //}
            if (JoypadApi != null)
            {
                IReadOnlyDictionary<string, object>[] dictionaries = new IReadOnlyDictionary<string, object>[] {
                        JoypadApi.Get(null),
                        JoypadApi.GetImmediate(null),
                        JoypadApi.GetWithMovie(null)
                };

                bool breakLoop = false;
                foreach (IReadOnlyDictionary<string, object> dict in dictionaries)
                {
                    if (dict.Count > 0)
                    {
                        if (dict.ContainsKey("Power"))
                        {
                            object hardReset;
                            bool got = dict.TryGetValue("Power", out hardReset);
                            if (got && hardReset is bool)
                            {
                                if ((bool)hardReset)
                                {
                                    if (labelsWriter != null)
                                    {
                                        labelsWriter.Write("hardReset:");
                                        labelsWriter.WriteLine(hardReset);
                                    }
                                    breakLoop = true;
                                }
                            }
                        }

                        if (dict.ContainsKey("Reset"))
                        {
                            object softReset;
                            bool got = dict.TryGetValue("Reset", out softReset);
                            if (got && softReset is bool)
                            {
                                if ((bool)softReset)
                                {
                                    if (labelsWriter != null)
                                    {
                                        labelsWriter.Write("softReset:");
                                        labelsWriter.WriteLine(softReset);
                                    }
                                    breakLoop = true;
                                }
                            }
                        }
                    }
                    if(breakLoop)
                    {
                        break;
                    } 
                }
            }
            base.UpdateAfter();
        }`

PR #3789, opened 2023-10-04
Libretro Settings problem by a Libretro core...

lhvista2007:
Good... I ask you to solve the problem of Libretro Settings by a core of Libretro worked in BizHawk that the window of this is all white, so that it has all the settings in it. Thank you all...


lhvista2007:
The problem of Libretro Settings by a core of Libretro worked in BizHawk that the window is all white, that if you solve to have all the settings in it, thank you ...

PR #3802, opened 2023-10-12
Add setting to control what happens when double-clicking RAM watch 

### LINKED ISSUE #3688: [Feature Request] Double click RAM watch to poke
Personally I tend to poke watches far more often than I edit their properties, so I'd like to open the poke window with a double click (instead of opening the edit window as it currently does). 

I can create a PR for this but wanted to ask for feedback first. I figure the way to do this would be either

1. Add a setting that changes what the double click does to the ram watch menu
2. Open poke when double clicking the value cell, open edit on the other cells
I prefer this because it's more flexible, but I don't know how intuitive this is to other users
3. Both (add settings for always edit/always poke/depending on cell)
4. Always poke

kalimag:
- Add a new setting to the `RamWatch` dialog menu that changes whether double-clicking a watch in the list will open the Edit or the Poke dialog. 
- Setting defaults to poke (previous behavior was edit)
- Pressing enter also obeys this setting 

Resolves #3688

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.Designer.cs
@@ -668,7 +668,7 @@ namespace BizHawk.Client.EmuHawk
             this.DoubleClickActionSubMenu.DropDownItems.AddRange(new System.Windows.Forms.ToolStripItem[] {
             this.DoubleClickToEditMenuItem,
             this.DoubleClickToPokeMenuItem});
-            this.DoubleClickActionSubMenu.Text = "Double-Click Action";
+            this.DoubleClickActionSubMenu.Text = "On Double-Clicking a Watch";
             this.DoubleClickActionSubMenu.DropDownOpening += new System.EventHandler(this.DoubleClickActionSubMenu_DropDownOpening);
             // 
             // DoubleClickToEditMenuItem
--- a/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.Designer.cs
@@ -107,6 +107,9 @@ namespace BizHawk.Client.EmuHawk
             this.OriginalMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
             this.WatchesOnScreenMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.WatchListView = new InputRoll();
+            this.DoubleClickActionSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
+            this.DoubleClickToEditMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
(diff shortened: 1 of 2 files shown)


src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.Designer.cs
```diff
@@ -659,6 +663,24 @@ private void InitializeComponent()
             this.WatchListView.KeyDown += new System.Windows.Forms.KeyEventHandler(this.WatchListView_KeyDown);
             this.WatchListView.MouseDoubleClick += new System.Windows.Forms.MouseEventHandler(this.WatchListView_MouseDoubleClick);
             // 
+            // DoubleClickActionSubMenu
+            // 
+            this.DoubleClickActionSubMenu.DropDownItems.AddRange(new System.Windows.Forms.ToolStripItem[] {
+            this.DoubleClickToEditMenuItem,
+            this.DoubleClickToPokeMenuItem});
+            this.DoubleClickActionSubMenu.Text = "Double-Click Action";
```
YoshiRulz, line 671:
```suggestion
            this.DoubleClickActionSubMenu.Text = "On Double-Clicking a Watch:";
```
Though if Enter has the same effect, I'm not sure...

src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.cs
```diff
@@ -494,6 +507,22 @@ private void EditWatch(bool duplicate = false)
 			}
 		}
 
+		private void PokeAddress()
```
YoshiRulz, line 510: I think the diff would be smaller if this was beside `PokeAddressMenuItem_Click`.

src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.cs
```diff
@@ -433,6 +435,17 @@ private void FullyUpdateWatchList()
...
 
+		/// <summary>
+		/// Open Edit or Poke window for selected watch depending on user setting
+		/// </summary>
+		private void OpenWatch()
+		{
+			if (Settings.DoubleClickToPoke)
+				PokeAddress();
+			else
+				EditWatch();
```
YoshiRulz, line 446:
```diff
- if (Settings.DoubleClickToPoke)
- 	PokeAddress();
- else
- 	EditWatch();
+ if (Settings.DoubleClickToPoke) PokeAddress();
+ else EditWatch();
```
(or use braces)

YoshiRulz (changes requested):

src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.Designer.cs
kalimag: I figured keyboard navigation should also have access to that feature, but that most users wouldn't care or would figure it out. Explicitly referring to double-clicking seemed more intuitive overall than trying to find an all-encompassing term, and it didn't seem worth it to over-complicate the menu or have a super long description somewhere

YoshiRulz (approved):
LGTM now.

Issue #3813, opened 2023-10-28
Value from mem address is not passed to read callback


VelpaChallenger:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"
[//]: # "One more thing: if you're on Linux, please comment on #1430 instead of opening an issue so we don't annoy the other devs."

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
When running a lua script which registers a memory callback function to the on_bus_read hook, and the core calls the callback function, the value from the memory address is not passed.

So far I found it happens with BSNES and Genplus-gx, but it could potentially happen with other cores as well. Also with every game, for example Virtual Bart (Sega Genesis version, that is core Genplus-gx) and Contra 3: The Alien Wars (for BSNES). Both memory addresses are for the RNG. Contra 3: The Alien Wars' RNG starts advancing as soon as the Konami presentation starts (that is, before the title screen) and Virtual Bart's after the roulette starts spinning (start the game, skip the cutscene, then wait a few moments until the roulette starts spinning, you can also go to the practice area which makes the RNG advance as well).

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Run Lua script which registers a memory callback function to on_bus_read hook (see below Lua script)
2. Go to section of the game which reads to the memory address passed to hook

Lua script:

```
function my_callback(addr, val, flags)
    print("addr being read is" .. string.format("%x", addr))
    print("value is" .. string.format("%x", val))
    print("flags is" .. flags)
    client.pause()
end
```

event.on_bus_read(my_callback, 0xFFFFC676) --FFFFC676 for Virtual Bart, E6 for Contra 3: The Alien Wars

### Output
[//]: # "Paste the contents of the error dialog if there is one (try Ctrl+C, it usually works), or paste the output from the Lua Console, or delete this section."
The value printed to the console is always 0.

### Host env.
[...]

PR #3821, opened 2023-11-05
Memory hook improvements (GPGX)

VelpaChallenger:

~~For BSNES core, signatures were updated and data value was passed to the read callback after reading from the bus.~~
For genplus-gx, since all callbacks (read/write/exec) use the same signature from `mem_cb`, I had to pass the value to all three of them. And since this could potentially cause some unwanted behavior, I also updated the values passed to the write and exec callbacks inside the core (I divided each update in a separate commit so it's easier to follow and understand).

Check if completed:

see #3813


Morilli (approved):
Changes look good from what I can see without actually testing them.

VelpaChallenger:
Something important to keep in mind is that there's a trade-off to these changes, and that is that a `memory.writebyte` inside a memory read callback function will not write the value before the reading **by the core,** so if someone wanted to try to force a value to be read **by the game** (as a form of stronger freezing if you will), it won't be possible anymore, since now the callback is called **after** the bus is read (albeit still before the game reads it). I don't know how many people actually do this, but just as a heads-up.

Morilli:
That's actually a valid point, and it reinforces the point that there IS a "before" and "after" state when it comes to reading values, and the decision on where the callback is called has an actual impact on how that callback can be used.

[Quoting myself from earlier](https://github.com/TASEmulators/BizHawk/issues/3813#issuecomment-1783861520):
> I'd definitely appreciate an update to the docs to clarify for both bizhawk devs and lua devs how exactly those functions should work.

@YoshiRulz I do believe the "before" part for the callback is correct and matches expectations, we just need to decide what to do with the `val` parameter and how it should get populated: https://github.com/TASEmulators/BizHawk/blob/6e77c7348f3c29722fbdae8040085390c9f03a17/src/BizHawk.Client.Common/lua/LuaHelperLibs/EventsLuaLibrary.cs#L206

YoshiRulz:
>> I'd definitely appreciate an update to the docs to clarify for both bizhawk devs and lua devs how exactly those functions should work.
>
> [...] we just need to decide what to do with the val parameter and how it should get populated

I don't understand your question.
> Fires immediately before the given address is read by the core. [...] `val` is the value read.

...seems pretty clear to me. Are you saying we should make that even clearer, maybe "`val` is the value to be read"? Or is it that existing implementations don't match the docs?

---

> so if someone wanted to try to force a value to be read by the game (as a form of stronger freezing if you will), it won't be possible anymore

In our discussions on Discord re: possible implementations, this never came up. Freezing needs to remain functional—and I believe if you've broken it for Lua then it must also be broken for cheats.

Morilli:
> Are you saying we should make that even clearer, maybe "val is the value to be read"? Or is it that existing implementations don't match the docs?

Both, probably. First of all yes, `val is the value to be read` is the only way that makes sense to me.
Also, I assume no core will actually implement both conditions "correctly" at the moment, aka call the callback *before* the value is read (and allow for manipulation of the value *to be read* before it IS read), and also provide the value *to be read* in the callback.

Making it clearer in the docs what's important would help here.

VelpaChallenger:
> In our discussions on Discord re: possible implementations, this never came up. Freezing needs to remain functional—and I believe if you've broken it for Lua then it must also be broken for cheats.

I think there's a misunderstanding: freezing still works as usual and hasn't been affected in any way. What I meant with the wording "stronger freezing" is that when you freeze a memory address using cheats (traditional way), you are not actually freezing the memory address in the strict sense of the word, or at least not the way I'd expect. Whenever a frame advances, if the game's code updates the memory address, it will be updated. Say for example that you freeze `0xC676` when it has stored a `0xC` value, if there's a line like `move.b D0, ($C676)`, and then there's another line `move.b ($C676), ($C284)`, then `0xC284` is not going to be populated with the value `0xC`, rather it's going to be populated with whatever value was in `D0`. That's what I mean when I say it's not a freeze in the strict sense of the word. It's only going to get updated to `0xC` at the end of the frame, but during it, freezing has no effect. However, when using memory callbacks, you can do something like `memory.writebyte(0xC676, 0xC)` inside your callback and then when `move.b ($C676), ($C284)` is executed, the value `0xC` is going to be passed to `0xC284` because that is the value that is going to be read at `0xC676`.

Also, the way I see it, there are 3 key moments to have in mind:

1. The moment the bus is read
2. The moment the game reads the value
3. The moment the callback is run

The moment the bus is read is for BSNES when this line `auto data = bus.read(address, r.mdr);` is executed. The moment the game reads the value is after the function used for accessing memory values returns (for BSNES, that would be when `CPU::read` returns). And the moment the callback is run is, well, when it's called by the core.

My point being that as I mentioned in the issue comments, while I agree with Morilli that the decision on where the callback is called has an actual impact on how that callback can be used, I still don't think there is really a "before" and an "after" state. The value still hasn't been sent when the callback is run, in fact, one possible way to go around this is to call `bus.read` again after calling the callback and then send that as the return value of `CPU::read` (for BSNES). That would be possible because the game's code still didn't receive the value at that point in time, so it's still before. 

I do believe `before the given address is read by the core` should be changed to `before the given address is read by the game's code`, though. `val is the value to be read` sounds more accurate, but it's still not fully accurate given what we are discussing here and how memory callbacks cannot be used for the stronger freezing I was talking about.

Also, speaking of which. This detail slipped my mind earlier, but for the existing implementations, like here: https://github.com/TASEmulators/BizHawk/blob/b98942dcfdb7acbb39acc2864253d2bf55132fda/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/NES.Core.cs#L1077-L1084 **memory callbacks already cannot be used for the stronger freezing**. So in that sense, nothing to really worry about. 
(As a positive of this implementation, you can know how many bytes are actually being read by the game (so val will be 1 byte if just 1 byte is being read, 2 bytes if 2 bytes are being read, etc.). This has been very useful to me for reverse engineering.)

YoshiRulz:
This discussion is continuing on Discord, but I want to apologise for my earlier assertion that you'd broken freezing. Freezing is, of course, already "broken" in that sense, and has been forever since it's implemented as pokes.
https://github.com/TASEmulators/BizHawk/blob/bb0d8f231d802dc6e4d3b09426e74435c9e23c37/src/BizHawk.Client.Common/tools/Cheat.cs#L168-L176
(This method is called twice in `StepRunLoop_Core`, both [before](https://github.com/TASEmulators/BizHawk/blob/bb0d8f231d802dc6e4d3b09426e74435c9e23c37/src/BizHawk.Client.EmuHawk/MainForm.cs#L3136) and [after](https://github.com/TASEmulators/BizHawk/blob/bb0d8f231d802dc6e4d3b09426e74435c9e23c37/src/BizHawk.Client.EmuHawk/MainForm.cs#L3214) `FrameAdvance`.)

YoshiRulz:
Please rebase. Since GPGX got a big update, it may not be trivial to recreate those changes, in which case you can drop them and we'll leave the linked issue open to track that. Or maybe it was magically fixed in the update? (GPGX also now includes a stronger version of freezing, if you were interested in that.)

VelpaChallenger:
Definitely interested, thanks for letting me know. Will check the changes and see if I can recreate what I did in this PR (if still necessary).

VelpaChallenger:
Maybe things changed in this last month, but still, to make the PR discussion easy to track, I'd like to include what we recently talked on Discord, namely that after bringing in the changes from gpgx upstream, it was in fact sort of "magically fixed", because now the order is the same as it is in this PR (first read the value, then call the callback). The problem is the same as before and that is that this way, deep freezing is not possible using `on_bus_read`. Maybe once deep freezing is stable enough (such that it can replace freezing by means of `on_bus_read`), we can leave the order of this PR and then replicate it to the BSNES and other cores. Though there was also discussion on updating the documentation on the expected behavior: first read, then callback with the value read, or first callback, just before the read, and then read. Once that is decided, I think I could move on updating the PR accordingly.

vadosnaprimer:
> The problem is the same as before and that is that this way, deep freezing is not possible using `on_bus_read`.

While we were working on deep freeze for gpgx I pondered a few approaches. The current one is simply freezing individual bytes by restoring their values after every write. We discussed changing _the value to be written_ right before it's written, also from inside the write callback.

The problem with changing it while inside the write callback is different widths of written values. If in a given scenario we only need to freeze 1 byte because we know the game uses that address that way, we won't be able to easily overwrite it because callbacks can access that address using a different width, and we don't want to freeze adjacent addresses, so we'll only need to change a part of the bigger value that goes to that one byte address. Or we may be freezing a 32-bit value, but the game may be writing to it using smaller sizes.

And overwriting it right after it's written while inside the write callback, changes the order of things, which may be problematic due to inconsistency.

YoshiRulz:
Please rebase again. It might be wise to split this into separate PRs for each core.

Morilli:
Have we ever decided on whether this approach is correct?

The fundamental question to answer is whether memory callbacks should happen before or after read/write. Write hooks currently usually happen before write, and if that's the case read hooks should too.

YoshiRulz:
Read hooks need to be before the read to allow for intercept (freeze).
Write hooks should be before write so the hook can see both the old and new values.

YoshiRulz:
From the above I gather it's not working in master, so I would like for this PR to fix it.

VelpaChallenger:
I will check this more carefully tomorrow, but I believe as Morilli said, we never really decided what the approach should be. And this PR makes it such that it is possible to pass the read value, but by means of first reading, _then_ sending the value. So the changes here are in contradiction with what is (I think) expected.

VelpaChallenger:
@YoshiRulz There are two different issues here:

- Memory callback should happen before read
- Core should pass read value to memory callback

Those two are in direct contradiction of each other: if memory callback happens before read, the core cannot pass the read value, as the read hasn't happen yet. If the core passes the read value, then the read should happen first.

Also, I checked the related GitHub issues, they're about the second. That's the "not working in master" that you're referring to I believe.

This PR fixes the second issue, but if it gets merged, we will have the first. I think we should first decide if we want both (memory callback before read _and_ core passes value read to callback), and then think how to get that working. I'm pretty sure one alternative we discussed in Discord back then was to not have memory callback happen before read, and instead have the callback send a value that says "read again" to the core, or something of the sort. That way, we both send the value to the callback and we preserve deep freezing functionality. 

But again, it is something that we need to decide.

YoshiRulz:
I don't see the contradiction.
> Read hooks need to be before the read to allow for intercept (freeze).

If that means ~~reading~~ peeking twice, so be it. Other cores have a fast path without the callback logic for when no callbacks have been added.

What I don't want is for the dummy implementation (return 0) to be replaced with an incorrect implementation.
This would be clearer if the service actually captured any of this nuance, but of course it doesn't. Not even the doc comments are that helpful.

VelpaChallenger:
Alright, sounds good to me. I will rebase and add the twice peeking, if nobody sees an issue with that. And split the PR, one for each core.

VelpaChallenger:
Done. This PR is for GPGX, will create another for BSNES.

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.IDebuggable.cs
```diff
@@ -72,30 +72,30 @@ public IMemoryCallbackSystem MemoryCallbacks
...
 		private void InitMemCallbacks()
 		{
-			ExecCallback = a =>
+			ExecCallback = new LibGPGX.mem_cb((a, val) =>
 			{
 				if (MemoryCallbacks.HasExecutes)
 				{
-					const uint flags = (uint)MemoryCallbackFlags.AccessExecute;
-					MemoryCallbacks.CallMemoryCallbacks(a, 0, flags, "M68K BUS");
+					uint flags = (uint)MemoryCallbackFlags.AccessExecute;
```
YoshiRulz, line 79: These are still `const` no?

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.IDebuggable.cs
```diff
@@ -72,30 +72,30 @@ public IMemoryCallbackSystem MemoryCallbacks
 
 		private void InitMemCallbacks()
 		{
-			ExecCallback = a =>
+			ExecCallback = new LibGPGX.mem_cb((a, val) =>
```
YoshiRulz, line 75: Shouldn't need explicit delegate constructors ever

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.IDebuggable.cs
VelpaChallenger: Yes, I'm not sure why code that I never added/removed was, well added/removed. Something seems to have gone wrong during rebasing. Either way, I just reinserted it.

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.IDebuggable.cs
VelpaChallenger: Right, this was code from 2023 and I realize now those don't have to be explicit. Updated as well.

YoshiRulz:
Is the submodule change part of this?

VelpaChallenger:
> Is the submodule change part of this?

Yes. I'm not sure why those changes are not shown here, I suppose it's because they're part of a different commit in a different repo. I guess I will have to create a PR in that repo as well?

YoshiRulz:
Yes, you'll need to PR against [our fork](https://github.com/TASEmulators/Genesis-Plus-GX).

---

On Discord, feos suggested this work is redundant with the `genesis.add_deepfreeze_value` API. That may be true, but it doesn't excuse the inconsistencies in the r/w/x hook APIs. They have other uses besides freezing values. And I don't want to proliferate core-specific freeze APIs.

vadosnaprimer:
Here are the patches to the core that make it possible to return something back to the core when a callback fires and that value could be used to override whatever is read or written by the core. Value read/written is not yet passed to hawk, it'd be up to this PR.

<details><summary>submodule</summary>

```diff
 core/debug/cpuhook.c |  4 ++--
 core/debug/cpuhook.h |  5 +++--
 core/m68k/m68kcpu.h  | 12 ++++++------
 3 files changed, 11 insertions(+), 10 deletions(-)

diff --git a/core/debug/cpuhook.c b/core/debug/cpuhook.c
index 1e58835..0bc4775 100644
--- a/core/debug/cpuhook.c
+++ b/core/debug/cpuhook.c
@@ -43,9 +43,9 @@
 #include <stdio.h>
 #include "cpuhook.h"
 
-void(*cpu_hook)(hook_type_t type, int width, unsigned int address, unsigned int value) = NULL;
+uint32(*cpu_hook)(hook_type_t type, int width, unsigned int address, unsigned int value) = NULL;
 
-void set_cpu_hook(void(*hook)(hook_type_t type, int width, unsigned int address, unsigned int value))
+void set_cpu_hook(uint32(*hook)(hook_type_t type, int width, unsigned int address, unsigned int value))
 {
 	cpu_hook = hook;
 }
diff --git a/core/debug/cpuhook.h b/core/debug/cpuhook.h
index 1fea3ec..e3a53b9 100644
--- a/core/debug/cpuhook.h
+++ b/core/debug/cpuhook.h
@@ -43,6 +43,7 @@
 #ifndef _CPUHOOK_H_
 #define _CPUHOOK_H_
 
+#include "types.h"
 
 typedef enum {
   HOOK_ANY      = (0 << 0),
@@ -80,12 +81,12 @@ typedef enum {
 
 /* CPU hook is called on read, write, and execute.
  */
-void (*cpu_hook)(hook_type_t type, int width, unsigned int address, unsigned int value);
+uint32 (*cpu_hook)(hook_type_t type, int width, unsigned int address, unsigned int value);
 
 /* Use set_cpu_hook() to assign a callback that can process the data provided
  * by cpu_hook().
  */
-void set_cpu_hook(void(*hook)(hook_type_t type, int width, unsigned int address, unsigned int value));
+void set_cpu_hook(uint32(*hook)(hook_type_t type, int width, unsigned int address, unsigned int value));
 
 
 #endif /* _CPUHOOK_H_ */
\ No newline at end of file
diff --git a/core/m68k/m68kcpu.h b/core/m68k/m68kcpu.h
index d5fce91..88b3192 100644
--- a/core/m68k/m68kcpu.h
+++ b/core/m68k/m68kcpu.h
@@ -860,7 +860,7 @@ INLINE uint m68ki_read_8(uint address)
 
 #ifdef HOOK_CPU
   if (UNLIKELY(cpu_hook))
-    cpu_hook(HOOK_M68K_R, 1, address, val);
+    val = cpu_hook(HOOK_M68K_R, 1, address, val);
 #endif
 
   return val;
@@ -880,7 +880,7 @@ INLINE uint m68ki_read_16(uint address)
 
 #ifdef HOOK_CPU
   if (UNLIKELY(cpu_hook))
-    cpu_hook(HOOK_M68K_R, 2, address, val);
+    val = cpu_hook(HOOK_M68K_R, 2, address, val);
 #endif
 
   return val;
@@ -900,7 +900,7 @@ INLINE uint m68ki_read_32(uint address)
 
 #ifdef HOOK_CPU
   if (UNLIKELY(cpu_hook))
-    cpu_hook(HOOK_M68K_R, 4, address, val);
+    val = cpu_hook(HOOK_M68K_R, 4, address, val);
 #endif
 
   return val;
@@ -914,7 +914,7 @@ INLINE void m68ki_write_8(uint address, uint value)
 
 #ifdef HOOK_CPU
   if (UNLIKELY(cpu_hook))
-    cpu_hook(HOOK_M68K_W, 1, address, value);
+    value = cpu_hook(HOOK_M68K_W, 1, address, value);
 #endif
 
   temp = &m68ki_cpu.memory_map[((address)>>16)&0xff];
@@ -937,7 +937,7 @@ INLINE void m68ki_write_16(uint address, uint value)
 
 #ifdef HOOK_CPU
   if (UNLIKELY(cpu_hook))
-    cpu_hook(HOOK_M68K_W, 2, address, value);
+    value = cpu_hook(HOOK_M68K_W, 2, address, value);
 #endif
 
   temp = &m68ki_cpu.memory_map[((address)>>16)&0xff];
@@ -960,7 +960,7 @@ INLINE void m68ki_write_32(uint address, uint value)
 
 #ifdef HOOK_CPU
   if (UNLIKELY(cpu_hook))
-    cpu_hook(HOOK_M68K_W, 4, address, value);
+    value = cpu_hook(HOOK_M68K_W, 4, address, value);
 #endif
 
   temp = &m68ki_cpu.memory_map[((address)>>16)&0xff];
```
</details>

<details><summary>hawk repo</summary>

```diff
 waterbox/gpgx/cinterface/callbacks.h  |  6 +++---
 waterbox/gpgx/cinterface/cinterface.c | 16 +++++++++-------
 2 files changed, 12 insertions(+), 10 deletions(-)

diff --git a/waterbox/gpgx/cinterface/callbacks.h b/waterbox/gpgx/cinterface/callbacks.h
index 4ee2e5ee78..13f16b4898 100644
--- a/waterbox/gpgx/cinterface/callbacks.h
+++ b/waterbox/gpgx/cinterface/callbacks.h
@@ -7,9 +7,9 @@
 
 typedef ECL_ENTRY void (*CDCallback)(int32 addr, int32 addrtype, int32 flags);
 
-extern ECL_ENTRY void (*biz_execcb)(unsigned addr);
-extern ECL_ENTRY void (*biz_readcb)(unsigned addr);
-extern ECL_ENTRY void (*biz_writecb)(unsigned addr);
+extern ECL_ENTRY uint32 (*biz_execcb)(unsigned addr);
+extern ECL_ENTRY uint32 (*biz_readcb)(unsigned addr);
+extern ECL_ENTRY uint32 (*biz_writecb)(unsigned addr);
 extern CDCallback biz_cdcb;
 
 extern ECL_ENTRY void (*cdd_readcallback)(int lba, void *dest, int subcode);
diff --git a/waterbox/gpgx/cinterface/cinterface.c b/waterbox/gpgx/cinterface/cinterface.c
index 2fae99f0a6..6e81f6e2ce 100644
--- a/waterbox/gpgx/cinterface/cinterface.c
+++ b/waterbox/gpgx/cinterface/cinterface.c
@@ -61,9 +61,9 @@ static uint8_t brm_format[0x40] =
 	0x52,0x41,0x4d,0x5f,0x43,0x41,0x52,0x54,0x52,0x49,0x44,0x47,0x45,0x5f,0x5f,0x5f
 };
 
-ECL_ENTRY void (*biz_execcb)(unsigned addr);
-ECL_ENTRY void (*biz_readcb)(unsigned addr);
-ECL_ENTRY void (*biz_writecb)(unsigned addr);
+ECL_ENTRY uint32 (*biz_execcb)(unsigned addr);
+ECL_ENTRY uint32 (*biz_readcb)(unsigned addr);
+ECL_ENTRY uint32 (*biz_writecb)(unsigned addr);
 CDCallback biz_cdcb = NULL;
 ECL_ENTRY void (*cdd_readcallback)(int lba, void *dest, int subcode);
 uint8 *tempsram;
@@ -783,7 +783,7 @@ void CDLog68k(uint addr, uint flags)
 	}
 }
 
-void bk_cpu_hook(hook_type_t type, int width, unsigned int address, unsigned int value)
+uint32 bk_cpu_hook(hook_type_t type, int width, unsigned int address, unsigned int value)
 {
 	switch (type)
 	{
@@ -804,7 +804,7 @@ void bk_cpu_hook(hook_type_t type, int width, unsigned int address, unsigned int
 		case HOOK_M68K_R:
 		{
 			if (biz_readcb)
-				biz_readcb(address);
+				return biz_readcb(address);
 
 			break;
 		}
@@ -812,13 +812,15 @@ void bk_cpu_hook(hook_type_t type, int width, unsigned int address, unsigned int
 		case HOOK_M68K_W:
 		{
 			if (biz_writecb)
-				biz_writecb(address);
+				return biz_writecb(address);
 
 			break;
 		}
 
 		default: break;
 	}
+	
+	return value;
 }
 
 #endif // USE_BIZHAWK_CALLBACKS
@@ -1035,7 +1037,7 @@ GPGX_EX void gpgx_reset(int hard)
 		gen_reset(0);
 }
 
-GPGX_EX void gpgx_set_mem_callback(ECL_ENTRY void (*read)(unsigned), ECL_ENTRY void (*write)(unsigned), ECL_ENTRY void (*exec)(unsigned))
+GPGX_EX void gpgx_set_mem_callback(ECL_ENTRY uint32 (*read)(unsigned), ECL_ENTRY uint32 (*write)(unsigned), ECL_ENTRY uint32 (*exec)(unsigned))
 {
 	biz_readcb = read;
 	biz_writecb = write;
```
</details>

VelpaChallenger:
@vadosnaprimer Thanks, but I already made the changes to the core in my local, though I used unsigned int instead of uint32. Is there any significant difference in the context of cores? In any case, I will include core changes as a PR against [Bizhawk fork of gpgx](https://github.com/TASEmulators/Genesis-Plus-GX), as we discussed above with Yoshi.

vadosnaprimer:
I tried to check that change but the link is empty, so I have no idea what you changed in the core.

VelpaChallenger:
I didn't create it yet, I'm on that

VelpaChallenger:
Updated this PR with the Bizhawk changes. Here's the PR with the [submodule updates](https://github.com/TASEmulators/Genesis-Plus-GX/pull/8):

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
```diff
@@ -119,7 +130,7 @@ public void DetachFromScript()
 
 		public string Event { get; }
 
-		private Action<object[]> Callback { get; }
+		private Func<object[],uint?> Callback { get; }
```
Morilli, line 133: I think it would be better if this was `object[], object[]`. Apparently we never care about lua function return values to C# code currently but it would make more sense if lua functions could return anything and we only filtered by `uint?` in the memory callback, not any callback.

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
```diff
@@ -60,12 +60,18 @@ public NamedLuaFunction(LuaFunction function, string theEvent, Action<string> lo
 			{
 				try
 				{
-					_function.Call(args);
+					object[] LuaFunctionReturn = _function.Call(args);
+					if(LuaFunctionReturn?.Length != 0) // If the value returned is not specified in the callback, the array will not have any elements.
+					{
+						return Convert.ToUInt32(LuaFunctionReturn[0]);
```
Morilli, line 66: I don't like this. Just check if the return value is `long` and cast it to an uint and ignore everything else.

src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
```diff
@@ -66,47 +66,50 @@ public void Add(IMemoryCallback callback)
...
-		private static void Call(MemoryCallbackCollection cbs, uint addr, uint value, uint flags, string scope)
+		private static uint Call(MemoryCallbackCollection cbs, uint addr, uint value, uint flags, string scope)
 		{
+			uint cbReturn = value; //By default, if no callback is called, it will return the value sent by the core unmodified.
 			foreach (var cb in cbs)
 			{
 				if (!cb.Address.HasValue || (cb.Scope == scope && cb.Address == (addr & cb.AddressMask)))
 				{
-					cb.Callback(addr, value, flags);
+					cbReturn = cb.Callback(addr, value, flags); //If there are many callbacks registered to the same addr, only the last callback's return value will be sent to the core.
```
Morilli, line 76: I don't think this is correct. Let's say you have two callbacks and only the first one wants to change the value. Won't this still retain the value of the second callback even though it didn't want to change the value?

src/BizHawk.Emulation.Common/Interfaces/IMemoryCallbackSystem.cs
```diff
@@ -5,7 +5,7 @@
 namespace BizHawk.Emulation.Common
 {
 	/// <param name="value">For reads/execs, the value read/executed; for writes, the value to be written. Cores may pass the default <c>0</c> if write/exec is partially implemented.</param>
-	public delegate void MemoryCallbackDelegate(uint address, uint value, uint flags);
+	public delegate uint MemoryCallbackDelegate(uint address, uint value, uint flags);
```
Morilli, line 8: This should be `uint?` to allow callbacks to explicitly *not* return a value.

Morilli (changes requested):

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
VelpaChallenger: _function.Call(args) returns an empty array, even if the Lua function returns nil implicitly. That's why I check here that it is not empty/null, and only then do the cast.

src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
VelpaChallenger: Yes, that's what I asked in Discord yesterday: 

> question: what if many callbacks are registered to run when the same memory address is read/written/executed? should we return the value from the last callback?

>i guess so
>that should match what would happen with callbacks doing writes as well

We agreed to only retain the last value. Though if you ask me, it would make more sense to not retain it, and instead, if the first one changes it but the second one doesn't, the updated value still reaches the core. I can update the logic to be like this, just clarifying why I did it that way.

YoshiRulz:
Can y'all test the return value feature in a separate PR? I felt we were about to get the value parameter fix merged.

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
VelpaChallenger: Do you mean `object[], object`? `object[], object[]` breaks because `object[]` cannot be converted to `uint`.

src/BizHawk.Emulation.Common/Interfaces/IMemoryCallbackSystem.cs
VelpaChallenger: Makes sense, I'll be adding it in a new PR if you're fine with splitting this one (based on Yoshi's comment, and well yeah this PR is pretty big).

Morilli:
I don't mind.

VelpaChallenger:
There. Back to the last commit before the return value feature.

waterbox/gpgx/Genesis-Plus-GX
Morilli, line 1: This commit is not correct. If I'm seeing it correctly it should be 79c243dee53291d4a004406200275d11c9875473.

VelpaChallenger: Why not correct? 79c243dee53291d4a004406200275d11c9875473 still had some leftovers from a (presumably) mixed rebase. In this last commit 1013b8dfc0fad6fcd97467c6bb9e1bf5f2c8c195 I compared with master, and added/removed the lines that should/shouldn't be there.

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
YoshiRulz: Lua has multi-return. You'd need to assert the array is of length 1 and then that the element is of the desired type (`if (arr is [ uint n ]) {}`).

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.IDebuggable.cs
```diff
@@ -72,28 +72,28 @@ public IMemoryCallbackSystem MemoryCallbacks
 
 		private void InitMemCallbacks()
 		{
-			ExecCallback = a =>
+			ExecCallback = (a, val) =>
```
YoshiRulz, line 75: Since you're touching all these lines, might as well rename the param to `addr` while you're at it

waterbox/gpgx/cinterface/callbacks.h
```diff
@@ -7,9 +7,9 @@
 
 typedef ECL_ENTRY void (*CDCallback)(int32 addr, int32 addrtype, int32 flags);
 
-extern ECL_ENTRY void (*biz_execcb)(unsigned addr);
-extern ECL_ENTRY void (*biz_readcb)(unsigned addr);
-extern ECL_ENTRY void (*biz_writecb)(unsigned addr);
+extern ECL_ENTRY void (*biz_execcb)(unsigned addr, unsigned int value);
+extern ECL_ENTRY void (*biz_readcb)(unsigned addr, unsigned int value);
+extern ECL_ENTRY void (*biz_writecb)(unsigned addr, unsigned int value);
```
YoshiRulz, line 12: Should these be simply `unsigned`? Is there a difference?

YoshiRulz (approved):
Diff LGTM

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.IDebuggable.cs
VelpaChallenger: Makes sense, renamed

PR #3827, opened 2023-11-12
Add a sound for making a successful submission to an RA leaderboard

BigJ64:

Prior to this change the `Leaderboard Attempt Complete` and `Unlock Achievement` sounds were using the exact same source. This was causing confusion for myself and others on whether an achievement was being unlocked or a leaderboard time/score was being posted. The sound is now split to allow a differentiation in these two events. The sound file included was made by myself but, as sound effects are highly subjective, I understand if another would rather be used.  If a place holder/duplicating the existing `overlay/unlock.wav` is preferred please let me know.

Check if completed:


CasualPokePlayer:
To be clear, these sounds mirror the ones used in RAIntegration, which has the sound effects [here](https://github.com/RetroAchievements/RAInterface/commit/45e820e1ae3189dde0a5b2338aac38f7a19b05d5) (and the overlay folder is in part intended to have shared usage amount BizHawk's rcheevos impl and RAIntegration).

Granted, there's probably no reason why we can't just diviate here, RAIntegration wouldn't care about an extra file in this folder, so this is probably fine as is (the worst is just the sound being maybe off compared to the other ones which are mostly just ripped from Mario).

BigJ64:
A better implementation within Bizhawk would be that the sounds would be fully customizable via a user interface but that is outside of my current abilities. If there is a concern in deviating from the RAIntegration causing user confusion due to a disparate sound, reusing `unlock.wav` copied and renamed to `lbcomplete.wav` would grant the functionality without changing the user experience as it is today.  There would still be an extra file but if RAIntegration were to add a sound to the overlay folder upstream and cause a conflict then this could just be tossed in favor of the upstream.

Let me know if I should adjust the PR in some way.  Thanks.

CasualPokePlayer:
The user can change up the .wavs in any way they want, although the pathing isn't something customization since RAIntegration is locked to the /overlay path.

I'm not sure if the PR should be adjusted or not, if there aren't any objections I'm fine with merging this as-is.

BigJ64:
Updated to remove file conflict.

Issue #3238, opened 2022-05-07
Auto-mute when lagging


YoshiRulz:
For example, when frame times exceed 19 ms (< 90% speed for a 60 fps game) for N consecutive frames. Then a similar heuristic would trigger an unmute if frame times go back down.

Meerkov:
It's an interesting idea. I do notice audio stuttering when I play at 54 FPS, but if anything, that warns me that I need to check my settings in order to get back to a proper 60 FPS.

I don't know much about how audio works, but would a more user-friendly answer be to buffer 1 frame of audio, and then stretch the audio for 2 frames worth of time if the next frame isn't ready? This solution sounds much more complicated, but I would assume would provide a better casual gaming experience than having the audio constantly muting/unmuting during variable CPU load.

YoshiRulz:
I wasn't thinking about casual play when I proposed this. You wouldn't expect frame times to vary much for casual play—as you elude to, a spike in frame times would probably be the result of misconfiguring EmuHawk or of another program hogging resources. I was mainly concerned about RAM Search hurting my ears while glitch hunting.

PR #3866, opened 2024-03-09
Audio lag fix

FredTheHunterProgrammer:

Issue : https://github.com/TASEmulators/BizHawk/issues/3238

DISCLAIMER : I have done this for a school project. I'm not going to think this is going to be accepted at all, but if you could leave some comments I would be grateful.

Added 2 settings in the Sound Settings menu for muting the app in certain circumstances. 

The first one is to mute the app when in the background or unfocused and can be toggled on and off.
The second on is to mute the app when the FPS reaches a threshold chosen by the user (by default 56). It can be toggled on and off and the threshold can also be modified in the settings.

image

Check if completed:
- [ X] I have run any relevant test suites
- [X ] I, the committer, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2022-07-15) and am compliant

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@
 /packages
 
 launchSettings.json
+
+/Dist/copyBIOS.sh
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -257,6 +257,10 @@ namespace BizHawk.Client.Common
 
 		public bool MuteInBG { get; set; } = false;
 
+		public bool MuteOnLag { get; set; } = false;
+
+		public decimal FPSThreshold { get; set; } = 100;
+
 		// Lua
 		public RecentFiles RecentLua { get; set; } = new RecentFiles(8);
 		public RecentFiles RecentLuaSession { get; set; } = new RecentFiles(8);
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -255,6 +255,8 @@ namespace BizHawk.Client.Common
 		public string SoundDevice { get; set; } = "";
 		public int SoundBufferSizeMs { get; set; } = 100;
 
+		public bool MuteInBG { get; set; } = false;
+
 		// Lua
 		public RecentFiles RecentLua { get; set; } = new RecentFiles(8);
 		public RecentFiles RecentLuaSession { get; set; } = new RecentFiles(8);
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -259,7 +259,7 @@ namespace BizHawk.Client.Common
 
 		public bool MuteOnLag { get; set; } = false;
 
-		public decimal FPSThreshold { get; set; } = 100;
+		public int FPSThreshold { get; set; } = 56;
 
 		// Lua
 		public RecentFiles RecentLua { get; set; } = new RecentFiles(8);
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -3302,7 +3302,10 @@ namespace BizHawk.Client.EmuHawk
 				UpdateToolsAfter();
 			}
 
-			if(Config.MuteInBG) {
+
+			// Mutes the game if EmuHawk is unfocused or minimized
+			// An option in the sound settings needs to be turned on for this to work.
+			if (Config.MuteInBG) {
 				if(WindowState == FormWindowState.Minimized || Form.ActiveForm == null) {
 					Sound.StopSound();
 				}
@@ -3312,6 +3315,20 @@ namespace BizHawk.Client.EmuHawk
 			} else { 
 				Sound.StartSound();
 			}
+
+			// Mutes the game if the current FPS reaches the threshold chosen by the user or lower
+			// An option in the sound settings needs to be turned on for this to work.
+			if(Config.MuteOnLag) {
+				if (_lastFps <= (double)Config.FPSThreshold)
+				{
+					Sound.StopSound();
+				}
+				else
+				{
+					Sound.StartSound();
+				}
+			}
+			
 
 			Sound.UpdateSound(atten, DisableSecondaryThrottling);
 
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -45,3 +45,5 @@
 /packages
 
 launchSettings.json
+
+/Dist/copyBIOS.sh
```
YoshiRulz, line 49: I presume this and `{firmware,lua}.zip` are things you were using for debugging?

src/BizHawk.Client.Common/config/Config.cs
```diff
@@ -255,6 +255,12 @@ public void ResolveDefaults()
 		public string SoundDevice { get; set; } = "";
 		public int SoundBufferSizeMs { get; set; } = 100;
 
+		public bool MuteInBG { get; set; } = false;
+
+		public bool MuteOnLag { get; set; } = false;
+
+		public int FPSThreshold { get; set; } = 56;
```
YoshiRulz, line 262: Needs to be a percentage; not all consoles run at 60 fps.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3302,7 +3302,36 @@ private void StepRunLoop_Core(bool force = false)
...
+			if(Config.MuteOnLag) {
+				if (_lastFps <= (double)Config.FPSThreshold)
+				{
+					Sound.StopSound();
+				}
+				else
+				{
+					Sound.StartSound();
+				}
+			}
```
YoshiRulz, line 3330: Assuming this logic is correct, it can be simplified *a lot*.

src/BizHawk.Client.EmuHawk/config/SoundConfig.Designer.cs
```diff
@@ -28,330 +28,409 @@ protected override void Dispose(bool disposing)
 		/// </summary>
 		private void InitializeComponent()
 		{
-			this.Cancel = new System.Windows.Forms.Button();
```
YoshiRulz, line 31: Sorry you had to deal with the Designer. If this feature is accepted, I'll rewrite this in a declarative style, then you can rebase and have a smaller diff.

src/BizHawk.Client.EmuHawk/Program.cs
```diff
@@ -31,8 +31,13 @@ internal static class Program
 		static Program()
 		{
 			// This needs to be done before the warnings/errors show up
+
```
YoshiRulz, line 34: Blank lines here and elsewhere are presumably holdouts from debugging? If this feature is accepted, you should remove whitespace changes.

src/BizHawk.Client.EmuHawk/config/SoundConfig.cs
```diff
@@ -145,5 +153,14 @@ private void nudRWFF_ValueChanged(object sender, EventArgs e)
...
 			tbRWFF.Value = (int)nudRWFF.Value;
 		}
+
+		private void muteOnLag_CheckedChanged(object sender, EventArgs e)
+		{
+			if(cbMuteOnLag.Checked) { 
+				FpsThresholdNumeric.Enabled = true;
+			} else { 
+				FpsThresholdNumeric.Enabled = false; 
+			}
```
YoshiRulz, line 163:
```suggestion
			FpsThresholdNumeric.Enabled = ((CheckBox) sender).Checked;
```

YoshiRulz (changes requested):
Thank you for working on this! When I opened the feature request, I had imagined it would be difficult to impossible under the current system, so seeing a proof-of-concept is a huge relief.

.gitignore
FredTheHunterProgrammer: Yes I was using copyBIOS.sh for testing but it wasn't useful anymore so I removed it. I will also remove it from gitignore. for ``{firmware,lua}.zip`` it's not something I added though so I'm not sure

src/BizHawk.Client.EmuHawk/MainForm.cs
FredTheHunterProgrammer: How is that?

src/BizHawk.Client.Common/config/Config.cs
FredTheHunterProgrammer: Just changed it now. Should I make it so the base FPS changes depending on the console? Because I'm not sure how to do that really. I still have 60fps hardcoded in for now

lua.zip
YoshiRulz, line 1: Remove these empty files too please.

src/BizHawk.Client.Common/config/Config.cs
YoshiRulz: You can get the loaded system's nominal framerate as a scalar like this:
https://github.com/TASEmulators/BizHawk/blob/5bf9d387b4fe342a9e9102e6a61cf3379affe7cb/src/BizHawk.Client.EmuHawk/MainForm.Events.cs#L953
(As the comment suggests, there's no helper method, so you can just copy-paste that. You probably don't even need to cache the result.)

FredTheHunterProgrammer: Should be fixed now, thanks for the help! And sorry for the delay, I'm very busy with school this week

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3302,6 +3302,28 @@ private void StepRunLoop_Core(bool force = false)
...
+				Sound.StopSound();
+			}
+			else
+			{
+				Sound.StartSound();
+			}
+
+			// Mutes the game if the current FPS reaches the threshold chosen by the user or lower
+			// An option in the sound settings needs to be turned on for this to work.
+			if (Config.MuteOnLag && _lastFps <= PlatformFrameRates.GetFrameRate(Emulator.SystemId, Emulator.HasRegions() && Emulator.AsRegionable().Region is DisplayType.PAL) * ((double)Config.FPSThresholdPercentage / 100))
```
CasualPokePlayer, line 3318: PlatformFrameRates are not intended to query the actual core's framerate per se (they are rather "political numbers"). The core's IVideoProvider provides this kind of information and should be used.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3302,6 +3302,28 @@ private void StepRunLoop_Core(bool force = false)
 				UpdateToolsAfter();
 			}
 
+			// Mutes the game if EmuHawk is unfocused or minimized
+			// An option in the sound settings needs to be turned on for this to work.
+			if (Config.MuteInBG && (WindowState == FormWindowState.Minimized || Form.ActiveForm == null))
```
YoshiRulz, line 3307: Does this account for a tool window having focus?

PR #3869, opened 2024-03-13
Replace `SortedSet` with `SortedList<T>` in `InputRoll`

YoshiRulz:
Following up on https://github.com/TASEmulators/BizHawk/commit/6d40c08c3c1404a217617c0bd237ba58f1ada76c#r139706953. Turns out I'd [already had that idea](https://github.com/TASEmulators/BizHawk/commit/67ba0e4f238a323a90a94dd3d25f7e9a08331830) some years ago (time flies) but never finished it.

I've tested this by trying a few of TAStudio's features and the assertion was never hit, but it deserves more scrutiny, and the `RemoveAll` call that was the impetus for all this will need reprofiling.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/Cell.cs
+++ b/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/Cell.cs
@@ -1,11 +1,14 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Diagnostics;
+
+using BizHawk.Common;
 
 namespace BizHawk.Client.EmuHawk
 {
 	/// <summary>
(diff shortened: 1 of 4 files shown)


YoshiRulz:
Cherry-picked recent changes from `master`, and made 1 more commit. I'm done messing with the selection implementation now.

re: nullability, I think it's worth trying to add NRT annotations, but after this is merged. I believe the comparison/order impl. currently doesn't assume its operands are non-null, but `IComparable` can.

adelikat (approved):
Approved, these are nice performance improvements

YoshiRulz:
Am I okay to rebase and merge this @Morilli?

Morilli (approved):
Have rebased this myself to be able to test the changes and performance looks very good.

PR #3880, opened 2024-03-27
Enable nullability for `Cell` and co. and clean up

YoshiRulz:
Starting on https://github.com/TASEmulators/BizHawk/pull/3869#issuecomment-1999636215. This builds upon that PR and the subsequent commits through to https://github.com/TASEmulators/BizHawk/commit/d67b1c1d237b151e6be8fa9dcaf2247768253f33, mainly https://github.com/TASEmulators/BizHawk/commit/4f144c74d47021018bc3252393db8685d98d58c7.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/Cell.cs
+++ b/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/Cell.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿#nullable enable
+
+using System;
 using System.Diagnostics;
 
 using BizHawk.Common;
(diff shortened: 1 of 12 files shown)


Morilli:
I do wonder, is this even the correct direction to go? As in, you made `Column` and `RowIndex` explicitly nullable, but could we make them explicitly not-null? What's even the realistic usecases for constructing a (non-null) `Cell` object with either null `RowIndex` or null `Column`?

If either of those (or both) could be non-null, it would most likely simplify the code.

YoshiRulz:
I agree, but in practice they both are set to null sometimes, for example:
https://github.com/TASEmulators/BizHawk/blob/2206571ab8c4a793f036af45dc137c27ced62c54/src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/InputRoll.cs#L1093-L1097

Morilli:
Yeah I've seen those cases in the code, but this one for example could just be replaced with `CurrentCell = null`, right?

YoshiRulz:
¯\\\_(ツ)\_/¯

YoshiRulz:
I think this should be merged, and if the simplification discussed above is possible, it can be done later.

PR #3883, opened 2024-03-28
Replace UniformInfos with init-only UniformInfo

Morilli:
Seems like a simple simplification of the code, but I'm not sure if it could break anything.

Check if completed:


CasualPokePlayer:
FWIW, this would break things for D3D9 where the vertex and pixel shaders both have uniforms of the same name (for OpenGL these are just considered the same, hence Sole existing, but for D3D land these are considered separate uniforms).

Isn't relevant anymore since the code was entirely refactored.

PR #3886, opened 2024-04-01
Only render on new frame

Morilli:
Not sure if it's safe to do this at all, but this gives a significant fps boost when watching bad apple.


YoshiRulz:
ApiHawk surfaces still need to be rendered, but `MainForm` doesn't know anything about that, only `DisplayManager`.

PR #3889, opened 2024-04-08
Update Genesis Plus GX Core

SergioMartin86:
This is a proposal to update the current GPGX core with one that is up-to-date with the current [upstream](https://github.com/ekeeke/Genesis-Plus-GX).  The new core source, containing minimal changes to interface with BizHawk, is stored in https://github.com/TASEmulators/Genesis-Plus-GX/

The goals are: 
* Update the core to the latest advances of upstream, including additional sound chip emulation (e.g., [OPLL](https://github.com/ekeeke/Genesis-Plus-GX/blob/master/core/sound/opll.h), [YM3438](https://github.com/ekeeke/Genesis-Plus-GX/blob/master/core/sound/ym3438.h))
* Make sure core synchronization with upstream is guaranteed both when running inside BizHawk, as well as natively compiled (my personal main motivation, as I plan to use this core for botting).
* Preserve all BizHawk features associated with this core
* Add any new features that might be useful (deep in-cpu freeze, sprite always-on-top, etc)
* Ensure future resyncs with upstream are as smooth as possible.

This is still a work in progress. At this moment (4/8/24), the waterboxed core is compiling correctly, but I still haven't even tried to run it in BizHawk. I open this pull request early for reviewers to make early comments and suggestions.

Check if completed:

----------------

**Update April 11th:**

Tasks Done:

- Abstracted out all file and CD access routines in the core so that both waterboxed and native variants can work.
- The native compilation passes tests for Genesis, Sega CD, SMS and GG (pass means: produces the same result as upstream. i.e., the work memory contains exactly the same values).
- In Bizhawk, manual Genesis and Sega CD tests pass (i.e., a given movie produces the same result as the native upstream core execution)
- Added setting to select the Nuke.YKT variants of the YM2413 and YM2612 sound cores
image
- Added functionality in the core for deep freezing variables in RAM (i.e., the freeze values will be re-poked every time a write is made in the m86k CPU)

To-dos:
- [x] Add LUA interface to set the values for deep freezing (useful for camhacks). I've got no idea how to do this. 
- [x] Add functionality and setting for "Sprites always on top" (useful for camhacks)
- [x] Figure out why the waterboxed core desyncs wrt the native build on the exact same inputs
- [x] Check new sound cores (do they work, do they sound as expected?, do the menu settings operate as expected?)


vadosnaprimer:
GUI for deep-freeze is not needed, because it will only be used in lua scripts, so it only needs to be exposed as lua API, tho I dunno how APIHawk may be involved. @YoshiRulz?

waterbox/gpgx/cinterface/cinterface.c
```diff
@@ -528,50 +542,155 @@ GPGX_EX int gpgx_init(const char* feromextension,
...
+	deepfreeze_list_size = 0;
+#endif
+
+	/**
+	 * Allocating large buffers
+	 */
+
+	// cart_hw/areplay.h
+
+	action_replay.ram = alloc_plain(sizeof(uint8) * 0x10000);
```
CasualPokePlayer, line 559: What's the point of all these alloc_plain calls? These seem to all make it so there's more unneeded changes in the submodule.

To be clear, alloc_plain is mostly useless, it came at a time before `malloc` actually worked. `alloc_sealed` similarly is less useful nowadays, due to dirty detection making it not useful for anything not actually changed during the core's lifetime (init time doesn't count here, since Seal() occurs after that).

src/BizHawk.Emulation.Cores/Consoles/Sega/gpgx64/GPGX.ISettable.cs
```diff
@@ -320,6 +328,8 @@ public LibGPGX.InitSettings GetNativeSettings(GameInfo game)
 					InputSystemB = SystemForSystem(ControlTypeRight),
 					Region = Region,
 					ForceSram = game["sram"],
+					YM2612SoundChip = YM2612SoundChip,
+			        YM2413SoundChip = YM2413SoundChip
```
CasualPokePlayer, line 332: Tabbing

waterbox/gpgx/cinterface/cinterface.c
SergioMartin86: Alright, so I will revert the stack->heap conversions back to the core originals. I thought this transformation was necessary from looking at the current port, but it seems not anymore.

CasualPokePlayer: To be clear, gpgx was the first waterboxed core, so it contains various practices which are antiqued by the modern waterbox setup.

Also variables in the global scope are *not* in the stack (*local* variables, i.e. variables local to the function they are declared within, are usually what's referred to for variables in the stack)

vadosnaprimer:
> * Add LUA interface to set the values for deep freezing (useful for camhacks). I've got no idea how to do this.

Just wire it up here https://github.com/TASEmulators/BizHawk/blob/d415bf4ca067dee88c29c8c0740a073259c020a9/src/BizHawk.Client.Common/lua/LuaHelperLibs/GenesisLuaLibrary.cs

nattthebear:
Thanks for looking at this!  It's a much requested update and sorely needed.

Everything mentioned so far is on point about the GPGX core and its current state.  Historically, it's been through a lot.  It was an early non-waterbox core, and had a lot of bad practices just from that, and then as both a C-only core and one with known savestate issues, the guinea pig for the first waterbox POC (which was actually an x86 thing, and also had no C++ support).  The POC showed that we probably needed amd64 for waterbox, specifically the extra address space, so it was then stuck in limbo for quite some time and the eventual waterbox core was created by resyncing the old waterbox POC with some other changes that had happened since then... all in all, quite a mess.

The rundown of the various allocation types was great too; there's not much need for `alloc_plain` anymore and even `alloc_sealed` is a bit of a hard sell as dirty detection ended up being much better than I originally anticipated.

YoshiRulz:
I guess a Lua-only helper is fine? Ideally, freeze cheats would just work like that for all cores, but they don't. The sanctioned workaround is to use a mem hook. I imagine a dedicated API would be much faster than that, so you're welcome to add one—and don't worry about ApiHawk. Consumers of that can get the core instance with `[RequiredApi] GPGX? _coreInstance;`, ~~exactly as the Lua library does~~ exactly as you'll need to do in `GenesisLuaLibrary`. And you also don't have to worry about getting the perfect API shape, since we can kick the can down the road until a second core gets the same functionality.

src/BizHawk.Client.Common/lua/LuaHelperLibs/GenesisLuaLibrary.cs
```diff
@@ -62,5 +67,19 @@ public void SetLayerBgW(bool value)
...
+		{
+			return gpgx.AddDeepFreezeValue(address, value);
+		}
+
+		[LuaMethodExample("genesis.clear_deepfreeze_list();")]
+		[LuaMethod("clear_deepfreeze_list", "Clears the list of deep frozen variables")]
+		public void ClearDeepFreezeList()
+		{
+			gpgx.ClearDeepFreezeList();
+		}
```
vadosnaprimer, line 82: Is it possible to allow removing freeze from individual frozen addresses? Also maybe a getter for entry count?

SergioMartin86: yes, absolutely, although my rudimentary knowledge of C# is the limiting factor for me here.

CasualPokePlayer:
Similarly on "unneeded changes," parts of the upstream repo seemed to be axed in our fork (e.g. gx/libretro/wiki/etc. folders). Please do not do this. Restore them back to matching upstream. Removing them just increases the diff size between us and upstream and will cause merge conflicts down the line whenever we want to sync back to upstream.

vadosnaprimer:
Yeah if some change looks too intrusive, maybe the hawk side could be tweaked to make it more gentle?

CasualPokePlayer:
Leaving in unused files in the submodule shouldn't require any Hawk-side changes in the first place: the unneeded files just aren't included within the Makefile so they just don't do anything, and it doesn't matter much if non-code files remain in place as-is.

src/BizHawk.Client.Common/lua/LuaHelperLibs/GenesisLuaLibrary.cs
```diff
@@ -72,14 +67,18 @@ public void SetLayerBgW(bool value)
 		[LuaMethod("add_deepfreeze_value", "Adds an address to deepfreeze to a given value. The value will not change at any point during emulation.")]
 		public int AddDeepFreezeValue(int address, byte value)
 		{
-			return gpgx.AddDeepFreezeValue(address, value);
+			// Here we'd need to get access to the gpgx core
```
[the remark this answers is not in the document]
YoshiRulz:
```csharp
[RequiredService]
private GPGX/*?*/ gpgx { get; set; }
```

vadosnaprimer:
I meant code changes.

SergioMartin86:
> I meant code changes.

At the moment I'm doing both: recover all original (unused) files **and** reverting all changes that ended up being unnecessary. For extreme `git diff` minimality, we could move the entire testbed to a new repo and include it only as a subproject.

SergioMartin86:
Alright, I'm done with reducing the number of diffs wrt upstream

vadosnaprimer:
> For extreme `git diff` minimality, we could move the entire testbed to a new repo and include it only as a subproject.

How often would tesbed change compared to the forked core?

SergioMartin86:
> > For extreme `git diff` minimality, we could move the entire testbed to a new repo and include it only as a subproject.
> 
> How often would tesbed change compared to the forked core?

Ideally, never. 

In practice, if a new file is ever added, one might need to modify the corresponding `meson.build` file.

vadosnaprimer:
I noticed upstream is already included into testbed as a submodule, and now we're kinda including all that into hawk, but with a tweaked core copy. I think it would be better to turn testbed into a specific repo that includes whatever version of the core is needed for your work, while also keeping the tasemus fork of the core even more clean.

SergioMartin86:
> I noticed upstream is already included into testbed as a submodule, and now we're kinda including all that into hawk, but with a tweaked core copy. I think it would be better to turn testbed into a new repo that includes whatever version of the core is needed for your work, while also keeping the bizhawk submodule even more clean.

I agree. Do I have permission to create a repo here or shall you do it?

vadosnaprimer:
If it's not needed for bizhawk or standalone general tasing, then a personal repo makes the most sense.

SergioMartin86:
> If it's not needed for bizhawk or standalone general tasing, then a personal repo makes the most sense.

Well, the tested is part of the CI pipeline. If we remove it, we lose the checks that makes sure the changes we made for BK did not affect emulation.    Of course we can remove it, but it's kind of a future risk

CasualPokePlayer:
Does the CI actually build an entire waterbox setup and contain the waterboxhost?

SergioMartin86:
> Does the CI actually build an entire waterbox setup and contain the waterboxhost?

No. It builds the new core natively, runs a series of movies and compares the resulting work RAM hash with the upstream core, to make sure that (a) it still works and (b) both cores produce the same result. This helps in making sure none of the changes I made affects emulation at all (at least as far as the set of tests allow).

src/BizHawk.Client.Common/lua/LuaHelperLibs/GenesisLuaLibrary.cs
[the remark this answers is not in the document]
SergioMartin86: Tried this and didn't compile. I guess it is supposed to be a placeholder until somebody implements it, right?

vadosnaprimer:
> Well, the tested is part of the CI pipeline. If we remove it, we lose the checks that makes sure the changes we made for BK did not affect emulation. Of course we can remove it, but it's kind of a future risk

If you have a personal repo of testbed with CI, and it submodules the hawk port and upstream, would it not work for comparing their states? Testbed files are just a top-level folder, so it shouldn't matter whether the core or the testbed is the submodule, and which of them is the main project.

SergioMartin86:
> > Well, the tested is part of the CI pipeline. If we remove it, we lose the checks that makes sure the changes we made for BK did not affect emulation. Of course we can remove it, but it's kind of a future risk
> 
> If you have a personal repo of testbed with CI, and it submodules the hawk port and upstream, would it not work for comparing their states? Testbed files are just a top-level folder, so it shouldn't matter whether the core or the testbed is the submodule, and which of them is the main project.

Yes, this would work for me. However, I consider it'd be a nice thing to have in general for bizhawk in the future: that all subcores are proofed against upstream.  For the time being I removed the testbed from the subproject so it's pretty clean now

vadosnaprimer:
Can you please reduce the amount of commits to those that can be combined by some generalized criterion? For example new features can in theory be separate commits, but I personally wouldn't mind if all the extra work is just one huge commit, to be able to see all the actual changes at once and evaluate them.

SergioMartin86:
> Can you please reduce the amount of commits to those that can be combined by some generalized criterion? For example new features can in theory be separate commits ~but I personally wouldn't mind if all the extra work is just one huge commit, to be able to see all the actual changes at once and evaluate them.~
> 
> EDIT: Nevermind the second sentence, github provides a way.

In summary, I did these four things:


- Adapted the core to BK
- Added deepfreeze
- Added sprites always on top
- Reverted unnecessary changes to minimize git diff

The individual commit messages and changes are kinda clear as to which of these belong to, yet they're all intertwined. It'd be a nightmare to bulk them and reorder them by concept, I wouldn't know where to start. I'm afraid "one huge commit" is the best clustering I (git) could ever come up with in this case.

vadosnaprimer:
I'll check them out tomorrow then and see if there's a nice way to group them.

YoshiRulz:
If the automated tests don't involve Waterbox, why not contribute them upstream?

SergioMartin86:
> If the automated tests don't involve Waterbox, why not contribute them upstream?

Allow me to explain the purpose of tests I added a bit more in detail. There are 3 projects involved in some way or another in this PR:

**Upstream GPGX Core**
- Build type: Native
- Run mode: Native
- Has CI testing? No.
- Should they have CI testing? Yes. To make sure the code builds and some example movies to make sure the different consoles/devices run properly and do not crash. Sync is not testable becuase this is the source emulator. 
- Will I contribute to this? Probably not. I had a couple attempts to contribute but I was always shot down. The current maintainer is not interested much in TASing affairs.

**TASEmulators/GPGX Core**
- Build type: Native
- Run mode: Native
- Has CI testing? Yes, the testbed I added (well, now I removed it for git diff cleanliness)
- Shoult it have CI Testing? Yes. For exactly the same reasons as the upstream one. Plus: to make sure the changes I introduced to support Bizhawk do not affect sync wrt the upstream. 
- Will I contribute to this? I already have.

**Waterboxed TASEmulators/GPGX Core**
- Build type: Waterboxed
- Run mode: Inside Bizhawk
- Has CI testing? No. (only the C# bizhawk part has build testing while the core is distributed pre-compiled)
- Should it have CI Testing? Yes. To make sure it builds correctly when interfaced with the waterbox. Probably could also use the same pipeline to do nighly builds and stop including compiled waterbox libs in the repository.
- Will I contribute to this? I'd be happy to help.

vadosnaprimer:
I tried running Gargoyles and the game resets after the first gameplay demo, while it should start the next level demo.

изображение

src/BizHawk.Client.Common/lua/LuaHelperLibs/GenesisLuaLibrary.cs
vadosnaprimer: While testing I got a couple more ideas. Function that returns current frozen addresses as a table (with frozen values), and more importantly ability to freeze the value that's currently in the cell without having to set it manually. I'll see what I can do...

SergioMartin86:
> I tried running Gargoyles and the game resets after the first gameplay demo, while it should start the next level demo. Left is updated core, right is current master. Upstream (latest libretro) seems to work fine.

Fixed in [201b486](https://github.com/TASEmulators/BizHawk/pull/3889/commits/201b486982ed110e0d9357951797061996432209)

vadosnaprimer:
I think SMS/GG doesn't have to be in this PR. Can happen later whenever ready.

SergioMartin86:
> I think SMS/GG doesn't have to be in this PR. Can happen later whenever ready.

Agreed. Removing from task list -- will create that PR as soon as we merge this one

SergioMartin86:
Someone that's more of an expert should check whether the Nuke sound core is being correctly setup/used.

SergioMartin86:
I'd like to post this question for discussion. In the hypothetical case we never find why waterboxing this core causes it to desync wrt the native core, would you consider allowing me to add a sync setting that allows the core to run natively? 

- The option will be set waterboxed by default. 
- All changes are made in bk (add the setting, add the makefile and msvc project, and the `if` statement to choose wbx or native)
- I already done this work in a branch and it works perfectly
- Only people who need absolute sync (me, when botting) can choose the native run, and only for reproduction / casual play purposes (not rerecording)

This would relieve me from having to create and maintain my own fork of BK just for the sake of avoiding this desync.

YoshiRulz:
I think that's a question for TASVideos judges, since if the 2 builds differ, then I'd expect that *at most 1* of them would produce console-verifiable movies.

SergioMartin86:
Update on the desync issue. It seems like the wbx core syncs perfectly with the native one. We haven't found the reason one it didn't before, but the main hypothesis is that I did not properly rebuild Bizhawk on windows and was observing older behavior. This is strange because I was able to reproduce the `Gargoyles` problem AND its solution on windows while the `Dino Runner` desync persisted. Could it have been an issue with the BK side (libGPGX.cs)?Anyway, it seems to have cleared up now.

SergioMartin86:
> Someone that's more of an expert should check whether the Nuke sound core is being correctly setup/used.

Adding this as a task, so we don't forget.

[ ] Check new sound cores

eugene-s-nesdev:
Can you add more audio options like chip types? 
https://docs.libretro.com/library/genesis_plus_gx/#audio

Sega Master System FM:
-MAME (YM2413)
-Nuked (YM2413)

Mega Drive/Genesis FM:
-MAME (YM2612)
-MAME (ASIC YM3438)
-MAME (Enhanced YM3438)
-Nuked (YM2612)
-Nuked (YM3438)

SergioMartin86:
> Can you add more audio options like chip types? https://docs.libretro.com/library/genesis_plus_gx/#audio
> 
> Sega Master System FM: -MAME (YM2413) -Nuked (YM2413)
> 
> Mega Drive/Genesis FM: -MAME (YM2612) -MAME (ASIC YM3438) -MAME (Enhanced YM3438) -Nuked (YM2612) -Nuked (YM3438)

Thanks for the info. The way sound chip selection is done in the code is super confusing and I had no idea there were this many options. I'll look through libRetro's implementation and adapt it to this core.

waterbox/gpgx/cinterface/cinterface.c
```diff
@@ -657,10 +661,70 @@ GPGX_EX int gpgx_init(const char* feromextension,
...
+        break;
+
+	  case YM2413_NUKED: 
+	    config.opll = 1;
+		config.ym2413 = 0;
+        break;
+	}
+
+	// Selecting FM Sound chip to use for Genesis / Megadrive / CD emulation
+    switch (settings->SMSFMSoundChip)
```
vadosnaprimer, line 700: Copypaste error?

SergioMartin86: I have an issue with tabbing/spaces with my IDE I still haven't been able to figure out. Feel free to fix

SergioMartin86:
> Can you add more audio options like chip types? https://docs.libretro.com/library/genesis_plus_gx/#audio
> 
> Sega Master System FM: -MAME (YM2413) -Nuked (YM2413)
> 
> Mega Drive/Genesis FM: -MAME (YM2612) -MAME (ASIC YM3438) -MAME (Enhanced YM3438) -Nuked (YM2612) -Nuked (YM3438)

Implemented 
image

eugene-s-nesdev:
For unknown reasons all 3438 cores (mame and nuked) works like 2612 cores, and has audible "ladder effect", but 3438 shouldn't have it. Retroarch's genplusgx core works correctly.
Please test it on EarthWorm Jim 2 (Debug menu - bgm track 0 "Moonlite sonata")
Debug menu: Run game, enter 1'st level, set pause, then А, С, С, А, В, А, В, Left

https://youtu.be/PuK_Rjf20iE

SergioMartin86:
> For unknown reasons all 3438 cores (mame and nuked) works like 2612 cores, and has audible "ladder effect" (YM2612 DAC noise), but 3438 shouldn't have it. https://en.wikipedia.org/wiki/Yamaha_YM2612#Low_volume_distortion_of_the_built-in_DAC Retroarch's genplusgx core works correctly.
> 
> Please test it on EarthWorm Jim 2 (Debug menu - bgm track 0 "Moonlight sonata") Debug menu: Run game, enter 1'st level, set pause, then А, С, С, А, В, А, В, Left
> 
> https://youtu.be/PuK_Rjf20iE

thanks, now I have a clear test to decide whether the 3438 is correctly implemented. I'll come back when I fixed the issue

SergioMartin86:
> For unknown reasons all 3438 cores (mame and nuked) works like 2612 cores, and has audible "ladder effect" (YM2612 DAC noise), but 3438 shouldn't have it. https://en.wikipedia.org/wiki/Yamaha_YM2612#Low_volume_distortion_of_the_built-in_DAC Retroarch's genplusgx core works correctly.
> 
> Please test it on EarthWorm Jim 2 (Debug menu - bgm track 0 "Moonlight sonata") Debug menu: Run game, enter 1'st level, set pause, then А, С, С, А, В, А, В, Left
> 
> https://youtu.be/PuK_Rjf20iE

ok this should be fixed now. It was a problem with the a BK -> C interface settings mismatch. I tested Moonlight sonata and it sounds like in your video

eugene-s-nesdev:
This is error:



It should be:
YM3438_MAME_DISCRETE
YM3438_MAME_ENCHANCED
YM3438_NUKED
YM2612_MAME
YM2612_NUKED

https://docs.libretro.com/library/genesis_plus_gx/#audio

SergioMartin86:
> This is error:
> 
> 
> 
> It should be: YM3438_MAME_DISCRETE YM3438_MAME_ENCHANCED YM3438_NUKED YM2612_MAME YM2612_NUKED
> 
> https://docs.libretro.com/library/genesis_plus_gx/#audio

I'll leave this fix to somebody more knowledgeable than me. The way these are all named in GPGX and what's being said in libretro makes everything very confusing for me

eugene-s-nesdev:
Current dc52cc09b build has strange issues:

YM2612_MAME_DISCRETE - sounds like 2612, strong ladder effect
YM2612_MAME_ENCHANCED - sounds like 3438, weak ladder effect
YM2612_NUKED  - sounds like 3438, weak ladder effect
YM3438_MAME - sounds like 3438, weak ladder effect
YM2612_NUKED - sounds like 3438, weak ladder effect

YoshiRulz:
Works on NixOS. Not sure what you wanted out of the Lua functions, they seem fine.

vadosnaprimer:
Listing frozen addresses and their values, also unfreezng individual ones.

PR #3890, opened 2024-04-11
make default.nix and shell.nix pure by adding checksums and system

bugQ:

The only things keeping both `default.nix` and `shell.nix` from evaluating in [pure evaluation mode](https://nixos.org/manual/nix/stable/command-ref/conf-file#conf-pure-eval) are the lack of checksums on calls to `builtins.fetchTarball`, and the lack of a `system` attribute to be passed down to their corresponding `import` statements. The associated URLs have not been changed in 5 months and thus don't seem likely to change in the near future either (except for the first, when 24.05 drops; I'd be happy to amend this PR if that happens before it is merged).

Purity becomes much more important if and when BizHawk gets published to [nixpkgs](/NixOS/nixpkgs) or as a standalone flake, but in the meantime would still make users like me who consume and evaluate these build scripts directly happier, and at a trivial maintenance cost.

Check if completed:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,8 +1,8 @@
 { system ? builtins.currentSystem
 , pkgs ? import (fetchTarball {
-		url = "https://github.com/NixOS/nixpkgs/archive/23.11.tar.gz";
-		sha256 = "1ndiv385w1qyb3b18vw13991fzb9wg4cl21wglk89grsfsnra41k";
-	}) { inherit system; }
+	url = "https://github.com/NixOS/nixpkgs/archive/23.11.tar.gz";
+	sha256 = "1ndiv385w1qyb3b18vw13991fzb9wg4cl21wglk89grsfsnra41k";
+}) { inherit system; }
 , lib ? pkgs.lib
 , stdenv ? pkgs.stdenvNoCC
 # infrastructure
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -1,4 +1,7 @@
-{ pkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/23.11.tar.gz") {}
+{ pkgs ? import (fetchTarball {
+		url = "https://github.com/NixOS/nixpkgs/archive/23.11.tar.gz";
+		sha256 = "1ndiv385w1qyb3b18vw13991fzb9wg4cl21wglk89grsfsnra41k";
+	}) {}
```
YoshiRulz, line 4: Why is this marked as a draft?

Unindent here and for `nixGLChannel` and I'll merge this immediately.

bugQ: I marked it as a draft after doing some local testing and finding that this change alone does not fully satisfy purity. I'll update the PR shortly.

What kind of indentation did you have in mind? The double indent does look a bit odd, but to me, it seemed more odd to put the ending punctuation at the same level as the top-level opening brace.

YoshiRulz: That's the convention I've been following, as you can see with `hawkSourceInfoDevBuild`. Once the official style guide is finished I'll consider adopting it.

PR #3935, opened 2024-06-03
Refactor IControlMainform handling in Mainform

### LINKED ISSUE #3870: Loading a tasproj does not dispose the rewind buffer
### Summary
When loading a tasproj from within TAStudio, the current rewind buffer is not disposed, meaning there will now be a rewind savestate buffer and the buffers from TAStudio loaded at the same time. That has no advantage whatsoever as rewind is disabled while TAStudio is active, and the buffer won't get filled with states until TAStudio is closed. Having that rewind buffer loaded the entire time is a waste of RAM.
This does not happen when loading TAStudio for the first time; the current rewind buffer is correctly disposed then.

Here's the order of operations in regards to `MemoryBlock` allocations + disposals when **opening TAStudio**, assuming rewind is currently enabled:
1. A new `MemoryBlock` is allocated for the `Current`, `Recent` and `Gaps` areas respectively.
2. Those 3 `MemoryBlock`s are then immediately disposed.
3. A new `MemoryBlock` is allocated for the `Current`, `Recent` and `Gaps` areas respectively.
4. The current rewind buffer is disposed.
5. A new rewind buffer is allocated.
6. The current (new) rewind buffer is disposed.

And what happens when a tasproj is loaded from within TAStudio (for example by just pressing `New`):
1. A new `MemoryBlock` is allocated for the `Current`, `Recent` and `Gaps` areas respectively.
2. Those 3 `MemoryBlock`s are then immediately disposed.
3. A new `MemoryBlock` is allocated for the `Current`, `Recent` and `Gaps` areas respectively.
4. The current rewind buffer is disposed.
5. The current `MemoryBlock`s for the `Current`, `Recent` and `Gaps` areas are disposed.
6. A new rewind buffer is allocated.

Clearly, the only problem here is that the newly created rewind buffer is not disposed again.
There is so much code involved in doing all these steps I honestly am afraid to touch any of it. Hopefully this documentation can help somehow.

Morilli:
- closes #3870
 
This changes the handling of `IControlMainform` tools loaded in the following way:
There is no longer a way for a tool to take or relinquish control over all mainform actions specified in that interface, instead, each of the `WantsToControlX` fields are queried for all loaded tools, and if one matches, that tool will be controlling this action.
That means that there could now be different loaded tools handling different actions, for example one tool controlling rewind, a different tool controlling savestates, etc.

Additionally, `CreateRewinder` now returns immediately if any tool has `WantsToControlRewind` set to true, for example TAStudio.
This still feels a bit arbitrary, every tool wanting to control rewind will have to manually disable rewind on load and re-enable it on close as TAStudio does, but this isn't a new issue introduced here, so...

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs
@@ -473,6 +473,19 @@ namespace BizHawk.Client.EmuHawk
 			return Load<T>(false);
 		}
 
+		public T FirstOrNull<T>(Predicate<T> condition) where T : class
+		{
+			foreach (var tool in _tools)
+			{
+				if (tool is T specialTool && condition(specialTool))
+				{
+					return specialTool;
+				}
+			}
+
+			return null;
+		}
+
 		/// <summary>
 		/// returns the instance of <paramref name="toolType"/>, regardless of whether it's loaded,<br/>
 		/// but doesn't create and load a new instance if it's not found
@@ -482,7 +495,7 @@ namespace BizHawk.Client.EmuHawk
 		/// you may pass any class or interface
 		/// </remarks>
 		public IToolForm/*?*/ LazyGet(Type toolType)
-			=> _tools.Find(t => toolType.IsAssignableFrom(t.GetType()));
+			=> _tools.Find(toolType.IsInstanceOfType);
 
 		internal static readonly IDictionary<Type, (Image/*?*/ Icon, string Name)> IconAndNameCache = new Dictionary<Type, (Image/*?*/ Icon, string Name)>
 		{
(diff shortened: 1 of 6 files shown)


src/BizHawk.Client.EmuHawk/tools/ToolManager.cs
```diff
@@ -473,6 +473,19 @@ public bool Has(Type toolType)
 			return Load<T>(false);
 		}
 
+		public T FirstOrNull<T>(Predicate<T> condition) where T : class
```
YoshiRulz, line 476: Type-check + `Predicate` seems weird. The caller should include `t is T &&` if that's desired.

Morilli: I agree this currently feels weird, but the function was specifically tailored to fit the needs of the callsite while staying somewhat generic. Having to cast twice, once in the predicate and once for casting the return value felt weird also ( like `(IControlMainform)Tools.FirstOrNull(static t => t is IControlMainform tool && tool.WantsToControlRewind);`).

YoshiRulz: You could also make `ToolManager` implement `IEnumerable<IToolForm>` and use regular LINQ at the callsites.

Issue #3505, opened 2022-12-28
Offer MAME core as a download, like FFmpeg


YoshiRulz:
is real big

vadosnaprimer:
Offering the whole MESS part is impossible because of the crazy infrastructure that we'd have to implement, and most MESS cores are worse than currently accepted on tasvideos. Particular MESS cores could work, like game and watch, or other things we don't have better options for. But they don't get too much demand.

PR #3937, opened 2024-06-04
Implement a rudimentary core/dll download form

Morilli:
- (resolves) #3505
 
This implements a very simple download form that can be used to download additional files into the `dll` directory, mainly intended to be used for large core downloads, such as the mame or encore cores. This, in addition to excluding those files from the release zips can help bring the initial download size of bizhawk down.

How it currently looks like (accessable from Tools->Download cores):
image

PRing for visibility and feedback. I think this could work as is, but probably needs additional support from the excluded cores to not feel bad when a user tries to load a ROM for a core that isn't downloaded and is then presented with an undecipherable error message.

Disclaimer: I do not take responsibility for the diff in `Mainform.Designer.cs`.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
@@ -57,6 +57,7 @@ namespace BizHawk.Client.EmuHawk
 			this.toolStripSeparator21 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.AutoloadLastSlotMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.SaveSlotSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
+			this.SelectSlot0MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.SelectSlot1MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.SelectSlot2MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.SelectSlot3MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
@@ -66,7 +67,6 @@ namespace BizHawk.Client.EmuHawk
 			this.SelectSlot7MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.SelectSlot8MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.SelectSlot9MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.SelectSlot0MenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.PreviousSlotMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.NextSlotMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.toolStripSeparator5 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
@@ -134,7 +134,6 @@ namespace BizHawk.Client.EmuHawk
 			this.toolStripMenuItem4 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.DisplayStatusBarMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.DisplayMessagesMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
-			this.toolStripSeparator8 = new BizHawk.WinForms.Controls.ToolStripSeparatorEx();
 			this.DisplayLogWindowMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.ConfigSubMenu = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.ControllersMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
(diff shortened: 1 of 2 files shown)


src/BizHawk.Client.EmuHawk/CoreDownloaderForm.cs
```diff
@@ -0,0 +1,108 @@
...
+using BizHawk.Common;
+using BizHawk.Common.PathExtensions;
+
+namespace BizHawk.Client.EmuHawk
+{
+	public partial class CoreDownloaderForm : Form
+	{
+		// TODO edit
+		private static readonly string BaseUrl = $"https://github.com/Morilli/BizHawk-extra/raw/{VersionInfo.MainVersion}/";
+		private static readonly string OutputPath = Path.Combine(PathUtils.DataDirectoryPath, "dll");
```
YoshiRulz, line 19: This is correct, but the dynamic loader for each core has to be changed to look here. (This is for NixOS where `BIZHAWK_HOME` and `BIZHAWK_DATA_HOME` aren't the same.)

src/BizHawk.Client.EmuHawk/CoreDownloaderForm.cs
```diff
@@ -0,0 +1,108 @@
...
+
+using BizHawk.Common;
+using BizHawk.Common.PathExtensions;
+
+namespace BizHawk.Client.EmuHawk
+{
+	public partial class CoreDownloaderForm : Form
+	{
+		// TODO edit
+		private static readonly string BaseUrl = $"https://github.com/Morilli/BizHawk-extra/raw/{VersionInfo.MainVersion}/";
```
YoshiRulz, line 18: Having FFmpeg in https://github.com/TASEmulators/ffmpeg-binaries makes sense because it doesn't change. Before hosting any core binaries in a GitHub repo like that, we need to plan ahead and decide on a versioning scheme, a way to check compatibility (in case it's downloaded manually), and an expiration date for URIs (#2235). We also need to decide how dev builds should behave.

src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
```diff
@@ -2758,7 +2766,8 @@ private void InitializeComponent()
 		private BizHawk.WinForms.Controls.ToolStripMenuItemEx AmstradCPCNonSyncSettingsToolStripMenuItem;
 		private BizHawk.WinForms.Controls.ToolStripSeparatorEx toolStripSeparator8;
 		private System.Windows.Forms.ToolStripMenuItem CaptureLuaMenuItem;
-		private System.Windows.Forms.ToolStripMenuItem AutosaveLastSlotMenuItem;
 		private ToolStripSeparatorEx toolStripSeparator24;
+		private System.Windows.Forms.ToolStripMenuItem DownloadCoresMenuItem;
```
YoshiRulz, line 2770: Revert unrelated changes and/or insert the menu item in `MainForm_Load`.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -4917,6 +4917,12 @@ private void SingleInstanceProcessArgs(string[] args)
 
 		private IRetroAchievements RA { get; set; }
 
+		private void DownloadCoresMenuItemClick(object sender, EventArgs e)
+		{
+			CoreDownloaderForm coreDownloaderForm = new CoreDownloaderForm();
```
YoshiRulz, line 4922:
```suggestion
			using CoreDownloaderForm coreDownloaderForm = new();
```

src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj
```diff
@@ -9,10 +9,10 @@
 		<ApplicationManifest>app.manifest</ApplicationManifest>
 		<Nullable>disable</Nullable>
 		<ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
+		<UseWindowsForms>true</UseWindowsForms>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="System.Net.Http" />
-		<Reference Include="System.Windows.Forms" />
```
YoshiRulz, line 15: Revert unrelated changes.

YoshiRulz (changes requested):

src/BizHawk.Client.EmuHawk/CoreDownloaderForm.cs
Morilli: What is `BIZHAWK_DATA_HOME` intended to be used for? In the core I only see references from the ffmpeg and RAIntegration downloader dialogs, and the gamedb loading code. I'm assuming `PathUtils.DllDirectoryPath` would be the correct one to use here, but what IS supposed to be put in `PathUtils.DataDirectoryPath`? Are the existing downloader forms also outputting into an incorrect location?

src/BizHawk.Client.EmuHawk/CoreDownloaderForm.cs
Morilli: Right, this is one of the things that need to be decided before this can work. In this implementation I was thinking of having a separate repository (just like ffmpeg-binaries) that hosts downloadable dlls for each released bizhawk version in a separate branch, so for release `2.9.2` you would have a `2.9.2` branch with the binaries from that release. Those would of course stay static after the release.

It would probably also be possible to fetch the binaries from the BizHawk repo itself using the release tag, but that feels a little more volatile in regards to future-proofness.
I see your point in that potentially breaking this download in the future because we don't control the url anymore would be bad, but I don't see how this case here differs from the ffmpeg download case.
Compatibility checking could be implemented but I personally wouldn't bother, because as long as we serve proper downloads compatibility should be guaranteed, and if people download stuff on their own, well, that's their own problem (you can already replace any non-bizhawk dlls and cores with different versions).

> We also need to decide how dev builds should behave.

I say ignore that case completely, dev builds can stay being full-size downloads and one dev `2.9.2` has little meaning when it could span an entire year of changes, so those downloads don't need to work (or be valid working files) for dev builds.

src/BizHawk.Client.EmuHawk/MainForm.cs
Morilli: Yeah funny story, `Show()` is non-blocking (versus the blocking `ShowDialog()`), so `using` the form disposes it immediately after it gets shown. (~~and I just realized I did the same with the `EditSubtitlesForm` and never tested it apparently~~)

src/BizHawk.Client.EmuHawk/CoreDownloaderForm.cs
YoshiRulz: `PathUtils.DllDirectoryPath` is read-only on NixOS.
Ideally, these optional components would be built from source and installed at the same time as EmuHawk, but I haven't looked into that for FFmpeg. (I had no idea RAIntegration used it, but IIRC that's a Windows-only component?)

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: For non-modal dialogs, inherit `IToolForm` and use `Tools.Load`

Morilli: I don't think that's necessary here or in general, the Form gets disposed on close anyway.

src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
Morilli: Dealing with the designer is giving me mild forms of permanent damage, I'll just quote from the file itself `do not modify the contents of this method with the code editor`. See also the disclaimer in the PR description.

src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj
Morilli: I'll extract this change to master for all projects

src/BizHawk.Client.EmuHawk/CoreDownloaderForm.cs
Morilli: Reading this again, is `BIZHAWK_DATA_HOME` shared between different versions of bizhawk? In that case this would actually pose a problem if non-static dlls were written to it.

YoshiRulz: Out-of-the-box, it points to e.g. `~/.local/share/emuhawk-monort-2.9.1`, so it's shared between instances of the same version.
I think you should use `DllDirectoryPath` and block the downloader on Linux. Non-NixOS distros can ship a lite and full package if there's demand.

PR #3941, opened 2024-06-08
PUAE core

### LINKED ISSUE #2003: Amiga core
There have been a few ports of the original UAE emulator, the only 2 active ones are WinUAE and FS-UAE.

WinUAE is Windows only and is developed in a sporadic way, so probably not the best option for porting (haven't checked if it separates UI and core).

FS-UAE is actively developed and has a community. Its author also pulls core updates from WinUAE every once in a while (it only happens in one direction I think, WinUAE isn't interested in co-op), so it kinda combines the best parts from the scene. Core separation is [probably there](https://github.com/FrodeSolheim/fs-uae/tree/master/libfsemu).

It may feel hard to set up amiga emulation, but here's what we've been doing with FS-UAE+libTAS:
http://tasvideos.org/forum/viewtopic.php?t=20837

Framebuffer is interesting, because most amiga games have small res, but some programs may enable hires. the emu renders at 752x572, and it's not pixel perfect iirc. All in all: to be figured out.

It had [issues](https://github.com/FrodeSolheim/fs-uae/pull/195) with libTAS savestates due to threading, but I haven't tested its internal savestates.

vadosnaprimer:
libretro-puae waterboxed

largely still wip, but ready for testing, so I want to be able to provide testers with automatic builds, and to let them create tickets

todo later:
- stereo slider
- resolution switching
- multiple floppies and swapper
- CD drive
- hard drive
- whdload
- controller type options for both ports
- cd32
- more crazy options

fix #2003


waterbox/uae/bizhawk.c
```diff
@@ -0,0 +1,91 @@
...
+	memcpy(f->base.VideoBuffer, retro_bmp, sizeof(retro_bmp) / sizeof(retro_bmp[0]));
+	last_mouse_x = f->MouseX;
+	last_mouse_y = f->MouseY;
+	sound_buffer = NULL;
+}
+
+ECL_EXPORT void GetMemoryAreas(MemoryArea *m)
+{
+	m[0].Data  = chipmem_bank.baseaddr;
+	m[0].Name  = "chipmem";
```
CasualPokePlayer, line 71: "chipmem" -> "Chip RAM"
"bogomem" -> "Slow RAM"
"fastmem" -> "Fast RAM"

waterbox/uae/Makefile
```diff
@@ -0,0 +1,274 @@
...
+	$(EMU)/archivers/lha/larc.c \
+	$(EMU)/archivers/lha/lhamaketbl.c \
+	$(EMU)/archivers/lha/lharc.c \
+	$(EMU)/archivers/lha/shuf.c \
+	$(EMU)/archivers/lha/slide.c \
+	$(EMU)/archivers/lha/uae_lha.c \
+	$(EMU)/archivers/lha/util.c \
+	$(EMU)/archivers/mp2/kjmp2.c
+
+ifeq ($(HAVE_MPEG2), 1)
```
Morilli, line 211: this `HAVE_MPEG2`, as well as the `STATIC_LINKING` below don't seem to get set anywhere, so you should probably remove those checks and keep/remove the branches based on the value. It looks like these libmpeg2 sources are not intended to be linked in, but all the below sources are.

vadosnaprimer: `HAVE_MPEG2` is used for cd32. Unsure what exactly `STATIC_LINKING` changes in the end result.

Morilli: So is this supposed to be passed to the makefile? I don't see this being set anywhere except for the makeile in `libretro-uae`, which isn't used.

vadosnaprimer: I don't plan to have such options. If the feature itself works, I'll just always enable it.

src/BizHawk.Emulation.Cores/Computers/Amiga/PUAE.cs
```diff
@@ -0,0 +1,261 @@
...
+			});
+
+			ControllerDefinition = InitInput();
+
+			if (!_puae.Init(args.Count, args.ToArray()))
+				throw new InvalidOperationException("Core rejected the rom!");
+
+			foreach (var s in filesToRemove)
+			{
+				//_exe.RemoveReadonlyFile(s);
```
nattthebear, line 73: Is this supposed to be commented out?

src/BizHawk.Emulation.Cores/Computers/Amiga/PUAE.cs
```diff
@@ -0,0 +1,261 @@
...
+		{
+			_syncSettings = lp.SyncSettings ?? new();
+
+			_puae = PreInit<LibPUAE>(new WaterboxOptions
+			{
+				Filename                   = "puae.wbx",
+				SbrkHeapSizeKB             = 5 * 512,
+				SealedHeapSizeKB           = 10 * 1024,
+				InvisibleHeapSizeKB        = 10 * 1024,
+				PlainHeapSizeKB            = 10 * 1024,
```
nattthebear, line 44: Although not strictly necessary, if possible we should trim the various heap sizes before release as it does improve performance a bit.

nattthebear (commented):
Looks good overall, left a few comments.

src/BizHawk.Emulation.Cores/Computers/Amiga/PUAE.cs
vadosnaprimer: The core seems to need some of those files during emulation, I wanted to figure out which ones I can release later.

src/BizHawk.Emulation.Cores/Computers/Amiga/PUAE.cs
vadosnaprimer: These are temporary until I figure out the maximum I need for the most beefy machine config.

PR #3946, opened 2024-06-11
`ArgParser` rewrite

YoshiRulz:
~~I'd still like to improve the documentation further, but this is functional already.~~ Documentation improved. If you have scripts for e.g. encoding please test them.

Don't merge without checking the commit hash in `README.md` (it should be of the commit immediately before the readme change).

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,8 @@ tl;dr:
 
 #### Passing command-line arguments
 
-EmuHawk takes some command-line options which aren't well-documented; you might be able to figure them out from [the source](https://github.com/TASEmulators/BizHawk/blob/78daf4913d4c8e47d24fc14d84ca33ddef913ed4/src/BizHawk.Client.Common/ArgParser.cs).
+EmuHawk takes some command-line options which are documented in [the source](https://github.com/TASEmulators/BizHawk/blob/2d37fc1f13afb0774629f16ffea5ff86d9b47951/src/BizHawk.Client.Common/ArgParser.cs).
+On Linux starting from 2.10, these can also be viewed offline with the usual `--help`.
 
 On Windows 8.1/10, it's easiest to use PowerShell for this. For example, to pass `--lua=C:\path\to\script.lua` as the first argument and `C:\path\to\rom.n64` as the second, navigate to the BizHawk install folder and run:
 ```pwsh
@@ -210,6 +211,7 @@ On Windows 8.1/10, it's easiest to use PowerShell for this. For example, to pass
 On Linux, you can pass arguments to `EmuHawkMono.sh` as expected and they will be forwarded to `mono`. (You can also `export` env. vars.) All the arguments work as on Windows, with some caveats:
 * file paths must be absolute (or relative to the install dir, `EmuHawkMono.sh` changes the CWD to there);
 * `--mono-no-redirect`: if you pass this flag *as the first argument*, it will be eaten by the script itself, and stdout/stderr will *not* be redirected to a file. (It's redirected by default.)
+** From 2.10, this will no longer be necessary.
 
 The same example as above would be `./EmuHawkMono.sh --lua=/path/to/script.lua /path/to/rom.n64`.
 
--- a/src/BizHawk.Client.Common/ArgParser.cs
+++ b/src/BizHawk.Client.Common/ArgParser.cs
@@ -23,47 +23,47 @@ namespace BizHawk.Client.Common
 				: base(name: name, description: description) {}
 		}
 
-		private static readonly Argument<string?> ArgumentRomFilePath = new("rom", () => null);
+		private static readonly Argument<string?> ArgumentRomFilePath = new(name: "rom", () => null, description: "path; if specified, the file will be loaded the same way as it would be from `File` > `Open...`");
 
 		private static readonly IReadOnlyList<Option> GeneratedOptions;
 
 		private static readonly BespokeOption<string?> OptionAVDumpAudioSync = new(name: "--audiosync", description: "bool; `true` is the only truthy value, all else falsey; if not set, uses remembered state from config");
 
-		private static readonly BespokeOption<int?> OptionAVDumpEndAtFrame = new("--dump-length");
+		private static readonly BespokeOption<int?> OptionAVDumpEndAtFrame = new(name: "--dump-length", description: "int; frame index at which to stop A/V dumping (encoding)");
 
 		private static readonly BespokeOption<string?> OptionAVDumpFrameList = new("--dump-frames"); // desc added in static ctor
 
 		private static readonly BespokeOption<string?> OptionAVDumpName = new("--dump-name"); // desc added in static ctor
 
-		private static readonly BespokeOption<bool> OptionAVDumpQuitWhenDone = new("--dump-close");
+		private static readonly BespokeOption<bool> OptionAVDumpQuitWhenDone = new(name: "--dump-close", description: "pass to quit completely after A/V dumping (encoding) finishes");
 
 		private static readonly BespokeOption<string?> OptionAVDumpType = new("--dump-type"); // desc added in static ctor
 
-		private static readonly BespokeOption<string?> OptionConfigFilePath = new("--config");
+		private static readonly BespokeOption<string?> OptionConfigFilePath = new(name: "--config", description: "path of config file to use");
 
-		private static readonly BespokeOption<string?> OptionHTTPClientURIGET = new("--url_get");
+		private static readonly BespokeOption<string?> OptionHTTPClientURIGET = new(name: "--url_get", description: "string; URI to use for HTTP 'GET' IPC (Lua `comm.http*Get*`)");
 
-		private static readonly BespokeOption<string?> OptionHTTPClientURIPOST = new("--url_post");
+		private static readonly BespokeOption<string?> OptionHTTPClientURIPOST = new(name: "--url_post", description: "string; URI to use for HTTP 'POST' IPC (Lua `comm.http*Post*`)");
 
-		private static readonly BespokeOption<bool> OptionLaunchChromeless = new(name: "--chromeless", description: "chrome is never shown, even in windowed mode");
+		private static readonly BespokeOption<bool> OptionLaunchChromeless = new(name: "--chromeless", description: "pass and the 'chrome' (a.k.a. GUI) will never be shown, not even in windowed mode");
 
-		private static readonly BespokeOption<bool> OptionLaunchFullscreen = new("--fullscreen");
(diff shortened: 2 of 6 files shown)


src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -2,221 +2,227 @@
...
+
+		private static readonly BespokeOption<int?> OptionAVDumpEndAtFrame = new("--dump-length");
+
+		private static readonly BespokeOption<string?> OptionAVDumpFrameList = new(name: "--dump-frames", description: $"implies `--{OptionAVDumpEndAtFrame.Name}=<end>` where `<end>` is the highest frame listed");
+
+		private static readonly BespokeOption<string?> OptionAVDumpName = new("--dump-name");
+
+		private static readonly BespokeOption<bool> OptionAVDumpQuitWhenDone = new("--dump-close");
+
+		private static readonly BespokeOption<string?> OptionAVDumpType = new("--dump-type");
```
Morilli, line 44: This actually requires `--dump-name` to be specified as well, which should probably be documented. Otherwise this option is silently not applied.

Morilli:
If we do this, is there a way to get rid of the `ParsedCLIFlags` struct as well? Feels like unnecessary duplication.

YoshiRulz:
> If we do this, is there a way to get rid of the `ParsedCLIFlags` struct as well? Feels like unnecessary duplication.

Yes, by returning the `ParserResult` and moving the `GetValueForOption` calls throughout `MainForm`. I will try to do that for this PR.

YoshiRulz:
Should `--help` group the flags semantically? (Re-ordering the list is trivial, unfortunately adding headers isn't.) I chose to list them alphabetically since that's what I'm used to from Unix tools.

src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -1,221 +1,235 @@
...
-	/// <remarks>
-	/// If a flag is given multiple times, the last is taken.<br/>
-	/// If a flag that isn't recognised is given, it is parsed as a filename. As noted above, the last filename is taken.
-	/// </remarks>
+	/// <summary>Parses command-line flags into a <see cref="ParsedCLIFlags"/> struct.</summary>
 	public static class ArgParser
 	{
-		/// <exception cref="ArgParserException"><c>--socket_ip</c> passed without specifying <c>--socket_port</c> or vice-versa</exception>
-		public static void ParseArguments(out ParsedCLIFlags parsed, string[] args)
+		private sealed class BespokeOption<T> : Option<T>
```
Morilli, line 17: Can you explain why this class is necessary?

README.md
```diff
@@ -199,7 +199,8 @@ tl;dr:
 
 #### Passing command-line arguments
 
-EmuHawk takes some command-line options which aren't well-documented; you might be able to figure them out from [the source](https://github.com/TASEmulators/BizHawk/blob/78daf4913d4c8e47d24fc14d84ca33ddef913ed4/src/BizHawk.Client.Common/ArgParser.cs).
+EmuHawk takes some command-line options which, starting from 2.9.2, can be listed with `--help`.
```
Morilli, line 202: This doesn't actually work on windows release builds due to the `WinExe` target, so this will need some consideration.

src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -1,221 +1,235 @@
...
+				.UseHelp()
+//				.UseEnvironmentVariableDirective() // useless
+				.UseParseDirective()
+				.UseSuggestDirective()
+//				.RegisterWithDotnetSuggest() // intended for dotnet tools
+//				.UseTypoCorrections() // we're only using the parser, and I guess this only works with the full buy-in
+//				.UseParseErrorReporting() // we're only using the parser, and I guess this only works with the full buy-in
+//				.UseExceptionHandler() // we're only using the parser, so nothing should be throwing
+//				.CancelOnProcessTermination() // we're only using the parser, so there's not really anything to cancel
+				.Build();
```
Morilli, line 133: Not sure if it's necessary to comment all options that you're not setting, but whatever. What do the `Use*Directive` calls actually accomplish though? From the very useful documentation I found they supposedly enable certain commandline syntax, but also it seems to require `EnableDirectives` and even when setting that, I couldn't get the syntax to work, so... ?

Also, what's with the version option?

src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -1,221 +1,235 @@
...
+			if (triggeredGeneratedOption is not null)
+			{
+				// means e.g. `./EmuHawkMono.sh --help` was passed, run whatever behaviour it normally has...
+				var exitCode = result.Invoke();
+				// ...and maybe exit
+				if (exitCode is not 0
+					|| triggeredGeneratedOption.Name is "help") // `Name` may be localised meaning this won't work? I can't grok the source for `HelpOption`
 				{
-					cmdRom = arg;
+					return exitCode;
```
Morilli, line 163: This looks kinda ugly... is there no way to standardize this into a single `Invoke()` call? I'm assuming not because the way bizhawk's argument parsing system works is so fundamentally different from what the CommandLine library expects.

Maybe a first step could be to separate long-term options that must stay for the entirety of the program's lifetime and could be stored in a struct like what already exists, and options that could be passed to the mainform directly. I don't really know, this just feel weird.

I mainly dislike the fact that trying to figure out how certain arguments are used and passed through takes multiple redirection steps: Search for `--dump-name` -> find `OptionAVDumpName` -> it gets passed to new ParsedCLIFlags as `cmdDumpName` argument -> the constructor sets the field `this.cmdDumpName` -> check where that field is used.

Morilli (commented):
This is probably an improvement for end-users (well, assuming you get `--help` to actually display stuff on windows) and about neutral for understanding code flow (it's a complete mess still).

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: > I mainly dislike the fact that trying to figure out how certain arguments are used and passed through takes multiple redirection steps: Search for `--dump-name` -> find `OptionAVDumpName` -> it gets passed to new ParsedCLIFlags as `cmdDumpName` argument -> the constructor sets the field `this.cmdDumpName` -> check where that field is used.

I could rename the options to match the `ParsedCLIFlags` fields, or better yet, rename the fields to these more descriptive names. I was hesitant to do that initially because I wanted to avoid churn in `MainForm`.

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: To explain, this list is [the expansion of `UseDefaults`](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.builder.commandlinebuilderextensions.usedefaults?view=system-commandline#remarks). I commented the ones I didn't use so that, if and when the list changes, there'll be a record of what didn't work in the past.
I'll be honest, I didn't play with the directives. I left them in just to follow the .NET convention and in the hope that someone would find a use for them.
We already have a `--version` which works well enough. When I (deleted that and) enabled `UseVersionOption`, it failed because it saw `--config=.../config.json --version` due to the launch script I use.

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: https://github.com/TASEmulators/BizHawk/blob/dfe4a72960a5c92555667671f9cd70d1130dcf74/src/BizHawk.Client.Common/ArgParser.cs#L127-L131
(essentially `root.Options.Where(o => o is not BespokeOption<>)`; this is necessary for `--help` as you already noted in another comment)
The other way, possibly the intended way, is to positively identify the `HelpOption`, but that type isn't exposed so you'd be comparing qualified names.

Morilli: In my testing checking for `Option<>` seems to filter out the `HelpOption`, did you try that?

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: > This looks kinda ugly... is there no way to standardize this into a single `Invoke()` call? I'm assuming not because the way bizhawk's argument parsing system works is so fundamentally different from what the CommandLine library expects.

That's it exactly. The library assumes it will be your `Program` implementation, and devolve *its execution privilege* to your peasant code. And also it heavily uses `async`. See the example here: https://learn.microsoft.com/en-us/dotnet/standard/commandline/handle-termination

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: Sorry, I missed this comment. What exactly did you try?
```diff
-var t = o.GetType();
-return !t.IsGenericType || t.GetGenericTypeDefinition() != typeof(BespokeOption<>);
+return o is Option<>;
```
...doesn't compile.
```csharp
return typeof(Option<>).IsInstanceOfType(o);
```
...compiles but matches nothing.

Morilli:
```diff
-                               return !t.IsGenericType || t.GetGenericTypeDefinition() != typeof(BespokeOption<>); // no there is no simpler way to do this
+                               return !t.IsGenericType || t.GetGenericTypeDefinition() != typeof(Option<>); // no there is no simpler way to do this
```

YoshiRulz: Oh... `HelpBuilder` just happens to inherit `Option` and not `Option<T>`. The latter seems to be only a thin subclass of `Option` purely for type safety, so I don't think it's a good idea to rely on `HelpBuilder`'s inheritance tree in this case.

Morilli: Even if it inherited `Option<T>` this code would work, because it compares exact types. Of course if you don't want to rely on `HelpOption` *not* being `Option<T>` instead then yeah, making your own wrapper type just for type comparison is the only way.

README.md
YoshiRulz: I've fixed this by caveating it with "on Linux".

YoshiRulz:
Rebased, only differences are:
- readme edits now reflect `--help` being Linux-only, and that the next release will be 2.10; and
- NuGet dep was made non-propagating.

There are no actionable review comments left, so I'll leave about a day for final comments before merging this. (I'd like this in 2.10-RC1 to give the greatest chance of catching regressions. Yesterday on Discord we discussed getting to an RC soon-ish, which I believe we can do!)

PR #3954, opened 2024-06-18
Refactor RamSearch to increase speed and improve functionality

Morilli:
Main changes:
- removes `IMiniWatchDetailed`, there is just `IMiniWatch` now which has full functionality.
- switching between `Detailed` and `Fast` search modes don't require starting a new search now.
- removes `PreviousType.LastSearch`, what that did was effectively set previous = current on every search, which is not helpful at all as you just delete the previous value information every time you perform a search. You can still set previous to current using the existing button.

This should improve speed in the default case and while using auto-search.


YoshiRulz:
One big commit makes this harder to review, but just from skimming it I think this is a good idea.

Morilli:
Alright, you'll know who's fault it was when something breaks.

PR #3964, opened 2024-07-07
Neshawk Tracelogger no longer prints incorrect addresses if a branch would overflow/underflow the PC

### LINKED ISSUE #3963: NesHawk: Tracelogger shows a 32 bit address when a branch wraps around address $FFFF
### Summary
If a branch instruction wraps around, either from $FFzz to $00zz, or from $00zz to $FFzz, the tracelogger incorrectly shows the destination of the branch as more than a 16 bit value.

TraceloggerBugExample

### Repro
I created a custom cartridge for the purposes of reproducing this bug, (for the screenshot shared above)
The source code just writes 0xF0 and 0xEC at address $0009 and $000A to form a branch that underflows, then takes a branch from address $FFF5 that overflows.

100thCoin:
This casts the value being printed to a ushort in order to prevent it from reading $10000 or $FFFFFFFF

This fixes #3963 

Before:
TraceloggerBugExample

After:
TraceloggerBugFix


YoshiRulz (approved):
I don't think this is the "correct" way of doing that, but it's not like we've ever handled integer overflows the correct way, so no harm done.

nattthebear:
This seems correct per the answers to https://stackoverflow.com/questions/15674560/6502-relative-address-mode-wrapping, and the referenced visual 6502 behavior.

> I don't think this is the "correct" way of doing that

What exactly do you mean?

YoshiRulz:
It's being implicitly widened by the addition, then truncated. I believe if the operands were truncated to u16 first then the addition would also overflow (`unchecked` is still the default...), producing the same result.

nattthebear:
> I believe if the operands were truncated to u16 first then the addition would also overflow

The result of `ushort + ushort` in C# is `int`, and cannot carry, overflow, wrap, or anything like that, so I'm not sure what you're proposing.

Morilli (approved):
Have played around a bit with this and I haven't found a better way to do this. If this calculation can under-/overflow as ushort and is supposed to be a ushort, we need to cast to ushort.

PR #3965, opened 2024-07-10
Simplify `Range<T>` usage and implementation, switching to half-open (end is exclusive)

YoshiRulz:
Done, including the change to half-open. The implementation was messy before 6ceb0c900f76514193187772507ee7dbe91f4c24 and messy afterwards. (IMO closed is the lesser of two evils, but the BCL chose half-open, so half-open it is.) Call-sites are free to choose whichever is simpler.
Branch name ended up being a misnomer since the generic interface is still there, ready for float intervals should the need arise. But all call-sites use either `Int32HalfOpenRange` or `Int64HalfOpenRange`.

After merge, need to remember to update https://github.com/TASEmulators/BizHawk/wiki/Available-C%23-and-.NET-features

### DIFF
--- a/src/BizHawk.Common/Ranges.cs
+++ b/src/BizHawk.Common/Ranges.cs
@@ -1,102 +1,212 @@
+using System.Runtime.CompilerServices;
+
 namespace BizHawk.Common
 {
-	/// <summary>represents a closed range of <typeparamref name="T"/> (class invariant: <see cref="Start"/> ≤ <see cref="EndInclusive"/>)</summary>
-#pragma warning disable CA1715 // breaks IInterface convention
-	public interface Range<out T> where T : unmanaged, IComparable<T>
-#pragma warning restore CA1715
+	/// <summary>
+	/// represents a closed interval (a.k.a. range) of <typeparamref name="T"/>
+	/// <br/>(class invariant: <see cref="INumericClosedIntervalContra{T}.Start"/> ≤ <see cref="INumericClosedIntervalContra{T}.EndInclusive"/>)
+	/// </summary>
+	public interface INumericClosedInterval<T> : INumericClosedIntervalContra<T>,
+		IEquatable<INumericClosedIntervalContra<T>>
+		where T : IComparable<T>
+	{
+		/// <summary>
+		/// leaves <paramref name="value"/> unchanged if it's contained in the interval,
+		/// or else sets it to whichever bound is closest
+		/// </summary>
+		void Clamp(ref T value);
+
+		/// <returns>
+		/// <see langword="true"/> iff <paramref name="value"/> is contained in this interval
+		/// (<paramref name="value"/> is considered to be in the interval if it's exactly equal to either bound)
+		/// </returns>
+		bool Contains(in T value);
+	}
+
+	/// <summary>see <see cref="INumericClosedInterval{T}"/>; this is a type parameter variance hack</summary>
+	public interface INumericClosedIntervalContra<out T>
 	{
 		T Start { get; }
 
 		T EndInclusive { get; }
 	}
 
-	internal class MutableRange<T> : Range<T> where T : unmanaged, IComparable<T>
+	/// <summary>
+	/// represents a closed range (a.k.a. interval) of <see cref="int"><c>s32</c>s</see>
+	/// <br/>(class invariant: <see cref="NumericClosedRangeBase{T}.Start"/> ≤ <see cref="NumericClosedRangeBase{T}.EndInclusive"/>)
+	/// </summary>
+#pragma warning disable MA0077 // already implements `IEquatable<TSelf>`
+	public sealed class Int32ClosedRange : NumericClosedRangeBase<int>
+#pragma warning restore MA0077
 	{
-		private (T Start, T EndInclusive) r;
+		private string? _ser = null;
 
-		public T Start
-		{
-			get => r.Start;
-		}
+		/// <inheritdoc cref="NumericClosedRangeBase{T}(T,T)"/>
+		public Int32ClosedRange(int start, int endInclusive)
+			: base(start, endInclusive) {}
 
-		public T EndInclusive
-		{
-			get => r.EndInclusive;
-		}
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public override bool Contains(in int value)
+			=> Start <= value && value <= EndInclusive;
 
-		/// <exception cref="ArgumentException"><typeparamref name="T"/> is <see langword="float"/>/<see langword="double"/> and either bound is <see cref="float.NaN"/></exception>
-		/// <exception cref="ArgumentOutOfRangeException"><paramref name="endInclusive"/> &lt; <paramref name="start"/></exception>
-		internal MutableRange(T start, T endInclusive)
-		{
-			if (endInclusive.CompareTo(start) < 0) throw new ArgumentOutOfRangeException(nameof(endInclusive), endInclusive, "range end < start");
-			if (start is float fs)
-			{
-				if (float.IsNaN(fs)) throw new ArgumentException("range start is NaN", nameof(start));
-				if (endInclusive is float fe && float.IsNaN(fe)) throw new ArgumentException("range end is NaN", nameof(endInclusive));
-			}
-			else if (start is double ds)
-			{
-				if (double.IsNaN(ds)) throw new ArgumentException("range start is NaN", nameof(start));
-				if (endInclusive is double de && double.IsNaN(de)) throw new ArgumentException("range end is NaN", nameof(endInclusive));
-			}
-			r = (start, endInclusive);
-		}
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public bool Contains(int value)
+			=> Start <= value && value <= EndInclusive;
 
+		/// <remarks>beware integer overflow when this range spans every value</remarks>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public uint Count()
+			=> (uint) ((long) EndInclusive - Start) + 1U;
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public bool Equals(Int32ClosedRange other)
+			=> Start == other.Start && EndInclusive == other.EndInclusive;
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public override string ToString()
-			=> $"{Start}..={EndInclusive}";
+			=> _ser ??= $"{Start}..={EndInclusive}";
 	}
 
-	/// <summary>contains most of the logic for ranges</summary>
-	/// <remarks>
-	/// non-generic overloads are used where the method requires an increment or decrement
-	/// </remarks>
-	public static class RangeExtensions
+	/// <summary>
+	/// represents a closed range (a.k.a. interval) of <see cref="long"><c>s64</c>s</see>
+	/// <br/>(class invariant: <see cref="NumericClosedRangeBase{T}.Start"/> ≤ <see cref="NumericClosedRangeBase{T}.EndInclusive"/>)
+	/// </summary>
+#pragma warning disable MA0077 // already implements `IEquatable<TSelf>`
+	public sealed class Int64ClosedRange : NumericClosedRangeBase<long>
+#pragma warning restore MA0077
 	{
 		private const ulong MIN_LONG_NEGATION_AS_ULONG = 9223372036854775808UL;
 
+		private ulong? _count = null;
+
+		private string? _ser = null;
+
+		/// <inheritdoc cref="NumericClosedRangeBase{T}(T,T)"/>
+		public Int64ClosedRange(long start, long endInclusive)
+			: base(start, endInclusive) {}
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public override bool Contains(in long value)
+			=> Start <= value && value <= EndInclusive;
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public bool Contains(long value)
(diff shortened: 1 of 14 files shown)


Morilli:
What's the WIP part about this? Could this just be merged as is or is something missing?

YoshiRulz:
~~It could be merged as-is. (It might be a good idea to have a unit test, at least before the refactor from closed to half-open, since that one changes some logic.)~~ The part that's missing is to make `Int32Interval` half-open. It may also be possible to make them value types, in which case some or all uses of `Int32Interval` could be the BCL `Range` instead.

SuuperW:
What is the advantage of this?
How is this simpler than the previous implementation, aside from removing unused code?
~~Why is `Int32Interval` closed while `Int64Interval` is half-open?~~ EDIT: You said above that changing that is "missing". I say change it before merging (if it's going to ever be merged).

SuuperW:
If you want this merged prior to my PR, address my comments and get it merged.

src/BizHawk.Common/Ranges.cs
```diff
@@ -1,253 +1,227 @@
...
+			start = range.Start;
+			endExclusive = range.EndExclusive;
+		}
 
-		/// <inheritdoc cref="MutableRange{T}(T,T)"/>
-		public static Range<T> RangeTo<T>(this T start, T endInclusive) where T : unmanaged, IComparable<T> => start.MutableRangeTo(endInclusive);
+		/// <inheritdoc cref="Int32HalfOpenRange(int,int)"/>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static Int32HalfOpenRange RangeTo(this int start, int endInclusive)
+			=> new(start: start, endExclusive: checked(endInclusive + 1));
```
Morilli: This will throw for core usages like [DSDA](https://github.com/TASEmulators/BizHawk/blob/161d5d4b2697354a57d00a00cc0df8a69ae04676/src/BizHawk.Emulation.Cores/Computers/Doom/DSDA.Controller.cs#L120) that pass `int.MaxValue` as `endInclusive` value. I guess that's the inherent problem of a half-open interval, so why make it half-open in the first place?

SuuperW: Uses of this method are confusing. `Int32HalfOpenRange _rangeX = 0.RangeTo(0);` is explicit about the range being half open, and about the ends being 0 and 0. This makes it look like the range is [0, 0) which is invalid.
The parameter name does specify that it is inclusive, but it's weird to specify a closed range and get an open range back. I think the name should be changed to `RangeToInclusive`.

src/BizHawk.Common/Ranges.cs
```diff
@@ -1,253 +1,227 @@
...
+		where T : IComparable<T>
 	{
-		T Start { get; }
-
-		T EndInclusive { get; }
+		/// <summary>
+		/// leaves <paramref name="value"/> unchanged if it's contained in the interval,
+		/// or else sets it to whichever bound is closest
+		/// </summary>
+		void Clamp(ref T value);
```
SuuperW, line 17: `T Clamp(T value)` please. Using `ref` here is not normal and will make code more difficult to read in most cases, including the only currently existing uses: `range.Clamp(ref value); return value;` becomes `return range.Clamp(value);`.

src/BizHawk.Common/Ranges.cs
```diff
@@ -1,253 +1,227 @@
...
+
+		/// <returns>
+		/// <see langword="true"/> iff <paramref name="value"/> is contained in this interval
+		/// (<paramref name="value"/> is considered to be in the interval if it's exactly equal to either bound)
+		/// </returns>
+		bool Contains(in T value);
```
SuuperW, line 23: This should not use `in`. This interface is going to be used with basic integer types, so using a reference provides no benefit. (Microsoft's docs in the `in` modifier also say this: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters#in-parameter-modifier)

src/BizHawk.Common/Ranges.cs
```diff
@@ -1,253 +1,227 @@
...
+		/// <see langword="true"/> iff <paramref name="value"/> is strictly contained in this interval
+		/// (<paramref name="value"/> is considered to be OUTSIDE the interval if it's exactly equal to either bound)
+		/// </returns>
+		bool StrictlyContains(in T value);
 	}
 
-	/// <summary>represents a closed range of <typeparamref name="T"/> which can be grown or shrunk (class invariant: <see cref="Start"/> ≤ <see cref="EndInclusive"/>)</summary>
-	public class MutableRange<T> : Range<T> where T : unmanaged, IComparable<T>
+	/// <summary>see <see cref="INumericHalfOpenInterval{T}"/>; this is a type parameter variance hack</summary>
+	public interface INumericHalfOpenIntervalContra<out T>
```
SuuperW, line 27: Why does this exist? The code compiles without this hack.

src/BizHawk.Common/Ranges.cs
```diff
@@ -1,253 +1,227 @@
...
+		public bool Equals(Int32HalfOpenRange other)
+			=> Start == other.Start && EndExclusive == other.EndExclusive;
 
-		/// <inheritdoc cref="Count(Range{int})"/>
-		public static ulong Count(this Range<long> range) => (range.Contains(0L)
-			? (range.Start == long.MinValue ? MIN_LONG_NEGATION_AS_ULONG : (ulong) -range.Start) + (ulong) range.EndInclusive
-			: (ulong) (range.EndInclusive - range.Start)
-		) + 1UL;
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public override bool StrictlyContains(in int value)
```
SuuperW: This method's name doesn't help understand what it does. "Strictly contains" has meaning when comparing two sets but this is comparing a potential member to a set. I think this method should be deleted, especially since you have removed the only uses of it.

src/BizHawk.Common/Ranges.cs
```diff
@@ -1,253 +1,227 @@
...
+	}
 
-		/// <inheritdoc cref="RangeToExclusive(int,int)"/>
-		public static MutableRange<short> MutableRangeToExclusive(this short start, short endExclusive) => endExclusive == short.MinValue
-			? throw ExclusiveRangeMinValExc
-			: new MutableRange<short>(start, (short) (endExclusive - 1));
+	public static class RangeExtensions
+	{
+		private static ArithmeticException ExclusiveRangeMinValExc
+			=> new("exclusive range end is min value of integral type");
```
SuuperW: This seems like a redundant check. If you omit this, won't it catch the error by finding that the inclusive end is less than the inclusive start?

src/BizHawk.Common/Ranges.cs
YoshiRulz: I was trying to minimise changes at call-sites, though I am already changing the type name on that line. I'll try it and see exactly how many extra lines would need changing. edit: `RangeTo{,Exclusive}`: +387/-253 across 14 files; `RangeTo{Inclusive,Exclusive}`: +522/-388 across 67 files. So a lot. We can always rename it later.

src/BizHawk.Common/Ranges.cs
YoshiRulz: There is a copying `Clamp` as an extension method:
https://github.com/TASEmulators/BizHawk/blob/6ceb0c900f76514193187772507ee7dbe91f4c24/src/BizHawk.Common/Ranges.cs#L175-L180

src/BizHawk.Common/Ranges.cs
YoshiRulz: Why do those axes have 4 billion possible values? Surely no TASer is expecting to use all of them? https://github.com/TASEmulators/BizHawk/commit/161d5d4b2697354a57d00a00cc0df8a69ae04676 @feos Maybe expose a smaller, arbitrary range and scale up the values, something like `(aShort << 16) | aShort`.

Per OP, half-open because everything in the BCL from `Range` to simple array bounds is half-open.

src/BizHawk.Common/Ranges.cs
YoshiRulz: I wanted it to be extensible to larger types.

src/BizHawk.Common/Ranges.cs
YoshiRulz: I wanted it to be extensible, and this should allow otherwise conflicting subclasses to be equated/compared (maybe they're Source Generated in multiple projects, for example), though I don't recall if I tested that.

src/BizHawk.Common/Ranges.cs
vadosnaprimer: I asked Doom TASers and they said it's better to not have arbitrary limits on camera movement so it could instantly jump across the whole level. If it's absolutely required to have a smaller value, we can sacrifice a few higher bits. Scaling is absolutely not an option, all small movements should be available.

src/BizHawk.Common/Ranges.cs
YoshiRulz: Yes, but checking explicitly lets me give a better error message than `System.ArgumentOutOfRangeException: range end < start\nParameter name: endExclusive\nActual value was -2147483648.`

src/BizHawk.Common/Ranges.cs
SuuperW: What is the benefit of having a `ref` overload, and what is the benefit of not having a non-ref overload in the interface (making it an extension instead)?

YoshiRulz: Though I haven't done it in this PR, it allows for a micro-optimisation in cases like these:
https://github.com/TASEmulators/BizHawk/blob/b9993c61e264060d817bfa7cc45f5e3f15fbf91c/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs#L1338
This too would be more relevant with larger numeric types.

The extension method isn't merged into the interface so that it can be used with all implementations with no copy-pasting or extra work. (In .NET Core there's the confusingly-named 'default interface method' feature which has the same effect, but we don't have access to that.)

src/BizHawk.Common/Ranges.cs
SuuperW: I guess having all those lines would make reviewing more tedious. Renaming it "later" should be "immediately after this PR is merged" imo.

src/BizHawk.Common/Ranges.cs
SuuperW: I don't see the problem with that message, except it should be `end <= start` not just `<`. But I guess it doesn't really hurt to have that special case.

src/BizHawk.Common/Ranges.cs
SuuperW: Do you have plans to use larger types? I don't think we should be making code that's weird (and potentially less optimal; passing references to `int` seems like it would not be optimal) to use with what we have, for the sake of some theoretical future use case.

src/BizHawk.Common/Ranges.cs
SuuperW: I don't understand. Is the code you gave an optimization of something? `ConstrainWithin` is a method which calls `Clamp(ref T)`. Surely that is less efficient than calling a `Clamp(T)` that is implemented directly.

> we don't have access to that

I wish we did, it would be nice to have.

SuuperW:
> but the BCL chose half-open, so half-open it is.

Do we need to follow the implementation details of the BCL?

I also noticed that with the changes you've made, there are no references to the new range interface, except extension methods and the one implementing class. I wonder if there's any reason to keep the interface.

src/BizHawk.Common/Ranges.cs
YoshiRulz: `ControllerType.Axes[AxisEditColumn].Range.Clamp(ref value);` would be the optimised version, but without benchmarks I can't be certain that it's actually faster.

I have a [very old PR](https://github.com/TASEmulators/BizHawk/pull/3393) which adds something similar to the DIM feature using only a Source Generator if you're interested.

src/BizHawk.Common/Ranges.cs
YoshiRulz: Because of the monomorphisation, I can simply add copying overloads for `int` and `long`.

src/BizHawk.Common/Ranges.cs
YoshiRulz: I've fixed the typo, and a bunch more issues with these helpers' error cases.

src/BizHawk.Common/Ranges.cs
SuuperW: I suspect that would be slower, not faster. Although it might actually depend on the specifics of how it's used. If the only thing being done with the clamped value is putting it in memory somewhere, `ref` might make it faster. If it's never going to leave the stack, a non-ref version might have it never even actually get onto the stack, just in a register, and could be faster.

But even if it is faster, I'd prefer not using it because modifying a value type without an assignment operator is weird C#. And this code isn't performance critical.

src/BizHawk.Common/Ranges.cs
SuuperW: I'm not sure I understand. Are you saying that you can add an overload without `in` to the implementing classes `IntXXHalfOpenRange`? I see in your force push you have already done this.

That's good enough for me, but I still don't understand why you wanted to write it for a use case that doesn't exist.

src/BizHawk.Common/Ranges.cs
SuuperW: I'd prefer leaving this out until it is needed. It makes the code more difficult to understand, and code supporting a non-existent use case is more likely to cause issues in the future as code gets updated or modified than it is to be useful.

PR #3975, opened 2024-07-22
Fix Uzebox framerate

YoshiRulz:
Copying from commit message:
> These numbers are derived from the 2010 whitepaper: NTSC's 4*(315/88 MHz) is doubled to give the master clock rate, which is then divided by 1820 cycles per line and 262 lines per (progressive) frame.


nattthebear:
This looks perfectly correct.  I'm at a loss to explain why it was ever wrong; I was aware of the type of crystal used (it's a common NTSC crystal) and the other details of the video system at the time.

nattthebear (approved):

PR #3981, opened 2024-07-30
movies: normalize loaded hashes to upper case

### LINKED ISSUE #3979: Hashes in movie headers should not be case-sensitive
Pretty self-explanatory title.

If you manually edit a (`.bk2` or `.tasproj`) movie file to change the game hash in the header but happened to copy in a lower case hash, it will spit out `Warning: Movie hash does not match the ROM` when you load the movie:
https://github.com/TASEmulators/BizHawk/blob/7a8b9b13ffdaae4b23edb253bc8a9baf25cc661e/src/BizHawk.Client.EmuHawk/MainForm.Movie.cs#L66-L69
(Same deal probably applies to firmware hashes and other hashes).

Hashes in general are not case-sensitive and I think BizHawk should either normalize the hashes upon movie load or use case-insensitive comparisons in relevant code.

(Certain movie metadata such as the game hash can't be edited without manually opening it as a `.zip` file and editing and such editing is useful -- thus it seems reasonable that the movie loading should be resilient to some amount of reasonable variation especially in something like a hash.)

see also #2336

RetroEdit:
fixes #3979

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/movie/BasicMovieInfo.cs
+++ b/src/BizHawk.Client.Common/movie/BasicMovieInfo.cs
@@ -101,7 +101,7 @@ namespace BizHawk.Client.Common
 
 		public virtual string Hash
 		{
-			get => Header[HeaderKeys.Sha1];
+			get => Header[HeaderKeys.Sha1].ToUpperInvariant();
 			set => Header[HeaderKeys.Sha1] = value;
 		}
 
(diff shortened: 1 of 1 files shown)


YoshiRulz (commented):
Should you do this during parsing i.e. in `LoadFields`?

RetroEdit:
I considered doing it during parsing, but I wanted a narrow solution that was relatively unproblematic. Reworking all the checksum code right now might be more ambitious and error-prone. (In that case it would also maybe be worthwhile to have the hashes be in a container that includes hash type).

Morilli (approved):
Probably fine. I looked at how this value is written and read and it's all just strings, so there's no real better way to do this, at least not easily.

PR #3987, opened 2024-08-05
Make external tools path independent of working directory

kalimag:
Fixes default external tools directory not being found if BizHawk is launched with a working directory other than the exe directory.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/ExternalToolManager.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/ExternalToolManager.cs
@@ -88,7 +88,7 @@ namespace BizHawk.Client.EmuHawk
 				DirectoryMonitor.Created -= DirectoryMonitor_Created;
 				DirectoryMonitor.Dispose();
 			}
-			var path = _config.PathEntries.AbsolutePathFor(_config.PathEntries[PathEntryCollection.GLOBAL, "External Tools"].Path, null);
+			var path = _config.PathEntries.ExternalToolsAbsolutePath();
 			if (Directory.Exists(path))
 			{
 				DirectoryMonitor = new FileSystemWatcher(path, "*.dll")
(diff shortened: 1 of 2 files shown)


src/BizHawk.Client.EmuHawk/tools/ExternalToolManager.cs
```diff
@@ -88,7 +88,7 @@ public void Restart(Config config)
 				DirectoryMonitor.Created -= DirectoryMonitor_Created;
 				DirectoryMonitor.Dispose();
 			}
-			var path = _config.PathEntries[PathEntryCollection.GLOBAL, "External Tools"].Path;
+			var path = _config.PathEntries.AbsolutePathFor(_config.PathEntries[PathEntryCollection.GLOBAL, "External Tools"].Path, null);
```
YoshiRulz, line 91: I don't see this pattern anywhere else in the codebase; maybe the `AbsolutePathFor` call belongs elsewhere?

kalimag: `MacroInputTool` does it

https://github.com/TASEmulators/BizHawk/blob/494fe90bfcb340e247d62df827106ec1b618c301/src/BizHawk.Client.EmuHawk/tools/Macros/MacroInput.cs#L258-L263

Everywhere else does it in `PathEntryCollectionExtensions`, e.g.

https://github.com/TASEmulators/BizHawk/blob/494fe90bfcb340e247d62df827106ec1b618c301/src/BizHawk.Client.Common/config/PathEntryCollectionExtensions.cs#L136-L140

I do think the public API probably should be designed differently if everything needs to do an additional `AbsolutePathFor` call anyway, but that seems like it'd require a cumbersome refactoring.

YoshiRulz: Ah that's why I couldn't find it. Please add a method to `PathEntryCollectionExtensions` instead.

Issue #3999, opened 2024-08-12
Prior RAM search workflow is broken in 2.9.2 dev


RetroEdit:
I don't know when this was broken, but I was trying to test an unrelated bug, and RAM search doesn't seem to be working properly in the latest dev build ( ab7f379e745bb241e537ee6fe0212f98ab421dad ).

Specifically, `Compare To`: `Previous Value` with `Comparison Operator`: `Not Equal To` dismissed all the addresses even if you visibly observe ones that have changed in the hex editor.

Tested with the game Libbet and the Magic Floor v0.07 for GB, which is my go-to testing game because it's homebrew. I moved the character, which should change some address's values (specifically: `0400`, `0404`, `0408`, `04EB`, `056F`, `0594`), but all the addresses were removed from the search as if none of them had changed.

I would venture to guess it's just broken in general, but that's the extent of my testing so far.

----

Okay, I think I've figured out the deal: 1a9e5e52f1f12cab55348c67eca8a9bb20d49329 (#3954) says the following:
> - removes `PreviousType.LastSearch` due to its questionable use case

I think this was the prior default? @Morilli care to explain the rationale here? To me, an intuitive search methodology is to anchor the list of addresses when you start RAM search by pressing new, then doing an action and searching for addresses that change as a result of that action. And then do another action that changes the state and iteratively do that until you've narrowed down to the most relevant addresses.

I don't think the other RAM search options supply the relevant functionality for effective repeated searches of this nature, so removing this option seems to reduce RAM search's usability. At the very least, `Previous Frame` seems like a bad default to me because it's only useful in a frame-by-frame context, not real-time.

RetroEdit:
I tried that, and I don't believe it actually serves as a proper replacement: it doesn't discern the vital detail that the value is actually different. Values can merely change coincidentally during the time you do an action, making the search less discerning when you're looking for what's different at the end of an action.

Morilli:
What you can do is set the `PreviousType` to original, and update the previous values manually with the UI button after every search. How often do you expect to need to search the value? I feel like it's not necessary to have this functionality builtin.

RetroEdit:
Using the `<-` "Copy Value to Previous" button in combination with setting `PreviousType` to "Original" sort of works. It's less convenient, because it's easy to accidentally forget to press the button, and now it's split what was previously one synchronized action of updating search into two discrete operations that you can't do simultaneously unless you pause.

[...]

PR #4004, opened 2024-08-16
Refactor RamSearch to increase speed and improve functionality

YoshiRulz:
(recreation of #3954, as per discussion around https://github.com/TASEmulators/BizHawk/issues/3999#issuecomment-2289237080)


src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
```diff
@@ -303,7 +305,7 @@ public void WriteFloat(long addr, float value, string domain = null)
 				LogCallback($"Warning: Attempted write {addr} outside memory size of {d.Size}");
 				return;
 			}
-			d.PokeUint(addr, BitConverter.ToUInt32(BitConverter.GetBytes(value), 0), _isBigEndian);
+			d.PokeUint(addr, Unsafe.As<float, uint>(ref value), _isBigEndian);
```
Morilli, line 308: Is there a reason you didn't make this into a global common function (like [before](https://github.com/TASEmulators/BizHawk/commit/0f22ef2ad05fe8f72f06fcb2b6a932008eba69a4#diff-cf06fd3b2597d6a73503739b41c844d1e0738f241b69f19de4e9865629e82a0fL160))?

Morilli:
Looks generally fine even though the individual commits probably don't compile. From the discussion in #3999 I'm assuming [97adf1a](https://github.com/TASEmulators/BizHawk/pull/4004/commits/97adf1a87dfbed9fef30aef790600d62e5b0f63d) should not be included now?

src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
YoshiRulz: I realised that having an allowlist of helpers wouldn't prevent you from using `Unsafe.As` directly and typo-ing the type arguments, so I'd have to use an Analyzer, and then there's no benefit to those explicit helpers.

Morilli: It doesn't prevent developers from using it, no, I just thought it might be practical as this uint<->float reinterpreting seems to be somewhat common around the watch area and having our own specific function for it could prevent someone using something like `BitConverter` instead.

YoshiRulz:
Some of this was merged as https://github.com/TASEmulators/BizHawk/commit/3c218c664205785691d1e2c0fc87402fbdfb9384 through https://github.com/TASEmulators/BizHawk/commit/f09e72e7e0397e30cf45a4d76006ef7efdb30ca2.

Morilli:
Looking at it again and I can't seem to be able to properly separate the changes more than in <https://github.com/Morilli/BizHawk/tree/xd>. Main commit is 3a63ec7da8395837b14eb82fd959df2c83d7a46c, I've tried to add a decently sized explanation to it but I feel like there's no natural way to deconstruct that commit any further while keeping the commits logical.

To be fair the commit looks bigger than it is; lots of `GetValue(w.Address)` -> `w.Current` and rewrite of the Mini*WatchDetailed classes.

YoshiRulz:
It looks like you already have it broken down into smaller conceptual pieces in the commit description?
- cache `Current`
- replace `GetValue` with `Current`
- (this has just come to me: can you make the `*WatchDetailed` classes extend `*Watch`? if so, do that first)
- move `IMiniWatchDetails` features up to `IMiniWatch`
- merge `IMiniWatchDetails` and `IMiniWatch`

Morilli:
> cache Current

Actually, that seems to already be done in the detailed watch, the variable is just called `_prevFrame` and only used internally.

> replace GetValue with Current

That can't be done until `Current` has a meaningful value at all times. And even if that is done, you still need to query the value for the non-detailed watches or update their code to store the current value as well (making them an exact copy of the detailed implementation, just without the ChangeCount++ line).

> (this has just come to me: can you make the *WatchDetailed classes extend *Watch? if so, do that at this point or maybe first)

Yeah, reasonable. But again feels like wasted effort when the classes are to be merged anyways.

> move IMiniWatchDetails features up to IMiniWatch

At this point, that is effectively only the change count logic. That is the only remaining difference between the detailed and non-detailed watch.

Morilli:
The parts currently in master should already resolve the slowdown incurred from `MiniWatchDetails`, so this change is not really necessary anymore. There's more that could be done, but the main goal with my changes was to make the detailed watch type not as stupidly slow as it was and fix some issues in the process, and I believe all of that is in master already.

src/BizHawk.Client.Common/Api/Classes/MemoryApi.cs
YoshiRulz: Resolved with https://github.com/TASEmulators/BizHawk/commit/10ebea4dda87cbc65dd3dd74cd83655b1a69dff1 (except in Cores for now)

PR #4008, opened 2024-08-18
Change OnMovieEnd default to 'Pause' instead of 'Finish'

### LINKED ISSUE #4007: Change default for `OnMovieEnd` to "Pause" instead of "Finished"
I think the default `On Movie End` state should be `Pause`, not `Switch to Finished`.

I believe the basic argument in favor is relatively simple and compelling: if a movie is paused, you merely have to unpause and it will be in the "Finished" state. **In contrast:** if a movie plays into Finished state, it will continue playing indefinitely and you probably have to replay the entire movie from the start to reach the exact end, which is time-consuming for long movies.

## Extra notes
1. As a side-note, I'm actually not sure what Finished's use-case is in modern times. Rarely if ever is there utility in playing a movie indefinitely past the movie end. You can barely do anything in Finished mode as far as I'm aware: you cannot modify or continue the current movie so anything you do afterwards won't be saved. It does serve to let you know you were previously playing a movie, and it does accept inputs, but those inputs are not going to be saved if you're trying to do serious testing.
2. I realize this is a long-standing default (introduced in b53cc908b8e2543e8a8ca4869121c96e73e27ed9 on Jul 14, 2014 for 1.7.1), maybe even inherited from FCEUX, but I do still think it should be changed. Finished mode was the *only* movie end option prior to the other options being introduced, so it seems to have been chosen as the default because of that.
3. (Tangentially, at time of writing, I actually think there might be a change to the emulator behavior where it prevents savestate saves on the last frame in 2.9.2 dev, which I don't think was the behavior prior and disrupts traditional movie recording (where a workflow for continuing a movie can be playing the whole movie and saving a savestate at the end, then loading the savestate). I vaguely recall seeing that commit. I'll try and move this to another issue when I have time to confirm.)

Semi-related: #3989

RetroEdit:
Fixes #4007

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/config/MovieConfig.cs
+++ b/src/BizHawk.Client.Common/config/MovieConfig.cs
@@ -12,7 +12,7 @@
 
 	public class MovieConfig : IMovieConfig
 	{
-		public MovieEndAction MovieEndAction { get; set; } = MovieEndAction.Finish;
+		public MovieEndAction MovieEndAction { get; set; } = MovieEndAction.Pause;
 		public bool EnableBackupMovies { get; set; } = true;
 		public bool MoviesOnDisk { get; set; }
 		public int MovieCompressionLevel { get; set; } = 2;


YoshiRulz (approved):

vadosnaprimer:
What happens after you unpause? Does the movie unload?

YoshiRulz:
It's using the existing setting. This only changes which is the default.

vadosnaprimer:
Is my question impossible to answer?

RetroEdit:
@feos It changes to Finished. Thus, you lose no functionality with this new functionality, and gain substantial convenience.

vadosnaprimer (approved):

PR #4010, opened 2024-08-21
Fix movie header values being potentially outdated upon starting a movie record

### LINKED ISSUE #3980: GB: Default settings don't properly save the boot ROM in movie settings
Follow-up to #3763

Basically, the boot ROM's info is not properly saved into the movie. Instead, BizHawk silently seems to ~~choose an arbitrary boot ROM if one is available~~ infer the boot ROM from the GB mode sync settings(?). This leads to movies where the sync settings don't properly indicate the firmware hash and don't have `"EnableBIOS":true` even though the boot intro plays.

Expected behavior: BizHawk saves a specific firmware hash in the movie and sets `"EnableBIOS":true` in the sync settings.

Morilli:
- closes #3980

Main behavior change here is to set minimal required header values into the movie when queueing it up, and setting all default values after the core has successfully rebooted. This ensures that all set values are accurate even if sync settings change on reboot.


CasualPokePlayer (approved):
Seems to work, LGTM :thumbsup:

YoshiRulz:
I'm late, but diffs LGTM. Though it would have been nice to have this as at least 2 commits on the trunk.

Issue #3410, opened 2022-10-08
Nymashock changes emulator window size between each game resolution change, no matter the settings


TizianoMattei:
### Summary
Whenever i try to manually resize bizhawk window during TASing or gameplay to fit my screen and also have space for tools like TAStudio to the side, the emulator window will resize to the default size. The issue doesnt happen when changing window sizes from the emulator settings in View-->Window Size (x2, x3, x4 etc)

I tried different settings (obviusly config -> display -> check "maintain aspect ratio" and "use system's recommendation") and whatnot, but noone of the settings in the "Display"config made a difference. 
What (partially) worked was changing the default constant framebuffer size from FALSE to TRUE in Nyma Core Non-Sync settings. This however produces some other graphical glitches and issues when TASing (specifically when dumping, pausing the emulator on resolution change will.. not pause it till the loading screen and/or console bios screen is gone, and using "toogle throttle" to speed up the emulator.

I've checked on actual console with the title i had this issue most frequently, "Bugs Bunny Lost in Time NTSC-U", connected directly to a hd monitor using retroThink from the RGB console signal, and it does indeed change resolution each loading screen. 
However the emulator window doesnt reset to default on Octoshock on each game resolution change, as opposed to Nymashock.

Behaviour on Nymashock: 
https://www.youtube.com/watch?v=7sxhxRZEKfw

### Repro
1. Open a psx game in Nymashock
2. Change manually the emulator window size (even before boot will work, as from the psx bios and the emulated game a resolution change will occur anyway
3. Bizhawk window will reset to its original size, despite the previous manual resize

### Host env.
Bizhawk 2.8, Win10 22H2, Amd cpu, Nvidia gpu

vadosnaprimer:
`Config -> Display -> Use custom size` will force the emulator window to anything you set.

TizianoMattei:
> `Config -> Display -> Use custom size` will force the emulator window to anything you set.

That is true, however when doing that, setting for example 1366x968, then moving the window (not resizing) to fit just right on the screen, will force the emulator window to be moved AGAIN in its starting position after each game resolution change. This would be just a minor inconvenience, but the problem is that after setting such resolution, when reopening bizhawk, the top bar of settings and such will spawn out of bounds in windows, making inaccessible unless the user is the forced to actually manually resize. (I have windows snapping and such disabled btw)

[...]

Issue #3039, opened 2021-12-18
N64 Graphics Plugin's "graphics resolution" shouldn't govern the size of the application Window


Nicholas-Steel:
### Summary
The N64 Graphics Plugin's "graphics resolution" shouldn't govern the size of the Application Window when launching an N64 ROM or rebooting the N64 emulation core. It should maintain the window size that was being used before emulation began.

The current behaviour is very annoying when using the Graphics Resolution setting as a form of Super Sample Anti Aliasing and configuring it in excess of the desktop resolution. For example: I set the Graphics Resolution to 3840x2160 (or 2560x1440) while my desktop resolution is 1920x1080, when I launch an N64 game the application window will suddenly exceed the boundaries of my monitor.

I have to Shift Right Click the Bizhawk program on the Taskbar and select "Size" or "Move" and tap the keyboard arrow keys to get an edge of the application in to view so that I can adjust the window size.

(If you think it's weird defining a widescreen resolution for N64 games, I have the GlideN64 graphics plugin configured to adjust the View Port to 16:9, some effects in games won't render along the left and right edge of the screen beyond the 4:3 space but most stuff will render fine)

### Reproduction Steps
1) Navigate to Config > Cores > N64 Video Plugin Settings > Global tab > Video Resolution
2) Adjust this and then load an N64 ROM.

zeromus:
Maybe we could add an option to suppress automatic resizing of the window, maybe view > window size > auto resize (and default it to on)

YoshiRulz:
I think we're simply giving users the wrong slider. They don't care how many times pixels are scaled up, they only care about the resulting window size and whether the game is legible without being too big.

Nicholas-Steel:
> I think we're simply giving users the wrong slider. They don't care how many times pixels are scaled up, they only care about the resulting window size and whether the game is legible without being too big.

I use it for Super Sample Anti Aliasing.

vadosnaprimer:
> I think we're simply giving users the wrong slider. They don't care how many times pixels are scaled up, they only care about the resulting window size and whether the game is legible without being too big.

How do you know that? GUI size multiplier won't even get set if you try to go off-screen with it, the problem is only with N64 plugins forcing hawk to go off-screen.

[...]

Issue #3544, opened 2023-02-01
All windows closed in the secondary monitor will always appear on the secondary monitor, after they are dragged to the primary monitor and closed properly


tommai78101:
### Summary

I have a dual monitors Windows 10 setup. My primary monitor (Monitor 1) is my laptop monitor. My secondary monitor (Monitor 2) is an external monitor.

My display arrangement looks like this:

![image](https://user-images.githubusercontent.com/6444116/215937338-a57a1615-d803-4fa5-9b14-dc60c1507ee6.png)

If I run Emuhawk and drag it to Monitor 1, and then I opened the TAStudio and RAM Watch, both of them will appear on Monitor 2, all gathered together on one side. I have a GIF to show what is happening.

![GIF](https://user-images.githubusercontent.com/6444116/215938106-12f0e98f-fb4b-457a-9213-dab2cc2f6d41.gif)

I do not know how to explain this properly, but I will do my best. When TAStudio is closed, the X and Y coordinates, and the width and height of the window is saved somewhere in the `config.ini`. Somehow, when TAStudio is opened, the X coordinate somehow isn't correct, so it defaults to the minimum value of Monitor 2's width (X: 0). This happens with the same thing with RAM Watch, where the X coordinate value would be limited by the minimum width of Monitor 2.

I am assuming that, because TAStudio is closed in the secondary monitor, and then EmuHawk is closed, the `config.ini` would store the X and Y coordinates of TAStudio of the secondary monitor, but not taking into account the total width of 2 monitors in a dual monitors setup.

### Repro
1. Have a dual monitor setup ready.
2. Have the dual monitor set up so that the primary monitor is on the left side, and the secondary monitor is on the right side.
3. Open EmuHawk.
4. Open TAStudio.
5. Drag EmuHawk to the primary monitor.
6. Drag TAStudio to the secondary monitor.
7. Close TAStudio.
8. Close EmuHawk, so that the `config.ini` file is properly saved.
9. Open EmuHawk.
10. Open TAStudio. TAStudio should appear in the secondary monitor.
11. Drag TAStudio to the primary monitor.
12. Close TAStudio.
13. Open TAStudio again. Observe TAStudio appears in the secondary monitor.

### Output
There is no error output. EmuHawk, TAStudio, and RAM Watch windows all behave normally.

### Host env.
- BizHawk 2.8; Win10 Pro x64 21H2 Build 19044.2486; Intel/NVIDIA
[...]

PR #4011, opened 2024-08-23
Add option to disable automatic resize of main window

### LINKED ISSUE #1560: Window is spontaneously moved/recreated during Yoshi's island's opening
The main emulator window suddenly teleports to the middle of the screen when Yoshi's island switches from the island scene back to the storybook scenes.

kalimag:
- Add option to disable the automatic resizing of the main window whenever the framebuffer size changes
  - Window can be resized arbitrarily and stays that size
  - Selecting a standard scale through the menu/keybinds/API still works
- Restore last window size after restart (if automatic resize is disabled)
- Minor changes to saved window position
  - Make config values nullable instead of magic `-1` value
  - Don't save window position if minimized (this didn't quite work as intended) or maximized

This feature is useful for streamers who want a consistent capture size, as well as for people who want to fit the BizHawk window into a fixed layout with other windows (corner snapping etc.).

This is not the same functionality as the custom resolution in the display settings, which stretches the framebuffer to the specified size and fits the window to that size exactly.

With the feature in this PR, the window/client area stays a consistent size while automatic letterboxing/pillarboxing, integer resizing, 1:1 pixels etc. still work. This is preferable for working with multiple games, resolution changes mid-game, changing DS/WS screen rotation/layouts and similar, where squeezing everything into the same resolution is undesirable. It's also just more convenient than figuring out a resolution in the right aspect ratio and entering that in the display settings all the time.

Check if completed:

resolves #1560, #3039, and #3410

### DIFF
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -136,8 +136,8 @@ namespace BizHawk.Client.Common
 		public bool MainFormStayOnTop { get; set; }
 		public bool StartPaused { get; set; }
 		public bool StartFullscreen { get; set; }
-		public int MainWndx { get; set; } = -1; // Negative numbers will be ignored
-		public int MainWndy { get; set; } = -1;
+		public int? MainWndx { get; set; }
+		public int? MainWndy { get; set; }
 		public int? MainWindowWidth { get; set; }
 		public int? MainWindowHeight { get; set; }
 		public bool RunInBackground { get; set; } = true;
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Drawing;
 using System.IO;
 
 using BizHawk.Bizware.Graphics;
@@ -136,10 +137,8 @@ namespace BizHawk.Client.Common
 		public bool MainFormStayOnTop { get; set; }
 		public bool StartPaused { get; set; }
 		public bool StartFullscreen { get; set; }
-		public int? MainWndx { get; set; }
-		public int? MainWndy { get; set; }
-		public int? MainWindowWidth { get; set; }
-		public int? MainWindowHeight { get; set; }
+		public Point? MainWindowPosition { get; set; }
+		public Size? MainWindowSize { get; set; }
 		public bool RunInBackground { get; set; } = true;
 		public bool AcceptBackgroundInput { get; set; }
 		public bool AcceptBackgroundInputControllerOnly { get; set; }
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -274,6 +274,8 @@ namespace BizHawk.Client.Common
 		public int DispCropRight { get; set; } = 0;
 		public int DispCropBottom { get; set; } = 0;
 
+		public bool ResizeWithFramebuffer { get; set; } = true;
+
 		// Sound options
 		public ESoundOutputMethod SoundOutputMethod { get; set; } = HostCapabilityDetector.HasXAudio2 ? ESoundOutputMethod.XAudio2 : ESoundOutputMethod.OpenAL;
 
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -138,6 +138,8 @@ namespace BizHawk.Client.Common
 		public bool StartFullscreen { get; set; }
 		public int MainWndx { get; set; } = -1; // Negative numbers will be ignored
 		public int MainWndy { get; set; } = -1;
+		public int? MainWindowWidth { get; set; }
+		public int? MainWindowHeight { get; set; }
 		public bool RunInBackground { get; set; } = true;
 		public bool AcceptBackgroundInput { get; set; }
 		public bool AcceptBackgroundInputControllerOnly { get; set; }
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -274,6 +274,9 @@ namespace BizHawk.Client.Common
 		public int DispCropRight { get; set; } = 0;
 		public int DispCropBottom { get; set; } = 0;
 
+		/// <summary>
+		/// Automatically resize main window when framebuffer size changes (default behavior)
+		/// </summary>
 		public bool ResizeWithFramebuffer { get; set; } = true;
 
 		// Sound options
--- a/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
@@ -628,7 +628,7 @@ namespace BizHawk.Client.EmuHawk
 		private void WindowSize_Click(object sender, EventArgs e)
 		{
 			Config.SetWindowScaleFor(Emulator.SystemId, (int) ((ToolStripMenuItem) sender).Tag);
-			FrameBufferResized(true);
+			FrameBufferResized(forceWindowResize: true);
 		}
 
 		private void DisableResizeWithFramebufferMenuItem_Click(object sender, EventArgs e)
--- a/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
@@ -625,16 +625,16 @@ namespace BizHawk.Client.EmuHawk
 			DisableResizeWithFramebufferMenuItem.Checked = !Config.ResizeWithFramebuffer;
 		}
 
-		private void WindowSize_Click(object sender, EventArgs e)
-		{
-			Config.SetWindowScaleFor(Emulator.SystemId, (int) ((ToolStripMenuItem) sender).Tag);
-			FrameBufferResized(forceWindowResize: true);
-		}
-
 		private void DisableResizeWithFramebufferMenuItem_Click(object sender, EventArgs e)
 		{
 			Config.ResizeWithFramebuffer = !DisableResizeWithFramebufferMenuItem.Checked;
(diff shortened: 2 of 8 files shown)


src/BizHawk.Client.EmuHawk/MainForm.Events.cs
```diff
@@ -615,15 +615,25 @@ private void ViewSubMenu_DropDownOpened(object sender, EventArgs e)
 		private void WindowSizeSubMenu_DropDownOpened(object sender, EventArgs e)
 		{
 			var windowScale = Config.GetWindowScaleFor(Emulator.SystemId);
-			foreach (ToolStripMenuItem item in WindowSizeSubMenu.DropDownItems)
+			foreach (ToolStripMenuItem item in WindowSizeSubMenu.DropDownItems.OfType<ToolStripMenuItem>())
```
YoshiRulz, line 618: Did you mean
```suggestion
			foreach (var item in WindowSizeSubMenu.DropDownItems.Cast<ToolStripMenuItem>())
```
?
You shouldn't have to change it though.

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
```diff
@@ -615,15 +615,25 @@ private void ViewSubMenu_DropDownOpened(object sender, EventArgs e)
...
+			DisableResizeWithFramebufferMenuItem.Checked = !Config.ResizeWithFramebuffer;
 		}
 
 		private void WindowSize_Click(object sender, EventArgs e)
 		{
 			Config.SetWindowScaleFor(Emulator.SystemId, (int) ((ToolStripMenuItem) sender).Tag);
+			FrameBufferResized(forceWindowResize: true);
+		}
+
+		private void DisableResizeWithFramebufferMenuItem_Click(object sender, EventArgs e)
```
YoshiRulz, line 634: I think if you inserted this method above the other one the diff would be cleaner? Not that important

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -60,23 +60,17 @@ private void MainForm_Load(object sender, EventArgs e)
 		{	
 			UpdateWindowTitle();
 			
-			ToolStripItem[] CreateWindowSizeFactorSubmenus()
```
YoshiRulz, line 63: Unrelated removal of this local method?

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -619,9 +613,17 @@ _argParser.SocketAddress is var (socketIP, socketPort)
 				Config.MainWndy = 0;
 			}
 
-			if (Config.MainWndx != -1 && Config.MainWndy != -1 && Config.SaveWindowPosition)
+			if (Config.SaveWindowPosition)
 			{
-				Location = new Point(Config.MainWndx, Config.MainWndy);
+				if (Config.MainWndx is int x && Config.MainWndy is int y)
```
YoshiRulz, line 618: Was it necessary to make these nullable to implement this new feature? If you could continue using `-1` as the sentinel value instead of `null`, I think that would be preferable. Otherwise, try `Point?` instead of 2 `int?`s.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1372,8 +1374,13 @@ public void TakeScreenshot(string path)
 			AddOnScreenMessage($"{fi.Name} saved.");
 		}
 
-		public void FrameBufferResized()
+		/// <param name="forceWindowResize">Override <see cref="Common.Config.ResizeWithFramebuffer"/></param>
+		public void FrameBufferResized(bool forceWindowResize = false)
```
YoshiRulz, line 1378: Omit docs and default argument on implementation

YoshiRulz (dismissed):

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
kalimag: No, it's filtering out the separator

src/BizHawk.Client.EmuHawk/MainForm.cs
kalimag: It's not necessary to the PR, no. I just made this change because negative values including -1 are valid window positions

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
YoshiRulz: I think the type check on `Tag` is enough for that, but you could also do something like:
```csharp
foreach (var item in WindowSizeSubMenu.DropDownItems.Cast<ToolStripItem>())
{
	if (item is not ToolStripSeparator && item.Tag is int itemScale)
	{
		// ...
	}
}
```
The confusing type hierarchy means it's not clear what `.OfType<ToolStripMenuItem>()` does just from looking at it; normally it'd be fine.

src/BizHawk.Client.EmuHawk/MainForm.cs
kalimag: I had to change this method to insert the items at the top of the menu (above the separator and new item added by this PR), and since there's no `InsertRange` method or similar the simplest way was inserting the items directly and dropping the whole array -> `AddRange` step.

Since theres nothing to return, the local method would just be

```csharp
void CreateWindowSizeFactorSubmenus() { ... }
CreateWindowSizeFactorSubmenus();
```

YoshiRulz: Oh right. I suggest you add those new items programmatically rather than in the Designer. Alternatively, just add a "scoping block" around this bit and re-indent it, that will make the diff readable.

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
kalimag: The items need to be cast to `ToolStripMenuItem` to access the `Checked` property so there needs to be some kind of additional cast. I can change it to `item is ToolStripMenuItem menuItem` or something inside the loop if that's clearer

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: Possibly related to https://github.com/TASEmulators/BizHawk/issues/3544?

kalimag: Not directly, tool window position saving is handled elsewhere

src/BizHawk.Client.EmuHawk/MainForm.cs
kalimag: Are you sure? The majority of calls are from within `MainForm` itself, not through the interfaces

src/BizHawk.Client.EmuHawk/MainForm.Events.cs
YoshiRulz: Yes, do that

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: IIRC if you omit the default argument you're sometimes forced to do `((IIntf) this).FrameBufferResized()` nonsense. I'm of the opinion that it should be explicit at every callsite anyway, but for this PR having it duplicated isn't a problem, it'll get fixed later.

Definitely drop the doc comment though.

YoshiRulz:
Adding
```diff
 typeof(object),
+typeof(Point),
 typeof(Queue<>),
 typeof(/*Single*/float),
+typeof(Size),
 typeof(string),
```
to
https://github.com/TASEmulators/BizHawk/blob/dc582a4e7a1bc688a6eb6fdf8427fb9736bf10c9/src/BizHawk.Tests/Client.Common/config/SerializationStabilityTests.cs#L34-L37
should fix the test failure.

src/BizHawk.Client.Common/config/Config.cs
```diff
@@ -136,8 +137,8 @@ public void SetWindowScaleFor(string sysID, int windowScale)
 		public bool MainFormStayOnTop { get; set; }
 		public bool StartPaused { get; set; }
 		public bool StartFullscreen { get; set; }
-		public int MainWndx { get; set; } = -1; // Negative numbers will be ignored
-		public int MainWndy { get; set; } = -1;
+		public Point? MainWindowPosition { get; set; }
+		public Size? MainWindowSize { get; set; }
```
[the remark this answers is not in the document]
YoshiRulz: snipe'd :P

kalimag:
`Point` and `Size` serialization will have to be re-tested on #3932

kalimag:
> LGTM now, so I've made a note to merge this before the upcoming RC.
> 
> Thanks for all the contributions recently.

Well, I finally had to stop procrastinating to get any of this into a release before 2.10.1 in December 2025 ;)

kalimag:
I have a few more changes that I didn't PR yet because of the merge conflicts with(out) this PR: https://github.com/kalimag/BizHawk/compare/pr/disable-automatic-resize...kalimag:BizHawk:pr/disable-automatic-resize-plus

Should I wait until this one is merged and open a new PR, or just push them into this PR?

YoshiRulz:
I don't follow. Merge conflicts with `master`? You're welcome to rewrite history and force push, just comment with what you changed compared to the old HEAD.

kalimag:
What I mean is that those changes are not strictly related to this PR, but the commits touch the same code and are based on this PR's branch rather than master.

So I can't trivially open a separate PR with just those changes without (probably) creating a bunch of merge conflicts between the two, and the easier option is to either have one PR (this one) containing all changes, or wait until this one is merged to open a second PR.

skywind3000:
Thanks, but I still can't find the new option to keep the window size:

图片

I am using the latest `BizHawk-2.10-rc1-win-x64.zip`

YoshiRulz:
`View` > `Window Size` > `Static Size`

skywind3000:
thanks, that’s really helpful. Why not make it enable by default?

PR #4022, opened 2024-09-05
Disk Drive Activity LED ToolTip enhancement

Asnivor:
A number of cores are using the IDriveLight interface, but many of them aren't using it for disk drive activity. We have disks, CDs, carts and other things hooked up.

This PR adds a DriveLightIconDescription string to the IDriveLight interface, allowing for cores to set their own ToolTip on the drive light icon.

I was planning on using a default implementation in the interface (so that cores didn't have to explicitly set a value and we could use the default), but it looks like this might require a higher version of C#.

Thoughts?

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -2679,6 +2679,7 @@ namespace BizHawk.Client.EmuHawk
 			if (Emulator.HasDriveLight() && Emulator.AsDriveLight() is { DriveLightEnabled: true } diskLEDCore)
 			{
 				LedLightStatusLabel.Image = diskLEDCore.DriveLightOn ? _statusBarDiskLightOnImage : _statusBarDiskLightOffImage;
+				LedLightStatusLabel.ToolTipText = Emulator.AsDriveLight().DriveLightIconDescription;
 				LedLightStatusLabel.Visible = true;
 			}
 			else
--- a/src/BizHawk.Emulation.Common/Interfaces/Services/IDriveLight.cs
+++ b/src/BizHawk.Emulation.Common/Interfaces/Services/IDriveLight.cs
@@ -16,9 +16,7 @@
 		/// </summary>
 		bool DriveLightOn { get; }
 
-		/// <summary>
-		/// Gets a value indicating the description of the drive light icon (that will be displayed as a ToolTip on MainForm
-		/// </summary>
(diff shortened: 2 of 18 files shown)


src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -2688,6 +2688,8 @@ private void HandleToggleLightAndLink()
 					LedLightStatusLabel.Image = Emulator.AsDriveLight().DriveLightOn
 						? _statusBarDiskLightOnImage
 						: _statusBarDiskLightOffImage;
+
+					LedLightStatusLabel.ToolTipText = Emulator.AsDriveLight().DriveLightIconDescription;
```
YoshiRulz, line 2692: Memoise: `var diskLEDCore = Emulator.AsDriveLight();` (why that wasn't done already I don't know)

Also there might be some way to avoid updating the label when it hasn't changed? `GetHashCode`? Or maybe WinForms does that internally already. I don't think any part of the app handles that at the moment though so don't worry about it.

YoshiRulz (approved):
I like this idea.

> I was planning on using a default implementation in the interface (so that cores didn't have to explicitly set a value and we could use the default), but it looks like this might require a higher version of C#.

Almost right: it depends on the *Runtime* target. https://github.com/TASEmulators/BizHawk/wiki/Available-C%23-and-.NET-features

Asnivor:
It feels like we could do something better here. Being able to instantiate multiple activity lights on by-core basis, all with custom tooltips and even different icons. Being able to have separate LEDs for disk, tape, CD or whatever might be quite a nice improvement, rather than ramming lots of things down into one indicator.

I'm just not entirely sure how to implement it yet.

src/BizHawk.Client.EmuHawk/MainForm.cs
Asnivor: Like this? https://github.com/TASEmulators/BizHawk/pull/4022/commits/1857557e86fe731cd1ddf9bed96b17f5b2a5e8cc

YoshiRulz:
Fixed on `master` as 6d05b9874c3dc7d0946055d3181745294c447267 and rebased this onto that.

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: Fixed on `master` as 6d05b9874c3dc7d0946055d3181745294c447267 and rebased this onto that.

Issue #3989, opened 2024-08-06
TAStudio modifies global "On Movie End" setting


RetroEdit:
(Split from #3735 ; originally written 2023-09-22)
I recently realized Play Movie will automatically override your `On Movie End` setting to be `Record`. **EDIT 2024-08-18**: See below: it's probably not directly Play Movie related.

~~This is extremely undesirable behavior for me... it's definitely possible to change the toggle inside `Play Movie` to pause on a specific frame, but the current behavior of silently altering user-selected settings seems extremely bad practice to me. It's sort of a UI choice more than a bug per se, though.~~
~~At the very least, it should remember if you wanted to end on a specific frame (unfortunately, this alone may have downsides because pausing on a non-movie end frame via play movie is sort of specific to the movie). Personally, I would like the a direct equivalent to the `On Movie End` options selection visible within Play Movie if it's going to permanently alter those settings.~~

Morilli:
Can you explain when this happens? Just playing a movie using the play movie dialog does respect the "On Movie End" settings for me.

RetroEdit:
Sadly, after numerous attempts, I don't seem to be able to reproduce this. 
Not sure if I was just confused 11 months ago when I wrote it or it's a weird edge case.

I will say: there seem to only be two places in the entire codebase where the `OnMovieEnd` setting is changed implicitly: TAStudio and Macros.

https://github.com/TASEmulators/BizHawk/blob/d9331c5a2808901c1aadfd91be90125f6649a13e/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs#L290-L292

https://github.com/TASEmulators/BizHawk/blob/d9331c5a2808901c1aadfd91be90125f6649a13e/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs#L741

https://github.com/TASEmulators/BizHawk/blob/963b925a1a65b73afe234a7a077d642ab70c516f/src/BizHawk.Client.EmuHawk/tools/Macros/MovieZone.cs#L200-L204

To me, implicit config changes to OnMovieEnd seem kind of hacky, especially when it's changing it to "Record", which can append junk to the end of your movie. I did some testing just now, and if you open TAStudio and it doesn't clean up properly (e.g., maybe it crashed for some reason), it will indeed permanently and silently save If I did encounter some issue with Play Movie, TAStudio seems the most likely culprit.

[...]

Issue #2622, opened 2021-02-13
Make "Autopause at end of Movie" work in TAStudio


vadosnaprimer:
The option has been there forever and never seemed to work. I tried to set it up in MovieSession and tastudio settings and failed.

see also #3989

RetroEdit:
This cannot be fixed without first fixing #2292 since currently the end of the movie is not reachable during normal playback in TAStudio (is reachable in recording mode though).

Morilli:
How is this supposed to work? Because of the way TAStudio works, when you're at the end of the movie, the movie will continually extend. So effectively you're always "at the end of the movie".

The way this works for normal movies is by detecting whether or not the movie is recording mode: In recording mode, the movie is silently extended, and autopause only has an effect while in play mode. However TAStudio does not enter recording mode upon reaching the end of the movie, so you'd get a pause every frame.

YoshiRulz:
Pause the frame before the end of the movie then.

vadosnaprimer:
Here's how taseditor does it (reliable source since it's been done there first, and it has no bugs)

https://github.com/TASEmulators/fceux/blob/f980ec2bc7dc962f6cd76b9ae3131f2eb902c9e7/src/drivers/win/taseditor/playback.cpp#L177-L182

PR #4023, opened 2024-09-07
Cleanup before/after frame logic in MovieSession

### LINKED ISSUE #2292: TAStudio records extra frames at the end during playback

Morilli:
-closes #2292.

This does change some behavior, but hopefully the new structure is more logical and more correct. Calling `HandleFrameAfter` from `HandleFrameBefore` definitely didn't feel proper.

This is in preparation for #3989, as this changed logic should be much clearer in what specific changes TAStudio wants: If the current playback is at the end of the movie,
1. the `HandlePlaybackEnd` function should not be called and instead the movie should just stay in the mode it was previously in
2. on frame advance, input should still be recorded, even in playback mode.

The 2. point is the biggest question mark to me still. There is this line `MovieController.SetFromSticky(StickySource);` which can currently only be hit on the very first frame or when enabling recording mode, advancing to the end of the tas movie, disabling recording mode and then advancing a single frame without entering input in the input roll.
If I knew how to get sticky/autohold input I would test it, although I'm assuming the current logic is incorrect if anything.

Could maybe also help with #2622, although that will require additional consideration.

### DIFF
--- a/src/BizHawk.Client.Common/movie/MovieSession.cs
+++ b/src/BizHawk.Client.Common/movie/MovieSession.cs
@@ -81,32 +81,30 @@ namespace BizHawk.Client.Common
 			else if (Movie.IsPlaying())
 			{
 				LatchInputToLog();
-
-				if (Movie.IsRecording()) // The movie end situation can cause the switch to record mode, in that case we need to capture some input for this frame
+				// if we're at the movie's end and the MovieEndAction is record, just continue recording in play mode
+				// TODO change to TAStudio check
+				if (Movie is ITasMovie && Movie.Emulator.Frame == Movie.FrameCount && Settings.MovieEndAction == MovieEndAction.Record)
 				{
-					HandleFrameLoopForRecordMode();
+					Movie.RecordFrame(Movie.Emulator.Frame, MovieOut.Source);
 				}
 			}
 			else if (Movie.IsRecording())
 			{
-				HandleFrameLoopForRecordMode();
+				LatchInputToUser();
+				Movie.RecordFrame(Movie.Emulator.Frame, MovieOut.Source);
 			}
 		}
 
-		// TODO: this is a mess, simplify
 		public void HandleFrameAfter()
 		{
 			if (Movie is ITasMovie tasMovie)
 			{
 				tasMovie.GreenzoneCurrentFrame();
-				if (tasMovie.IsPlayingOrFinished() && Settings.MovieEndAction == MovieEndAction.Record && Movie.Emulator.Frame >= tasMovie.InputLogLength)
-				{
-					HandleFrameLoopForRecordMode();
-					return;
-				}
+				// TODO change to TAStudio check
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Which of the things described in #2292 does it fix?

Morilli:
> Which of the things described in #2292 does it fix?

The part where TAStudio always keeps an extra frame at the end in playback mode. The behavior is now effectively the same as record mode. When pressing frame advance after creating a new tasproj, the movie will have a 1 frame length.

vadosnaprimer:
Alright yep, I think this fix makes it possible to properly set movie duration on its last frame in tastudio. When we've emulated the last frame of the tasproj, we're in the same state as if we've just recorded a regular movie. Obviously drawing on last frame in the InpuRoll adds one more row which is correct, but having playback cursor on that frame should not have added yet another frame to the movie, because the fact of emulating it does not require any new frames.

vadosnaprimer:
> on frame advance, input should still be recorded, even in playback mode.

I disagree with this if it's not optional.

Morilli:
> > on frame advance, input should still be recorded, even in playback mode.
> 
> I disagree with this if it's not optional.

This is just how TAStudio has worked the last couple years. The alternative would be to respect the "on movie end" setting completely, but then you'd get a switch to record mode at the end, which may or may not be what people expect when using TAStudio.

Issue #3760, opened 2023-08-25
Many PSX ID Hashes Are Probably Wrong


CasualPokePlayer:
Somewhat recently (around 2021?) it seems that redump has placed a DCP flag on PSX dumps. While this doesn't actually affect the actual disc contents (so the full disc hash is unchanged), it does affect the TOC, and so the ID hashes will be wrong and good dumps by redump standards will be marked bad by BizHawk. `Castlevania - Symphony of the Night (USA)` is a recent example of this change brought up on Discord when trying to figure out why someone's dump was being marked as bad, which turned out that the gamedb hashes assume no DCP flag is present.

These are all the games that will probably need to have the ID hashes rechecked (based on redump cues with DCP present)
```
Afraid Gear (Japan)
Aitakute... - Your Smiles in My Heart (Japan) (Disc 1)
Aitakute... - Your Smiles in My Heart (Japan) (Disc 2)
Aitakute... - Your Smiles in My Heart (Japan) (Disc 3)
Aitakute... - Your Smiles in My Heart (Japan) (Disc 4)
Aitakute... - Your Smiles in My Heart - Oroshitate no Diary - Introduction Disc (Japan)
Akumajou Dracula X - Gekka no Yasoukyoku (Japan) (Rev 1)
Akumajou Dracula X - Gekka no Yasoukyoku (Japan)
Azure Dreams (USA)
Breath of Fire IV (USA) (Beta)
Breeding Stud 2 (Japan)
Bugi (Japan)
Castlevania - Symphony of the Night (Asia)
Castlevania - Symphony of the Night (USA)
Cheesy (Japan)
Crypt Killer (Europe)
Dance Dance Revolution (Japan)
Dance Dance Revolution (USA)
Dance Dance Revolution - Disney's Rave (Japan)
Dance Dance Revolution - Extra Mix (Japan)
Dance Dance Revolution 2nd Remix (Japan)
Dance Dance Revolution 2nd Remix - Append Club Version Vol. 1 (Japan, Asia)
Dance Dance Revolution 2nd Remix - Append Club Version Vol. 2 (Japan)
Dance Dance Revolution 5th Mix (Japan)
Dance Dance Revolution Konamix (USA)
Dancing Stage - Euromix (Europe) (Beta)
Dancing Stage - Party Edition (Europe)
Dancing Stage Fever (Europe)
Dancing Stage Fusion (Europe) (En,Fr,De,Es,It)
[...]

PR #4024, opened 2024-09-08
Fix Sega CD/Saturn disc hashing

kalimag:
Fix `DiscHasher.OldHash()` method used to hash non-PSX discs
- Don't hash generated (empty) lead-in/lead-out tracks
- Limit to track length, not absolute LBA of next track

Previously, Saturn and Sega CD discs are all hashed to the same value `D41D8CD98F00B204E9800998ECF8427E` (empty sequence). This happens because the method is trying to hash the non-existent lead-in track and `track.NextTrack.LBA` is `0` (the LBA of the "real" first track).

Hashing PC Engine discs already mostly worked because the first track is usually an audio track and the lead-in track copies the mode of the first track and is skipped. A handful of discs where the first data track is <512 sectors may generate a different hash now, because the old implementation was reading past the end of the track. The only examples of that I could find were some unlicensed games that weren't in the gamedb.

Jaguar CD hashing still doesn't work because all tracks are marked as audio on those discs.

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
+++ b/src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
@@ -88,7 +88,7 @@ namespace BizHawk.Emulation.DiscSystem
 			var buffer = new byte[512 * 2352];
 			var dsr = new DiscSectorReader(disc);
 			// don't hash generated lead-in and lead-out tracks
-			for (int i = 1; i < disc.Session1.Tracks.Count - 1; i++)
+			for (int i = 1; i <= disc.Session1.InformationTrackCount; i++)
 			{
 				var track = disc.Session1.Tracks[i];
 
--- a/src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
+++ b/src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
@@ -87,12 +87,15 @@ namespace BizHawk.Emulation.DiscSystem
 		{
 			var buffer = new byte[512 * 2352];
 			var dsr = new DiscSectorReader(disc);
-			foreach (var track in disc.Session1.Tracks)
+			// don't hash generated lead-in and lead-out tracks
+			for (int i = 1; i < disc.Session1.Tracks.Count - 1; i++)
 			{
+				var track = disc.Session1.Tracks[i];
+
 				if (track.IsAudio)
 					continue;
 
-				var lba_len = Math.Min(track.NextTrack.LBA, 512);
+				var lba_len = Math.Min(track.NextTrack.LBA - track.LBA, 512);
 				for (var s = 0; s < 512 && s < lba_len; s++)
 					dsr.ReadLBA_2352(track.LBA + s, buffer, s * 2352);
 
--- a/src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
+++ b/src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
@@ -1,3 +1,7 @@
+#nullable enable
+
+using System.Collections.Generic;
+using System.Text;
 using BizHawk.Common;
 using BizHawk.Common.BufferExtensions;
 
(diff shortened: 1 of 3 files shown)


YoshiRulz:
Does this resolve #1075 then? That would be huge.

kalimag:
Kind of? I tested a fair number of Saturn, Sega CD and PC Engine discs, and didn't find any collisions between different games. However, different revisions/bad dumps do occasionally produce the same hash, which is expected since it's only hashing the first track.

BizHawk now has another set of console-specific disc hashing functions in the RA code. I don't know how well they deal with revisions, but perhaps reusing those for general hashing would also be an option. They currently live in the `EmuHawk` project, but other than swapping out the RA `ConsoleID` for BizHawks `DiscId` or `SystemId` it probably wouldn't be hard to move them into the disc or common project.

CasualPokePlayer:
> BizHawk now has another set of console-specific disc hashing functions in the RA code. I don't know how well they deal with revisions, but perhaps reusing those for general hashing would also be an option. They currently live in the EmuHawk project, but other than swapping out the RA ConsoleID for BizHawks DiscType or SystemId it probably wouldn't be hard to move them into the disc or common project.

FWIW, doing this could also help solve the issue of 3DS hashing, as rcheevos has their own limited hashing method for it. The downside of course, is it requires firmware (aes_keys.txt / seeddb.bin) in the case the game is encrypted (although if the firmware is not present the game couldn't be loaded anyways).

The disc hashing code too is also meant to be more a secondary implementation compared to the rcheevos implementation, although I've been mulling considering just removing it and deferring the hashing code entirely to rcheevos, like debug hashing code does (it has callbacks set in place so it will end up using our own disc system just fine). For general hashing however, it could perhaps still be better to have our own implementation so we are not directly subject to rcheevos in normal use (i.e. most users).

CasualPokePlayer:
> Does this resolve #1075 then? That would be huge.

Jaguar CD still won't work with this approach. Fundementally, Jaguar CDs are "audio" CDs, they never have "data tracks" (this was a design decision by Atari to squeeze out more space, at of course the obvious cost of no error correction data and so a much higher risk of disc read failures). You'd want a specialized disc hashing approach here, likely doing what [RA does](https://github.com/TASEmulators/BizHawk/blob/a72b0f7e8c181f3b140ab401238c4f77f08ad0f7/src/BizHawk.Client.EmuHawk/RetroAchievements/RetroAchievements.GameVerification.cs#L168-L275).

src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
```diff
@@ -87,12 +87,15 @@ public string OldHash()
 		{
 			var buffer = new byte[512 * 2352];
 			var dsr = new DiscSectorReader(disc);
-			foreach (var track in disc.Session1.Tracks)
+			// don't hash generated lead-in and lead-out tracks
+			for (int i = 1; i < disc.Session1.Tracks.Count - 1; i++)
```
CasualPokePlayer, line 91: Use `i <= disc.Session1.InformationTrackCount;` rather here.

CasualPokePlayer (changes requested):

src/BizHawk.Emulation.DiscSystem/DiscHasher.cs
YoshiRulz:
```suggestion
			for (int i = 1; i <= disc.Session1.InformationTrackCount; i++)
```
?

CasualPokePlayer (approved):

kalimag:
- Is there any reason not to also hash the TOC for non-PSX discs? (apart from #3760)
- Should there be special handling for old movies saved with the meaningless `D41D8CD98F00B204E9800998ECF8427E` hash? 
- Is it alright to just duplicate the Jaguar CD hashing code from the RA integration into the DiscSystem project to cover that system as well?

YoshiRulz:
I'm afraid this didn't end up making it into RC1. I think it's because I pushed `master:release` without pulling `master`.

PR #4026, opened 2024-09-08
Improve Lua REPL handling of statements

kalimag:
Rework the way the Lua console REPL executes commands

The old way first tries to execute `$"console.log({rawCommand})"`, which prints out the values of an entered expression, but results in a syntax error if the command is a statement such as `foo = 1`. If the first attempt errors, it executes the raw command without the log call, so the command still works, but it's printing out confusing syntax error messages anyway. If the expression itself causes an error like `"a" + 1` or `string.format("%error")`, it tries to run it twice and prints out two different errors.

The PR changes it so `LuaLibraries.ExecuteString` attempts to first parse the modified command without running it, and then only executes either the modified command or the raw command. This avoids unnecessary error messages, double execution of certain commands etc.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
@@ -1388,18 +1388,14 @@ namespace BizHawk.Client.EmuHawk
 
 					LuaSandbox.Sandbox(null, () =>
 					{
-						LuaImp.ExecuteString($"console.log({rawCommand})");
-					}, () =>
-					{
-						LuaSandbox.Sandbox(null, () =>
+						var results = LuaImp.ExecuteString(rawCommand);
+						// empty array if the command was e.g. a variable assignment or a loop without return statement
+						// "void" functions return a single null
+						// if output didn't change, Print will take care of writing out "(no return)"
+						if (results is not [ ] and not [ null ] || OutputBox.Text == consoleBeforeCall)
 						{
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -1388,18 +1388,14 @@ private void InputBox_KeyDown(object sender, KeyEventArgs e)
...
 					{
-						LuaImp.ExecuteString($"console.log({rawCommand})");
-					}, () =>
-					{
-						LuaSandbox.Sandbox(null, () =>
+						var results = LuaImp.ExecuteString(rawCommand);
+						// empty array if the command was e.g. a variable assignment or a loop without return statement
+						// "void" functions return a single null
+						// if output didn't change, Print will take care of writing out "(no return)"
+						if (results is not [ ] and not [ null ] || OutputBox.Text == consoleBeforeCall)
```
YoshiRulz, line 1395: nitpick: `is not ([] or [ null ])`
also `ReferenceEquals` for string comparison maybe

CasualPokePlayer: ReferenceEquals should not be used. The getter for Output.Text has no guarantee to actually be the same reference with 2 calls to it (especially in this case, getting the text from a native window is going to be a p/invoke call, which means creating a new string, i.e. a new reference)

CasualPokePlayer (approved):

PR #4029, opened 2024-09-13
Main window auto-size fixes

### LINKED ISSUE #2850: Window Size & position incorrectly adjusted when exiting Fullscreen
### Summary
A certain display configuration results in the application window being shifted up-left until the title bar and menu bar are displayed out of view. I must shift right-click the application on the taskbar and use the Move function to move the application and regain view of these application elements.

### Reproduction Steps
1) Open Bizhawk and the N64 video plugin settings.
2) Set the resolution on the Main tab to 1920x1080 (same resolution as the computer monitor).
3) On the GlideN64 tab navigate to the Display settings and set Aspect Ratio to Adjust.
4) In the emulator Display settings set the Aspect Ratio to 1:1 pixel perfect mode.
5) Load an N64 game and enter Fullscreen Mode.
6) Exit Fullscreen mode.
7) Application window is shifted upwards and marginally to the left.

### Computer specifications
Bizhawk v2.6.2
Windows 10 21H1
Nvidia Geforce 1070Ti (465.89)
Ryzen 3700X
1920x1080 monitor

### Screenshots
After exiting fullscreen:
image

kalimag:
- Ensure that the top left corner of the window remains on screen if the entire window cannot fit at 1x scale
  - Fix title bar/menu moving off-screen with unusually large framebuffer sizes
  - The bottom/right edge may move off-screen instead, but that seems like the lesser evil
- When `MainWindow.FrameBufferResized` is called while the window is minimized or maximized, defer the resize until the window is restored to normal again
  - Fix window being disproportioned or moving erratically because the resize logic was working off of the minimized/maximized window sizes
- Remember window maximized state after restart
  - This was intended to be part of #4011, but it depends on the previous fix to work properly

Resolves #2850

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -142,6 +142,7 @@ namespace BizHawk.Client.Common
 		public bool StartFullscreen { get; set; }
 		public Point? MainWindowPosition { get; set; }
 		public Size? MainWindowSize { get; set; }
+		public bool MainWindowMaximized { get; set; }
 		public bool RunInBackground { get; set; } = true;
 		public bool AcceptBackgroundInput { get; set; }
 		public bool AcceptBackgroundInputControllerOnly { get; set; }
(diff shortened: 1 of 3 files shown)


YoshiRulz (approved):
LGTM. Just to clarify, does this fix the linked issue for fullscreen or just maximised? Our fullscreen isn't real fullscreen anyway so I expect it's both.

kalimag:
Yeah it works for "fullscreen" as well.

PR #4033, opened 2024-09-16
Auto-generate preferred cores from source

### LINKED ISSUE #3960: Preferred core picker is intransparent and inflexible
The current core preferences picker (`Config`->`Preferred Cores`) is not fully transparent in how it works, and does not allow configuring everything. On top of that, it is also manually maintained and needs to be updated when cores get added or modified, which does / did not always happen.

Let's take this example: 
image
What happens when `Gambatte` is selected here? The preferred core for `GB` and `GBC` will be set to `Gambatte`. The fact that both `GB` and `GBC` are affected is not made clear anywhere and can only be determined by testing it out.
One might argue that this could be "user-friendly" by not overloading them with redundant options, but I'd say having preferences made explicit is preferable over implicitly assuming `GB` implies `GBC`.
We can already see that it is impossible to set different core preferences for `GB` and `GBC` games, and the same is true for a number of other options.

One thing I was only recently made aware of in the context of #3635: It is currently impossible to choose the `SubBSNESv115+` core for Satellaview games, even though it does have the corresponding `CoreConstructor` attribute, because it isn't exposed in the preferred core picker. Adding it to the `SNES` group is problematic because then the preference for `SNES` is applied to `Satellaview`, even though some SNES cores don't support Satellaview at all!

My suggestion is to make choices for all SystemIDs explicit and automatically generated. That would guarantee that 
- all SystemIDs can have a preferred core set via config
- all cores can be selected without having to manually add them to the config
- `CoreConstructor` attributes have user-facing meaning 

I have #4033 that shows my suggested approach.

see also #2289

Morilli:
- closes #3960

RFC.

This auto-generates the UI options for the core preferences picker from the list of available cores. This also requires changing the core attribute for many cores to signal that they are the preferred core for a specific system.
Basically, instead of having the core preference list stored entangled with the UI picker data and having to be manually maintained, the list is now based purely on the available cores.

This and the added tests ensure that all system ids with multiple core choices have an entry in the core picker (preventing mistakes like what d74b130c22a8e877cbf6ecec45a1ab714bb2f77f fixed) and that the default preference is always known, even when the user has configured a different preferred core for a given system id.

This will also make it impossible for fa08002fbf2d8481b6558630c2f57f51953b771e to occur, where quick(er)nes was both the preferred core and had lower than default priority. In reality this doesn't really show itself, but it's best to fix it nonetheless.

### DIFF
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -18,50 +18,15 @@ namespace BizHawk.Client.Common
 	{
 		public static string ControlDefaultPath => Path.Combine(PathUtils.ExeDirectoryPath, "defctrl.json");
 
-		/// <remarks>
-		/// <c>CoreNames[0]</c> is the default (out-of-the-box) core.<br/>
-		/// <c>AppliesTo</c> are concatenated to make the submenu's label, and
-		/// <c>Config.PreferredCores[AppliesTo[0]]</c> (lookup on global <see cref="Config"/> instance) determines which option is shown as checked.<br/>
-		/// The order within submenus and the order of the submenus themselves are determined by the declaration order here.
-		/// </remarks>
-		public static readonly IReadOnlyList<(string[] AppliesTo, string[] CoreNames)> CorePickerUIData = new List<(string[], string[])>
-		{
-			([ VSystemID.Raw.A26 ],
-				[ CoreNames.Atari2600Hawk, CoreNames.Stella ]),
-			([ VSystemID.Raw.Satellaview ],
-				[ CoreNames.Bsnes115, CoreNames.SubBsnes115 ]),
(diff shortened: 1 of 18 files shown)


YoshiRulz:
So this makes `CoreConstructor.Priority` the single-source-of-truth? I like a SSoT but I think we should remove the priority system completely.

Morilli:
> So this makes CoreConstructor.Priority the single-source-of-truth?

Yes, effectively. I'm myself not 100% sold on it, but it felt better than the current system at least. If we can get something set up that enforces that all systems that have multiple cores emulating it also have a preference choice in the UI, that would probably be enough already.

> I think we should remove the priority system completely.

Oh? So just have preferences set manually and don't allow cores to alter their own priority?

YoshiRulz:
Yeah, either in `Config` or `CoreInventory`.

Morilli:
Going to close this as it's not required and not agreed on the direction.

PR #4047, opened 2024-09-20
Refactor sticky controller logic to separate it from user input

Morilli:
I've refactored the entire sticky controller / adapter thing because it felt bad and inconsistent, and made it so sticky input is now set into a controller instead of being inherently overlayed on top of an existing `IController` source.

This has a couple of advantages, mainly TAStudio being able to use sticky input directly while ignoring user input. There's tons of bugs with autofire/autohold in TAStudio still, but this should be a step in the right direction.

Previous logic:
Userinput is fed into the StickyXorAdapter, where bool buttons are xor'd with the sticky value, and axis values are OR'd with the source, prioritising sticky axis value over user input axis value.
Then that output is fed into the AutofireStickyXorAdapter, where bool buttons are xor'd with the autofire value and axis values are OR'd with the (StickyXorAdapter) source, prioriting the autofire sticky value.
That output is then used as movie input.

New logic:
There's a `StickyHoldController` and a `StickyAutofireController`, whose button and axes values are OR'd together (prioritising autohold over sticky autofire) and stored as `StickyController`. Then, userinput and `StickyController` are xor'd to generate the movie input source.

By having this `StickyController` not rely on user input, it's possible to extract sticky values independently of the current state of user pressed buttons. As I said there's still heaps of incorrect behavior in TAStudio, but technically this should allow TAStudio to get correct sticky values.

One immediate improvement this brings is getting correct autofire values when extending a movie in TAStudio without being in recording mode (unless you click on the input roll or you have multiple autofire columns XD).

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
@@ -22,7 +22,7 @@ namespace BizHawk.Client.Common
 
 		public IReadOnlyDictionary<string, object> Get(int? controller = null)
 		{
-			return _inputManager.AutofireStickyXorAdapter.ToDictionary(controller);
+			return _movieSession.MovieIn.ToDictionary(controller);
 		}
 
 		public IReadOnlyDictionary<string, object> GetWithMovie(int? controller = null)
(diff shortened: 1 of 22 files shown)


YoshiRulz (commented):
Rename `StickyControllers.cs` to `StickyAdapters.cs` to shrink the diff.

Morilli:
I've restructured the entire class, the fact that git doesn't pick up the rename automatically should tell you the diff is large. I don't think you'll get much value out of comparing line by line, rather compare function-by-function.

Morilli:
better?

zeromus:
I can't review this, I can't remember how this nightmare code works

YoshiRulz (approved):
Not understanding how the controller stack works isn't going to stop me.

Sorry about the haptics stuff BTW, I never finished splitting up `IController` and IIRC haptics were simplified on my branch.

Issue #3863, opened 2024-03-01
[PSX][Octoshock+Nymashock][Gex - Enter the Gecko] BizHawk 2.9.1 locks up when attempting to load Gex - Enter the Gecko for PSX for both included cores


boktai1000:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"
[//]: # "One more thing: if you're on Linux, please comment on #1430 instead of opening an issue so we don't annoy the other devs."

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."

When attempting to open / run Gex - Enter the Gecko for PS1, BizHawk will completely lock up.

![image](https://github.com/TASEmulators/BizHawk/assets/9054563/0601cb66-8c17-47ad-980f-fbc56771412b)

```
Algorithm : MD5
Hash      : 68F3483C474B8582D985FD6283FEC66B
File      : Gex - Enter the Gecko (USA).cue

Algorithm : MD5
Hash      : 078092D2434011A12D5C56DA4DE0EEB4
File      : Gex - Enter the Gecko (USA).bin

Algorithm : MD5
Hash      : 490F666E1AFB15B7362B406ED1CEA246
Path      : PSX_3.0(A).bin
```

Additionally afterwards I have installed the Libretro PSX core following steps here that I used to try some of my CHD formatted games, and can confirm that this core/configuration did work and load the game properly for what it's worth.
https://www.reddit.com/r/EmulationOnPC/comments/11pj0u3/comment/jc1fjfl/

I have also tested these bin/cue files with DuckStation and RetroArch and can confirm they loaded without issue there.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Setup BizHawk with firmware required for PS1.
2. Attempt to load a Redump verified copy of Gex - Enter the Gecko for PS1 using defaults / default core, or swapping to other PS1 core (both Octoshock and Nymashock exhibit lockup behavior)
### Output
[...]

PR #4085, opened 2024-10-10
Fixes for PlayStation RetroAchievements hashing

prenoeg:
Minor fix for games like Chrono Cross where the SYSTEM.CNF file contains a slightly different BOOT line than expected:
"BOOT\t= cdrom:" vs
"BOOT = cdrom:"

Additional fix for games like Tekken 3 and Wild Arms where the PlayStation executable is in a subdirectory.  For example:
BOOT = cdrom:\TEKKEN3\SLUS_004.02;1 or
BOOT = cdrom:\EXE\SCUS_946.08;1

In these cases, the application would get stuck in an infinite loop while looking for the executable filename in the same disc sector.  I believe this is related to a previously closed issue (#3863).

### DIFF
--- a/src/BizHawk.Client.EmuHawk/RetroAchievements/RetroAchievements.GameVerification.cs
+++ b/src/BizHawk.Client.EmuHawk/RetroAchievements/RetroAchievements.GameVerification.cs
@@ -114,6 +114,9 @@ namespace BizHawk.Client.EmuHawk
 							dsr.ReadLBA_2048(sector, buf2048, 0);
 							exePath = Encoding.ASCII.GetString(buf2048);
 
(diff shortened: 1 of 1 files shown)


CasualPokePlayer:
These "fixes" seems to be something which ends up making BizHawk mismatch rcheevos' hash behavior (rcheevos fails to find the executable for Chrono Cross, and just fails to hash Tekken 3 regardless without any other errors logged). Presumingly these issues just happen on their end too and should be fixed.

Otherwise these do appear to be more correct behavior (the hashes even seem to match what RetroAchievements has set for those games... which is odd since again rcheevos doesn't seem to be able to hash these games anyways).

PR #4087, opened 2024-10-12
order firmware groups in the dialog and in the code

vadosnaprimer:
there's some discrepancy between the 2 orders, because the GUI list relies on full names (of the groups) while in the code short names are used to add them to DB

### DIFF
--- a/src/BizHawk.Client.EmuHawk/config/FirmwaresConfig.cs
+++ b/src/BizHawk.Client.EmuHawk/config/FirmwaresConfig.cs
@@ -179,11 +179,13 @@ namespace BizHawk.Client.EmuHawk
 
 		private void FirmwaresConfig_Load(object sender, EventArgs e)
 		{
+			string GetSystemGroupName(string sysID) => SystemGroupNames.TryGetValue(sysID, out var name) ? name : "FIX ME (FirmwaresConfig.cs)";
+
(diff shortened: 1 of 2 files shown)


Morilli:
I like this. Being prompted for a firmware and then having to search through the (seemingly) unsorted list is very annoying.

vadosnaprimer:
Because it's a nightmare to go through unsorted mess in case you need to add/edit something.

YoshiRulz:
Ctrl+F?

vadosnaprimer:
Dude seriously? You actually need it to be unsorted, and searching it by trying to remember what the short name is is absolutely the only allowed way?

YoshiRulz:
IDRC, but it would have been nice to keep the diff clean.

YoshiRulz:
It clobbers the blame, and `.git-blame-ignore-revs` (#3349) can't see through it:
```
$> echo a08cb73a30f2e0877fdbd89d1f59040d3418818a >.git-blame-ignore-revs
$> git blame src/BizHawk.Emulation.Common/Database/FirmwareDatabase.cs | grep -F a08cb73a3 | wc -l
274
```

vadosnaprimer:
Yeah sorry. Cleanest diff is no diff. Maybe commit description will help in such a dare situation.

Issue #3930, opened 2024-05-30
(Option to) Use grey stroke colour for `0x00` in Hex Editor


YoshiRulz:
I believe the Hex Editor's main pane is implemented as a single `Label` with a monospace font, so this wouldn't be trivial to add.

Issue #3769, opened 2023-09-09
Use slightly different colour for zeroes in Hex Editor


YoshiRulz:
edit: #3930 is a duplicate of this (mb), but that one is more detailed so closing in favour of that.

PR #4098, opened 2024-10-25
HexEditor

Asnivor:
This implements a HexLabelEx control that replaces the LocaLabelEx 'AddressesLabel' control in HexEditor.cs.
It overrides OnPaint(), parses the incoming string and draws the string out using different colours depending on whether the hex value is '00' or not.
A new colour selector is also implemented within HexColor.cs so the user can choose which colour they want displayed for 00 values.

This addresses #3769 and the duplicate #3930 

Notes:

1. I could not find a way to have all this stuff directly within HexEditor.cs itself. Maybe it's just a lack of knowledge, but I needed to intercept the OnPaint control method and NOT call base.OnPaint(). But if I disable base.OnPaint() in the control, the PaintEventHandler is never called. *shrug*.
2. I'm only able to test this on Windows currently. The various arbtrary spacing floats used in HexLabelEx would have to be tested on linux and modified accordingly (there is already an OSTailoredCode.IsUnixHost check in the constructor ready to be added to).

I'm sure it can all be cleaned up, but I'm stuck at the moment on the 'how'.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/HexEditor/HexColor.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/HexEditor/HexColor.Designer.cs
@@ -41,7 +41,9 @@
             this.HexMenubar = new System.Windows.Forms.Panel();
             this.label2 = new BizHawk.WinForms.Controls.LocLabelEx();
             this.HexBackgrnd = new System.Windows.Forms.Panel();
-            this.colorDialog1 = new System.Windows.Forms.ColorDialog();
+			this.label7 = new BizHawk.WinForms.Controls.LocLabelEx();
+			this.Hex00 = new System.Windows.Forms.Panel();
+			this.colorDialog1 = new System.Windows.Forms.ColorDialog();
             this.groupBox1.SuspendLayout();
             this.SuspendLayout();
             // 
@@ -59,9 +61,11 @@
             this.groupBox1.Controls.Add(this.HexMenubar);
             this.groupBox1.Controls.Add(this.label2);
             this.groupBox1.Controls.Add(this.HexBackgrnd);
-            this.groupBox1.Location = new System.Drawing.Point(3, 2);
+			this.groupBox1.Controls.Add(this.label7);
+			this.groupBox1.Controls.Add(this.Hex00);
+			this.groupBox1.Location = new System.Drawing.Point(3, 2);
             this.groupBox1.Name = "groupBox1";
-            this.groupBox1.Size = new System.Drawing.Size(144, 192);
+            this.groupBox1.Size = new System.Drawing.Size(144, 222);
             this.groupBox1.TabIndex = 0;
             this.groupBox1.TabStop = false;
             // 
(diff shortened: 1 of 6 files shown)


Asnivor:
image

Asnivor:
image

vadosnaprimer (approved):
Approving the look, ~~just add more info the PR title~~.

src/BizHawk.Client.EmuHawk/tools/HexEditor/HexColor.cs
```diff
@@ -20,6 +20,7 @@ private void HexColors_Form_Load(object sender, EventArgs e)
 			HexFreeze.BackColor = _hexEditor.Colors.Freeze;
 			HexFreezeHL.BackColor = _hexEditor.Colors.HighlightFreeze;
 			HexHighlight.BackColor = _hexEditor.Colors.Highlight;
+			Hex00.BackColor = _hexEditor.Colors.Foreground00;
```
YoshiRulz, line 23: Maybe use `Zeroes` in identifiers instead of `00`

src/BizHawk.Client.EmuHawk/tools/HexEditor/HexColor.Designer.cs
```diff
@@ -41,7 +41,9 @@ private void InitializeComponent()
             this.HexMenubar = new System.Windows.Forms.Panel();
             this.label2 = new BizHawk.WinForms.Controls.LocLabelEx();
             this.HexBackgrnd = new System.Windows.Forms.Panel();
-            this.colorDialog1 = new System.Windows.Forms.ColorDialog();
+			this.label7 = new BizHawk.WinForms.Controls.LocLabelEx();
+			this.Hex00 = new System.Windows.Forms.Panel();
+			this.colorDialog1 = new System.Windows.Forms.ColorDialog();
```
YoshiRulz, line 46: Whitespace noise can be cleaned up. If you want I can do it immediately before merge.

src/BizHawk.Client.EmuHawk/tools/HexEditor/HexEditor.cs
```diff
@@ -219,11 +220,21 @@ private void HexEditor_Load(object sender, EventArgs e)
 
 		protected override void UpdateAfter()
 		{
+			if (AddressesLabel.ZeroColor != Colors.Foreground00)
+			{
+				AddressesLabel.ZeroColor = Colors.Foreground00;
+			}
```
YoshiRulz, line 226: I don't think this belongs here

src/BizHawk.WinForms.Controls/BizHawk.WinForms.Controls.csproj
```diff
@@ -10,4 +10,9 @@
 		<PackageReference Include="System.Drawing.Common" />
 		<ProjectReference Include="$(ProjectDir)../BizHawk.Common/BizHawk.Common.csproj" />
 	</ItemGroup>
+	<ItemGroup>
+	  <Compile Update="LabelEx\HexLabelEx.cs">
+	    <SubType>Component</SubType>
+	  </Compile>
```
YoshiRulz, line 16: Does this have any effect? If so then I'd expect the rest of the controls in this project to be missing their IDE icons or whatever, since they're obviously not listed here.

src/BizHawk.WinForms.Controls/LabelEx/HexLabelEx.cs
```diff
@@ -0,0 +1,128 @@
+using System.ComponentModel;
+using System.Drawing;
+using System.Windows.Forms;
+using BizHawk.Common;
+
+
+namespace BizHawk.WinForms.Controls
+{
+	/// <inheritdoc cref="Docs.LabelOrLinkLabel"/>
```
YoshiRulz, line 9: Should add a `<seealso/>` to those docs

src/BizHawk.WinForms.Controls/LabelEx/HexLabelEx.cs
```diff
@@ -0,0 +1,128 @@
...
+			char gap = ' ';
+
+			string[] lines = text.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None);
+
+			Color color = this.ForeColor;
+
+			foreach (string line in lines)
+			{
+				// split left and right panes
+				string[] panes = line.Split('|');
```
YoshiRulz, line 57: Really don't like this. The caller has full knowledge of the content that's being passed in and should be able to make separate calls before or instead of passing in a concatenated line.

YoshiRulz (changes requested):
Wrong approach IMO.

Not sure if you considered this, but when implementing this feature, make sure it's *as fast or faster* than the current version. (I imagine there's enough leeway in the stale code to allow for being faster overall despite doing more work.)

Morilli:
This does not work when the data size isn't 1 byte.

Asnivor:
Sounds like this needs a lot more work.

PR #4130, opened 2024-12-03
websocket server

austinmilt:
This PR adds a base websocket server implementation, to be used for two-way JSON-based communication with clients in both event broadcasting and request/response messaging, such as to send commands to the emulator. The implementation is meant to stand alongside the existing `SocketServer` (more performant, lower-level, less dev friendly) and the websocket "server" (renamed to `WebSocketClient`)  which is actually a client implementation for connecting to and sending messages to external websocket servers.

Some notes on the implementation:
- It uses a single channel for all communications and differentiates messages by a topic ID enum in request/response message wrappers. The primary advantage with this implementation is clients get the send/receive to/from a single URL.
- The message wrappers use composition for topic-specific messages (i.e. the `registration`-specific message lives under the `registration` field of the wrapper). The primary advantage here is that clients can use a single implementation to serialize/deserialize all messages in a single pass over the JSON.
- Clients may be registered for topics by registration requests or by force (for e.g. the error and registration topics). 
- Being registered for a topic does not preclude a request message being handled, e.g. a client can send an echo request without getting it echoed back.
- Topics can be broadcast - goes to all clients registered for that topic - or client-specific - goes to a specific client based on some other message filtering or request/response. The implementation here does not use any broadcast topics. Those can be added in future topics.

# Main Changes
- renamed the prior `WebSocketServer` to `WebSocketClient` to match its usage, and make room for the new websocket server added here (`WebSocketClient`, `CommApi`)
- added a websocket server implementation as described above including client registration and generic methods for messaging clients (`WebSocketServer`)
- added messages for three topics - topic registration (`Registration.cs`), echo for debugging (`Echo.cs`), and generic errors (`Error.cs`) - which are wrapped in wrappers described above (`MessageWrapper.cs`) and described by their topic enum (`MessageWrapper`).
- added CLI args for creating the websocket server (`ArgParser`, `IMainFormForApi`, `ParsedCLIFlags`, `MainForm`) and use these to initialize the server (`MainForm`)

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/WebSocketServer.cs
+++ b/src/BizHawk.Client.Common/Api/WebSocketServer.cs
@@ -5,12 +5,22 @@ using System.Threading.Tasks;
 using System.Net.WebSockets;
 using System.Threading;
 using System.Text;
+using System.Collections.Generic;
+using BizHawk.Client.Common.Websocket.Messages;
+using Newtonsoft.Json;
+using BizHawk.Common.CollectionExtensions;
+using Newtonsoft.Json.Serialization;
 
 namespace BizHawk.Client.Common
 {
 	public sealed class WebSocketServer
 	{
-		private readonly HttpListener httpListener;
+		private static readonly HashSet<Topic> forcedRegistrationTopics = [Topic.Error, Topic.Registration];
+		private readonly HttpListener clientRegistrationListener;
+		private CancellationToken _cancellationToken = default;
+		private bool _running = false;
+		private readonly Dictionary<string, WebSocket> clients = [];
+		private readonly Dictionary<Topic, HashSet<string>> topicRegistrations = [];
 
 		/// <param name="host">
 		/// 	host address to register for listening to connections, defaults to <see cref="IPAddress.Loopback"/>>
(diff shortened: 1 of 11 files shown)


src/BizHawk.Client.Common/Api/WebSocketServer.cs
```diff
@@ -1,14 +1,215 @@
...
+						break;
+				}
+
+			}
+			catch (Exception e)
+			{
+				// this could happen if, for instance, the client sent a registration request to the echo topic, such
+				// that we tried to access the wrong field of the wrapper
+				// TODO proper logging
+				Console.WriteLine("Error handling message {0}", e);
```
austinmilt, line 150: Would love a suggestion on how to properly do logging (both debug and otherwise)

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -496,11 +497,22 @@ byte[] NetworkingTakeScreenshot()
...
+
 				_argParser.SocketAddress is var (socketIP, socketPort)
 					? new SocketServer(NetworkingTakeScreenshot, _argParser.SocketProtocol, socketIP, socketPort)
+					: null,
+
+				_argParser.WebSocketServerAddress is var (websocketServerIP, websocketServerPort)
+					? new WebSocketServer(System.Net.IPAddress.Parse(websocketServerIP), websocketServerPort)
 					: null
 			);
+			// TODO better place to put the start of the server?
```
austinmilt, line 511: Would appreciate advice here

Issue #4151, opened 2024-12-31
[C64] DeltaSerializer allocates tons of memory with disk drive active


SaxxonPike:
### Summary

In the C64 core, there is degraded performance when a floppy drive is attached.

### Repro

1. Open a C64 image (disk preferably, PRG/D64/G64)
2. Using a memory analysis tool (dotPeek in my case) see that the GC is thrashed constantly

### Host env.

- Bizhawk Interim (master 2024-12-30)

SaxxonPike:
I have a fix incoming, will link a PR to this issue shortly.

SaxxonPike:
EasyFlash cartridge (ID 0x20) is also affected, fix will also address performance issues with that mapper.

PR #4152, opened 2024-12-31
[C64] Fix memory performance issues

SaxxonPike:
Addresses https://github.com/TASEmulators/BizHawk/issues/4151

Causes:

- "Save RAM modified" flag is always True
- Deltas are calculated each frame due to the aforementioned flag
- Excess array allocations in the C64 core during delta calculation

Solutions:

- Fewer runtime array allocations (re-use buffers)
- Proper implementation of "save RAM modified" flag (only set when the disk data has changed since the last save)
- Instead of tracking which disk tracks are used at all, track which disk tracks are modified
- Store deltas only for tracks that have been modified

Practical tests run:

- Performance monitoring
- Load and save state using C64 disk media (D64, PRG)
- Load and save state on some other cores

Before changes, using Remember's release of "Uridium":
before-changes
After changes, using the same disk:
after-changes

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.Cores/Computers/Commodore64/Media/Disk.cs
+++ b/src/BizHawk.Emulation.Cores/Computers/Commodore64/Media/Disk.cs
@@ -10,7 +10,6 @@ namespace BizHawk.Emulation.Cores.Computers.Commodore64.Media
 		public const int FluxBitsPerTrack = 16000000 / 5;
 		public const int FluxEntriesPerTrack = FluxBitsPerTrack / FluxBitsPerEntry;
 		private readonly DiskTrack[] _tracks;
-		private bool[] _usedTracks;
 		public bool Valid;
 		public bool WriteProtected;
 
@@ -21,7 +20,6 @@ namespace BizHawk.Emulation.Cores.Computers.Commodore64.Media
 		{
 			WriteProtected = false;
 			_tracks = new DiskTrack[trackCount];
-			_usedTracks = new bool[trackCount];
 			FillMissingTracks();
 			Valid = true;
 		}
@@ -37,7 +35,6 @@ namespace BizHawk.Emulation.Cores.Computers.Commodore64.Media
 		{
 			WriteProtected = true;
 			_tracks = new DiskTrack[trackCapacity];
-			_usedTracks = new bool[trackCapacity];
 			for (var i = 0; i < trackData.Count; i++)
 			{
 				var track = new DiskTrack();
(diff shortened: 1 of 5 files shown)


src/BizHawk.Common/DeltaSerializer.cs
```diff
@@ -20,8 +20,19 @@ namespace BizHawk.Common
 	/// </summary>
 	public static class DeltaSerializer
 	{
+		public static int GetMaxDeltaSize<T>(int count)
+			where T : unmanaged =>
+			Marshal.SizeOf<T>() * count;
```
CasualPokePlayer, line 25: sizeof() should be used, not Marshal.SizeOf<> (which has different semantics and is much slower)

CasualPokePlayer (dismissed):

src/BizHawk.Emulation.Cores/Computers/Commodore64/Media/Disk.cs
```diff
@@ -148,13 +148,21 @@ public void AttachTracker(bool[] usedTracks)
...
 		{
-			for (var i = 0; i < _tracks.Length; i++)
+			// This is made parallel particularly for disks with a large number of tracks.
+			// Worst case, this is called once per frame, and has to run a diff over 400kb
+			// per track.
+			//
+			// In the future, in addition to "used tracks" tracking, we should also track
+			// which tracks have changed since the last call.
+			
+			Parallel.ForEach(Partitioner.Create(0, _tracks.Length), range =>
```
YoshiRulz, line 158: Weird but okay

YoshiRulz (changes requested):
What part of this is using `ArrayPool`? If you scrapped all of this and simply swapped out allocations for `ArrayPool`, how effective would that be?

SaxxonPike:
Thanks for the feedback. Having stepped away and come back to it, I think the scope of these changes is larger than it needs to be. I'm going to put in refinements this week.

YoshiRulz (approved):
Rebase again and I think this is good. And please update the OP, including your estimate for the performance boost.

SaxxonPike:
Performance results added. The results are similar across all tested disks- I just used one as an example.

YoshiRulz:
Do you need help with rebasing? I was expecting
```
p 4f34320b3
p abf2014d3
p d6223e8ef
f ff92bddb5
p d9d629e1b
```

YoshiRulz:
Is [this](https://github.com/YoshiRulz/BizHawk_BranchBackup/commits/disk-delta-fixes-yoshi/) what you intended?

SaxxonPike:
Yeah, I think some of the branch history got a bit muddled. That looks fine to me. For posterity, the commit list should look more like this: https://github.com/SaxxonPike/BizHawk/tree/disk-delta-fixes-merge

YoshiRulz:
Would you like to force-push that here then?

SaxxonPike:
@YoshiRulz Thank you for taking care of this.

PR #4164, opened 2025-01-05
Libretro game name

YoshiRulz:
@CasualPokePlayer Is there a way to pull a name from the core, or is this the best method?


CasualPokePlayer:
> Is there a way to pull a name from the core, or is this the best method?

Libretro does not have any "game name" given by the core, it is always expected that be provided by the frontend (typically with some kind of database, or lazily with just the filename).

Issue #3214, opened 2022-04-13
[Request] Add Genesis Plus GX's SMS core


Sonia-7:
Genesis Plus GX's SMS core probably has the best emulation available for the system. The emulator claims 100% compatibility on all platforms it supports, which is probably true as I've never ran into any SMS bugs while using it.

The reason I'm requesting this is that SMSHawk is problematic. Until recently several games were bugged, those got fixed but it broke some stuff elsewhere. An example is issue #3187, which gives the impression the core is some sort of house of cards similar to PCEHawk. Another problem is that audio emulation is kind of bad. Originally it was too quiet before the #2814 fix, but that ended up making the FM a little too grating. Now there are some nasty popping sounds in some games (examples are Phantasy Star's town or dungeon themes). Those sound much more clean on GPGX.

By the way, that SMS core already works through libretro. I even made a test movie with one game (Shinobi) and it synced perfectly, so save states seem stable. Since the hawk already has all components necessary to run it, maybe it'd not be that hard to integrate it properly? (SMSHawk doesn't need to be deleted, we could have two options similar to Nyma/PCEHawk.)

CasualPokePlayer:
Savestates are non-consideration as our gpgx core is waterboxed.

Note too our gpgx core is rather old (with various cherry picked commits), so it may be a bit less accurate than you expect.

CasualPokePlayer:
The genesis emulation from meda is already just a genesis gx fork, and the sms core is some fork off "SMS Plus" which is a mystery in terms of accuracy (and very low priority for meda's maintenance).

Shinobi-1987:
> The genesis emulation from meda is already just a genesis gx fork, and the sms core is some fork off "SMS Plus" which is a mystery in terms of accuracy (and very low priority for meda's maintenance, so it isn't going to be getting better).

Gotcha, i was just thinking hey this is probably easy considering all the nyma going on, mind you if it made sense and was that easy you prob would be doing it already. Was just a thought from someone who doesnt know enough lol

getCursorsExe:
> our gpgx core is waterboxed

Doesn't need to be waterboxed anymore, the latest upstream release doesn't have any savestate (including desync) issues.

Sonia-7:
[...]

Issue #3879, opened 2024-03-25
Bizhawk Bug Filers need approved set of "Upstream" emulators


Meerkov:
When filing bugs in games, the first reply is inevitably "did you test in upstream". However people filing bugs that use BizHawk may have no idea what this means. Therefore, an approved set of links to upstream emulators are necessary to expedite the reproduction process.

YoshiRulz:
As you were told on Discord, ported cores have their name shown in the status bar, and *should* have a hyperlink in `Help` > `About...`. GPGX is unique in that there is no standalone desktop app to compare to.

Links could be added to [this table](https://github.com/TASEmulators/BizHawk#cores) in the readme, but I fear that would make it too information-dense. I think writing a guide specifically for reporting bugs in games would be better in the long-term; a good chunk of our bugs are that kind, but we have about 2 lines total of documentation for it at the moment.

vadosnaprimer:
Upstream column in that label would be exactly for telling people what to test on, so I don't see a problem in one more column.

vadosnaprimer:
Like in the GPGX's case, we can backport individual fixes if we can't afford updating the whole thing. And people telling us how everything works upstream is all the more nag to pull from it properly (which is why newly ported cores are organized to make it as easy as possible).

nattthebear:
>Ported cores have their name shown in the status bar, and should have a hyperlink in Help > About....

Given that this is the case, I don't feel we need to do anything else.  Is this ticket about frustration finding upstreams in general, or just GPGX-specific?

Meerkov:
This ticket is about all cores, not gpgx specific. GPGX does not appear in the bug title or description.

PR #4169, opened 2025-01-07
Post-2.10 readme updates

YoshiRulz:
[rendered](https://github.com/TASEmulators/BizHawk/blob/readme-updates/README.md)

- [x] Are we de-emphasising SMSHawk, or is it a competitor? [#3214](https://github.com/TASEmulators/BizHawk/issues/3214) suggests it's inferior.
- [ ] Anything I missed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,7 +339,7 @@ Game Boy / Color | **Gambatte**, **GBHawk**, **SameBoy**
 Game Boy Advance | **mGBA**
 Intellivision | **IntelliHawk**
 MSX | **MSXHawk**
-N64 | Ares64, **Mupen64Plus**
+N64 | **Ares64**, Mupen64Plus
 NDS | **melonDS**
 Neo Geo Pocket | **NeoPop**
 NES | **NesHawk**, quickerNES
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
I wouldn't care about which SMS core is better. Let users know about both and then in a few years judge by what is used in movies.

CasualPokePlayer:
> optionally DirectX

DirectX components we are using are part of Windows since at least Windows 7, there should be no case where the user needs to concern themselves about that unless their Windows install is outright broken.

YoshiRulz:
Will 2.10 run on Win7? I guess that part can be omitted regardless.

CasualPokePlayer:
Win7 should be the minimum in practice, although "officially" per policy it's Win10.

YoshiRulz:
That's all the changes I wanted to make.

re: SMSHawk, the line above reads "core names in **bold** are accuracy-focused and acceptable on TASVideos," so by that metric it should stay.
But that would also mean Atari2600Hawk should stay, unless you plan to block submissions made with that? We never added a core switch prompt for A2600 like there is for NES, SNES, and TG-16.

CasualPokePlayer:
As far as things actually being acceptable on TASVideos, we've pretty much ruled that anything released is acceptable. If anything, *acceptable* could be changed to *preferred*.

README.md
```diff
@@ -371,18 +378,40 @@ A short [FAQ](https://tasvideos.org/Bizhawk/FAQ) is provided on the TASVideos wi
...
+	* [ares](https://ares-emu.net) for several consoles
+	* [Cemu](https://cemu.info/) for Wii U
+	* [Flycast](https://github.com/flyinghead/flycast) for Dreamcast
+	* [Ruffle](https://ruffle.rs) for Flash
+	* RIP Ryujinx :(
+* Ported to BizHawk:
+	* [BSNES](https://github.com/bsnes-emu/bsnes) for SNES
+	* [Genesis Plus GX](https://github.com/ekeeke/Genesis-Plus-GX) for Sega's 8- and 16-bit consoles
+	* [GSE](https://github.com/CasualPokePlayer/GSE) for GB/C
+	* [MAME](https://www.mamedev.org) for countless arcades, plug-and-plays, etc.
```
vadosnaprimer, line 405: Our mame only runs arcades.

YoshiRulz: These are links to upstream. In the cores table, MAME is listed beside "Arcade", which is what we use for both the sysID and long name. We don't really need all these though; #3879 would probably be more beneficial overall.

Issue #4176, opened 2025-01-08
[C64, 6502] Incorrect behavior for "undocumented" CPU functions


SaxxonPike:
### Summary

The test suite by Wolfgang Lorenz was used to verify the proper operation of undocumented 6502 functions in the C64 core, which uses the M6502X core. Some of the 6502 tests (for functionality common to all systems that use this core) did not pass:

- Undocumented opcodes ANE, ARR, SHX, SHY, SHA
- Decimal mode in SBC for input values that are not BCD representable (e.g. $9A)

There are some separate issues specific to the 6510's I/O port emulation but that will be considered a separate issue as it will only affect C64 (and not other emulator cores)

### Repro

1. Download `tsuit215_d64.zip` from the PC64 test suite from this page: https://www.zimmers.net/anonftp/pub/cbm/crossplatform/emulators/msdos/pc64/
2. Load `disk1.d64` or `disk2.d64` into Bizhawk
3. Run the commands `LOAD"*",8` and then `RUN`
4. Wait until the test halts with a register readout during tests such as "sbcb", "aneb" or "shaay"

### Output

An example of some of the failed tests:
![Disk1 2025-01-07 13 59 17](https://github.com/user-attachments/assets/9adab25f-6213-4505-933e-2b1caad01ade)

What it is supposed to look like (this demonstrates some of the fixes in an upcoming related PR):
![Disk2 2025-01-07 23 50 13](https://github.com/user-attachments/assets/6c0cee19-6002-4a8e-94f9-06473f412767)

### Host env.

- BizHawk dev 2025-01-07

### Considerations

The behavior of some opcodes (certainly SHA/SHX/SHY and ANE) are not consistent between systems of different models, but are more or less consistent among the same model - enough that they can be reasonably emulated with some confidence. On the C64, these opcodes are seen "in the wild" from commercial releases in the late 80s, for certain. (Wizball and Turrican are cited in the "No More Secrets" document which can be used as the authoritative source on how to properly emulate these for the C64 at least.)

We will need to know what to look for in the *other* systems that use the M6502X core - we have A7800 and NES at least, probably Lynx and possibly others. It is unlikely that this is a one-size-fits-all situation, but we do have concrete evidence of the behavior of these functions on all first-party and licensed 6502 CPUs.

I'd like to hear from other core devs before moving forward with this one.

PR #4177, opened 2025-01-08
[6502] Pass Lorenz C64 tests

SaxxonPike:
This change makes the 6502 opcode tests (not specific to 6510 I/O behavior) pass on the M6502X CPU core:

- Fix decimal mode SBC implementation (flags improperly set)
- Add ANE implementation
- Implement SHA/SHX/SHY page correction behavior and RDY pin interaction
- Allow system cores to configure the "constant" values for ANE and LXA, this is implemented like a register that can be "read". Not all manufactured 6502 derivatives will have identical values, but should mostly be consistent among the same make and model of CPU

See more in this issue: https://github.com/TASEmulators/BizHawk/issues/4176

This is a draft because we will want input from other system core developers who maintain cores that use the M6502X CPU core.

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.Cores/CPUs/MOS 6502X/MOS6502X.cs
+++ b/src/BizHawk.Emulation.Cores/CPUs/MOS 6502X/MOS6502X.cs
@@ -245,6 +245,18 @@ namespace BizHawk.Emulation.Cores.Components.M6502
 			private set => P = (byte)((P & ~0x80) | (value ? 0x80 : 0x00));
 		}
 
+		/// <summary>
+		/// For the unsupported opcode, ANE.
+		/// If your core requires a specific constant for this opcode, set it here.
+		/// </summary>
+		public Func<byte> AneConstantFunc { get; set; }
+
+		/// <summary>
+		/// For the unsupported opcode, LXA.
+		/// If your core requires a specific constant for this opcode, set it here.
+		/// </summary>
+		public Func<byte> LxaConstantFunc { get; set; }
+
 		public long TotalExecutedCycles;
 
 		// SO pin
--- a/src/BizHawk.Emulation.Cores/CPUs/MOS 6502X/MOS6502X.cs
+++ b/src/BizHawk.Emulation.Cores/CPUs/MOS 6502X/MOS6502X.cs
@@ -249,13 +249,13 @@ namespace BizHawk.Emulation.Cores.Components.M6502
 		/// For the unsupported opcode, ANE.
 		/// If your core requires a specific constant for this opcode, set it here.
 		/// </summary>
-		public Func<byte> AneConstantFunc { get; set; }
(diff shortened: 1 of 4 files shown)


YoshiRulz:
I believe Atari2600Hawk, A7800Hawk, and NesHawk would be the other relevant cores, and they're all maintainerless since alyosha left.

SaxxonPike:
@YoshiRulz I will do my best to verify that these changes still satisfy existing tests that still exist (SHX/SHY/SHA is a really rare case and not super useful, doubt much commercial software uses them at all much less full on system-specific test suites)

Some links for consideration:
- A2600:
  - https://forums.atariage.com/topic/291749-seeking-emulator-test-programs/ (good list there)
- A7800:
  - https://emulation.gametechwiki.com/index.php/Atari_7800_emulators (Rikki & Vikki note, honestly there is not much for this system)
- NES:
  - https://www.nesdev.org/wiki/Emulator_tests (lots here)
  - Blargg's tests as cited by the relevant NEShawk source itself

YoshiRulz:
https://gitlab.com/tasbot/libre-roms-ci/-/snippets

SaxxonPike:
This makes the 1541 disk drive CPU test suite pass.
pc64 1 2025-01-11 13 05 03

SaxxonPike:
instr_test-v5/all_instrs.nes results on NESHawk
all instrs 2025-01-16 13 25 34

SaxxonPike:
@YoshiRulz I would consider this no longer draft; I am confident in my testing that these changes are correct.

src/BizHawk.Emulation.Cores/CPUs/MOS 6502X/MOS6502X.cs
```diff
@@ -245,6 +245,18 @@ public bool FlagN
 			private set => P = (byte)((P & ~0x80) | (value ? 0x80 : 0x00));
 		}
 
+		/// <summary>
+		/// For the unsupported opcode, ANE.
+		/// If your core requires a specific constant for this opcode, set it here.
+		/// </summary>
+		public Func<byte> AneConstantFunc { get; set; }
```
YoshiRulz, line 252: If these are per-console constants, do they need to be delegates?

SaxxonPike: Hmm, perhaps not - some docs say that these can vary as a system is running, so it was implemented in a way that this could be accomplished. Some docs call them "constants" but it's not a great name, it's said to be a kind of entropic value. However, it is probably safe to assume for a "pristine" emulator state that these can stay constant. They don't seem to vary in any of the real hardware I have, for what it's worth.

SaxxonPike: Resolved in commit 3553760

PR #4178, opened 2025-01-08
[A26][Stella] Adds support for the console Select switch

### LINKED ISSUE #4175: [A26][Stella] Select switch is not implemented
### Summary
The "Select" input doesn't do anything while using the new Stella core. This input corresponds to the Select switch on the console itself, akin to the reset switch.

Upon reviewing the code, it appears the input was never hooked into Stella? Stella's [Switches class](https://github.com/TASEmulators/stella/blob/48734d5c38c12e3988959ebcb6efee4a570268e3/src/emucore/Switches.hxx) doesn't even provide a function for setting this switch's state oddly enough. However, according to [this section of code](https://github.com/TASEmulators/stella/blob/48734d5c38c12e3988959ebcb6efee4a570268e3/src/emucore/Switches.cxx#L100-L107) the internal `mySwitches` variable does appear to contain a bit for the select switch state.

Having this input is pretty critical. I suspect adding a `Switches::setSelect(bool)` function on Stella's side, and then hooking into it in [here](https://github.com/TASEmulators/BizHawk/blob/master/src/BizHawk.Emulation.Cores/Consoles/Atari/Stella/Stella.IEmulator.cs#L19-L29) and [here](https://github.com/TASEmulators/BizHawk/blob/master/waterbox/stella/BizhawkInterface.cxx#L160-L183) would be sufficient? Seems that's [what was done to support the reset switch](https://github.com/TASEmulators/stella/commit/2d655d8f49ecaa8a65ad8bb7c0147e70c7a062d1#diff-bf7ee3073538732f82630e44ef3de1f4ba70813dfdf44fa5469b69e9225dffdaR159-R176).

### Repro
1. Set preferred A26 core to Stella.
2. Load any Atari 2600 rom that makes use of the select switch.
   - [Video Chess](https://tasvideos.org/5645M) for example uses it to select the difficulty mode (represented by the number 1-8 at the top of the board).
   - Paul Slocum's "Testcart" test rom is also a good option.
4. Try pressing the Select switch using any input method (controller/keyboard, virtual pad, or TAStudio).

### Host env.
- BizHawk 2.10; Linux

see also #3203

bigbass1997:
In order to implement this, some arguments from `stella_frame_advance()` had to be condensed.

To keep things consistent, I combined together all of the switches that belong to the same memory register. The power switch remains separate. Like the Reset switch, the Select switch defaults to a high (1) state, and when pressed becomes low (0).

Since the waterboxed core already had a function for changing the TV color switch, I also added **_partial_** handling for that. But it will still have to be hooked up on the BizHawk side (in [Stella.IEmulator.cs](https://github.com/TASEmulators/BizHawk/compare/master...bigbass1997:stella-select-switch?expand=1#diff-a9ea1112346775177f65cf985a11dadbe31f006b0b1e5ecbe67ec0ec44b58518)) in order for BizHawk to be capable of switching it. For now, it's always set to the Color (1) state.

I've tested to confirm all of the inputs (both controllers and console switches) are working as intended.

Resolves #4175

Check if completed:


src/BizHawk.Emulation.Cores/Consoles/Atari/Stella/LibStella.cs
```diff
@@ -28,7 +28,7 @@ public abstract bool stella_init(
 			[In] InitSettings settings);
 
 		[BizImport(CallingConvention.Cdecl)]
-		public abstract void stella_frame_advance(int port1, int port2, bool reset, bool power, bool leftDiffToggled, bool rightDiffToggled);
+		public abstract void stella_frame_advance(int port1, int port2, int switchPort, bool power);
```
Morilli, line 31: This does not match the unmanaged signature and the original one doesn't match either, how does this even work? Anyway, might be cleaner to make this a `[Flags]` enum considering the core-side is also manually parsing this instead of relying on actual input port data.

Morilli: Apparently it works because these arguments are just put in registers which are 8 bytes, so you could also declare every argument as `ulong`, but it doesn't mean you should. To be fair this was wrong before you came here but it still needs to be fixed: `uint8_t, uint8_t, uint8_t, bool` is `byte, byte, byte, bool`.

bigbass1997: Oh, apologies! Yeah I was just working off what was already there; didn't realize `byte` was the type that should've been used.

PR #4180, opened 2025-01-10
Drop Octoshock

YoshiRulz:
Needed to keep some managed code around since the `.pjm` importer uses the sync settings type for serialisation.


CasualPokePlayer:
"Debugger to work in Nymashock" isn't really the point of the "standard debugging features." It's more for Lua stuff and the tracelogger, which was infinitely more useful than that crap debugger.

CasualPokePlayer:
1) The source code is not untouched, it's just (intended to be, at least) minimally modified. Just having lag stuff requires some touching of the code, arguably just as much as mem hooks would be.
2) mednafen has some debugging code in it already, it might be possible to rather just wire that up if anything (although again, if not, the debugging hooks are minimal regardless)

YoshiRulz:
> This should probably force the priority in getting the debugger to work in Nymashock to be more important then, right? Otherwise referring to "please use Octoshock in 2.10" would kinda suck years down the line (and versions) if it's not complete.

My opinion is that this shouldn't be blocked by Nymashock missing features. We just released our least buggy version yet, so if a few power users need to hold on it for a couple releases, that wouldn't be so bad. And perhaps regressing on those features would motivate working on Nymashock? I don't actually know who's maintaining it atm...

And the same goes for BSNES, though it's much easier there since DiztinGUIsh makes up for our lack of CDL.

PR #4188, opened 2025-01-15
Add a method to the `comm` Lua library to allow setting `ExpectContinue`

### LINKED ISSUE #4187: Lua Comm.httpPost fails when posting to servers that don't support Expect: 100-Continue
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"

### Summary

The C#.NET HTTP library defaults to sending an "Expect: 100-Continue" header with POST requests and waiting for a 100 Continue response before sending the body of the request. This causes POSTs to fail when sent to servers that don't support it, and can cause slowdown and latency in pretty much any other case.

The specific case I'm looking at is sending a POST to [SAMMI](https://sammi.solutions/), which is automation software used for streaming. It doesn't support Expect: 100-Continue and so never received the body of the request.

### Repro
Set up SAMMI to listen for webhook and run this LUA script in BizHawk:
```
comm.httpPost("https://localhost:9450/webhook", '{"trigger": "myWebhookTrigger", "customData": "Some cool data", "customObjectData": {"key": "Hello World", "anotherKey": "Bye"}, "customArrayData": ["Silver", "Sugar"]}')
```

### Output
Essentially nothing happens.

### Host env.
This is a general problem with C#.NET.

I have a patch ready to add another method to the comm LUA library to set the value of ExpectContinue.

reversefold:

Fixes #4187 

Add a method to allow setting the ExpectContinue functionality in the C#.NET to off or on explicitly, allowing compatibility with servers that don't support the default "Expect: 100-Continue" header that C#.NET defaults to sending for POST requests.

I have done a test build but not directly tested the functionality as of yet.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/HttpCommunication.cs
+++ b/src/BizHawk.Client.Common/Api/HttpCommunication.cs
@@ -20,6 +20,8 @@ namespace BizHawk.Client.Common
 
 		public int Timeout { get; set; }
 
+		public bool ExpectContinue { get; set; } = true;
+
 		public HttpCommunication(Func<byte[]> takeScreenshotCallback, string getURL, string postURL)
 		{
 			_takeScreenshotCallback = takeScreenshotCallback;
@@ -49,6 +51,7 @@ namespace BizHawk.Client.Common
 		public async Task<string> Post(string url, FormUrlEncodedContent content)
 		{
 			_client.DefaultRequestHeaders.ConnectionClose = true;
+			_client.DefaultRequestHeaders.ExpectContinue = ExpectContinue;
 			HttpResponseMessage response;
 			try
 			{
--- a/src/BizHawk.Client.Common/Api/HttpCommunication.cs
+++ b/src/BizHawk.Client.Common/Api/HttpCommunication.cs
@@ -20,7 +20,7 @@ namespace BizHawk.Client.Common
 
 		public int Timeout { get; set; }
 
-		public bool ExpectContinue { get; set; } = true;
+		public int ExpectContinueThreshold { get; set; } = 1024 * 1024; // 1MB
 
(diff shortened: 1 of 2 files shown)


reversefold:
Hmmmm, `TryComputeLength` is protected, we don't have access to it directly. Instead I based it on the payload length.

src/BizHawk.Client.Common/Api/HttpCommunication.cs
```diff
@@ -30,10 +32,14 @@ public HttpCommunication(Func<byte[]> takeScreenshotCallback, string getURL, str
 
 		public string ExecGet(string url = null) => Get(url ?? GetUrl).Result;
 
-		public string ExecPost(string url = null, string payload = "") => Post(
-			url ?? PostUrl,
-			new FormUrlEncodedContent(new Dictionary<string, string> { ["payload"] = payload })
-		).Result;
+		public string ExecPost(string url = null, string payload = "")
+		{
+			_client.DefaultRequestHeaders.ExpectContinue = payload.Length > ExpectContinueThreshold;
```
YoshiRulz, line 37:
```suggestion
			_client.DefaultRequestHeaders.ExpectContinue = payload.Length >= ExpectContinueThreshold;
```

YoshiRulz, line 37: I think putting this here instead of in `Post` where you had it before means that the setting will affect subsequent GET requests (or POST requests which use `Post` directly). Fixed in https://github.com/TASEmulators/BizHawk/commit/88c54d11b6d888182cac96f54679db9a68a34e56.

reversefold: Well, ExpectContinue only applyies to POST, not to GET, it is only for when you're sending a payload.

PR #4189, opened 2025-01-15
comm.httpPostRaw

reversefold:

Adds `comm.httpPostRaw` which supports setting the `Content-Type` and sends the supplied payload string verbatim instead of forcing it into a form field like httpPost.

Check if completed:

Note: Also includes https://github.com/TASEmulators/BizHawk/pull/4188

### DIFF
--- a/src/BizHawk.Client.Common/Api/HttpCommunication.cs
+++ b/src/BizHawk.Client.Common/Api/HttpCommunication.cs
@@ -65,10 +65,14 @@ namespace BizHawk.Client.Common
 		}
 #pragma warning restore DOC105
 
-		public string ExecPostRaw(string url = null, string payload = "", string contentType = "text/plain") => Post(
-			url ?? PostUrl,
-			new StringContent(payload, Encoding.UTF8, contentType)
-		).Result;
+		public string ExecPostRaw(string url = null, string payload = "", string contentType = "text/plain")
+		{
+			return Post(
+				url ?? PostUrl,
+				new StringContent(payload, Encoding.UTF8, contentType),
+				payload.Length > ExpectContinueThreshold
+			).Result;
+		}
 
 		public async Task<string> Get(string url)
 		{


YoshiRulz:
Please rebase.

reversefold:
Rebased and updated

src/BizHawk.Client.Common/Api/HttpCommunication.cs
```diff
@@ -64,6 +65,15 @@ public string ExecPostAsForm(string url = null, string payload = "")
 		}
 #pragma warning restore DOC105
 
+		public string ExecPostRaw(string url = null, string payload = "", string contentType = "text/plain")
+		{
+			return Post(
+				url ?? PostUrl,
+				new StringContent(payload, Encoding.UTF8, contentType),
```
YoshiRulz, line 72: I set up `ContentObjectFor` above for you—it should be simply
```diff
-#pragma warning disable BHI1005 // exception type
-				_ => throw new NotImplementedException()
+				/*MIME_PLAINTEXT,*/ _ => new StringContent(payload, Encoding.UTF8, mimeType)
-#pragma warning restore BHI1005
```
but I'm now realising it shouldn't have `[ConstantExpected]` and also that the method can be `public static`. I'll fix that in a minute.

YoshiRulz (commented):
I'm not sure if the name "raw" is apt, since it's taking text rather than a bytestream.

src/BizHawk.Client.Common/Api/HttpCommunication.cs
reversefold: I chose "raw" because it was all I could think of to differentiate from the existing one that submits a form. Technically that one takes text as well. This one just posts the text as the payload directly instead of forcing it into a form value.

reversefold: Ah, I hadn't seen the extra method. Not sure why it's really needed since there would still be different methods to pass in the different mimeTypes anyway....and for the raw post method I'm exposing the contentType header not to allow encoding the string differently but to allow the script author to set the header as appropriate. This library doesn't need to create a different content object if it's "text/plain" or "application/json". The incoming string is still just a string that's send as the body of the request as StringContent.

PR #4192, opened 2025-01-18
[C64] 1541 disk drive emulation facelift

SaxxonPike:
Alright friends, this one's been on my to-do list for a long time and I've made a push to fix some of the more notoriously picky disk loaders. The disk changes from here on out will likely be far less substantial (though there will be a few odd cases yet to fix- those will get their own work later.)

Disk checking utilities will now work and you can also run format commands without it erroring out now. This came down to how sparse we were putting the bits on the magnetic media. There's a neat little "feature" of the drive hardware where if you space your bits too far apart, you can get ghost bits clocked into the shift register. It's a bug that has been emulated, but it was being triggered on media that was properly formatted in the first place.

The quick and dirty changelog:

[C64] Disk: 1541 drive saveram now works properly when rewind is disabled
[C64] Disk: Use raw track capacity values per the G64 file format specification, should fix some disk loaders that are expecting data not to be so sparse (Spindle demos in D64 format particularly)
[C64] Disk: fix 1541 drive address decodes (RAM and timers are supposed to be mirrored every 0x2000 bytes in the 0x0000-0x7FFF range)
[C64] VIA: 6522 core facelift, shift register and use of low-order timer latches implemented, should fix some disk loaders ("Sprite B*****e 2" demo plays, yes they called it that, and it's not what it sounds like)
[C64] CIA: fix PB6/PB7 outputs when enabled on CRA/CRB

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.Cores/Computers/Commodore64/MOS/Cia.Registers.cs
+++ b/src/BizHawk.Emulation.Cores/Computers/Commodore64/MOS/Cia.Registers.cs
@@ -46,13 +46,13 @@
 			switch (addr)
 			{
 				case 0x0:
-					return _port.ReadPra(_pra, _ddra, _prb, _ddrb);
+					return _port.ReadPra(_praOut, _ddraOut, _prbOut, _ddrbOut);
 				case 0x1:
-					return _port.ReadPrb(_pra, _ddra, _prb, _ddrb);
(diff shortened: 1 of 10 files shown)


YoshiRulz:
This has no merge conflicts with #4177 right? And I'm good to merge both of them?

SaxxonPike:
@YoshiRulz Should be good, yes. I developed the feature with both of them in the working tree, and cherry picked this PR specifically. The two PRs should be compatible.

PR #4196, opened 2025-01-26
Adding DSDA as Doom/Hexen/Heretic core

### LINKED ISSUE #4149: [Core Port Req.] Chocolate Doom
https://github.com/chocolate-doom/chocolate-doom

The most accurate of the source ports.

Can't tell if it even has threads, but it seems to need to be detached from frontend and turned headless.

SergioMartin86:

Adding https://github.com/TASEmulators/dsda-doom as port for emulating Doom/Hexen/Heretic

- [x] Compile/Link DSDA into the waterbox
- [x] Add the new core into Bizhawk
- [x] Enable wad loading via Bizhawk
- [x] Add WAD signature detection (initial 4 bytes being IWAD or PWAD)
- [x] Add check exactly 1 IWAD is loading (not more no less) at the C# level (not wbx)
- [x] Enable frame advancing
- [x] Enable passing of the video buffer
- [x] Add TAS input parsing
- [x] Add normal key bindings for casual playthrough / manual recording
- [x] Add support for Hexen inputs
- [x] Add support for Heretic inputs
- [x] Add import of Doom demo movies
- [x] Add import of Heretic demo movies
- [x] Add import of Hexen demo movies
- [x] Add configuration (skill, compat level, nomonsters, etc)
- [x] Enable sound/music
- [x] Support Multiplayer
- [x] Support multi-wad (Iwad + pwads -- testcase: TNT map 31 fix)

Movie exporter will be developed as an external tool, so not part of this PR.

closes #4149


SergioMartin86:
> Original Doom, as well as other ports including DSDA, allow binding mouse to horizontal (turning) and vertical (forward/backward) movement, I guess by using mouse delta for a given frame. Currently exposed analog values can't be used for this because they set movement speed based on mouse position if you bind it, but we need per-frame delta (I assume it's how dsda does it).
> 
> Can you expose those 2 analog inputs too?

I don't think there is a need to expose 2 extra inputs. You can bind, as you say, the mouse axes to the forward and turning values directly in BK. The problem remains, of course, that BK provides the absolute value of mouse position as the new value, instead of the delta. However, I don't think this is supported natively in BK.

So I believe this is a question for the BizHawk main devs: is there a way to configure mouse bindings to take delta mouse position, not absolute?

src/BizHawk.Client.Common/movie/import/HexenLmpImport.cs
```diff
@@ -0,0 +1,132 @@
...
+
+			// Getting input file stream
+			var input = SourceFile.OpenRead().ReadAllBytes();
+			Stream sr = new MemoryStream(input);
+
+			// Parsing header
+			byte skillLevel = (byte)sr.ReadByte();
+			byte episode = (byte) sr.ReadByte();
+			byte map = (byte) sr.ReadByte();
+			byte player1Present = (byte) sr.ReadByte();
```
YoshiRulz, line 31: These can definitely be cleaned up.

SergioMartin86:
> It's easy to convert them to deltas before sending inputs to the core, I did it for Amiga.
> 
> https://github.com/TASEmulators/BizHawk/blob/08266f9450d9e9b685cfebb562b97b01fcd59c7e/waterbox/uae/bizhawk.c#L79-L80

I am told in discord that this is not going to work for doom because we have no way to capture and hide the mouse. So, as soon as it leaves the window you're going to stop being able to turn

vadosnaprimer:
That reply was about doing it on the hawk side, which is not needed because it can be converted just like I show. Amiga only accepts delta mouse inputs, and it works perfectly.

SergioMartin86:
> That reply was about doing it on the hawk side, which is not needed because it can be converted just like I show. Amiga only accepts delta mouse inputs, and it works perfectly.

Even if it leaves the window? 

I mean, I can implement deltas no problem. But as soon as that mouse is out of the box, you'll stop turning.

vadosnaprimer:
I think that'd still be better than nothing, and it'd always be in-bounds during fullscreen.

SergioMartin86:
Alright, added support for mouse running/turning. Unfortunately, the mouse runs out window, even if fullscreen.

src/BizHawk.Client.Common/movie/import/HexenLmpImport.cs
```diff
@@ -0,0 +1,132 @@
...
+			syncSettings.CompatibilityMode = DSDA.CompatibilityLevelEnum.C0;
+
+			var hexenController = new HexenController(0);
+			var controller = new SimpleController(hexenController.Definition);
+			controller.Definition.BuildMnemonicsCache(Result.Movie.SystemID);
+
+			bool isFinished = false;
+			while (!isFinished)
+			{
+				if (player1Present > 0) parsePlayer(controller, sr, 0);
```
YoshiRulz, line 77: Why is this indexed from 0?

src/BizHawk.Emulation.Cores/Computers/Doom/LibDSDA.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace BizHawk.Emulation.Cores.Computers.Doom
+{
+	public abstract class CInterface
+	{
+		[UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+		public delegate int load_archive_cb(string filename, IntPtr buffer, int maxsize);
+
+		[StructLayout(LayoutKind.Sequential)]
+		public class InitSettings
```
YoshiRulz, line 13: Can these be `struct`s?

src/BizHawk.Emulation.Cores/Computers/Doom/DSDA.ISettable.cs
```diff
@@ -0,0 +1,316 @@
...
+			[Display(Name = "17 - PrBoom Latest Default")]
+			C17 = 17,
+			[Display(Name = "21 - MBF21")]
+			C21 = 21
+		}
+
+		public enum SkillLevelEnum : int
+		{
+			[Display(Name = "1 - I'm too young to die")]
+			S1 = 1,
```
YoshiRulz, line 57: If `.lmp` serialises these as 0 though 4, we should too.

SergioMartin86: 1-5 is the [canonical numeration](https://doom.fandom.com/wiki/Skill_level#Doom,_Doom_II_and_Final_Doom_skill_levels). This is also how you pass the desired skill level to DSDA and Vanilla Doom as argument (-skill 1 -> easiest). It's only in the lmp that this value is normalized to zero, but this would be cause for confusion for anybody else using the interface in a normal way.

src/BizHawk.Emulation.Cores/Computers/Doom/LibDSDA.cs
SergioMartin86: Not sure I understand

YoshiRulz: I would hope you're not relying on reference identity for these, so why declare them as `class`es?

src/BizHawk.Emulation.Cores/Computers/Doom/DSDAControllers.cs
```diff
@@ -0,0 +1,248 @@
...
+			{
+				BoolButtons = BaseDefinition
+				.Select(b => $"P{PortNum} " + b)
+				.ToList()
+			}.AddAxis($"P{PortNum} Run Speed", (-50).RangeTo(50), 0)
+				.AddAxis($"P{PortNum} Strafing Speed", (-127).RangeTo(127), 0)
+				.AddAxis($"P{PortNum} Turning Speed", (-128).RangeTo(128), 0)
+				.AddAxis($"P{PortNum} Weapon Select", (0).RangeTo(7), 0)
+				.AddAxis($"P{PortNum} Mouse Running", (-128).RangeTo(128), 0)
+				.AddAxis($"P{PortNum} Mouse Turning", (-128).RangeTo(128), 0)
```
Morilli, line 41: Assuming these are stored as a byte, shouldn't all of these be `[-128, 127]`? (same for the other controllers below)

SergioMartin86: fixed

src/BizHawk.Emulation.Cores/Computers/Doom/LibDSDA.cs
SergioMartin86: I simply imitated what other cores did. If I change any of these into a struct, I get a runtime error: 

System.NotImplementedException: 'Unrecognized parameter type!'

YoshiRulz:
Should IWADs be firmware, similar to Satellaview or Vectrex *Minestorm*? I'm thinking that could save a lot of Multi-disk Bundler use.

SergioMartin86:
> Should IWADs be firmware, similar to Satellaview or Vectrex _Minestorm_? I'm thinking that could save a lot of Multi-disk Bundler use.

I considered this option. However, there are two reasons why it's not possible:
- There exist many custom IWADs outside the standard ones. These wouldn't be supported if we consider them firmware.
- In the majority of cases, the expectation is that only the IWAD (and no other WAD) is used. This would require creating a movie with no "rom", only firmware. Is this even supported in Bizhawk?

YoshiRulz:
It would need to support loading an IWAD through the standard path as well, yeah.

src/BizHawk.Client.Common/movie/import/HexenLmpImport.cs
SergioMartin86: I don't have a good answer for this. For some reason, I need to pass playerId 0 for P1 inputs. If I use [1-4] indexing, it'll be shifted by one. I haven't done anything in DSDA to warrant this, it just works like that, so I assume it's internal to BK?

YoshiRulz: https://github.com/TASEmulators/BizHawk/blob/699c61a131798cfd3275272bf2ab5fc6890490ee/src/BizHawk.Emulation.Cores/Computers/Doom/DSDA.ControllerDeck.cs#L14
https://github.com/TASEmulators/BizHawk/blob/699c61a131798cfd3275272bf2ab5fc6890490ee/src/BizHawk.Client.Common/movie/import/HexenLmpImport.cs#L70
?

YoshiRulz:
Please double-check the fields are still being parsed in the right order after ^.

SergioMartin86:
> Please double-check the fields are still being parsed in the right order after ^.

checked. they work.

src/BizHawk.Emulation.Cores/CoreNames.cs
```diff
@@ -19,6 +19,7 @@ public static class CoreNames
 		public const string ColecoHawk = "ColecoHawk";
 		public const string CPCHawk = "CPCHawk";
 		public const string Cygne = "Cygne/Mednafen";
+		public const string DSDA = "DSDA-Doom";
```
YoshiRulz, line 22: Reminder to (rebase and) add settings to `MainForm.VSystem.cs`.

vadosnaprimer:
@nattthebear if a Doom port were added using the new wbx setup, it'd break on https://github.com/TASEmulators/BizHawk/blob/b66672f97f148452044f2a014bd5f2ea9b4c5a2d/src/BizHawk.Emulation.Cores/Waterbox/WaterboxCore.cs#L77 due to no memory "domains" existing inside the game engine, so nothing to really expose for such cases. Would it make sense to let this to be empty? Otherwise some hacky fake mem domain would have to be used and it may be confusing to users.

nattthebear:
@vadosnaprimer I don't think this problem is unique to waterbox, is it?  A lot of general purpose tooling will fail without useful memory domains.

I have no objection if we change that to `SingleOrDefault` so long as the fallout isn't too bad.

vadosnaprimer:
Decided to do a wiser thing and expose useful objects as an artificial memory domain. Exposing them as dedicated lua library would arguably be more work and it'd be less usable, since other tool wouldn't have access to it.

YoshiRulz (approved):

vadosnaprimer:
@SergioMartin86 something's wrong with weapon selection via buttons (fine in demos). 1 is indeed fist, but for pistol you need to press 3 instead of 2, for shotgun you need to press 5 instead of 3, and for chaingun you need to press 7. Plasma and BFG can't be selected.

src/BizHawk.Emulation.Cores/Computers/Doom/DSDA.ControllerDeck.cs
```diff
@@ -0,0 +1,73 @@
...
+		public byte ReadPort2(IController c)
+			=> Port2.Read(c);
+
+		public byte ReadPort3(IController c)
+			=> Port3.Read(c);
+
+		public byte ReadPort4(IController c)
+			=> Port4.Read(c);
+
+		public int ReadPot1(IController c, int pot)
```
vadosnaprimer, line 44: Port?

PR #4216, opened 2025-02-09
Tastudio Lua Branch Undo Callbacks

c7fab:
Implements proper callbacks for branch load/update/remove undos.
Previously it didn't work, since it wasn't possible to receive in Lua whether the callback came from a regular action or an undo.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -618,5 +618,18 @@ namespace BizHawk.Client.EmuHawk
 				};
 			}
 		}
+
+		[LuaMethodExample("function RemoveUndone(index)\r\nconsole.log(\"You did a branch remove undo for branch \"..index)\r\nend\r\ntastudio.onbranchundoremove(RemoveUndone)")]
+		[LuaMethod("onbranchundoremove", "called whenever a branch removal is undone. luaf must be a function that takes the integer branch index as a parameter")]
+		public void OnBranchUndoREmove(LuaFunction luaf)
+		{
+			if (Engaged())
+			{
+				Tastudio.BranchRemoveUndoneCallback = index =>
+				{
+					luaf.Call(index);
+				};
+			}
+		}
 	}
 }
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -592,5 +592,18 @@ namespace BizHawk.Client.EmuHawk
 				};
 			}
 		}
+
+		[LuaMethodExample("function LoadUndone()\r\nconsole.log(\"You did a branch load undo\")\r\nend\r\ntastudio.onbranchundoload(LoadUndone)")]
+		[LuaMethod("onbranchundoload", "called whenever a branch load is undone. luaf must be a function without parameters")]
+		public void OnBranchUndoLoad(LuaFunction luaf)
+		{
+			if (Engaged())
+			{
+				Tastudio.BranchLoadUndoneCallback = () =>
+				{
+					luaf.Call(); 
+				};
+			}
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


c7fab:
Missing callback for branch reordering

c7fab:
Should the branch indices be +1?
For now every index is off by one compared to the indices in the branches list. Would make it more userfriendly.

src/BizHawk.Client.EmuHawk/tools/TAStudio/BookmarksBranchesBox.cs
```diff
@@ -358,7 +366,7 @@ private void UndoBranchToolStripMenuItem_Click(object sender, EventArgs e)
 			if (_branchUndo == BranchUndo.Load)
 			{
 				LoadBranch(_backupBranch);
-				LoadedCallback?.Invoke(Branches.IndexOf(_backupBranch));
+				LoadUndoneCallback?.Invoke();
```
YoshiRulz, line 369: Quick glance at this PR and it seems you're adding a bunch more callbacks when ones already exist? Would it not be suitable to subscribe to those?

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -592,5 +604,57 @@ public void OnBranchRemove(LuaFunction luaf)
 				};
 			}
 		}
+
+		[LuaMethodExample("function LoadUndone()\r\n\tconsole.log(\"You did a branch load undo\")\r\nend\r\ntastudio.onbranchundoload(LoadUndone)")]
+		[LuaMethod("onbranchundoload", "called whenever a branch load is undone. luaf must be a function without parameters")]
```
YoshiRulz, line 609: Don't use existing docs as inspiration because they suck

src/BizHawk.Client.EmuHawk/tools/TAStudio/BookmarksBranchesBox.cs
c7fab: Currently it only properly works with undoing branch loads, because there it's possible to get that a branch action undo has been occured, since the the index of the backup branch is -1 when doing an undo load.
For Saving and Removing it's not possible to figure out in lua whether it was a normal action or an undo. The index for undoing saving is just the same as doing a regular branch save, so no chance to figure out which occured. And remove undo is just like adding a new branch.
The user might want to do different things in lua depending on what has happened with the branches.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
c7fab: What needs rewrite? LuaMethodExample or LuaMethod?

src/BizHawk.Client.EmuHawk/tools/TAStudio/BookmarksBranchesBox.cs
YoshiRulz: I think I understand—you're changing this callsite to differentiate it from the callsite in `PrepareHistoryAndLoadSelectedBranch`? I think adding an argument would be a better approach. @feos Thoughts?

c7fab: That's something I thought about doing too, but only have one onbranchaction method and make luaf a function that takes an integer for the index and a string for every action "save", "load", "remove", "undoload" "undoupdate" and "undoremove" and make the current deprecated. Reordering then still has it's own method.

YoshiRulz: Because then the script author is forced to care. If you add a parameter to the existing callback, existing scripts will continue to work, and authors can opt-in to caring.

c7fab: Doing with string would mean the user can then just make a lua table with functions and make each table index a string. Then there wouldn't be many if-statements in the callback function.
Either way, I don't mind how it's done as long as it finally works

c7fab:
I had some time to have a look at this again.
First thing, I solved the issue for getting the selected (or in my case, a) branch index by writing a costum dropdown gui in EmuHawk client, so I don't think `GetSelectedBranch()` is not much of an use.

OnBranchReorder can probably avoided by using the branch Id, either passing it as a parameter or from `tastudio.getbranches()[index].Id`

`tastudio.getbranches()` doesn't contain the -1 index for backup branches, so the user can't get Id, Text and Frame for it. 

`OnBranchRemoved()` is called after the branch has been removed, so any fields get by `tastudio.getbranches()[index]` returns the ones of the next branch or nil.

Issue #4138, opened 2024-12-07
PUAE core - Lack of Amiga drive info / Disk Switching Messages


ScHlAuChii:
### Summary
It is currently completely unclear what disk is in what drive.
You dont know if a disk was actually inserted or if the disk you inserted is broken.

This is how it currently is:
![CurrentState](https://github.com/user-attachments/assets/3b0a6584-7304-41f5-9cdc-0aa6e750054a)

This is how I suggest it should be:
![FutureState](https://github.com/user-attachments/assets/f8f964dc-0bf3-40cb-b2f9-9aeece471469)

Additionally when you insert or eject a disk, Bizhawk should say something like:
Turrican2AGADisk1 inserted to FD0
Turrican2AGADisk1 ejected from FD0

vadosnaprimer:
Do you really need to know which drive each slot is in? You can iterate through drives, and I made it print what file is inserted, and the max drive number is 4, so that's kinda quicker?

ScHlAuChii:
Yep knowing which drive each slot is in is very useful information.
Currently Amigahawk only supports unclean ADF dumps which often were badly cracked.
So even if you inserted a slot it doesnt guarantee that it works. 
Also some Amiga games are picky in how they support those drive.
So any extra info to the user is useful.

ScHlAuChii:
Is that in the current build?

vadosnaprimer:
Yes.

ScHlAuChii:
Yep its perfectly fine and works as intended - but:
If you set 4 drives and load in an XML with 3 slots, it crashes once you switch to the 4th drive.

-------------------------------------------------------------
System.ArgumentOutOfRangeException: Der Index lag außerhalb des Bereichs. Er darf nicht negativ und kleiner als die Sammlung sein.
Parametername: index
   bei System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   bei BizHawk.Emulation.Cores.Computers.Amiga.PUAE.FrameAdvancePrep(IController controller, Boolean render, Boolean rendersound) in /src/BizHawk.Emulation.Cores/Computers/Amiga/PUAE.cs:Zeile 139.
   bei BizHawk.Emulation.Cores.Waterbox.WaterboxCore.FrameAdvance(IController controller, Boolean render, Boolean rendersound) in /src/BizHawk.Emulation.Cores/Waterbox/WaterboxCore.cs:Zeile 207.
[...]

Issue #3206, opened 2022-04-04
 [Feature Request] Nyma Cores Selecting the Disk and Nubmer the Disk


Spikestuff:
Currently the Nyma cores only have "previous disk" and "next disk" as its options but it doesn't state which disk you're currently using.

It also doesn't allow the freedom for instance when you need to go from Disk 1 to Disk 4 where instead you're forced to skim over Disk 2 and 3.

CasualPokePlayer:
Fixed in 58c28aefbaaa74e1c3a2a5b9bae4b1d5724a365d

PR #4217, opened 2025-02-11
Adding DOSBox as core for DOS TASing

SergioMartin86:

- [x] Add floppy disk mounting
- [x] Add CD-ROM mounting
- [x] Add proper CD-ROM callbacks for reading
- [x] Add disk swapping
- [x] Add mem-based R/W Hard Disk
- [x] Add multiple pre-formatted disk images
- [x] Save Hard Disk to SaveRAM
- [x] Load Hard Disk from SaveRAM
- [x] Remove SDL dependency
- [x] Add video buffer getting
- [x] Add sound
- [x] Add keyboard
- [x] Add mouse
- [x] Use proper BK mouse deltas, instead of manually calculated ones
- [x] Add joystick
- [x] Create machine presets
- [x] Add setting overriding
- [x] Add file type association
- [x] Add input mnemonics
- [x] Add drive access light callbacks
- [x] Add mem domains
- [x] Add basic sound management settings
- [x] Add basic video management settings
- [x] Add basic cpu management settings
- [x] Add check for saveRAM (if hd size is different, do not load it)
- [x] Add default inputs for mouse (‎Assets/defctrl.json)
- [x] Add variable framerates

Lag frame detection cannot be implemented since these machines are IRQ, not polling based

Pending Fixes:

- [x] CDRom images should be able to be passed as plain files not as DiscAsset. -> Implemented a pseudo-fix: forcing cdrom images passed to DOSBox to be renamed to extension e.g., '.dosbox-iso', instead of .iso. Is there a better fix?
- [x] Timing seems to be a bit off (too fast)
- [x] Audio seems to be clicking (is this related to timing)
- [x] Figure out why windows 3.1 doesn't detect the mouse: requires a [special driver](https://dosbox-x.com/wiki/Guide%3AInstalling-Windows-3.1x#_mouse_support). Otherwise it works correctly.
- [x] Implement support for .cue+.bin multimedia CDROM images

Desirable features for future work (not in this PR):

- Implement support for .hdi files (PC98). Test game: Tohou
- Add firmware support for YM2608, FM Towns, PC98, etc

Test Tools:

Used this to test joystick and mouse support:
https://www.vogons.org/viewtopic.php?p=187168#p187168

Check if completed:


src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,336 @@
...
+
+				var writableHDDImageData = Zstd.DecompressZstdStream(new MemoryStream(writableHDDImageFile)).ToArray();
+				_exe.AddReadonlyFile(writableHDDImageData, FileNames.WHD);
+				configString += "imgmount c " + FileNames.WHD + ".img\n";
+			}
+
+			//// CPU (core) configuration
+			configString += "[cpu]\n";
+			if (_syncSettings.CPUCycles != -1) configString += $"cycles = {_syncSettings.CPUCycles}";
+			if (_syncSettings.CPUCycles != -1) configString += $"cycles = {_syncSettings.CPUCycles}";
```
YoshiRulz, line 190: Copy-paste error?

SergioMartin86: yes. removed

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -66,7 +67,7 @@ public DOSBox(CoreLoadParameters<object, SyncSettings> lp)
 			// Parsing rom files
 			foreach (var file in _roms)
 			{
-				var ext = file.RomPath.Substring(startIndex: file.RomPath.LastIndexOf('.')).ToLowerInvariant();
+				var ext = file.RomPath.Substring(startIndex: file.RomPath.LastIndexOf('.')).ToLowerInvariant().RemovePrefix('.');
```
YoshiRulz, line 70:
```suggestion
				var ext = file.RomPath.Substring(startIndex: file.RomPath.LastIndexOf('.') + 1).ToLowerInvariant();
```
My bad there

SergioMartin86: Actually what happens if the file has no extension?

YoshiRulz: I guess this would treat the whole string as the extension, but that's a weird edge case to handle here. I don't think the file would ever reach the core ctor.

Morilli: Can't you guys just use `Path.GetExtension` please?

SergioMartin86: fixed

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1937,7 +1938,7 @@ private void LoadSaveRam()
 					byte[] sram;
 
 					// some cores might not know how big the saveram ought to be, so just send it the whole file
-					if (Emulator is AppleII or C64 or MGBAHawk or NeoGeoPort or NES { BoardName: "FDS" })
+					if (Emulator is AppleII or C64 or MGBAHawk or NeoGeoPort or NES { BoardName: "FDS" } or DOSBox)
```
YoshiRulz, line 1941: Opened #4242.

src/BizHawk.Emulation.Common/Base Implementations/Bk2MnemonicLookup.cs
```diff
@@ -802,6 +802,122 @@ public static string LookupAxis(string button, string systemId)
...
+				["Joystick Button 2"] = '2',
+				["Joystick Up"] = 'U',
+				["Joystick Down"] = 'D',
+				["Joystick Left"] = 'L',
+				["Joystick Right"] = 'R',
+				["Mouse Left Button"] = 'l',
+				["Mouse Middle Button"] = 'm',
+				["Mouse Right Button"] = 'r',
+				["Next Floppy Disk"] = '>',
+				["Next CDROM"] = '>',
```
vadosnaprimer, line 817:
```suggestion
				["Next CDROM"] = '}',
```
It's better to use slightly different "arrows" for different operations, and cdrom is more advanced so it gets a curly brace :)

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,359 @@
...
+		private bool _nextCDROMPressed = false;
+		private List<IRomAsset> _floppyDiskImageFiles = new List<IRomAsset>();
+		private List<IRomAsset> _CDROMDiskImageFiles = new List<IRomAsset>();
+		private int _floppyDiskCount = 0;
+		private int _CDROMCount = 0;
+		private int _currentFloppyDisk = 0;
+		private int _currentCDROM = 0;
+		private string GetFullName(IRomAsset rom) => rom.Game.Name + rom.Extension;
+
+		public override int VirtualWidth => LibDOSBox.SVGA_MAX_WIDTH;
```
vadosnaprimer, line 46: `VirtualWidth` represents how the frame buffer should be stretched by forntend to look authentic to how the user saw the final thing back in the day. For IBM PC that means it's just a 4:3 monitor. Now since height can't be stretched (because it represents scanlines), we should just tell frontend to resize width to be  4/3 of height.

vadosnaprimer:
Currently loading .dosbox-iso (the only thing I tried) opens this dialog

0cbcd2b824f19688

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
SergioMartin86: Thanks, I made the corresponding change and it looks correct now

vadosnaprimer:
Mouse movement is now recognized! But is there a way to send deltas directly to the core without working with absolute values? Ideally relative mouse input would be exposed to user in addition to absolute position, so they could pick which of them to use for their particular game. In Doom core mouse movement ended up being relative-only, I think it would work for this core too as an option.

SergioMartin86:
> Mouse movement is now recognized! But is there a way to send deltas directly to the core without working with absolute values? Ideally relative mouse input would be exposed to user in addition to absolute position, so they could pick which of them to use for their particular game. In Doom core mouse movement ended up being relative-only, I think it would work for this core too as an option. This is valuable because absolute movement is very limited, like we saw with Doom, but relative movement allows unlimited turning, which is required for comfortable play.

Yes, we can pass deltas, but we also need the absolutes for games that use it (Simcity) since it preserves the exact same position of the pointer as in the host window.

vadosnaprimer:
That's what I mean by it being optional: just expose both mouse controls for users to decide which to use.

SergioMartin86:
> That's what I mean by it being optional: just expose both mouse controls for users to decide which to use.

deltas being passed now.
Also mouse pressed / released events (what DOSBox cares about), instead of whether the mouse button is pressed or not at BK's TAStudio piano roll.

SergioMartin86:
> * That bin/cue, iso, chd and mdf are functional and can be automatically mounted as a "cd-rom".

.cue+bin working now

SergioMartin86:
> [DosBox-X] Supaplex crashing after exit in controls menu Video:
> On DosBox-X working
Hi, can you please specify the contents of your .conf file in your working dosbox-x? You may use pastebin.com or similar.

Danik2343:
> > [DosBox-X] Supaplex crashing after exit in controls menu Video:
> > On DosBox-X working
> 
> Hi, can you please specify the contents of your .conf file in your working dosbox-x? You may use pastebin.com or similar.

[dosbox-x.zip](https://github.com/user-attachments/files/19159916/dosbox-x.zip)

Danik2343:
NVIDIA_Overlay_Xu94qXRm91
In DosBox-X a: drive working

SergioMartin86:
> It works on C: drive, when installing from CD game, but not working in discette a: (BizHawk)

ah, one important thing: diskettes are read-only). The setup is probably trying to save your changes to a file and failing to do so.  Normally the game should tell you it failed, but seems this one has a catastrophic response.

I'll consider this as solved

src/BizHawk.Client.Common/RomLoader.cs
```diff
@@ -990,6 +990,8 @@ private static class RomFileExtensions
 
 			public static readonly IReadOnlyCollection<string> Doom = new[] { "wad" };
 
+			public static readonly IReadOnlyCollection<string> DOS = new[] { "ima", "img", "conf", "dosbox-iso", "dosbox-cue", "dosbox-bin", "dosbox-mdf", "dosbox-chf", "xdf", "dmf", "fdd", "fdi", "nfd", "d88" };
```
CasualPokePlayer, line 993: Similar to https://github.com/TASEmulators/BizHawk/pull/4264#discussion_r1992915231, these "dosbox-*" CD extensions are unacceptable. The core must use BizHawk's CD interface, it is not allowed to handle CDs themselves. This is a blocker for merge.

CasualPokePlayer (dismissed):

vadosnaprimer:
I compared 1993 and 1994 presets with upstream using doom. Regardless of framerate set in the settings, there's tearing visible on every other frame (because the game runs at half DOS framerate) that is not there in upstream (I dumped video from dosbox-x).


https://github.com/user-attachments/assets/56917a35-db54-4adb-b8a8-01d388ec6c7f

https://github.com/user-attachments/assets/a404f591-4c6f-40ab-99ec-20261d78885a

src/BizHawk.Client.Common/RomLoader.cs
SergioMartin86: Fixed

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,492 @@
...
+            _discIndex--;
+            if (_discIndex < 0) _discIndex = _discAssets.Count - 1;
+            CoreComm.Notify($"Selected CDROM {_discIndex}: {_discAssets[_discIndex].DiscName}", _messageDuration);
+        }
+
+        private void CDRead(string cdRomName, int lba, IntPtr dest, int sectorSize)
+        {
+			// Console.WriteLine($"Reading from {cdRomName} : {lba} : {sectorSize}");
+
+			if (! _cdRomFileToReaderMap.ContainsKey( cdRomName ) ) throw new InvalidOperationException($"Unrecognized CD File with name: {cdRomName}");
```
YoshiRulz, line 325: `TryGetValue`, also what happened to the spacing on this line?

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,492 @@
...
+			fi.joystick1.button1 = _syncSettings.EnableJoystick1 && controller.IsPressed($"P1 {Inputs.Joystick} {JoystickButtons.Button1}") ? 1 : 0;
+			fi.joystick1.button2 = _syncSettings.EnableJoystick1 && controller.IsPressed($"P1 {Inputs.Joystick} {JoystickButtons.Button2}") ? 1 : 0;
+			fi.joystick2.up      = _syncSettings.EnableJoystick2 && controller.IsPressed($"P2 {Inputs.Joystick} {JoystickButtons.Up}") ? 1 : 0;
+			fi.joystick2.down    = _syncSettings.EnableJoystick2 && controller.IsPressed($"P2 {Inputs.Joystick} {JoystickButtons.Down}") ? 1 : 0;
+			fi.joystick2.left    = _syncSettings.EnableJoystick2 && controller.IsPressed($"P2 {Inputs.Joystick} {JoystickButtons.Left}") ? 1 : 0;
+			fi.joystick2.right   = _syncSettings.EnableJoystick2 && controller.IsPressed($"P2 {Inputs.Joystick} {JoystickButtons.Right}") ? 1 : 0;
+			fi.joystick2.button1 = _syncSettings.EnableJoystick2 && controller.IsPressed($"P2 {Inputs.Joystick} {JoystickButtons.Button1}") ? 1 : 0;
+			fi.joystick2.button2 = _syncSettings.EnableJoystick2 && controller.IsPressed($"P2 {Inputs.Joystick} {JoystickButtons.Button2}") ? 1 : 0;
```
YoshiRulz, line 367: This might still function if you remove the `_syncSettings.EnableJoystick1` checks, but to keep it fast, wrap the 6 lines in an `if`.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,492 @@
...
+
+			// Updating mouse state
+			_mouseState.posX = fi.mouse.posX;
+			_mouseState.posY = fi.mouse.posY;
+			if (fi.mouse.leftButtonPressed    > 0) _mouseState.leftButtonHeld = true;
+			if (fi.mouse.middleButtonPressed  > 0) _mouseState.middleButtonHeld = true;
+			if (fi.mouse.rightButtonPressed   > 0) _mouseState.rightButtonHeld = true;
+			if (fi.mouse.leftButtonReleased   > 0) _mouseState.leftButtonHeld = false;
+			if (fi.mouse.middleButtonReleased > 0) _mouseState.middleButtonHeld = false;
+			if (fi.mouse.rightButtonReleased  > 0) _mouseState.rightButtonHeld = false;
```
YoshiRulz, line 402: Idea for cleaning this up: create and populate a local `nextState` above and then assign it to `_mouseState` here

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,492 @@
...
+
+			// Processing floppy disks swaps
+			fi.driveActions.insertFloppyDisk = -1;
+			if (_floppyDiskCount > 1)
+			{
+				if (controller.IsPressed(Inputs.NextFloppyDisk))
+				{
+					if (!_nextFloppyDiskPressed)
```
YoshiRulz, line 410: Combine conditions: `if (_floppyDiskCount > 1 && !_nextFloppyDiskPressed && controller.IsPressed(Inputs.NextFloppyDisk))`

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,492 @@
...
+					{
+						_currentFloppyDisk = (_currentFloppyDisk + 1) % _floppyDiskCount;
```
YoshiRulz, line 412: This wraps, but `SelectNextDisc` clips?

src/BizHawk.Client.Common/RomLoader.cs
YoshiRulz: They're still mentioned in `FileID.cs`

src/BizHawk.Emulation.Cores/Waterbox/WaterboxCore.cs
```diff
@@ -122,7 +122,7 @@ private void AdvanceRtc()
 		private WaterboxMemoryDomain[] _saveramAreas;
 		private int _saveramSize;
 
-		public unsafe bool SaveRamModified
+		public virtual unsafe bool SaveRamModified
```
YoshiRulz, line 125: I did this to accommodate your overrides (`DOSBox.ISaveRam.cs`), but I'm not sure whether you should be doing that. Obviously all the other Waterbox cores make do with the base implementations.

YoshiRulz (dismissed):

SergioMartin86:
> I compared 1993 and 1994 presets with default config in upstream using doom. Regardless of framerate set in the settings, there's tearing visible on every other frame (because the game runs at half DOS framerate) that is not there in upstream (I dumped video from dosbox-x). Framerate is exactly as in dosbox-x capture, but tearing was the same even with a slightly different value in core settings.

fixed, please retest

src/BizHawk.Emulation.Cores/Waterbox/WaterboxCore.cs
Morilli: Exposing the save ram as memory domain instead would most likely be the correct way to handle saveram.

src/BizHawk.Emulation.Cores/Computers/DOS/LibDOSBox.cs
```diff
@@ -0,0 +1,238 @@
...
+		[BizImport(CC)]
+		public abstract void pushCDData(int cdIdx, int numSectors, int numTracks);
+
+		[BizImport(CC)]
+		public abstract void pushTrackData(int cdIdx, int trackId, CDTrack data);
+
+		// CD Management Logic END
+
+		[BizImport(CC, Compatibility = true)]
+		public abstract bool Init(bool joystick1Enabled, bool joystick2Enabled, bool mouseEnabled, ulong hardDiskDriveSize, ulong fpsNominator, ulong fpsDenominator);
```
YoshiRulz, line 64: I assume `fpsNominator` is a typo for `fpsNumerator`

YoshiRulz: It's on the unmanaged side too (for some reason I thought you'd have to change both at the same time—you don't, but you probably should).

src/BizHawk.Client.Common/movie/PlatformFrameRates.cs
```diff
@@ -35,6 +35,7 @@ public static class PlatformFrameRates
 			["ChannelF_PAL"] = 15625.0 / 312.0, // 4000000 / (256 * 312)
 			["Coleco"] = 59.9227510135505,
 			["Doom"] = 35.0,
+			["DOS"] = 70.086302895322945,
```
YoshiRulz, line 38: Please include your source for this value as a comment.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
YoshiRulz: I just noticed the assignments after the `if`s, those controller reads should be deduped.

src/BizHawk.Emulation.Cores/Waterbox/WaterboxCore.cs
YoshiRulz: Don't need to change this class anymore.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.Controllers.cs
```diff
@@ -0,0 +1,120 @@
...
+
+			// Adding mouse inputs
+			if (settings.EnableMouse)
+			{
+				controller.BoolButtons.Add(Inputs.Mouse + " " + MouseInputs.LeftButton);
+				controller.BoolButtons.Add(Inputs.Mouse + " " + MouseInputs.MiddleButton);
+				controller.BoolButtons.Add(Inputs.Mouse + " " + MouseInputs.RightButton);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.PosX, (0).RangeTo(LibDOSBox.SVGA_MAX_WIDTH), LibDOSBox.SVGA_MAX_WIDTH / 2);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.PosY, (0).RangeTo(LibDOSBox.SVGA_MAX_HEIGHT), LibDOSBox.SVGA_MAX_HEIGHT / 2);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.SpeedX, (-LibDOSBox.SVGA_MAX_WIDTH / 2).RangeTo(LibDOSBox.SVGA_MAX_WIDTH / 2), 0);
```
vadosnaprimer, line 65: Speed (delta) is a signed short per axis.

src/BizHawk.Emulation.DiscSystem/DiscIdentifier.cs
```diff
@@ -185,6 +190,16 @@ public DiscType DetectDiscType()
...
 
+			// Detects ISO9660 / Joliet CD formats (target for DOS / Windows)
+			// Source: https://en.wikipedia.org/wiki/ISO_9660
+			if (DetectISO9660())
+				return DiscType.DOS;
+
+			// Detects UDF
+			// Source: https://www.cnwrecovery.com/manual/HowToRecogniseTypeOfCDDVD.html
+			if (DetectUDF())
+				return DiscType.DOS;
```
YoshiRulz, line 201: Leave the source on the method implementing the algorithm i.e.
```csharp
/// <remarks><see href="https://en.wikipedia.org/wiki/ISO_9660"/></remarks>
private bool DetectISO9660()
{
```

src/BizHawk.Client.Common/movie/PlatformFrameRates.cs
```diff
@@ -35,6 +35,7 @@ public static class PlatformFrameRates
 			["ChannelF_PAL"] = 15625.0 / 312.0, // 4000000 / (256 * 312)
 			["Coleco"] = 59.9227510135505,
 			["Doom"] = 35.0,
+			["DOS"] = 70.086303, //  Video dump from DOSBox-X has 70086303/1000000 (70.086303) framerate when launching into DOS.
```
YoshiRulz, line 38: This is suspicious to me—DOSBox-X just happens to use a *decimal* approximation of the JPC-rr ratio `226575000/3232800`? Though when I looked in the JPC-rr source for that ratio, I only found [the numerator](https://gitlab.com/TASVideos/JPC-rr/-/blob/bf3a341ad233439e21a32f33dffc1e730668121d/org/jpc/emulator/pci/peripheral/VGACard.java#L95), while the denominator appeared to be set to [a different value](https://gitlab.com/TASVideos/JPC-rr/-/blob/bf3a341ad233439e21a32f33dffc1e730668121d/org/jpc/emulator/pci/peripheral/VGACard.java#L748) which gives exactly 60 Hz.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.Controllers.cs
```diff
@@ -0,0 +1,116 @@
...
+			// Adding mouse inputs
+			if (settings.EnableMouse)
+			{
+				controller.BoolButtons.Add(Inputs.Mouse + " " + MouseInputs.LeftButton);
+				controller.BoolButtons.Add(Inputs.Mouse + " " + MouseInputs.MiddleButton);
+				controller.BoolButtons.Add(Inputs.Mouse + " " + MouseInputs.RightButton);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.XAxis, (0).RangeTo(LibDOSBox.SVGA_MAX_WIDTH), LibDOSBox.SVGA_MAX_WIDTH / 2);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.YAxis, (0).RangeTo(LibDOSBox.SVGA_MAX_HEIGHT), LibDOSBox.SVGA_MAX_HEIGHT / 2);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.XDelta, (-32769).RangeTo(32768), 0);
+				controller.AddAxis(Inputs.Mouse + " " + MouseInputs.YDelta, (-32769).RangeTo(32768), 0);
```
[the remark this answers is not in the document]
vadosnaprimer: The answer is yes

https://github.com/joncampbell123/dosbox-x/blob/36888dc377cac09e13516705c985d04483794e17/src/ints/mouse.cpp#L857-L860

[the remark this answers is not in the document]
vadosnaprimer: Looking at it more, I actually can't find any kind of range limit. Whatever comes from SDL is sent to the core, and I can't even find where `mouse.ps2x` values are used... also they look absolute at the second glance.

The fact that with this range minimal delta that hawk gets is 272, something is probably not perfectly right in my commit, even tho it made doom turning work (and doom clamps turns to 50).

vadosnaprimer:
Okay so in the current state, movement in doom seems to work great actually! The only problem is that mouse button release doesn't always register, and I'm sure it worked great before.

vadosnaprimer:
Also given frontend automatically allows configuring sensitivity for every analog input, maybe it's not needed as a core setting?

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.Controllers.cs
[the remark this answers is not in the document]
YoshiRulz: The code you linked limits (by wrapping, in a questionable way) the range to s16, `-32768..=32767`. Which I think is what you said in some other comment.

SergioMartin86:
> Okay so in the current state, movement in doom seems to work great actually! The only problem is that mouse button release doesn't always register, and I'm sure it worked great before.
> 
> EDIT
> 
> Seems to be caused by [355dce2](https://github.com/TASEmulators/BizHawk/commit/355dce2004b1f3727e278bd33efec426450ac6d1)

Discussed in Discord, but worth reporting for others to see. This mouse issue was reproduced and fixed in [55e1f4b](https://github.com/TASEmulators/BizHawk/pull/4217/commits/55e1f4bfeb980d34cfb2029a5a3d3b75ad10e34f)

SergioMartin86:
> Also given frontend automatically allows configuring sensitivity for every analog input, maybe it's not needed as a core setting?

I hesitate on removing this syncsetting. The sensitivity in BK can be adjusted up to 200%, but there might be DOS games where going beyond may be necessary.  The default is set at 1 anyway, so no harm no foul?

vadosnaprimer:
> The sensitivity in BK can be adjusted up to 200%, but there might be DOS games where going beyond may be necessary. The default is set at 1 anyway, so no harm no foul?

Sounds fair, on the other hand maybe we need more variety in the frontend range?

vadosnaprimer:
I want to merge this PR, but my final request is core sending a callback to hawk when **rendering FPS** value changes (and passing num and denom):
https://github.com/TASEmulators/dosbox-x/blob/324193b96393571b9a335f39a92aa619a37550bd/src/hardware/vga_draw.cpp#L7019
That way we'll be able to tell users which **input FPS** to use for their movies (since that is configurable), so they could match input FPS to in-game render FPS value.

DOSBox also allows to force a different render FPS value but we won't need that.

YoshiRulz:
I still have some unresolved comments in here, and I'm not sure if all of Spike's https://github.com/TASEmulators/BizHawk/pull/4217#issuecomment-2708328922 was addressed.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.Controllers.cs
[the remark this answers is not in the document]
vadosnaprimer: This was changed as well, also see [this comment](https://github.com/TASEmulators/BizHawk/commit/93bc50288f30d3d093ae3dba98aa2c44ce0b6fe7) about raw delta specifics in hawk that make exact ranges not incredibly important (as long as minimal possible value is 1).

src/BizHawk.Client.Common/movie/PlatformFrameRates.cs
YoshiRulz: With no value in the dictionary it falls back to 60. I want to see some value given with some sort of reasoning, whether that's experimental or based on a spec sheet.

The reference to JPC-rr values are from an older revision of this file.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
YoshiRulz: Yes. [rebased](https://github.com/YoshiRulz/BizHawk_BranchBackup/commit/8b2e9f1130e974ff889cd9bb55037956ff3686e1)

SergioMartin86:
> Hi. Still borked on reboot, so creating a verification movie of a 3.1/3.11 setup is still not possible.
> 
> image

it's not necessary to reboot for win3.1, simply go to the installation folder and run win.

Also have you tried rebooting like this? 
https://dosbox-x.com/wiki/Guide:DOS-Installation-in-DOSBox%E2%80%90X#_reboot_dos

src/BizHawk.Client.Common/movie/PlatformFrameRates.cs
vadosnaprimer: When dosbox-x reboots, [initial values](https://github.com/TASEmulators/dosbox-x/blob/324193b96393571b9a335f39a92aa619a37550bd/src/hardware/vga_draw.cpp#L7019) are:

| Name | Value | Type
| -- | -- | --
| clock | 3146888 | unsigned __int64
| vtotal*htotal | 44900 | unsigned __int64
| fps | 70.086592427616921 | double

which then gets written to video [like this](https://github.com/TASEmulators/dosbox-x/blob/324193b96393571b9a335f39a92aa619a37550bd/src/hardware/hardware.cpp#L1091-L1092).

Then immediately

| Name | Value | Type
| -- | -- | --
| clock | 3146875 | unsigned __int64
| vtotal*htotal | 52500 | unsigned __int64
| fps | 59.940476190476190 | double

and back to `3146888 / 44900` when DOS launches.

When DOOM launches, `3146888 / 44900` is explicitly set once again.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -0,0 +1,492 @@
...
+        private const int _messageDuration = 4;
+
+		// Drive management variables
+		private bool _nextFloppyDiskPressed = false;
+		private bool _nextCDROMPressed = false;
+		private List<IRomAsset> _floppyDiskImageFiles = new List<IRomAsset>();
+		private int _floppyDiskCount = 0;
+		private int _currentFloppyDisk = 0;
+		private int _currentCDROM = 0;
+		private string GetFullName(IRomAsset rom) => rom.Game.Name + rom.Extension;
```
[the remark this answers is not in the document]
vadosnaprimer: Hello?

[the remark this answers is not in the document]
YoshiRulz: The sole callsite:
https://github.com/TASEmulators/BizHawk/blob/942c708081dea35d38825b690dbd6dad40bdbed9/src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs#L415
Is `Path.GetFileName(rom.RomPath)` equivalent?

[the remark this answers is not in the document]
vadosnaprimer: > Is `Path.GetFileName(rom.RomPath)` equivalent?

Yes! Thanks, switched.

> Having the core print the host filename of a rom is a strange thing to do—consider that roms may not be sourced from any disk on the host, they could for example be downloaded from the Internet ephemerally by an ext. tool.

Discussion here: #4138. If someone can automate rom downloads (which is mostly illegal) they'd still have an idea what the filename is, otherwise how'd they even make an xml? And even if someone is so clueless that they have no idea which file is which, it doesn't mean it's completely useless info for normal users which file they've just inserted into guest. When swapping disks it's essential, they don't care how we call those files internally but it's clear which host file is being used.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
vadosnaprimer: Pulled.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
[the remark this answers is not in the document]
CasualPokePlayer: Path.GetFileName(rom.RomPath) is not equivalent, you'd need to at least do rom.RomPath.SubstringAfter('|') here for such.

SergioMartin86:
> > it's not necessary to reboot for win3.1, simply go to the installation folder and run win.
> > so creating a **verification movie**
> 
> Why is it something that has been showcased as something that can be done with [MAME](https://tasvideos.org/6082M) or [PCem](https://tasvideos.org/6314M) not be feasible for BizHawk?
> 
> > Have you tried rebooting like this?
> 
> The shutdown command disappears into the sunset when I tried.

Rebooting should be fixed now.

`shutdown /r`

->

all images reloaded correctly

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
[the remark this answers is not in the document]
YoshiRulz: > If someone can automate rom downloads (which is mostly illegal)...

Where the rom bytestream is read from is orthogonal to whether you are licensed to use or copy it. Case in point: I recently added the first [homebrew to the repo](https://gitlab.com/tasbot/libre-roms-ci/-/tree/master/homebrewAndDemos) we pull testroms from.

> ...they'd still have an idea what the filename is, otherwise how'd they even make an xml?

I hadn't thought of that, but I believe you're right that the bundle `.xml` loader necessitates there being a filename. It may not be a good design though.

> [...] it doesn't mean it's completely useless info for normal users which file they've just inserted into guest. When swapping disks it's essential, they don't care how we call those files internally but it's clear which host file is being used.

Of course.

waterbox/dosbox/bizhawk.cpp
```diff
@@ -124,6 +128,10 @@ ECL_EXPORT bool Init(InitSettings* settings)
 	ticksTarget = 0.0;
 	ticksElapsed = 0;
 
+	// Setting VGA refresh rate initial value, just in case
+	_vgaRefreshRateNumerator = settings->fpsNumerator;
+    _vgaRefreshRateDenominator = settings->fpsDenominator;
```
vadosnaprimer, line 133: Please initialize them as 0, so the first actual value is reported too.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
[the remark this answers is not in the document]
vadosnaprimer: > Where the rom bytestream is read from is orthogonal to whether you are licensed to use or copy it. Case in point: I recently added the first [homebrew to the repo](https://gitlab.com/tasbot/libre-roms-ci/-/tree/master/homebrewAndDemos) we pull testroms from.

Which is why I said "mostly".

>  It may not be a good design though.

What?

src/BizHawk.Emulation.DiscSystem/DiscIdentifier.cs
```diff
@@ -185,6 +190,9 @@ public DiscType DetectDiscType()
 			if (DetectWii())
 				return DiscType.Wii;
 
+			if (DetectISO9660()) return DiscType.DOS;
```
CasualPokePlayer, line 193: This should be moved much lower, after other ISO checks, DetectISO9660 is what the below [iso.Parse](https://github.com/TASEmulators/BizHawk/blob/6dccb759cba8c00d3cdbbb7d397649eadab0ed15/ExternalProjects/iso-parser/ISOVolumeDescriptor.cs#L220) does (along with CDI so it's not strictly ISO9660 anyways)

CasualPokePlayer (dismissed):

vadosnaprimer:
Update on framerates: they're reported by the core for those who want to force something specific, but the whole thing is now switching to ICycleCount and variable framerate unless something is forces, which won't happen often. So movie durations will be independent from internal framerates.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.ICycleTiming.cs
```diff
@@ -0,0 +1,15 @@
+﻿using BizHawk.Emulation.Common;
+
+namespace BizHawk.Emulation.Cores.Computers.DOS
+{
+	public partial class DOSBox : ICycleTiming
+	{
+		// DOSBox emulates the internal core at stable 1ms increments, regardless of the configuration. Therefore, this is basis for time calculation.
+		double ICycleTiming.ClockRate => 1000.0;
+
+		// How many ms have elapsed in the previous emulated frame
```
vadosnaprimer, line 10: Does this mean "during the previous frame only" or " overallup to the current frame overall"?

Morilli: Looking at the implemention this seems to be the ticks from one frame only, which is not what the frontend expects.

vadosnaprimer:
An axis whose range is all floppy indices, should be both easy enough to add and useful enough to be exhaustive for users. Might also consider this for Amiga. The only question is, is such a swapper also relevant for CDs?

SergioMartin86:
> Oh I see... what's a good way to make that a stand out then?

The reason I introduced this option is to prevent Bizhawk from constantly storing the entire hard disk (slow) when in the majority of use cases the users won't need to create verification movies. We'll provide tasvideos-sanctioned verif movies for all the possible windows version anyway.

> Also just for asking sake-- even though I believe it's _not_ necessary for this system. Is there a way to dynamically select what floppy disk is inserted?
> Cause all it is, is "next floppy disk" which is reminiscent to [an issue the Nyma cores had](https://github.com/TASEmulators/BizHawk/issues/3206).

> An axis whose range is all floppy indices, should be both easy enough to add and useful enough to be exhaustive for users. Might also consider this for Amiga. The only question is, is such a swapper also relevant for CDs?

Implemented for both floppy and cd. Not with an axis but with a prev/next/swap buttons. _Swap_ swaps the current media with the selected one. Soft-eject is not implemented in dosbox as it entails unmounting, which is irreversible (from a BK standpoint). Please test.

waterbox/dosbox/bizhawk.cpp
```diff
@@ -210,7 +210,7 @@ ECL_EXPORT void FrameAdvance(MyFrameInfo* f)
 	if (f->driveActions.insertCDROM >= 0)
 	{
 		printf("Swapping to CDROM: %d\n", f->driveActions.insertCDROM);
-		swapInDrive(3, f->driveActions.insertFloppyDisk + 1); // 3 is D:
+		swapInDrive(3, f->driveActions.insertCDROM + 1); // 3 is D:
```
vadosnaprimer, line 213: Is it fixed which letter each type of medium is inserted into?

src/BizHawk.Emulation.Common/Base Implementations/Bk2MnemonicLookup.cs
```diff
@@ -904,7 +904,7 @@ public static string LookupAxis(string button, string systemId)
 				["RightBracket"] = ']',
 				["Semicolon"] = ';',
 				["Quote"] = '\'',
-				["Period"] = '.',
+				["Period"] = 'p', // because '.' represents no input
```
YoshiRulz, line 907: We should have a unit test for this.

vadosnaprimer: Agreed, I most certainly did it too.

waterbox/dosbox/bizhawk.cpp
vadosnaprimer: Looks like yes. My final wish is to have an enum for those letters so we don't have to pass magic numbers coupled with comments. A `#define` is acceptable too.

.gitmodules
```diff
@@ -85,6 +85,9 @@
 [submodule "ExternalProjects/FlooohChips/chips"]
 	path = ExternalProjects/FlooohChips/chips
 	url = https://github.com/floooh/chips.git
+[submodule "waterbox/dosbox/dosbox-x"]
+	path = waterbox/dosbox/dosbox-x
+	url = https://github.com/TASEmulators/dosbox-x.git
```
vadosnaprimer, line 90: Oh and please add the branch here.

src/BizHawk.Emulation.Common/Base Implementations/Bk2MnemonicLookup.cs
YoshiRulz: I added an assert instead https://github.com/TASEmulators/BizHawk/commit/a41817b7030fec528aff5da57b88e543a8c5ea68

vadosnaprimer (approved):
THANKS FOR TREMENDOUS WORK!

I think this is ready for merging, and matters of saveram and weird mouse controls in some games will need to be resolved separately before we mark the core as released.

nattthebear:
> THANKS FOR TREMENDOUS WORK!

Agreed, thanks a million @SergioMartin86.

For historical context, DOSBox was a core that was sought after for years but deemed impossible because no good savestates existed, and could not be made due to some emulation core details (maybe host stack problems?)

After 15+ years, but you proved them wrong 🎉.

PR #4218, opened 2025-02-12
(Experimental Tool) TAStudioMPR - For Local Multiple Player Controllers

### LINKED ISSUE #899: TAStudio Feature request: Visual cue for seperating P1 from P2
So, I see this on TAStudio:
bug1

It would be much better for me to have a visual cue that seperate P1 from P2. Maybe a thicker and darker line?

Concept screen:

bug2

### LINKED ISSUE #2349: TAStudio Multiplayer TASing: clear frames for one player
Colin brought this up, and I figure it's a decent feature request, even if it turns out to be annoying to do and never gets done.

### LINKED ISSUE #2510: (issue title -- summarise the summary)
### Summary
In the fceux TASEditor manual, it mentions Multitrack recording.
http://fceux.com/web/help/taseditor/index.html?Toolbox.html#multitracking

It shows in the UI, the ability to pick a player, and only changes the desired inputs. http://fceux.com/web/help/taseditor/PianoRoll.html

Does TAStudio have this feature? Or a way to "lock" inputs so they don't get rewritten?

My usage case is to do a quick record of a player's movement across a stage then, go back and tighten up the route, then repeat with the next player.

ShinobiWannabe:
[dev build for branch | ShinobiWannabe:master](https://nightly.link/ShinobiWannabe/BizHawk/workflows/ci/master?preview)

This is a version of TAStudio splitting out the various "controllers" for a controller definition for a system into multiple InputRolls.  This is specifically targeting multiple player controller tases (think a local game with multiple controllers).   Basically with this you can do the main Insert, Delete, Clone, Copy/Paste within a single controller.  You can also copy input from one controller to another.

Youtube Example of TAStudioMPR working with a multiplayer game.
https://www.youtube.com/watch?v=qcw37tiDtzM

image

How It Works
Upon load it looks at the Emulator.ControllerDefinition with specific attention to the fields ControlsOrdered and the ControlsOrdered.Count.  It dynamically creates InputRolls (after the first InputRoll) based on the number, and then creates columns for each based on their inputs.  Once the the InputRolls are created they are re-sized and re-located to make it look nicer.  While visually the InputRolls are separate per controller the movie file is the same full size.  When operations such as Insert, Delete, Clone, Cut, Copy, Paste are performed they go to the TasMovie.Editing.cs with their own functions.  The startOffset and the controller Length for the current controller are passed in to most functions.  Here character and string manipulations are performed so that operations only effect the intended controller.  Here are some brief descriptions of the various important operations.

Clear: Will empty out the inputs for the selected frames of the specified InputRoll.

Insert Frame : First it adds an additional empty frame to the Log file at the end.  Then creates a copy of the current frame, but sets the inputs for the current controller to nothing aka ".". Then it shifts all the inputs down only within the relevant controller.

Insert # Frames : Similar to above, but it inserts them in batches so it does not do the entire shifting of inputs X number of times for inputs.

Clone: Like the Insert Frame operation, but instead of making the inserted frames have no input, it uses the selected input to clone.  

Clone X Times: This repeats the Clone operation X number of times.

Copy: This operation copies the selected rows and puts it into the clipboard.

Cut: This copies takes the inputs from within a controller and moves them to the clipboard then Deletes them from that controller.

Paste: This operation goes through the copied inputs and overwrites the current InputRoll's controller inputs starting from the selected frame. Note that this Paste only works within the same controller. For controller to controller look at the right click "Paste Selected Rows From".

Paste Insert: This is a combination of Insert and Paste where the inputs are inserted and the other inputs are shifted down. Note that this Paste only works within the same controller. For controller to controller look at the right click "Paste Selected Rows From".

Paste Selected Rows From: A new one that makes sense for multiple controllers.  Here if you select rows from a InputRoll you can right click on an InputRoll that you want to paste to, then select a "TasView" from the sub to paste the selected rows.  While originally I wanted to use the CurrentTasView (TasView and InputRoll refer to the same pianoroll object) and just freely copy and paste, I am not exactly certain how to do this cleanly.  With the copy you can also copy from some other string outside the program, you can also activate the Copy event from the TAStudio menu, and if you move the mouse around the focus can change as well.  If there is a cleaner way than using the dropdown menu to paste between InputRolls then I would be for it.

Truncate: This one takes the current frame and starting from there empties the inputs all the way to the end of the load for that controller.  Then it goes back to find the first frame where the rest of the inputs across all controllers are empty and will Remove Range from there to the end of the Log.

Please inform me of any thoughts, issues, or possible improvements.  This is pretty experimental, but it should be useful for tasers wanting an easier way to tas games with multiple controllers.  I had been waiting for years for something like this, but nothing was close to what I wanted for multiplayer games so I decided to give it a shot myself.


vadosnaprimer:
If frame column is included into each player's input roll, is the frame-column-only view really needed?

vadosnaprimer:
> What is the frame-column-only view? Not sure what you are referring to here.

[From the video.](https://github.com/user-attachments/assets/e174b9ef-85fb-4f4b-ad24-d47c5b8d8d02)

> When you say where the view belongs, what do you mean? Like adding a label above each view with a controller number or something?

Yeah.

adelikat:
I want this!

SuuperW:
Auto-fire significantly complicates the insert and delete operations when editing a subset of columns. I think that the best solution for this at the moment is to ignore the problem and assume auto-fire (and auto-hold) will not be used alongside split views. Unless someone intends to actually use these features together and can tell us how it should behave, I don't think we'll be able to make good decisions for that use case. And while ignoring auto-fire, there is no meaningful difference between a movie with 1,000 frames vs a movie with 1,001 frames where the last frame is empty; so we can allow insert to always extend the movie while delete never shortens the movie. The user will be able to use the truncate feature if they care about empty frames once the TAS is complete.

I also think that binding markers to input could result in unexpected behavior with split views. Which view's inputs are they bound to? Would you get different results from inserting a frame before a marker for player 1 then inserting one for player 2 vs. inserting a frame without a split view? We might want to warn users who enable both options (splitting into multiple input rolls should be optional) that they might have a bad time.

SuuperW:
@adelikat @ShinobiWannabe I have implemented something like this in #4661. Post bugs or other feedback there and I'll try to address them.

PR #4228, opened 2025-02-14
Refactor equality check in `MDS_Format.MountBlobs`

YoshiRulz:
This goes back to the initial implementation https://github.com/TASEmulators/BizHawk/commit/6d87be1396eb6b9ba41370bce76a7019774d826e. @Asnivor Can you say whether this matches the original intent? Or just whether it still functions correctly? That's probably more important.

(Throwback to my comment 6 years ago in https://github.com/TASEmulators/BizHawk/commit/1d5a097a75f93ad8898717835a2b6a419967441e.)


Morilli:
I sadly can't test this, but I find it highly likely that both this and the original code fail horribly in case a duplicate is found.

That said, I looked at how this dictionary is used and don't even see how that makes any sense and works, so...

PR #4264, opened 2025-03-08
Adding Opera (3DO) as emulation core in Bizhawk

SergioMartin86:

- [x] Add temporary .3do-bin based CD-ROM loading
- [x] Add proper CD reading callbacks
- [x] Add video buffer getting
- [x] Add sound
- [x] Add gamepads
- [x] Add mouse
- [x] Add settings
- [x] Add input mnemonics
- [x] Add drive access light callbacks
- [x] Add mem domains
- [x] Add SaveRAM
- [x] Add IRegionable
- [x] Add FlightStick
- [x] Add Lightgun
- [x] Add Arcade Lightgun
- [x] Add Trackball
- [x] Add Flight Stick
- [x] Load Font ROMS
- [x] Add input poll reading (red lag frames vs input read frames)
- [x] Allow any bios loading (currently limited to what the core recognizes)
- [x] Use proper mouse deltas
- [x] Add default inputs for mouse (‎Assets/defctrl.json)
- [x] Add multi-disc support
- [x] Rename all instances of _3DO to Panasonic3DO
- [x] Add Opera submenu to console configuration menu 
- [x] Fix joystick mapping 

Check if completed:


src/BizHawk.Client.Common/RomLoader.cs
```diff
@@ -978,6 +978,8 @@ private static class RomFileExtensions
 
 			public static readonly IReadOnlyCollection<string> A78 = new[] { "a78" };
 
+			public static readonly IReadOnlyCollection<string> _3DO = new[] { "3do-iso" };
```
CasualPokePlayer, line 981: Absolutely not. 3DO is a CD console, and looks to just be accepting normal CD files. Users should not be renaming their CD files to "3do-iso", they should be loading them up as normal and that goes through the standard BizHawk CD interface, which this core MUST use to be considered acceptable for merge.

CasualPokePlayer (changes requested):

src/BizHawk.Client.Common/RomLoader.cs
CasualPokePlayer: Discussed this on Discord, this was apparently used more for testing purposes before the CD interface was actually hooked up (done last).

Still, doing this is a dealbreaker, and makes this unacceptable for merge until that is done at least.

SergioMartin86: fixed

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
```diff
@@ -0,0 +1,159 @@
...
+			public int right;
+			public int start;
+			public int select;
+			public int buttonA;
+			public int buttonB;
+			public int buttonX;
+			public int buttonY;
+			public int buttonL;
+			public int buttonR;
+		}
```
Morilli, line 72: Is there a reason these are all `int`s instead of a type more closely resembling their data range, like `bool`?

SergioMartin86: I did try using `bool` but I keep getting false at the core side, even when I send trues from the bk side. I looked high and low for a discrepancy in the apis, but everything looks correct.

CasualPokePlayer: When using bool, was the core actually bool too, or just ints? The struct sizes need to match up.

Also the `Compatibility = true` would also probably cause issues regardless if you did that, as that would go under "normal" marshalling rules (i.e. the struct is converted into one with 4 byte bools instead of just directly pinning the original struct). You generally shouldn't use Compatibility = true unless you need "normal" marshalling as that slower.

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System.Runtime.InteropServices;
+using BizHawk.BizInvoke;
+using BizHawk.Emulation.Cores.Waterbox;
+
+namespace BizHawk.Emulation.Consoles._3DO
```
YoshiRulz, line 5: Don't like the underscore in namespace.

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
```diff
@@ -0,0 +1,341 @@
...
+		private int _discIndex;
+		private readonly List<DiscSectorReader> _cdReaders = new List<DiscSectorReader>();
+		private static int CD_SECTOR_SIZE = 2048;
+		private readonly byte[] _sectorBuffer = new byte[CD_SECTOR_SIZE];
+
+		private void SelectNextDisc()
+		{
+			_discIndex++;
+			if (_discIndex == _discAssets.Count) _discIndex = 0;
+			CoreComm.Notify($"Selected CDROM {_discIndex}: {_discAssets[_discIndex].DiscName}", _messageDuration);
```
YoshiRulz, line 179: The intent of the duration param is for a human to step in and keep long strings on-screen longer. If you're not customising them individually, pass `null`, not `4`.

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,402 @@
...
+	 return frames;
+}
+
+void RETRO_CALLCONV retro_input_poll_callback()
+{
+	_isLagFrame = false;
+		// printf("Libretro Input Poll Callback Called:\n");
+}
+
+int16_t processController(const int portType, controllerData_t& portValue, const unsigned device, const unsigned index, const unsigned id)
```
Morilli, line 95: It looks like the entire function would be better using a `switch` instead of dozens of `if`s. Also, I assume you have some copy-paste errors in here as things like `if (RETRO_DEVICE_ID_JOYPAD_Y)` make no sense.

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,402 @@
...
+    void opera_lr_callbacks_set_audio_sample_batch(retro_audio_sample_batch_t cb);
+    void opera_lr_callbacks_set_environment(retro_environment_t cb);
+    void opera_lr_callbacks_set_input_poll(retro_input_poll_t cb);
+    void opera_lr_callbacks_set_input_state(retro_input_state_t cb);
+    void opera_lr_callbacks_set_log_printf(retro_log_printf_t cb);
+    void opera_lr_callbacks_set_video_refresh(retro_video_refresh_t cb);
+    RETRO_API void *retro_get_memory_data(unsigned id);
+    RETRO_API size_t retro_get_memory_size(unsigned id);
+				void retro_set_controller_port_device(unsigned port_, unsigned device_);
+				void retro_get_system_av_info(struct retro_system_av_info *info_);
```
Morilli, line 62: There are mismatched spaces and tabs everywhere, could you decide on one and fix the formatting for this file?

YoshiRulz (dismissed):

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,402 @@
...
+		f->base.Samples = _audioSamples;
+		memcpy(f->base.SoundBuffer, _audioBuffer, _audioSamples * sizeof(int16_t) * _CHANNEL_COUNT);
+}
+
+ECL_EXPORT void GetMemoryAreas(MemoryArea *m)
+{
+	int memAreaIdx = 0;
+
+	m[memAreaIdx].Data  = _memoryAreas.wram;
+	m[memAreaIdx].Name  = "Work RAM";
```
Morilli, line 370: Isn't this internally `DRAM`? Does it make sense to call this wram here instead of dram?

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,402 @@
...
+{
+	 if (NVRAM == NULL) return;
+		memcpy(sramBuffer, get_sram_buffer(), get_sram_size());
+}
+
+ECL_EXPORT void set_sram(uint8_t* sramBuffer)
+{
+	 if (NVRAM == NULL) opera_nvram_init(NVRAM,NVRAM_SIZE);
+		memcpy(get_sram_buffer(), sramBuffer, get_sram_size());
+}
```
Morilli, line 402: For some reason `retro_get_memory_data` does not seem to expose this NVRAM (via `RETRO_MEMORY_SAVE_RAM` for example) but perhaps that would be the better place to expose this to instead of writing an own function for it.

You could then also expose this NVRAM as memory domain with the `MEMORYAREA_FLAGS_SAVERAMMABLE` flag and I believe the frontend will then automatically handle saveram for you.

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,402 @@
...
+	//// Getting memory areas
+	/** 2 = wram, 3 = vram*/
+
+	// WRAM
+	_memoryAreas.wram = (uint8_t*)retro_get_memory_data(RETRO_MEMORY_SYSTEM_RAM);
+	_memorySizes.wram = retro_get_memory_size(RETRO_MEMORY_SYSTEM_RAM);
+
+	// VRAM
+	_memoryAreas.vram = (uint8_t*)retro_get_memory_data(RETRO_MEMORY_VIDEO_RAM);
+	_memorySizes.vram = retro_get_memory_size(RETRO_MEMORY_VIDEO_RAM);
```
Morilli, line 315: I don't think having these stored in global structs is necessary, just call these functions in `GetMemoryAreas` directly.

waterbox/opera/Makefile
Morilli, line 1: As with the bizhawk.cpp file, this has lots of unnecessary and mismatched whitespace. You don't need to use one line per CCFLAGS argument, just put a couple on the same line and split it once the line becomes too long.

waterbox/opera/Makefile
```diff
@@ -0,0 +1,107 @@
...
+												-D__LIBRETRO__ \
+												-DNDEBUG \
+												-DHAVE_CDROM \
+												-DHAVE_STDINT_H \
+												-DHAVE_STDLIB_H \
+												-DHAVE_SYS_PARAM_H \
+												-DINLINE=inline \
+												-DTHREADED_DSP
+
+CXXFLAGS := $(CCFLAGS) 	-std=c++20
```
Morilli, line 21: Is this just for the bizhawk.cpp? Might be easier to just make it a normal `.c` file and get rid of all the cpp handling.

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,402 @@
...
+  uint8_t* vram;
+};
+
+struct MemorySizes
+{
+	size_t wram;
+  size_t vram;
+};
+
+#define _MAX_AUDIO_SAMPLE_COUNT 4096
```
Morilli, line 23: unused?

Morilli (changes requested):

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
SergioMartin86: fixed

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
SergioMartin86: I couldn't come up with a better solution. Namespaces or variables cannot start with a number

CasualPokePlayer: ThreeDO

waterbox/opera/bizhawk.cpp
SergioMartin86: my IDE was playing tricks with me. Fixed now.

waterbox/opera/bizhawk.cpp
SergioMartin86: fixed. Also exposing non volatile ram (SaveRAM)

waterbox/opera/bizhawk.cpp
SergioMartin86: fixed

waterbox/opera/Makefile
SergioMartin86: fixed. again my ide playing games with me

waterbox/opera/Makefile
SergioMartin86: I do use std::string in bizhawk.cpp. It is much cleaner than C-style string manip. But I removed the c++20 and cleaned things up

waterbox/opera/bizhawk.cpp
SergioMartin86: yes. removed

waterbox/opera/bizhawk.cpp
SergioMartin86: This is the way I do it across all my projects. It helps me a lot to keep things tidy. I rather keep it consistent on my side

waterbox/opera/Makefile
SergioMartin86: I keep flags one per line because that helps in quickly removing  and adding one by one

src/BizHawk.Emulation.Common/Base Implementations/Bk2MnemonicLookup.cs
```diff
@@ -83,6 +83,11 @@ public static string LookupAxis(string button, string systemId)
 			["Run"] = 'R',
 			["RUN"] = 'R',
 
+			["A"] = 'A',
+			["B"] = 'B',
+			["X"] = 'X',
+			["Y"] = 'Y',
```
Morilli, line 89: Those should be unnecessary as the mnemonic fallback is already the string itself.

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
SergioMartin86: I tried again, making sure again that the structs coincide. You can find them here: 

[BK Side]
https://github.com/SergioMartin86/BizHawk/blob/b4836176a8df5d2eba2d77598412036767840d8d/src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs#L58

[WBX side] https://github.com/SergioMartin86/BizHawk/blob/b4836176a8df5d2eba2d77598412036767840d8d/waterbox/opera/bizhawk.hpp#L13

The values are simply not passed around:
image

Perhaps you will see the discrepancy -- I just don't see it

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
SergioMartin86: whatever you guys decide is better. I personally think _3DO is much closer both alphabetically and visually than ThreeDO.

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
Morilli: It's probably got something to do with the bullshit default bool marshalling rules that nobody wants, although `BizInvoke` should usually get rid of that...

CasualPokePlayer: In the case Compatibility = false, it does get "rid" of it, as Compatibility = false always just pins the struct and passes the pointer along (never any conversion, great as long as you only have blittable members). In the case of Compatibility = true, it has to go through standard marshalling code as you're just calling a delegate from GetDelegateForFunctionPointer (which is again much slower than using Compatibility = false)

Morilli: Yeah but there is no compatibility in the usage of this struct (just the frame advance call as far as I can see)

CasualPokePlayer: Ah, in that case it's getting placed into a class, which has some strange behavior in this regard, since this isn't strictly blittable behavior overall. For saving space and just using bools, `byte` would be used C# side (or maybe rather bit flags if ypu really want to conserve space, which might or might not be wanted since bigger frame info means more GC pressure)

waterbox/opera/bizhawk.cpp
```diff
@@ -0,0 +1,393 @@
...
+size_t RETRO_CALLCONV retro_audio_sample_batch_callback(const int16_t *data, size_t frames)
+{
+  memcpy(_audioBuffer, data, sizeof(int16_t) * frames * _CHANNEL_COUNT);
+  _audioSamples = frames;
+  return frames;
+}
+
+void RETRO_CALLCONV retro_input_poll_callback()
+{
+  _isLagFrame = false;
```
CasualPokePlayer, line 75: retro_input_poll is completely useless in figuring out lag information. It is simply a callback that tells the frontend that the *core* wants input updated, which in practice is something that happens every single retro_run call regardless, making it fairly redundant in the libretro API. It is *not* a callback that tells the frontend that the *game* is acquiring input (i.e. actual non-lag indication). You cannot rely on libretro handling lag information for you, BizHawk code must have its own implementation for that.

CasualPokePlayer (changes requested):

waterbox/opera/Makefile
vadosnaprimer: > I keep flags one per line because that helps in quickly removing and adding one by one

Same here.

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
Morilli: Adding a `uint64_t xdd;` at the start of `struct gamePad_t` makes it work, so apparently there's 8 extra padding bytes in the c# class somehow. There more I've read into marshalling and blitting the more confused I get as to how things even work in other cores. It's possible everything that is working is just working by pure chance.

CasualPokePlayer: Could be weird mismatch of C# vs C padding rules? In the perspective of marshalling, a lot of usual knowledge doesn't *strictly* apply per se, since the class instance just gets pinned down and the pointer to such is passed on, so it matters how C# is internally ordering the data.

waterbox/opera/bizhawk.cpp
SergioMartin86: fixed

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
SergioMartin86: fixed -> ThreeDO

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
```diff
@@ -0,0 +1,324 @@
...
+			DefaultHeight = LibOpera.PAL2_HEIGHT,
+			MaxWidth = LibOpera.PAL2_WIDTH,
+			MaxHeight = LibOpera.PAL2_HEIGHT,
+			DefaultFpsNumerator = LibOpera.PAL2_VIDEO_NUMERATOR,
+			DefaultFpsDenominator = LibOpera.PAL2_VIDEO_DENOMINATOR,
+		};
+
+		private readonly List<IDiscAsset> _discAssets;
+
+		private string GetFullName(IRomAsset rom) => rom.Game.Name + rom.Extension;
```
YoshiRulz, line 57: Ditto of https://github.com/TASEmulators/BizHawk/pull/4217#discussion_r1999430713

SergioMartin86:
> As someone who got the up-to-date build for testing, I may bring some more points:
> 
> 1. Both controller and TAStudio layout have wrong namings for some buttons:
>    Left column is a 3DO button. Right - button mapping in Bizhawk as of now.
>    A - Y
>    B - X
>    C - A
>    X (Stop) - s ("Select")
>    P (Play/Pause) - S ("Start"), X
> 
> 
Thanks so much for the info, I'll get the controller fixed. 

> I think, this might be a good candidate to put in "3DO Controller Configuration" window: https://www.deviantart.com/adrianoramosofht/art/Panasonic-3DO-Controller-2-845186266

The controller pic is nice, though I have no idea how to add it to the Bizhawk GUI.

> 2. (optional) There is no option to plug in Player 2-8
Player 2 should be functional, but I haven't tested it. I'll give it a lok

> 3. .chd/.ccd/.cue game images can only be launched as a XML Multi-Disk Bundler

Afaik .cue images ought to work, but the other devs here should be able to give you a better answer

vadosnaprimer:
For controller images we use actual photos.

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
vadosnaprimer, line 57:
```suggestion
		private string GetFullName(IRomAsset rom) => Path.GetFileName(rom.RomPath.SubstringAfter('|'));

```

SergioMartin86:
> As someone who got the up-to-date build for testing, I may bring some more points:
> 
> 1. Both controller and TAStudio layout have wrong namings for some buttons:
>    Left column is a 3DO button. Right - button mapping in Bizhawk as of now.
>    A - Y
>    C - A
>    X (Stop) - s ("Select")
>    P (Play/Pause) - S ("Start"), X

Fixed.

image

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
SergioMartin86: Removed this function. Wasn't called anywhere

vadosnaprimer:
> For controller images we use actual photos.

The only one in public domain I could find is this GoldStar one https://commons.wikimedia.org/wiki/File:3DO-101M-Controller.jpg

There are Panasonic controller photos but they are creative-commons-share-alike, which is not compatible with us.

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
vadosnaprimer: What's the state of this discussion?

SergioMartin86:
> > For controller images we use actual photos.
> 
> The only one in public domain I could find is this GoldStar one https://commons.wikimedia.org/wiki/File:3DO-101M-Controller.jpg
> 
> There are Panasonic controller photos but they are creative-commons-share-alike, which is not compatible with us.

I don't know how to add this in Bizhawk. Help?

vadosnaprimer:
Weird, no idea why I'm seeing this, it's not in this PR it seems

изображение

.gitmodules
```diff
@@ -93,3 +93,6 @@
 	path = waterbox/dsda/core
 	url = https://github.com/TASEmulators/dsda-doom.git
 	branch = wbx
+[submodule "waterbox/opera/opera-libretro"]
+	path = waterbox/opera/opera-libretro
+	url = https://github.com/TASEmulators/opera-libretro.git
```
vadosnaprimer, line 98: Branch please.

src/BizHawk.Client.Common/movie/PlatformFrameRates.cs
```diff
@@ -17,6 +17,7 @@ public static class PlatformFrameRates
 
 		private static readonly Dictionary<string, double> Rates = new Dictionary<string, double>
 		{
+			["Panasonic3DO"] = 60.0, // The emulator (Opera-Libretro) reports exact 60.0 for NTSC https://github.com/libretro/opera-libretro/blob/67a29e60a4d194b675c9272b21b61eaa022f3ba3/libopera/opera_region.c#L10
```
vadosnaprimer, line 20: Please add the PAL variant too (same key but with a `"_PAL"` suffix).

.gitmodules
SergioMartin86: added

src/BizHawk.Client.Common/movie/PlatformFrameRates.cs
SergioMartin86: added

SergioMartin86:
> Weird, no idea why I'm seeing this, it's not in this PR it seems
> 
> изображение
> 
> > Afaik .cue images ought to work, but the other devs here should be able to give you a better answer
> 
> Other devs halp!

Both these issues should be resolved now

vadosnaprimer:
I think for places where we show the core name as a string to the user, "3DO" is better than "Panasonic3DO". It wasn't exclusive to Panasonic, and there's no problem in starting strings with a digit.

src/BizHawk.Emulation.Cores/Consoles/3DO/LibOpera.cs
Morilli: I think the problem is that bool types don't count as blittable types for whatever garbage legacy reason in C# which is most likely the reason using bool in any struct at all no longer guarantees struct layout. So as unsatisfied as I am with the conclusion I say leave it as is.

Morilli: I don't think so. Applying atrributes to every bool would only change the layout in interop which we're not using here.

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
```diff
@@ -0,0 +1,320 @@
...
+					gameInput.gamepad.down = controller.IsPressed($"P{port} {JoystickButtons.Down}") ? 1 : 0;
+					gameInput.gamepad.left = controller.IsPressed($"P{port} {JoystickButtons.Left}") ? 1 : 0;
+					gameInput.gamepad.right = controller.IsPressed($"P{port} {JoystickButtons.Right}") ? 1 : 0;
+					gameInput.gamepad.buttonX = controller.IsPressed($"P{port} {JoystickButtons.ButtonX}") ? 1 : 0;
+					gameInput.gamepad.buttonP = controller.IsPressed($"P{port} {JoystickButtons.ButtonP}") ? 1 : 0;
+					gameInput.gamepad.buttonA = controller.IsPressed($"P{port} {JoystickButtons.ButtonA}") ? 1 : 0;
+					gameInput.gamepad.buttonB = controller.IsPressed($"P{port} {JoystickButtons.ButtonB}") ? 1 : 0;
+					gameInput.gamepad.buttonC = controller.IsPressed($"P{port} {JoystickButtons.ButtonC}") ? 1 : 0;
+					gameInput.gamepad.buttonL = controller.IsPressed($"P{port} {JoystickButtons.ButtonL}") ? 1 : 0;
+					gameInput.gamepad.buttonR = controller.IsPressed($"P{port} {JoystickButtons.ButtonR}") ? 1 : 0;
```
YoshiRulz, line 239: I feel like this method can be cleaned up further

YoshiRulz (approved):
Besides my one comment here, code LGTM.

YoshiRulz:
> I think for places where we show the core name as a string to the user, "3DO" is better than "Panasonic3DO". It wasn't exclusive to Panasonic, and there's no problem in starting strings with a digit.

Is that not the case already? The core name is `Opera`, the sysID (which I assume you meant) and long name are both `3DO`, and the firmware group name is `3DO / 3DO Arcade / 3DO M2`. The raw sysID should only appear in movies etc. and I think the Multi-Disk Bundler.

SergioMartin86:
> > I think for places where we show the core name as a string to the user, "3DO" is better than "Panasonic3DO". It wasn't exclusive to Panasonic, and there's no problem in starting strings with a digit.
> 
> Is that not the case already? The core name is `Opera`, ~the sysID (which I assume you meant)~ I was looking at the diff here but it doesn't add the sysID because it was added before! It is indeed `Panasonic3DO`. We should be able to change that and not break anything. Anyway the system long name is `3DO`, and the firmware group name is `3DO / 3DO Arcade / 3DO M2`, which are intended for display. The raw sysID should only appear in movies etc. and I think the Multi-Disk Bundler.

There was a remaining place that said "Panasonic3DO", but I now fixed it: 

image

src/BizHawk.Emulation.Cores/Consoles/3DO/Opera.cs
vadosnaprimer: Any suggestions?

YoshiRulz: Maybe:
```csharp
var pfx = $"P{port} ";
int IsPressed(string button) => controller.IsPressed(pfx + button) ? 1 : 0;
gameInput.gamepad.buttonB = IsPressed(JoystickButtons.ButtonB);
```

SergioMartin86: I see the point being made, but personally I value readability and maintainability pretty highly. Even if this comes at the price of some repetition.

The proposed change doesn't reduce the number of LoC, but in fact adds an indirection for the meaning of pfx making, imo, the code ever so slightly harder to read.

I'm not opposed to the change if the maintainers think it's better, but I'd personally leave this unchanged.

vadosnaprimer: Indeed, doesn't look too much simpler than the original. Doesn't feel like a qualitative improvement, but rather a matter of style, so I'd leave it as is I guess.

PR #4272, opened 2025-03-16
TAStudio marker lua 

### LINKED ISSUE #1161: TAStudio branches/markers lua functions
Same as in TAS Editor: https://fceux.com/web/help/taseditor/LuaAPI.html#markedframe
and...
+ getmarkerframe(int index)
+ restructure removemarker(int frame) to removemarker(int index) instead

c7fab:
Implements four new lua functions:

- tastudio.getmarkerabove(frame)
returns a table with fields Frame and Text for the marker that is at or above the given frame, where the field Frame is the actual frame where the marker is located
- tastudio.getmarkers()
returns a list with all markers for the loaded movie, for which each entry has a table with fields Frame and Text
- tastudio.getbranchmarkerabove(branchId, frame)
returns a table with fields Frame and Text for the marker that is at or above the given frame for the given branchId
- tastudio.getbranchmarkers(branchId)
returns a list with all markers for the loaded movies for the given branchId

Those functions returned the tables in with no time drawbacks in constant time.

With those function it's possible to do what's possible in FCEUX TAS Editor, thus this PR implements all functionality from and resolves #1161. Furthermore it's not needed to have a function that gets a "marker index", since the "Id" can just be set in tastudio.getmarkers()[index].
It also makes it possible to delete a marker by index with tastudio.removemarker(tastudio.getmarkers()[index].Frame).

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -475,12 +475,14 @@ namespace BizHawk.Client.EmuHawk
 
 		[LuaMethodExample("local marker = tastudio.getbranchmarkerabove(tastudio.getbranches()[1].Id, 100)")]
 		[LuaMethod("getbranchmarkerabove", "returns a table of the marker at or above the given frame for the given branchId with fields Frame and Text")]
-		public LuaTable GetBranchMarkerAbove(string branchId, int frame)
+		public LuaTable GetBranchMarkerAbove(int index, int frame)
 		{
 			var table = _th.CreateTable();
 			if (!Engaged()) return table;
 
-			var branch = Tastudio.CurrentTasMovie.Branches.FirstOrDefault(b => b.Uuid.ToString() == branchId);
+			if (index >= Tastudio.CurrentTasMovie.Branches.Count) return table;
+
+			var branch = Tastudio.CurrentTasMovie.Branches[index];
 			var marker = branch.Markers.PreviousOrCurrent(frame);
 			table["Frame"] = marker.Frame;
 			table["Text"] = marker.Message;
@@ -507,11 +509,13 @@ namespace BizHawk.Client.EmuHawk
 
 		[LuaMethodExample("local markers = tastudio.getbranchmarkers(tastudio.getbranches()[1].Id)\r\nfor i = 0, #makers, 1 do\r\n\tconsole.log(v[i].Text)\r\nend")]
 		[LuaMethod("getbranchmarkers", "returns a table of all markers with fields Frame and Text")]
-		public LuaTable GetBranchMarkers(string branchId)
+		public LuaTable GetBranchMarkers(int index)
 		{
 			if (!Engaged()) return _th.CreateTable();
 
-			var branch = Tastudio.CurrentTasMovie.Branches.FirstOrDefault(b => b.Uuid.ToString() == branchId);
+			if (index >= Tastudio.CurrentTasMovie.Branches.Count) return _th.CreateTable();
+						
+			var branch = Tastudio.CurrentTasMovie.Branches[index];
 
 			return _th.EnumerateToLuaTable(
 				branch.Markers.Select(m =>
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -459,6 +459,71 @@ public string GetMarker(int frame)
...
+
+			var branch = Tastudio.CurrentTasMovie.Branches.FirstOrDefault(b => b.Uuid.ToString() == branchId);
+			var marker = branch.Markers.PreviousOrCurrent(frame);
+			table["Frame"] = marker.Frame;
+			table["Text"] = marker.Message;
+
+			return table;
+		}
+
+		[LuaMethodExample("local markers = tastudio.getmarkers()\r\nfor i = 0, #makers, 1 do\r\n\tconsole.log(v[i].Text)\r\nend")]
```
Morilli, line 491: Everything beyond the getmarkers() doesn't even run as is, maybe just remove that part? Ideally the description is enough to understand the structure.

Morilli (changes requested):

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -459,6 +459,71 @@ public string GetMarker(int frame)
...
+				{
+					var table = _th.CreateTable();
+					table["Frame"] = m.Frame;
+					table["Text"] = m.Message;
+					return table;
+				}),
+				indexFrom: 0);
+		}
+
+		[LuaMethodExample("local markers = tastudio.getbranchmarkers(tastudio.getbranches()[1]\r\nfor i = 0, #makers, 1 do\r\n\tconsole.log(v[i].Text)\r\nend")]
```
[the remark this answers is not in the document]
c7fab: Shouldn't write the example in plain text... The `tastudio.getbranchinput` function also takes the branch uuid, so I just build it of that. It probably can be done with the index.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -459,6 +459,75 @@ public string GetMarker(int frame)
 			return "";
 		}
 
+		[LuaMethodExample("local markerframe = tastudio.getmarkerabove(100).Frame")]
+		[LuaMethod("getmarkerabove", "returns a table of the marker at or above the given frame with fields Frame and Text")]
+		public LuaTable GetMarkerAbove(int frame)
```
Morilli, line 464: Not entirely sure about the name containign "above", especially when it could also return a marker for the frame specified instead of the marker "above" that frame. Perhaps `GetMarkerFor(int frame)` would communicate this better, but I don't feel strongly about this; if you think this is fine keep it.

Morilli (commented):
Code looks fine and from some basic testing it also works well and as expected.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
c7fab: In TAS Editor this behaviour was just called getmarker, but in TAStudio that function name is already in use. I will leave the name as is.

YoshiRulz:
This API is garbage. Why are they all returning associative arrays (with duplicated code I might add)? Why is it 0-indexed? Why is an out-of-range index argument being silently ignored? Reverting.

Morilli:
> This API is garbage.

Do you have some alternative suggestions? Clearly getting access to markers via lua is something people want, so what is the problem here?

YoshiRulz:
I said exactly what the problems are. [This branch](https://github.com/YoshiRulz/BizHawk_BranchBackup/tree/markerlua-yoshi) fixes some of them but mostly keeps the 4 functions as they are. My suggestion for a better API would be to return only marker indices from the lookup functions and have a separate function for retrieving the name of a marker.

Issue #377, opened 2015-06-22
Evaluate pointers in RAM watch


GoogleCodeExporter:
```
I've been doing some work with N64 games using BizHawk 1.9.1 lately and have 
run into the issue of games storing data I want to monitor in different areas 
of memory each execution, locating them using a pointer + offset system.

An example is character position in Donkey Kong 64. The 4 byte pointer is 
located at 0x807FBB4C and X, Y & Z position values are 4 byte floats stored at 
[base + 0x7C], [base + 0x80] and [base + 0x84] respectively, where base is the 
location on the system bus that 0x807FBB4C points to.

Monitoring these values requires me to write some Lua to resolve the pointer 
and display the value on screen or hook BizHawk with an external tool that 
supports pointer evaluation such as Cheat Engine.

Needless to say both workarounds aren't as elegant as using BizHawk's built in 
tools for watching and poking values. This feature would make my work a lot 
quicker and allow distribution of tables to the TAS and glitch hunting 
communities without bundling Lua scripts that simply evaluate pointers.

Thanks in advance, Isaac.
```

Original issue reported on code.google.com by `partyatm...@gmail.com` on 21 Jan 2015 at 3:51

zeromus:
I think what he means is *(float*)(*((u8*)0x807FBB4C)+0x7C)
Not sure If I'd rather attempt to evaluate C-like pointer expressions like that or use some kind of ad-hoc language. We could use whatever cheat engine uses... but evaluating the C-like expression could be feasible, perhaps if combined with some conventions that make sense for us. Here's a suggestion. Oops, I guess it's kind of a custom language now:
*0x807FBB4C+0x7C
This says "dereference 0x807FBB4C to get a u8*. Add 0x7C.". The result is a pointer. The result is always a pointer, that's what this expression is supposed to be. And it's assumed to be of the needed type.

adam-larson-lee:
My guess is a lot of games use this approach. I'm working with Demon's Crest right now and each game object in WRAM has pointers to a ROM location for certain information (hitbox size, for example).

I think what would work well here is a stack calculator, because in my case I don't just want to do all my operations in WRAM. Something like this would well for my case and could be easily adapted for the original feature request:

'''
read( 'ROM', read('WRAM', 0x001300, u16), u16)
'''

Then your watch, instead of an address, has a this calculation string where the output is the value. Alternatively you could do:

'''
read('WRAM', 0x001300, u16)
'''

[...]

PR #4279, opened 2025-03-28
Watches with pointers

YoshiRulz:
With this PR, I aim to implement #377. My goal is to have a checkbox in the watch editor which swaps out the address box for a freeform textbox (that's all done), into which you can type semi-arbitrary expressions involving pointer arithmetic (see below). The editor will store this with a new `Watch` subclass, which holds onto an AST of the expression (should be done modulo parsing), re-evaluating the effective address each time RAM Watch calls `Update` (should be done modulo the peeks). Addresses are assumed to be in the sysbus, and the editor enforces this, disabling pointer mode when no sysbus exists (done).

For the expressions, I plan use a C-like syntax with:
- hex literals (and only hex literals) for addresses;
- '$'-prefixed decimal literals (or '$0x'-prefixed hex literals) for constants;
- a pointer dereference operator;
- grouping with parentheses;
- addition, subtraction, and multiplication operators for offsets;
- bitwise intersection and conjunction (AND and OR) operators and a right shift operator for masking.
- Example: `*(0x1234 + 2 * (*0xABCD & $7))`
- And expressions will be typed—either pointer- or non-pointer-coloured u32—and operators will only accept certain types of argument.
	- Actually I think only binary operations on two pointer arguments and dereferencing on non-pointers need to be blocked? No reason to block arithmetic with non-pointers.
- I think I'll just copy [C#'s operator precedence](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#operator-precedence).
- My plan was to distinguish multiplication and dereferencing by requiring whitespace around all binary operators. I believe it would be possible to have a consistent parser even without whitespace (like... C#'s), it's just simpler that way. Alternatively, we could change the sigil for dereferencing to something unique, maybe '@'.

### DIFF
--- a/src/BizHawk.Client.Common/tools/Cheat.cs
+++ b/src/BizHawk.Client.Common/tools/Cheat.cs
@@ -167,17 +167,24 @@ namespace BizHawk.Client.Common
 			{
 				if (ShouldPoke())
 				{
-					switch (_watch.Size)
+					try
 					{
-						case WatchSize.Byte:
-							_watch.Poke(((ByteWatch)_watch).FormatValue((byte)_val));
-							break;
-						case WatchSize.Word:
-							_watch.Poke(((WordWatch)_watch).FormatValue((ushort)_val));
-							break;
-						case WatchSize.DWord:
-							_watch.Poke(((DWordWatch)_watch).FormatValue((uint)_val));
-							break;
+						switch (_watch.Size)
+						{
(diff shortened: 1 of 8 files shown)


vadosnaprimer:
In asm code I've seen, $ is usually a prefix for hexadecimal and specifically for addresses. For constants (immediate values), # is used on top.

Issue #4206, opened 2025-02-02
[Genplus GX] Sega CD game, After Armageddon Gaiden malfunctions in 2.10


BlackPaladin2022:
Bizhawk 2.10 x64 Windows 10

For unknown reason, the Sega CD game, "After Armageddon Gaiden: Maju Toushouden Eclipse" will not run correctly.  After the BIOS is displaying and you hear the Sega jingle, the game freeze on the screen I'm showcasing here, but you'll hear the music coming from the game running about a minute later.  This issue does not occur in Bizhawk 2.91.

![Image](https://github.com/user-attachments/assets/8c286741-7eb2-4812-932e-ee9551489d61)

The first file in the BIN/CUE set uses this hash if it's relevant.  The game will run correctly on other emulators like Retroarch 1.20, but it will not run correctly on BizHawk 2.10 (Again, Bizhawk 2.91 doesn't have this issue.)

File/ROM SHA-1: 68D54D51960D357DE9E4018AA3324ECD13272135
File/ROM CRC32: EBFA34C8

Morilli:
I've bisected this to be caused by https://github.com/ekeeke/Genesis-Plus-GX/commit/a6002bb25431c6ce03cf10e67c317403990312bc which is literally the last commit included in our gpgx version.

PR #4283, opened 2025-04-01
Memory Hook Improvements (Return value)

VelpaChallenger:
Part 2 of https://github.com/TASEmulators/BizHawk/pull/3821. Includes MemoryCallbackDelegate return value update from `uint` to `uint?`.

Check if completed:


src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
```diff
@@ -60,12 +60,18 @@ public NamedLuaFunction(LuaFunction function, string theEvent, Action<string> lo
 			{
 				try
 				{
-					_function.Call(args);
+					object[] LuaFunctionReturn = _function.Call(args);
+					if(LuaFunctionReturn?.Length != 0) // If the value returned is not specified in the callback, the array will not have any elements.
+					{
+						return Convert.ToUInt32(LuaFunctionReturn[0]);
+					}
+					return null;
```
YoshiRulz, line 68: This should `return _function.Call(args);` and let the `MemCallback` implementation do the type check.

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
```diff
@@ -84,7 +90,12 @@ public NamedLuaFunction(LuaFunction function, string theEvent, Action<string> lo
 				LuaLibraries.IsInInputOrMemoryCallback = true;
 				try
 				{
-					Callback(new object[] { addr, val, flags });
+					uint? CallbackReturn = Callback(new object[] { addr, val, flags });
+					if (CallbackReturn is null)
+					{
+						return val;
+					}
+					return (uint) CallbackReturn;
```
YoshiRulz, line 98: Per above, `Callback` should return an `object?[]`, and this should be
```csharp
return Callback([ addr, val, flags ]) is [ long n ] ? unchecked((uint) n) : val;
```
Maybe print if something unexpected was returned.

src/BizHawk.Client.EmuHawk/tools/Debugger/BreakpointControl.cs
```diff
@@ -53,14 +53,15 @@ private void BreakPointView_QueryItemText(object sender, RetrieveVirtualItemEven
 				: Color.White;
 		}
 
-		private void BreakpointCallback(uint addr, uint value, uint flags)
+		private uint? BreakpointCallback(uint addr, uint value, uint flags)
 		{
 			MainForm.PauseEmulator();
 			UpdateValues();
 			MainForm.AddOnScreenMessage("Breakpoint hit");
+			return 0;
```
YoshiRulz, line 61: Shouldn't these be `null`?

src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
```diff
@@ -66,47 +66,50 @@ public void Add(IMemoryCallback callback)
...
-		private static void Call(MemoryCallbackCollection cbs, uint addr, uint value, uint flags, string scope)
+		private static uint Call(MemoryCallbackCollection cbs, uint addr, uint value, uint flags, string scope)
 		{
+			uint cbReturn = value; //By default, if no callback is called, it will return the value sent by the core unmodified.
 			foreach (var cb in cbs)
 			{
 				if (!cb.Address.HasValue || (cb.Scope == scope && cb.Address == (addr & cb.AddressMask)))
 				{
-					cb.Callback(addr, value, flags);
+					cbReturn = (uint) cb.Callback(addr, value, flags); //If there are many callbacks registered to the same addr, only the last callback's return value will be sent to the core.
```
YoshiRulz, line 76: From https://github.com/TASEmulators/BizHawk/pull/3821#discussion_r2023574499, this needs to be something like `if (cb.Callback(addr, value, flags) is uint n) cbReturn = n;` to get the desired behaviour.

Though I wonder how that will behave if one callback "vetos" by returning a value, and another callback uses the old method of poking memory.

src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
```diff
@@ -66,47 +66,50 @@ public void Add(IMemoryCallback callback)
...
+					value = Call(_reads, addr, value, flags, scope);
 				}
 			}
 
 			if (HasWrites)
 			{
 				if ((flags & (uint) MemoryCallbackFlags.AccessWrite) != 0)
 				{
-					Call(_writes, addr, value, flags, scope);
+					value = Call(_writes, addr, value, flags, scope);
```
YoshiRulz, line 101: Should w/x hooks be allowed to change the value, or only reads? Keep in mind that the same callback object can be registered for r/w/x.

YoshiRulz (commented):
Fun fact, GitHub now requires you to fill out this field!

VelpaChallenger:
What field?

src/BizHawk.Emulation.Common/Base Implementations/MemoryCallbackSystem.cs
VelpaChallenger: From our discussion in Discord, I believe we want w hooks to be allowed to change the value as well. Not sure why you would want to override in w hooks when you can do it in r hooks (and it's more powerful too I think, as it works even with unitialized values), but at the very least the possibility is there. Also, this is the quote I'm referring to:

>allowing a callback to modify the value would be the best solution as it would also allow changing write values in write hooks which i believe does not work currently

In such a scenario where the same callback object is registered for r/w/x, I would leave the responsibility to the user to decide whether to return a value, or not. Since it's the same callback, it will either override the same value for all r/w/x, or leave it as it is for all of them. Which I think would be the expected (otherwise, a different callback should be used).

src/BizHawk.Client.EmuHawk/tools/Debugger/BreakpointControl.cs
VelpaChallenger: Yes, I was actually wondering if they should be 0 or null. None should make much of a difference, but perhaps null is clearer on the intent, specially if in the future we want an integer to be returned and to mean something (right now, it is never consumed, so it's literally the same whatever you return).

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
VelpaChallenger: So the Lua function returns always a long, and we're casting it to an uint (if it is long) to keep consistency with what the core reads/writes/execs?

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
VelpaChallenger: Oh I see. We have two types of callback, input and memory. By doing the check in `Callback` instead of `MemCallback`, we're including the check to `Input` as well. Ok, I will move it in that case.

VelpaChallenger:
That should do it. My local tests were successful in this commit.

Morilli:
@YoshiRulz can we get this in or do you see something wrong with it? I wanna update gpgx to fix #4293 and #4206 before next release and would like to get this in first to avoid conflicts.

Morilli:
@VelpaChallenger can you rebase/merge latest master as well? You'll need to rebuild the waterbox core to include changes from both this PR and 2e539b3c471a411411978acb9b1151460d2f51a3 to resolve the conflict.

Morilli (approved):
Looks good to me.

Issue #3818, opened 2023-10-31
Organisation of subprojects


YoshiRulz:
tl;dr: I'm calling for a moratorium on copying third-party code into the repo, and I'm proposing that we move a few dirs around.

We all hate code duplication. So why is it okay for us to needlessly copy entire codebases?
~~snip~~ (Cut out a bunch of back-and-forth here since I feel none of you need to be convinced this is a good idea, only that it's worth the effort to do it.)

So let's assess the damage:
- `/ExternalCoreProjects/Virtu`, Virtu core, < 11 kLOC, upstream is https://github.com/digital-jellyfish/Virtu/tree/master/Virtu
- `/ExternalProjects/NLua`, Lua host, < 10 kLOC, upstream is https://github.com/NLua/NLua
- `/ExternalProjects/iso-parser`, disc system, < 2 kLOC, upstream was on Google Code
- `/blip_buf`, multiple cores, < 2 kLOC, upstream was on Google Code
- `/libmupen64plus`, Mupen64Plus and plugins, < 340 kLOC, upstreams are under https://github.com/mupen64plus
- `/lynx`, Handy core, < 12 kLOC, upstream is in tarball (or https://github.com/TASEmulators/mednafen/tree/master/src/lynx)
- `/psx`, Octoshock core, < 68 kLOC, upstream is in tarball (or https://github.com/TASEmulators/mednafen/tree/master/src/psx)
- `/quicknes`, QuickNes core, < 19 kLOC, upstream was presumably on Google Code but it looks like https://github.com/kode54/QuickNES_Core was used
- `/waterbox/ares64/ares`, Ares64 core, < 55 kLOC, upstream is https://github.com/ares-emulator/ares
- `/waterbox/bsnescore/bsnes`, (new) BSNES core, < 67 kLOC, upstream is https://github.com/bsnes-emu/bsnes
- `/waterbox/gpgx`, Genplus-gx core, < 88 kLOC, upstream is https://github.com/ekeeke/Genesis-Plus-GX
- `/waterbox/libsnes`, (old) BSNES core, < 63 kLOC, upstream is https://github.com/bsnes-emu/bsnes
- `/waterbox/nyma/zlib`, zlib for Nyma cores, < 19 kLOC, upstream is https://github.com/madler/zlib
- `/waterbox/picodrive`, PicoDrive core, < 79 kLOC, upstream is https://github.com/notaz/picodrive
[...]

Issue #2046, opened 2020-05-25
Add support for compressed optical disk images


nattthebear:
These are nifty things.  PCSX2 supports some, as well as other emulators presumably.  I wanted to add this some time ago, but was told not to.

Spikestuff:
"Nifty things"
You mean the "beautiful" loads that are different from a clean image of the game?
You mean, even though [we have the perfect example as to avoid this](http://tasvideos.org/forum/viewtopic.php?t=21609), you want to add something for the general community even though that it's not the focus of the TASing community?

If this somehow screws with loads compared to a clean image then damn on this implementation.

nattthebear:
> You mean the "beautiful" loads that are different from a clean image of the game?

No, this is not about GCN images with parts missing or any similar capability.  This is about bin files compressed with something like gz, with the emulator building an index file automatically and decompressing on the fly.

see also #2245

CasualPokePlayer:
Kind of a dupe with https://github.com/TASEmulators/BizHawk/issues/1482?

nattthebear:
I'm not sure; are choade files compressed?

CasualPokePlayer:
They are (also the most popular generic disk compression format).

PR #4284, opened 2025-04-06
Adding PPSSPP as ported emulation core for Sony PSP

SergioMartin86:
Adding https://github.com/TASEmulators/ppsspp/tree/ported as new ported core in Bizhawk

- [x] Core compiling and running on Bizhawk
- [ ] Add proper cd callbacks
- [ ] Correctly handle shutdown
- [ ] Add save/load
- [ ] Add inputs
- [ ] Add memory domains

Check if completed:
- [ ] I, the committer, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2024-06-22) and am compliant


inconsistent-dg:
I get this error when trying to load a rom (both ISO and CHD)
image

Spikestuff:
As of - https://github.com/TASEmulators/BizHawk/pull/4284/commits/3d3a12ab1277de0124992c0f8c23f49077c3fdba
All issues that were initially in this post has been resolved (check comment history).

**Regardless, hell yeag on getting titles to work eien86.**

image

SergioMartin86:
> Second thing to note as above is related. Some titles can crash out in BizHawk. Example title being used is Jetpack Joyride (minis) where the game will be able to boot

This should be fixed now.

SergioMartin86:
The core is now mostly functional and I reduced the diff wrt  upstream the most I could. 

Still, there seems to be some games (Crash Tag Team Racing (US).iso) that fail to load. These games DO work on my headless player<https://github.com/SergioMartin86/headlessPPSSPP>, so no idea why putting this into bizhawk could be a problem.

I haven't figured out how NVRAM is saved so that is still pending.

.gitmodules
```diff
@@ -89,3 +89,12 @@
 	path = waterbox/dsda/core
 	url = https://github.com/TASEmulators/dsda-doom.git
 	branch = wbx
+[submodule "waterbox/ppsspp/ppsspp"]
+	path = ppsspp/ppsspp
+	url = https://github.com/TASEmulators/ppsspp.git
```
vadosnaprimer, line 94: Which branch?

SergioMartin86: [ported2](https://github.com/TASEmulators/ppsspp/tree/ported2)

vadosnaprimer: It's better to always keep hawk specific changes rebased on upstream branch that we rely on. And each time we pull upstream and rebase anew we would increment our branch version like this: wbx-1, wbx-2, etc, so every rebase is a new branch, because it basically changes history, which we want to preserve on every update. This is better than merging upsream changes because it makes our changes harder to find over time.

And yeah please explicitly put the branch into this file, that way it's clear where the code is coming from. Only pointing to commit can result in detached head and that commit can belong to different branches, making future maintenance more annoying.

Spikestuff:
I truncated out the issues [from my above post](https://github.com/TASEmulators/BizHawk/pull/4284#issuecomment-2791408518).
Just read the History for what's been fixed (good shit eien).

In terms of new issues (by the autobuilds here). Using [f1a8d12](https://github.com/TASEmulators/BizHawk/pull/4284/commits/f1a8d120a77de3c39e4e229812de18fd8878c2ce).
The major one besies the wrong font is being called compared to PPSSPP for some reason.
Is that the "™" is missing from "Memory Stick™":
image

Compared to PPSSPP:
image

And according to KAGE-008 at the very least this font issue affects Ghost in the Shell, Assassin's Creed Bloodlines, Crimson Room: Reverse, Jeanne D'Arc and Dissidia.

I was able to validate this missing text issue with Crimson Room: Reverse.
image

Another title having issues beside Crash Tag Team Racing is Xyanide Resurrection according to KAGE-008 as well.
You're able to boot into the game, but you land into a loading loop before the gameplay starts.

And you're currently unable to access PSP > Settings.

(At least all these are issues with autobuilds)

YoshiRulz:
Missing glyphs aside, could the difference in font metrics cause any problems? Would a scrollbar appear if the message was slightly longer?

.gitmodules
SergioMartin86: added

SergioMartin86:
Regarding unstability for some games. I suspect there is some memory corruption that causes issues on Bizhawk+Windows. When running the core on my own driver (https://github.com/SergioMartin86/headlessPPSSPP), it works perfectly for the games that crash

.gitmodules
vadosnaprimer: I can see the `wbx` branch in the repo now but not in `.gitmodules`...

SergioMartin86: Uh, this is strange. This core isn't wbxed; it's ported. This every must be a residual from when I tried to wbx it

vadosnaprimer: Ah LOL indeed. Then the usual name for it is `bizhawk`.

InfamousKnight:
Can someone link a prebuilt zip file for linux?

YoshiRulz:
https://github.com/TASEmulators/BizHawk/actions/runs/14476049776/artifacts/2950288002

InfamousKnight:
error

And it doesn't load star wars battlefront 2.

InfamousKnight:
Eh, having some stability issues over here with it. Weird problems like it not working on an x86_64 machine but then "sometimes" works on some windows emulator for android. Limited to gdi+. 

I'm assuming it's not wbx because of performance? As a taser, performance is of no concern unless it's running at like 1fps.

InfamousKnight:
I'm curios why this is not waterboxed. I know waterbox would affect performance, but I don't like tasing while worrying about desyncs.

SergioMartin86:
> I'm curios why this is not waterboxed. I know waterbox would affect performance, but I don't like tasing while worrying about desyncs.

The interpreter crashes when waterboxed. We don't know exactly what, but for now a waterboxed implementation is out of the question. The core seems to have solid load/save state procedures so I'm confident it can still be used ported directly.

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.ISettable.cs
```diff
@@ -0,0 +1,65 @@
+using BizHawk.Common;
+using BizHawk.Emulation.Common;
+
+namespace BizHawk.Emulation.Cores.Consoles.Sony.PSP
+{
+	public partial class PPSSPP : ISettable<PPSSPP.Settings, PPSSPP.SyncSettings>
```
YoshiRulz, line 6: If there are no settings, use `ISettable<object, ...>`, and likewise for sync settings, and if there are neither then I think you can just omit the interface.

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IStatable.cs
```diff
@@ -0,0 +1,56 @@
...
+			writer.Write(LagCount);
+			writer.Write(Frame);
+		}
+
+		public void LoadStateBinary(BinaryReader reader)
+		{
+
+			var stateLen = reader.ReadInt32();
+
+			if (stateLen > _stateBuf.Length)
```
YoshiRulz, line 39: ditto

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IStatable.cs
```diff
@@ -0,0 +1,56 @@
...
+
+		public bool AvoidRewind => true;
+
+		public void SaveStateBinary(BinaryWriter writer)
+		{
+
+			var stateLen = _libPPSSPP.GetStateSize();
+			writer.Write(stateLen);
+
+			if (stateLen > _stateBuf.Length)
```
YoshiRulz, line 19: Should this be `!=`? Under what circumstances could this happen?

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IVideoProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+	public partial class PPSSPP : IVideoProvider
+	{
+		static int BW = 480;
+		static int BH = 270;
+		internal bool VideoDirty;
+
+		protected readonly IntPtr _context;
+
+		public PPSSPP(LibPPSSPP core, IntPtr context)
```
YoshiRulz, line 13: `core` argument is ignored? This constructor doesn't seem to be used anyway.

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.cs
```diff
@@ -0,0 +1,178 @@
...
+			// Setting CD callbacks
+			_libPPSSPP.SetCdCallbacks(_CDReadCallback, _CDSectorCountCallback);
+
+			//// Pre-loading emulator resources
+
+			// Getting compat.ini -- required to set game-specific compatibility flags
+			string resourceName = "";
+
+			resourceName = "compat.ini";
+			var compatIniData = new MemoryStream(Resources.PPSSPP_COMPAT_INI.Value).ToArray();
```
YoshiRulz, line 72: Would this work for these?
```suggestion
			var compatIniData = Resources.PPSSPP_COMPAT_INI.Value.ReadAllBytes();
```

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.cs
```diff
@@ -0,0 +1,178 @@
...
+			_syncSettings = lp.SyncSettings ?? new();
+			ControllerDefinition = CreateControllerDefinition(_syncSettings, _isMultidisc);
+
+			// Registering service provider
+			_serviceProvider = new(this);
+			_serviceProvider.Register<IVideoProvider>(this);
+			_serviceProvider.Register<ISoundProvider>(this);
+
+			// Loading LibPPSSPP
+			_resolver?.Dispose();
```
YoshiRulz, line 55: This is the constructor—this field is definitely `null` here.

YoshiRulz (changes requested):

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.ISettable.cs
SergioMartin86: There are no settings now, but there may be in the future. For this reason, I think it's better to preserve the structure to allow these, even if currently empty.

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.cs
SergioMartin86: ha, yes, removed

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.cs
SergioMartin86: unfortunately it didnt (compilation error)

Severity	Code	Description	Project	File	Line	Suppression State	Details
Error (active)	CS1061	'byte[]' does not contain a definition for 'ReadAllBytes' and no accessible extension method 'ReadAllBytes' accepting a first argument of type 'byte[]' could be found (are you missing a using directive or an assembly reference?)	BizHawk.Emulation.Cores	C:\Users\sergiom\Documents\BizHawk\src\BizHawk.Emulation.Cores\Consoles\Sony\PSP\PPSSPP.cs	71

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IVideoProvider.cs
SergioMartin86: removed constructor altogether

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IStatable.cs
SergioMartin86: fixed

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IStatable.cs
SergioMartin86: fixed

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.cs
YoshiRulz: If it's already a byte array, you don't need to `ReadAllBytes` -_-

edit: You might want to copy it, which would be just the `.ToArray()` without the stream, since `loadResource` is unmanaged and might mutate its argument.

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.ISettable.cs
YoshiRulz: Preserve the copy-pasted boilerplate? When you don't plan on touching the core for a while?

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.cs
SergioMartin86: fixed now. The whole MemoryStream wasn't necessary

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.ISettable.cs
SergioMartin86: removed

YoshiRulz:
Linux build doesn't work (`InvalidOperationException: got null pointer from dlopen, error: dlerror reported no error`).
```
$> ldd output/dll/libppsspp.so 
        linux-vdso.so.1 (0x00007fa7b2076000)
        libavcodec.so.57 => not found
        libavformat.so.57 => not found
        libavutil.so.55 => not found
        libswresample.so.2 => not found
        libswscale.so.4 => not found
        libSDL2-2.0.so.0 => not found
        libstdc++.so.6 => not found
        libm.so.6 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libm.so.6 (0x00007fa7b0317000)
        libgcc_s.so.1 => /nix/store/4y1jj6cwvslmfh1bzkhbvhx77az6yf00-xgcc-14.2.1.20250322-libgcc/lib/libgcc_s.so.1 (0x00007fa7b203e000)
        libc.so.6 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libc.so.6 (0x00007fa7b0000000)
        /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib64/ld-linux-x86-64.so.2 (0x00007fa7b2078000)
```

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IEmulator.cs
```diff
@@ -0,0 +1,120 @@
...
+		public void Dispose()
+		{
+			_libPPSSPP.Deinit();
+
+			if (_disposed)
+			{
+				return;
+			}
+
+			_disposed = true;
```
Morilli, line 114: I don't think this code does anything. Did you mean to put the disposed check at the top?

Also because this core is using disc assets, you need to dispose them here too.

SergioMartin86:
> Linux build doesn't work (`InvalidOperationException: got null pointer from dlopen, error: dlerror reported no error`).
> 
> ```
> $> ldd output/dll/libppsspp.so 
>         linux-vdso.so.1 (0x00007fa7b2076000)
>         libavcodec.so.57 => not found
>         libavformat.so.57 => not found
>         libavutil.so.55 => not found
>         libswresample.so.2 => not found
>         libswscale.so.4 => not found
>         libSDL2-2.0.so.0 => not found
>         libstdc++.so.6 => not found
>         libm.so.6 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libm.so.6 (0x00007fa7b0317000)
>         libgcc_s.so.1 => /nix/store/4y1jj6cwvslmfh1bzkhbvhx77az6yf00-xgcc-14.2.1.20250322-libgcc/lib/libgcc_s.so.1 (0x00007fa7b203e000)
>         libc.so.6 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libc.so.6 (0x00007fa7b0000000)
>         /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib64/ld-linux-x86-64.so.2 (0x00007fa7b2078000)
> ```

Fixed now for me
image

SergioMartin86:
> Is there a reason the linux version depends on the entirety of ffmpeg and the windows version... doesn't?

both depend on ffmpeg -- in windows the prebuilt files are provided by ppsspp; in Linux, only the statically compiled libs, so I had to build the relocatable ones.

src/BizHawk.Emulation.Cores/Consoles/Sony/PSP/PPSSPP.IEmulator.cs
SergioMartin86: I believe this is fixed now. Added disposing of disc data too

SergioMartin86:
After some testing with other games and re-recording extensibly wit TAStudio, seems like this port is more stable than I previously thought. It works correctly as ported for Linux and Windows. 

I believe this could be even released normally (i.e., not as experimental), if we can produce some movies and have them peer-synced

InfamousKnight:
Tested midnight club 3 and this showed up:

DB: hash 391844022ABD2FEBA85CDF48FD7543FD not in game database.
In PPSSP Core
[CD] Sector count: 81584

And opening the zip file just crashes bizhawk. I had to extract the zip to even get that far.

Assets/dll/libavcodec.so.57
```diff
@@ -0,0 +1 @@
+../../Assets/dll/libavcodec.so
```
YoshiRulz: Why symlink? You control the filename.

YoshiRulz (commented):
You missed `ln -s libSDL2-2.0.so libSDL2-2.0.so.0`, but otherwise it does load now.
Unfortunately I think I have a corrupted/bad dump T_T the disc indicator is blinking but it's stuck on a black screen.

YoshiRulz:
@InfamousKnight This branch doesn't add to the gamedb, and you can't load discs from archives, see #2046.

Assets/dll/libavcodec.so.57
SergioMartin86: This is because libavcodec.so is what we want to version control (even though it's binary, at least we'd know consistently when it was changed and by whom), whereas the .57 is the particular version(?) the linker decided it wanted to look for. The symlink is just to redirect it to the actual lib

YoshiRulz:
> After some testing with other games and re-recording extensibly wit TAStudio, seems like this port is more stable than I previously thought. It works correctly as ported for Linux and Windows.

Does this mean
> Find out why some games crash on load (whereas they don't for https://github.com/SergioMartin86/headlessPPSSPP)

was resolved? Or just that you haven't found any sync problems with games which do load?

SergioMartin86:
> > After some testing with other games and re-recording extensibly wit TAStudio, seems like this port is more stable than I previously thought. It works correctly as ported for Linux and Windows.
> 
> Does this mean
> 
> > Find out why some games crash on load (whereas they don't for https://github.com/SergioMartin86/headlessPPSSPP)
> 
> was resolved? Or just that you haven't found any sync problems with games which do load?

As far as I tested, all the games on my test bench work right and don't suffer from sync issues. 

Only exception: Crash Tag Team Racing (US).iso which gets stuck reading the disc on startup.

Spikestuff:
Alright, let's see what's different... and new things to mention.

Refer to the other two, which have a list of things that I have been editing if something has been corrected.
- https://github.com/TASEmulators/BizHawk/pull/4284#issuecomment-2791408518
- https://github.com/TASEmulators/BizHawk/pull/4284#issuecomment-2799913953
Everything in the second one still stands.

A new one is when loading the iso and having a cso in the same folder it'll notice the cso and complain about it.
image

Oddity: Much like with encore there is a first boot desync type thing going on when going TAStudio -> bk2.
When I did stumble on this, all I had to do was add a frame that corrected it.
This uses _Namco Museum Battle Collection (USA)_ as the test game.
The fail file was the fresh experience of it so what's typical to stumble on.
The "fix" made a frame adjustment after stumbling into this oddity.
[NMBC.zip](https://github.com/user-attachments/files/21705694/NMBC.zip)

There's this issue which **you can only avoid by going with frame advance**. For some reason in certain titles like _Namco Museum Battle Collection_ or _The Basics_ the emulator will flash green for a frame. This isn't present in PPSSPP.
greenflash
eien is using the Software Renderer.

kage2051:
> As pointed out by KAGE-008, we don't know where the games are saved on our system.

Took me about 10 minutes to figure it out. The folder with all save data is in C: or whichever your main drive is, in the root dir. To sync you might want to remove all savedata before proceeding, or backing up your own savedata if it was already started with it.

And yes, sync wise this core is way more stable than what I initially thought. Yesterday I did a 15 minute test of The 3rd Birthday with successful results. Same with a 5 minute test of Lumines, a game which supposedly gives 'random' blocks.

InfamousKnight:
Screenshot 2025-08-10 11 46 12 AM

This is what happens when I try to open the 3rd birthday... 

chromebook using crostini. I can load everything else but psp games.. I have all dependencies installed: mono, libc6-dev, libsdl2-dev, libopenal1 and ffmpeg.

Maybe debian bookworm has an issue with it? that's what the default Chromebook Linux subsystem uses

Guerrerosp:
Most of the games that I tested worked fine, the exceptions are Tekken DR, Assassins Creed, God of War, they don't present any error but simply stuck on a black screen with the frames running. Sillent hill origins I get a black screen after the start menu.
The only thing missing is the rewind function, I don't know if it's disabled intentionally or I messed up something
I did a 10 minutes movie test on Avatar and did not desync once

Assets/dll/libSDL2-2.0.so
Morilli, line 1: We already have `libSDL2.so`, can't you just use that? Probably need to `patchelf` the binary but I'd rather not have two SDL libs shipped.

Morilli:
Can the `ppsspp` folder be in *any* subfolder (submodules, ExternalProjects, ExternalCoreProjects) that isn't root? I would like to move away from polluting the BizHawk root folder with random core folders.

YoshiRulz (changes requested):
Don't write SaveRAM to `C:\`, figure out something like Encore.

Also move `/ppsspp` to `/ExternalCoreProjects/ppsspp` like Morilli said (re: #3818).

edit: At https://github.com/TASEmulators/BizHawk/pull/4284/commits/45935d6c685f195189e7837c40cc8ea0c647f237, it's giving "got null pointer from dlopen" again on Linux, but I don't want to block this any further.

YoshiRulz:
~~Almost!~~ edit: Morilli is right, this I think is pulled from Mono's closure? `ldd ${mono}/bin/mono` --> `libstdc++.so.6 => /nix/store/xvzz97yk73hw03v5dhhz3j47ggwf1yq1-gcc-13.2.0-lib/lib/../lib64/libstdc++.so.6`
```
$> LD_LIBRARY_PATH="$PWD/dll:$PWD:$LD_LIBRARY_PATH" ldd dll/libppsspp.so 
        linux-vdso.so.1 (0x00007f0637c31000)
        libSDL2.so => /home/yoshi/Documents/BizHawk/output/dll/libSDL2.so (0x00007f063663e000)
        libstdc++.so.6 => not found
        libm.so.6 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libm.so.6 (0x00007f0637b40000)
        libgcc_s.so.1 => /nix/store/4y1jj6cwvslmfh1bzkhbvhx77az6yf00-xgcc-14.2.1.20250322-libgcc/lib/libgcc_s.so.1 (0x00007f0637b12000)
        libc.so.6 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libc.so.6 (0x00007f0636400000)
        /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib64/ld-linux-x86-64.so.2 (0x00007f0637c33000)
        libdl.so.2 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libdl.so.2 (0x00007f0637b0b000)
        libpthread.so.0 => /nix/store/zdpby3l6azi78sl83cpad2qjpfj25aqx-glibc-2.40-66/lib/libpthread.so.0 (0x00007f0637b06000)
$> ls dll/libstdc++*
dll/libstdc++-6.dll
```

Morilli:
I think that's correct. A lot of current .so's rely on `libstdc++.so.6`.

ppsspp/SDL
Morilli, line 1: Is this used? I briefly checked and didn't see any reference.

InfamousKnight:
```
infamousbytes@penguin:~/BizHawk-dev-linux/dll$ ldd libppsspp.so
./libppsspp.so: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by ./libppsspp.so)
./libppsspp.so: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./libppsspp.so)
./libppsspp.so: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.31' not found (required by ./libppsspp.so)
./libppsspp.so: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by ./libppsspp.so)
./libppsspp.so: /lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.15' not found (required by ./libppsspp.so)
```

Still no luck with getting it to work on Chromebook Linux subsystem..

Any ideas? Sorry if I'm being a burden..

Morilli:
> Still no luck with getting it to work on Chromebook Linux subsystem

The .so was built with GLIBC version 2.42 so if yours is lower you (or someone else) will have to rebuild. I don't have a system with lower GLIBC right now so I can't do it.

InfamousKnight:
Finally got it to work! I updated debian 12 to 13.

Tested midnight club 3 and its a black screen...

And input seems to do nothing... I did configure the controls, but even though input osd is showing, its not really applying to the game.

Spikestuff:
\> Tested midnight club 3 and its a black screen...
Known issue. That game is above 1,500,000,000 bytes.

**Edit:** Yes, input works otherwise I'd point that out as an issue. No idea what you're doing wrong.

InfamousKnight:
Is input working for you spikestuff in the latest build?

InfamousKnight:
Tried again with latest build and input still doesn't register on Linux. In the osd, inputs still show. And L and R show up as !.  But they don't apply to the game. Tested star wars battlefront 2, the 3rd birthday, among others.

src/BizHawk.Emulation.Cores/Resources/ppsspp_compatvr.ini
```diff
@@ -0,0 +1,1376 @@
...
+# The uses cases are strict:
+#   * Enable fixes for things we can't reasonably emulate without completely ruining
+#     performance for other games, such as the screen copies in Dangan Ronpa
+#   * Disabling accuracy features like 16-bit depth rounding, when we can't seem to
+#     implement them at all in a 100% compatible way
+#   * Emergency game-specific compatibility fixes before releases, such as the GTA
+#     music problem where every attempted fix has reduced compatibility with other games
+#   * Enable "unsafe" performance optimizations that some games can tolerate and
+#     others cannot. We do not currently have any of those.
+#   * This file is only for VR (virtual reality). For non-VR purposes, use compat.ini.
```
YoshiRulz, line 17: ???

InfamousKnight:
Updates over here: can confirm the input issue is Linux only. I used wine to test input and it works there. Despite L and R showing as !, it still registers.

Midnight club la remix has a black screen too and that game is less than 600 mb. So I don't think file size is the issue here.

Also, is there a way to map analog controls to keyboard inputs? Like 'I' key is analog up? And 'L' key is analog left and so forth? My gamepad controller doesn't work...

Spikestuff:
> Midnight club la remix has a black screen too and that game is less than 600 mb. So I don't think file size is the issue here.

It's more of the issue that you got a bad dump. L.A. Remix is 894MB.
image

Spikestuff:
Just linking Part 2 & 3. Part 1 is essentially "done" it's just the larger filesize issue, which, whatever.
- https://github.com/TASEmulators/BizHawk/pull/4284#issuecomment-2799913953
- https://github.com/TASEmulators/BizHawk/pull/4284#issuecomment-3172419527

I'm making this as the unique post, as this doesn't make sense to things that we're aware of, because BizHawk freezes in this instance, and the only logical reasonings is the missing text, but as we established games still work... _right?_

Exit (USA/Europe) locks up failing to display the tutorial text, as what's below is what we should be getting.
Exit (USA/Europe)

Exit (Japan/Korea) locks up just before that... displaying the level text instead, again this is what we should be getting.
Exit (Japan/Korea)

But what I gathered from an RTA run from 5 years ago. Is that this shouldn't prevent the game from running.
Exit (USA/Europe)
Exit (Japan/Korea)
This is PPSSPP v1.8 (2019), which is able to deal with it-- with a different font, but it's able to deal with it.

Theoretically, Hawk shouldn't have this issue. But it unexpectedly locks up instead.
image

---

Next is Mercury Meltdown (USA/Europe).
This game refuses to boot, and the only thing that this can possibly be is Hawk failing to load the intro FMVs _somehow_.
image
This is well below the "large filesize" issue, so the only answer this points to is again, it failing to load the very first FMV.

dancrimpirate:
great ! now go for gamedb .
it's required for homebrew apps and especially pspkvm the j2me emulator

InfamousKnight:
And sync test update on midnight club la remix, tested 10 minutes of footage and it syncs! Dumping avi video works too and doesnt desync.

If it shows a black screen, turn off rewind, advance to next screen, and you can turn it back on.

I was surprised how fast it was despite rewind and using some low spec hardware. 

Although its unfinished, I'm still happy with this.

SergioMartin86:
CPP has resolved the game loading error (summary: it was bk's logic for loading discs that was incompatible with larger formats.)

Updated with upstream PPSSPP -- this core is up to date. Additionally, it has minimal changes so future reupstreaming should pretty straightforward

SergioMartin86:
reverting re-upstream as it breaks some games

Spikestuff:
> reverting re-upstream as it breaks some games

Just for context. It seems to be titles that are doing initial memory card checks and can't get past that.
Unless your name is Mercury Meltdown, that game is just broken, and still no idea why.

Just to showcase two titles that are outright explicitly saying it's that portion, where it previously worked (in Guilty Gear's case):

githubc00l:
received an error trying to use recording mode in TAStudio:

```
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at BizHawk.Client.Common.TasMovieChangeLog.MergeLastActions() in /src/BizHawk.Client.Common/movie/tasproj/TasMovie.History.cs:line 196
   at BizHawk.Client.EmuHawk.TAStudio.FastUpdateAfter() in /src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs:line 119
   at BizHawk.Client.EmuHawk.TAStudio.UpdateAfter() in /src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs:line 109
   at BizHawk.Client.EmuHawk.ToolManager.UpdateToolsAfter() in /src/BizHawk.Client.EmuHawk/tools/ToolManager.cs:line 769
   at BizHawk.Client.EmuHawk.MainForm.StepRunLoop_Core(Boolean force) in /src/BizHawk.Client.EmuHawk/MainForm.cs:line 3121
   at BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop() in /src/BizHawk.Client.EmuHawk/MainForm.cs:line 973
   at BizHawk.Client.EmuHawk.Program.SubMain(String[] args) in /src/BizHawk.Client.EmuHawk/Program.cs:line 377 
```
edit: opened #4439

Spikestuff:
[False.](https://github.com/TASEmulators/BizHawk/pull/4284#issuecomment-3245911097) Unfortunately.
Even before latest build there were issues here and that shined bright with Tekken 6.

I will immediately clarify out of the gate that [this TAS exists](https://youtu.be/k8Is3UAyYIQ) only due to the fact that I didn't clear Greenzone after spotting this issue very early.
Now originally back on https://github.com/TASEmulators/BizHawk/pull/4284/commits/8957808cb2c84482e1a6ceb0fca33aef2f48b898 (unless it was earlier) we had a form of the consistency, in the inconsistency, to explain I will use the original text I was going to have here and then pivot into what it's devolved to.

Do note that the tasproj file referred to is lost, and is technically not required as things have gotten worse (which I'll get into).

Spikestuff - 17th August
> Now here's an interesting one where I'm going to provide the tasproj to, and inform that converting this as a bk2, or playing it without the piano roll, will cause a desync.
> So the game is Tekken 6 (USA), what was *meant* to happen is that Yoshimitsu is *meant* to take on Devil Jin first. Ok, we agree here right? Ok, we're all wrong.
> When the Greenzone is cleared, the first fight is actually Kazuya... most of the time. You can still *somehow* generate a different character like Nina, or Armor King, or Bryan.
> 
> So what about the bk2 (or played without piano roll)? After all I did mention it will cause a desync.
> bk2 doesn't even get into the character select screen. So there isn't a consistent way of knowing who you're facing first, and there kinda doesn't seem like there's logic to this.
> 
> As for creating from a bk2, at face value, there doesn't seem to be an issue here.

At least back then, we had a form of consistency where we were able to consistently get Kazuya as our first opponent to fight after clearing Greenzone, and doing everything that could be done. It was Kazuya no matter what. The bk2 failed, it will get to the main menu with the wrong character and get stuck there, whatever no big deal compared to what is going to be mentioned.

Since then [I have a new tasproj file that I'm providing](https://github.com/user-attachments/files/22279141/tekken6.zip) which is using the latest build at the time of writing. Like the previous one before it, it mentions within the Markers what steps are being done and the Greenzone has been removed.

However, compared to the previous one's Markers I am unable to tell you who the first fight is against, cause it's wildly inconsistent when Greenzone is cleared, the bright news is bk2 syncs with the exact same issue.

When Yoshimitsu is selected. I am getting every valid character in the Tekken roster as Yoshi's opponent.
What I originally wrote on the 17th of August is gone. This threw it all out the window. There is absolutely no logic.

**It is now extremely inconsistent** and you cannot make a TAS with this anymore from start to end with a cleared Greenzone tasproj file (cause remember as outlined in the original comment bk2 failed outright, comparatively to where it works to a point now).

---

Alright. Now that a major problem has been pointed out, let's switch to lighter news.

RIIIIIIIIIIIIIIIIDGE RACER has graphical glitches.
The level is Sunset Drive, and pay attention to the scenery here on BizHawk:
image

And here on PPSSPP:
image

And now the second one, we have model glitches, which reminds me of older PPSSPP which used to struggle with _Tekken: Dark Resurrection_.

To make it obvious and to the point I chose Jack-5 because damn he's big and it's so obvious on him.

Here's the error in BizHawk:
Tekken_-_Dark_Resurrection_USA_EnFrDeEsIt 2025-09-11_16 28 27

And here's PPSSPP being fine with it:
ULUS10139_00000
(Yes there's two Tekken games on PSP. This one doesn't seem to have the desync issues which makes it more interesting/annoying.)

---

**Critical Edit**: @YoshiRulz can you quit changing the way I type things out.
You do it on everything I type and it's a very annoying overreach that you have.

Spikestuff:
> The only workaround I found is to create a save state in the main menu while recording the inputs.

How did you miss the part where I mentioned that it was previously **somewhat consistent** in the tasproj version compared to now?

---

**Edit:** And they deleted their comment... keeping up the section I quote them on cause creating a different takeaway is insane.

InfamousKnight:
Tested latest build and input still doesn't register on linux... just sucks I have to use wine... it adds so much overhead..

Is anyone else having that issue with linux? I'm using linux mint zara cinnamon

YoshiRulz:
I can't even load the core. I think this is a different error than I got last time though:
```
System.InvalidOperationException: got null pointer from dlopen, error: /nix/store/xvzz97yk73hw03v5dhhz3j47ggwf1yq1-gcc-13.2.0-lib/lib/../lib64/libstdc++.so.6: version `CXXABI_1.3.15' not found (required by /home/yoshi/Documents/BizHawk/output/dll/libppsspp.so)
  at BizHawk.Common.OSTailoredCode+LinuxLLManager.LoadOrThrow (System.String dllToLoad) [0x00026] in <fda7fbf47001476aa396e40d89ba93b6>:0 
  at BizHawk.Common.DynamicLibraryImportResolver..ctor (System.String dllName, System.Boolean hasLimitedLifetime) [0x0000e] in <fda7fbf47001476aa396e40d89ba93b6>:0 
  at BizHawk.Emulation.Cores.Consoles.Sony.PSP.PPSSPP..ctor (BizHawk.Emulation.Cores.CoreLoadParameters`2[TSettings,TSyncS] lp) [0x001a5] in <d72ea0d4cec24692bf9bacdef04beadc>:0 
```

InfamousKnight:
And the first desync I encountered: star wars battlefront 2. When I selected mos eisley assault mode, and then selected villains when the team select showed up, it didnt select villains on playback. I was going to do a jango fett run where he does all the kills and doesn't take damage and no missed shots. Would have been really cool to see.

I haven't tested thoroughly to see a good way to reproduce it.

InfamousKnight:
And ratchet and clank size matters:

Good performance. I'm getting 25+ fps. Sometimes over 40 fps.

Sync stable: yes

But it does have some weird requirements. You can't have rewind turned on on booting the game up. After you close save game, you have to turn off rewind until you reach control of ratchet. Do not have rewind enabled during cutscenes. So a lot of switching around if you want to use rewind.

InfamousKnight:
[Edit] nevermind. Most of my performance issues stem from having to use wine. Im sure once the linux build is fixed, its gonna make a big difference

I dont think removing threads was necessary:  https://github.com/hrydgard/ppsspp/issues/19404

Bringing back the threads would improve performance and probably wouldn't affect sync stability.

InfamousKnight:
*bump* upstream ppsspp has llvmpipe support.

From my understanding, llvmpipe Is a requirement for waterbox?

phoenix1291:
---  UPDATE ---
The message below is obsolete. I found the link for the version that works for PPSSPP [https://nightly.link/SergioMartin86/BizHawk/workflows/ci/ppsspp?preview](url).
It would be good to put the link at the top of the thread so that other people don't make the same mistake I did. 

----------------------------------------------------------------------------------------------------------------------------------------

Is it supposed to work on Windows?
I haven't been able to load a game.
I'm using the latest dev version of Bizhawk.

Screen Shot 12-04-25 at 03 36 PM

Spikestuff:
> I'm using the latest dev version of Bizhawk.

Wrong version. Use the link at the very top of this pull request.
It says (again at the very top):
[dev build for branch | SergioMartin86:ppsspp](https://nightly.link/SergioMartin86/BizHawk/workflows/ci/ppsspp?preview)

SergioMartin86:
This core is ready to be merged as experimental. It provides a minimal playable experience that gives RTA runners, glitch hunters, and casual players re-recording tools not available in the upstream emulator. 

I think the only hope to get this to submission-ready is to release it now as-is and hope for potential interested devs to jump in and polish the missing details I can't.

ppsspp/make/Makefile
```diff
@@ -0,0 +1,868 @@
...
+# ifeq ($(UNAME), Linux)
+# 	TARGET = libppsspp.so
+# else
+# 	TARGET = libppsspp.dll
+# endif
+
+LDFLAGS = -shared -s $(CXXFLAGS)
+
+DEST = ../../Assets/dll
+DESTCOPY = ../../output/dll
```
YoshiRulz: `git mv ppsspp ExternalCoreProjects` then update paths here

SergioMartin86: done

YoshiRulz:
Will merge tomorrow if nothing else comes up.

SergioMartin86:
> Will merge tomorrow if nothing else comes up.

ping :)

YoshiRulz:
I definitely didn't bookmark this in the wrong folder and forget. Couldn't be me.

Issue #4275, opened 2025-03-18
Verify downloaded binaries


YoshiRulz:
As of https://github.com/TASEmulators/BizHawk/commit/e0974e3c3092bd5f0027983ef863a7909f9c8d2f, the checksums for the FFmpeg binaries in [the ffmpeg-binaries repo](https://github.com/TASEmulators/ffmpeg-binaries) are hardcoded in `FFmpegService`. That will be checked before FFmpeg is ever executed.

---

As of the same commit, the RAIntegration download no longer blindly follows any address, it has to resolve to `retroachievements.org`.
RA's API seems to be down at the moment so I can't check if that includes a checksum, but even if it did, it would need to be signed as well (and we hardcode the pubkey).
Though as CPP points out below, the downloader isn't available out-of-the-box, so we don't have to worry about the endpoint coming back online and pushing broken (or malicious) updates.

CasualPokePlayer:
Also, the user is told the exact url which the download will be occurring on. If the user does not trust that URL, they can exit out of the form, and no download will take place.

vadosnaprimer:
I don't know what even needs to happen to get problems that this issue is about to manifest themselves. What are the regular user actions that require this "fix" so things don't blow up?

YoshiRulz:
Sorry, I forgot to come back and update the OP. I believe that I resolved everything re: FFmpeg, and all that's left is to secure RAIntegration.
For a user to get hacked via that downloader currently, they would have to manually install RAIntegration, and *then* get targeted by DNS poisoning or something. So we don't need to worry about it.
I'm leaving this open until either the RAIntegration downloader is confirmed to work or we decide to remove it.

PR #4287, opened 2025-04-10
Downloader dedup

YoshiRulz:
Does enough for #4275 that I'd be comfortable putting out a release with it.


src/BizHawk.Client.EmuHawk/DownloaderForm.cs
```diff
@@ -0,0 +1,191 @@
...
+					//last chance. exiting, don't dump the new file
+					if (_exiting) return;
+					exStream.CopyTo(fs);
+					fs.Position = 0L;
+					if (!PreChmodCheck(fs)) throw new Exception("download failed (pre-chmod validation)");
+					fs.Dispose();
+					if (OSTailoredCode.IsUnixHost)
+					{
+						OSTailoredCode.ConstructSubshell("chmod", $"+x {DownloadTo}", checkStdout: false).Start();
+						Thread.Sleep(50); // Linux I/O flush idk
```
Morilli, line 113: I realize this is mostly copy-pasted code but shouldn't this idelly be `Process.WaitForExit()` instead of an arbitrary sleep?

YoshiRulz:
@CasualPokePlayer This is blocking a release candidate.
edit: Ploughing ahead. I did say I tested FFmpeg, and per https://github.com/TASEmulators/BizHawk/issues/4275#issuecomment-2733955137 RAIntegration isn't used by default, plus the endpoint is still down.

PR #4326, opened 2025-05-20
Bump .NET SDK to 10

### LINKED ISSUE #4097: Migrate solution files to `.slnx`
This is a new format [only recently merged](https://github.com/dotnet/msbuild/commit/adb4394981ec9c44f59dfa9fbc903670eddf6be6) into the CLI and not supported by stable VS2022.
edit 2025-11: It's in stable VS2026 and possibly VS2022, and also Rider and VS Code.

see also https://gitlab.com/YoshiRulz/yoshis-hawk-thoughts/-/issues/216

YoshiRulz:
This should be delayed until at least 2026-01 so people have time to update their toolchains.

Resolves #4097.


ExternalProjects/BizHawk.Analyzer/BrokenCollectionExpressionAnalyzer.cs
Morilli, line 1: Have you tested whether this is even still necessary? I feel like I remember that `[ ]` for `ArraySegment`s was changed to now resolve to a useful (non-null) value, but I haven't checked.

YoshiRulz: If you revert https://github.com/TASEmulators/BizHawk/commit/83d04b0481ed23f524babd4ef671b3ae20deb22d and invoke with `-p:TestProjTargetFrameworkOverride=net8.0`, a couple Lua unit tests fail. That's still the case after updating to .NET 10 (passing `-p:TestProjTargetFrameworkOverride=net10.0`, and you'll also need to revert the `Lock` alias in `MainSlnCommon.props` because ¯\\\_(ツ)\_/¯).

I tried this as a sanity check, but it passes regardless of `TestProjTargetFrameworkOverride`, so the failure must be elsewhere in the BCL.
```csharp
ArraySegment<int> a = [ ];
Assert.IsNull(a.Array);
```

`ArraySegment<T>.Empty` is .NET Standard 2.1 or later, but it might be polyfillable now with extension props.

ExternalProjects/BizHawk.Analyzer/BrokenCollectionExpressionAnalyzer.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.Collections.Immutable;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class BrokenCollectionExpressionAnalyzer : DiagnosticAnalyzer
+{
+	private static readonly DiagnosticDescriptor DiagBrokenCollectionExpression = new(
+		id: "BHI1234",
+		title: "don't this",
+		messageFormat: "don't this",
```
YoshiRulz, line 11: TODO these were obviously meant to be placeholders

BizHawk.slnx
kasperk81, line 1: fyi: VS2026 supports `DefaultStartup="true"` in `<Project` element (in 2022 it's a no-op).

BizHawk.slnx
```diff
@@ -0,0 +1,24 @@
...
+	<Project Path="src/BizHawk.BizInvoke/BizHawk.BizInvoke.csproj" />
+	<Project Path="src/BizHawk.Bizware.Audio/BizHawk.Bizware.Audio.csproj" />
+	<Project Path="src/BizHawk.Bizware.Graphics.Controls/BizHawk.Bizware.Graphics.Controls.csproj" />
+	<Project Path="src/BizHawk.Bizware.Graphics/BizHawk.Bizware.Graphics.csproj" />
+	<Project Path="src/BizHawk.Bizware.Input/BizHawk.Bizware.Input.csproj" />
+	<Project Path="src/BizHawk.Common/BizHawk.Common.csproj" />
+	<Folder Name="/Client/">
+		<Project Path="src/BizHawk.Client.Common/BizHawk.Client.Common.csproj" />
+		<Project Path="src/BizHawk.Client.DiscoHawk/BizHawk.Client.DiscoHawk.csproj" />
+		<Project Path="src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj" />
```
YoshiRulz, line 11:
```suggestion
		<Project Path="src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj" DefaultStartup="true" />
```
[Apparently](https://www.meziantou.net/visual-studio-tips-and-tricks-default-startup-project.htm) changes which is shown in VS2026's launch toolbar when a dev first loads the solution.

Morilli: This does seem to work on VS2026, but only if the solution has't been opened before.

PR #3564, opened 2023-02-17
Make nix buildable from flake (sandbox enabled)

OmnipotentEntity:
Updated the nix files so that they build in a sandboxed environment, such as a nix flake.  An example of such a flake (that you may use to test this) is as follows:

```
{
  inputs = {
    flakeUtils.url = "github:numtide/flake-utils";
    nixpkgs.url = "github:nixos/nixpkgs/release-22.05";
    bizhawk.url = "github:OmnipotentEntity/BizHawk/master"; 
    bizhawk.flake = false;
  };
  outputs = { self, ... } @ inputs:
    inputs.flakeUtils.lib.eachSystem [ "x86_64-linux" ] (system:
      let
        pkgs = import inputs.nixpkgs {
          inherit system;
          config.allowUnfree = true;
        };
        bizhawk = import inputs.bizhawk {
          forNixOS = true;
          pkgs = pkgs;
          mono = pkgs.mono;
          lua = pkgs.lua5_4;
        };

      in rec {
        devShells = {
          # The activation script can be used as dev-shell
          default = pkgs.mkShell {
            buildInputs = with pkgs; [
              bizhawk.emuhawk
              bizhawk.discohawk
              gradle
              jdk11
              gtk2-x11
            ];
          };
        };

        devShell = devShells.default;
      }
    );
}
```

Tested as functional under NixOS.  In order to do this I had to:

1. Add a few packages to the main Nix file (such as speexdsp)
2. Change how nuget files were referenced (`name` was depreciated)
3. Update a handful of packages that had become stale (specifically OpenTK)
4. Remove references to nixpkgs's lib by simply porting the functions over into the file.

[//]: # "Apart from the mandatory license signature, these tasks are optional, but doing them could save reviewers some time and get the PR merged sooner."
Check if completed:
- [ ] I have run any relevant test suites
- [x] I, the committer, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2022-07-15) and am compliant

PR #4328, opened 2025-05-22
Add flake.nix

SignalWalker:
Adds a `flake.nix` (and its attendant `flake.lock`), containing outputs for all of the packages and devshells output by `default.nix` and `shell.nix`, an overlay output, and an `apps` output (to make it possible to run emuhawk with just `nix run github:TASEmulators/BizHawk`).

Notes:
- ~~The flake description is just copied straight from the github repo short description~~
- I set the `nixpkgs` input to match the default already present in `default.nix`
- ~~I also set a formatter output (which tells `nix fmt` what formatter to use), which may or may not be desired~~
- The default package and app are `emuhawk-latest-bin`, and the default devshell is `emuhawk-latest`, because that made sense to me, but it could also make sense to have them all be `emuhawk-latest`
- ~~The `apps` output may not be completely correct -- it works for the default, but I haven't checked that the generated program path is accurate for all packages~~
- `nix flake show` acts weird if you don't set `--option allow-import-from-derivation true`

Check if completed:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -174,9 +174,9 @@ in {
 		IDEs = {
 			kate = [ kate omnisharp-roslyn ];
 		};
-		shellHook = drv: ''
+		shellHook = isFlake: drv: ''
 			export BIZHAWKBUILD_HOME='${builtins.toString ./.}'
-			export BIZHAWK_HOME="$BIZHAWKBUILD_HOME/output/"
+			export BIZHAWK_HOME="${if isFlake then "$(pwd -P)" else "$BIZHAWKBUILD_HOME"}/output/"
 			ldLibPath='${lib.makeLibraryPath drv.buildInputs}' # for running tests
 			if [ -z "$LD_LIBRARY_PATH" ]; then
 				export LD_LIBRARY_PATH="$ldLibPath"
--- a/default.nix
+++ b/default.nix
@@ -150,9 +150,11 @@ in {
 			|| isVersionAtLeast "2.6" value.hawkSourceInfo.version))
 	];
 	latestVersionFrag = lib.head releaseFrags;
-	combined = pp // asmsFromReleaseArtifacts // releasesEmuHawkInstallables // {
-		inherit depsForHistoricalRelease populateHawkSourceInfo releaseTagSourceInfos;
-		inherit (emuhawk-local.assemblies) libretroCores;
+	combined = let
+		launchScriptsForLocalBuild = launchScriptsFor emuhawk-local.assemblies true;
+	in (pp // asmsFromReleaseArtifacts // releasesEmuHawkInstallables // {
+		inherit depsForHistoricalRelease populateHawkSourceInfo releaseTagSourceInfos launchScriptsForLocalBuild;
+		inherit (emuhawk-local.assemblies) libretroCores;		
 		bizhawkAssemblies = pp.buildAssembliesFor (fillTargetOSDifferences hawkSourceInfoDevBuild);
 		"bizhawkAssemblies-${latestVersionFrag}" = pp.buildAssembliesFor
 			(fillTargetOSDifferences releaseTagSourceInfos."info-${latestVersionFrag}");
@@ -172,8 +174,27 @@ in {
 		IDEs = {
 			kate = [ kate omnisharp-roslyn ];
 		};
-		launchScriptsForLocalBuild = launchScriptsFor emuhawk-local.assemblies true;
-	};
+		shellHook = drv: ''
+			export BIZHAWKBUILD_HOME='${builtins.toString ./.}'
+			export BIZHAWK_HOME="$BIZHAWKBUILD_HOME/output/"
+			ldLibPath='${lib.makeLibraryPath drv.buildInputs}' # for running tests
+			if [ -z "$LD_LIBRARY_PATH" ]; then
+				export LD_LIBRARY_PATH="$ldLibPath"
+			else
+				export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$ldLibPath"
+			fi
+			alias discohawk-monort-local='${launchScriptsForLocalBuild.discohawk}'
+			alias emuhawk-monort-local='${launchScriptsForLocalBuild.emuhawk}'
+			case "$-" in *i*)
+				pfx="$(realpath --relative-to="$PWD" "$BIZHAWKBUILD_HOME")/"
+				if [ "$pfx" = "./" ]; then pfx=""; fi
+				printf "%s\n%s\n" \
+					"Run ''${pfx}Dist/Build{Debug,Release}.sh to build the solution. You may need to clean up with ''${pfx}Dist/CleanupBuildOutputDirs.sh." \
+					"Once built, running {discohawk,emuhawk}-monort-local will pull from ''${pfx}output/* and use Mono from Nixpkgs."
+				;;
+			esac
+		'';
+	});
 in combined // lib.listToAttrs (lib.concatLists (builtins.map
 	(f: [
 		{ name = f "latest-bin"; value = combined.${f "${latestVersionFrag}-bin"}; }
--- a/flake.nix
+++ b/flake.nix
@@ -44,13 +44,5 @@
           default = self.devShells.${system}.emuhawk-latest;
         }
       ) nixpkgsFor;
-      overlays.default =
-        final: prev:
-        # filter derivations to only include `emuhawk` and `discohawk` ones (i.e. excluding `bizhawkAssemblies`)
-        std.filterAttrs (name: pkg: (std.hasPrefix "emuhawk" name) || (std.hasPrefix "discohawk" name)) (
-          # import `default.nix` with the overlayed package set
-          # (i don't know the circumstances under which `final` wouldn't have a `system` attribute, but we may as well account for it)
-          importDefaultDerivationsWith (final.system or "") final
-        );
     };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -27,7 +27,6 @@
         (std.filterAttrs (name: val: std.isDerivation val) (import ./default.nix { inherit system pkgs; }));
     in
     {
-      formatter = mapAttrs (system: pkgs: pkgs.nixfmt-rfc-style) nixpkgsFor;
       packages = mapAttrs (
         system: pkgs:
         (importDefaultDerivationsWith system pkgs)
--- a/flake.nix
+++ b/flake.nix
@@ -9,7 +9,7 @@
     let
       std = nixpkgs.lib;
       systems = [
-        # TODO :: can we build for aarch64-linux?
+        # this is currently the only supported system, according to https://github.com/TASEmulators/BizHawk/issues/1430#issue-396452488
         "x86_64-linux"
       ];
       nixpkgsFor = std.genAttrs systems (
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,7 @@
 {
   description = "EmuHawk is a multi-system emulator written in C#. As well as quality-of-life features for casual players, it also has recording/playback and debugging tools, making it the first choice for TASers (Tool-Assisted Speedrunners).";
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/release-24.05";
+    nixpkgs.url = "github:NixOS/nixpkgs?ref=24.05";
   };
   outputs =
     inputs@{ self, nixpkgs, ... }:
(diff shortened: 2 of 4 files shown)


SignalWalker:
I checked the "relevant test suites" box under the assumption that there aren't any such tests; is that correct?

YoshiRulz:
Actually I do have a regression test script for Nix, but adding a `flake.nix` would make no difference.

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+      nixpkgsFor = std.genAttrs systems (
+        system:
+        import nixpkgs {
+          localSystem = builtins.currentSystem or system;
+          crossSystem = system;
+          overlays = [ ];
+        }
+      );
+      # because for some reason the standard library doesn't include this (i think?)
+      startsWith = prefix: st: (substring 0 (stringLength prefix) st) == prefix;
```
YoshiRulz, line 24: `lib.hasPrefix`

flake.nix
```diff
@@ -0,0 +1,80 @@
+{
+  description = "BizHawk is a multi-system emulator written in C#. BizHawk provides nice features for casual gamers such as full screen, and joypad support in addition to full rerecording and debugging tools for all system cores.";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-24.05";
+  };
+  outputs =
+    inputs@{ self, nixpkgs, ... }:
+    with builtins;
+    let
+      std = nixpkgs.lib;
```
YoshiRulz, line 10: Can't you just add `lib` to the function paramset of `outputs`?

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+      systems = [
+        # this is currently the only supported system, according to https://github.com/TASEmulators/BizHawk/issues/1430#issue-396452488
+        "x86_64-linux"
+      ];
```
YoshiRulz, line 14: I set a bunch more on `emuhawk-latest.meta`, if you can re-use that.

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+        // {
+          default = self.packages.${system}.emuhawk-latest-bin;
+        }
+      ) nixpkgsFor;
+      devShells = mapAttrs (
+        system: pkgs:
+        # ./shell.nix outputs some non-derivation attributes and some extraneous derivations, so we have to filter those out
+        (std.filterAttrs (
+          name: val: std.isDerivation val && name != "stdenv" && name != "out" && name != "inputDerivation"
+        ) (import ./shell.nix { inherit system pkgs; }))
```
YoshiRulz, line 45: Note that the shells come from `avail` (equivalent to `packages` here):
https://github.com/TASEmulators/BizHawk/blob/27e711dbe1c4528b2e1fa8c4785a9a03730f1703/shell.nix#L49-L58
Ideally this could do the same rather than trying to filter derivations.

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+        // {
+          default = self.devShells.${system}.emuhawk-latest;
+        }
+      ) nixpkgsFor;
+      overlays.default =
+        final: prev:
+        # filter derivations to only include `emuhawk` and `discohawk` ones (i.e. excluding `bizhawkAssemblies`)
+        std.filterAttrs (name: pkg: (startsWith "emuhawk" name) || (startsWith "discohawk" name)) (
```
YoshiRulz, line 53: What is this for? I'm not exporting any overlays.

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+      apps =
+        let
+          toApps =
+            app: pkgs:
+            # filter packages to only include ones whose name starts with `app`, then map them to app definitions
+            mapAttrs (name: pkg: {
+              type = "app";
+              # this seems to be correct, but I'm not entirely sure
+              program = "${pkg}/bin/${pkg.name}";
+            }) (std.filterAttrs (name: val: startsWith app name) pkgs);
```
YoshiRulz, line 67: This is a lot of code for 2 things

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+      # because for some reason the standard library doesn't include this (i think?)
+      startsWith = prefix: st: (substring 0 (stringLength prefix) st) == prefix;
+      # import the derivations from default.nix for the given system & package set
+      importDefaultDerivationsWith =
+        system: pkgs:
+        # ./default.nix outputs some non-derivation attributes, so we have to filter those out
+        (std.filterAttrs (name: val: std.isDerivation val) (import ./default.nix { inherit system pkgs; }));
+    in
+    {
+      formatter = mapAttrs (system: pkgs: pkgs.nixfmt-rfc-style) nixpkgsFor;
```
YoshiRulz, line 32: Is this required? I'm definitely not following the Nixpkgs coding conventions.

flake.nix
```diff
@@ -0,0 +1,80 @@
+{
+  description = "BizHawk is a multi-system emulator written in C#. BizHawk provides nice features for casual gamers such as full screen, and joypad support in addition to full rerecording and debugging tools for all system cores.";
```
YoshiRulz, line 2: Copy from readme instead (the GitHub repo's desc is ancient but I can't edit it).

YoshiRulz (dismissed):

flake.nix
SignalWalker: i swear i ctrl+f'd `prefix` in the nixpkgs manual and didn't find it ;-;

flake.nix
SignalWalker: I don't think so? As far as I can tell, the only inputs to `outputs` are `self` and whatever inputs you define in the `inputs` attrset.

flake.nix
SignalWalker: Not really, no; it just feels like good practice to define one (but, thinking about it, it's probably better to just leave it out to preserve the state of the existing files).

flake.nix
SignalWalker: It's just for convenience when using `nix run`, so it's not necessary if you'd prefer to avoid adding to the maintenance burden.

flake.nix
SignalWalker: I don't know if it's possible to get the `meta` attribute from the derivation without first importing `nixpkgs` (to get a `stdenv` with which to make the derivation), and we can't do that without specifying a system (and `builtins.currentSystem` isn't available in flakes, last I checked).

The system set itself can be generated without importing `nixpkgs`, though, so it would be possible to generate it there, if you're fine with either copying the platform set function into `flake.nix` or splitting it off into its own file that could be imported as needed.

flake.nix
SignalWalker: I mean, yeah, someone consuming the flake could just make the overlay themself if they really wanted to.

flake.nix
YoshiRulz: `nix run` docs say you don't need `apps` at all, it falls back to `packages.x86_64-linux.default`.

flake.nix
YoshiRulz: Did some more reading and apparently this is a workaround for Flakes' lack of cross-compilation support. Since there's only 1 supported platform atm, I think it's safe to remove.

flake.nix
SignalWalker: On mine that gives me `error: cannot find flake 'flake:lib' in the flake registries`; I think that's adding an implicit flake input like:
 ```nix
 inputs.lib = {
    type = "indirect";
    id = "lib";
};
```
and, for some reason, you have `lib` in your registry and I don't.

File: flake.nix [hunk 9]
```diff
@@ -0,0 +1,48 @@
+{
+  description = "EmuHawk is a multi-system emulator written in C#. As well as quality-of-life features for casual players, it also has recording/playback and debugging tools, making it the first choice for TASers (Tool-Assisted Speedrunners).";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-24.05";
```
YoshiRulz, line 4: I've been using the tags e.g. `24.05`, which I think are different

flake.nix
SignalWalker: So, like in [6be3da2](https://github.com/TASEmulators/BizHawk/pull/4328/commits/6be3da2aeb2694e86280918ed407504221ad174d)?

YoshiRulz:
Is https://github.com/numtide/flake-utils considered good practice or nah?

default.nix
```diff
@@ -184,8 +186,27 @@ in {
 		IDEs = {
 			kate = [ kate omnisharp-roslyn ];
 		};
-		launchScriptsForLocalBuild = launchScriptsFor emuhawk-local.assemblies true;
-	};
+		shellHook = drv: ''
+			export BIZHAWKBUILD_HOME='${builtins.toString ./.}'
```
YoshiRulz, line 190: Wait does this work in Flakes?

SignalWalker: Oh, when evaluating as part of a flake, that `./.` gets interpreted as a nix store path, so `BIZHAWKBUILD_HOME` is set to something like `/nix/store/1igqk0g3q3bj6zhcfmrgzirzjsqbgpx9-source` instead of something like `/home/tasbot/bizhawk`. That's fine for reading data, but the store is (usually) read-only, so `$BIZHAWK_HOME` doesn't work as-is (I assume).

I'm not really sure what the best way to solve this would be (especially in the case of someone doing something like `nix develop github:tasemulators/bizhawk`, since that would mean that the source dir *only* exists in the store).

`nix develop` sets environment variables for each of the derivation outputs (ex. `$out`), which might be useful. They're relative to the directory in which you invoke `nix develop`, though (ex. if you invoke `nix develop ./.#emuhawk-latest` in `BizHawk/Dist`, `$out` is set to `BizHawk/Dist/outputs/out` rather than `BizHawk/outputs/out`).

SignalWalker:
> Is https://github.com/numtide/flake-utils considered good practice or nah?

Honestly, I just don't use it because downloading it adds a bit of time to flake commands and because nix error messages are so obtuse that I'd rather have as little abstraction as possible. As far as I know, though, it's not considered bad practice; that's just my own preference.

default.nix
YoshiRulz: IMO it's fine to assume `nix develop` will only be called from the repo root, already checked-out.

`BIZHAWKBUILD_HOME` is used at build-time to locate the repo root (not for the main solution though). `BIZHAWK_HOME` is used at run-time, and it was mainly added for NixHawk so I don't think anything expects it to be writable, BUT it can't point to `${BizHawk-source}/output` since that doesn't exist.

YoshiRulz:
Fair enough, I just don't like Flakes (as you've probably gathered) and would prefer to minimise the amount of code used for it.

SignalWalker:
Fair. I could rewrite it to use flake-utils, if you'd prefer that.

default.nix
YoshiRulz: You can do that, but again, it needs to exist and contain the assemblies and other assets. The launch script is essentially `cd $BIZHAWK_HOME; mono ./EmuHawk.exe`.

No-one is going to be using EmuHawk's dev shell to build WaterboxHost. It can get its own shell when I eventually package it.

YoshiRulz:
Pinging @OmnipotentEntity who was working with Flakes at some point (#3564).

bugQ:
> Is https://github.com/numtide/flake-utils considered good practice or nah?

It's at least very commonplace. If it weren't intended to be used with or without nixpkgs, it might as well already be a nixpkgs lib module.

bugQ:
That being said, since there is only one supported system at the moment, I don't know how much "boilerplate" it would actually eliminate.

YoshiRulz:
I would like to see some form of this merged. Flakes might not offer many benefits over what we already have, but the instant onboarding for new Nix users is a killer feature.
Please rebase. Per https://github.com/TASEmulators/BizHawk/commit/474fe58d9f4ca38624bfb424973a57a4ed922892, we're unfortunately stuck on (my fork of) Nixpkgs 24.05 for the foreseeable future.

SignalWalker:
> I would like to see some form of this merged. Flakes might not offer many benefits over what we already have, but the instant onboarding for new Nix users is a killer feature. Please rebase. Per [474fe58](https://github.com/TASEmulators/BizHawk/commit/474fe58d9f4ca38624bfb424973a57a4ed922892), we're unfortunately stuck on (my fork of) Nixpkgs 24.05 for the foreseeable future.

I honestly kinda forgot that I made this, sorry! Rebased and committed an extra change adding an `isFlake` parameter to `shellHook`.

holly-hacker:
Hi, has there been any progress on this PR? I'm trying to switch to bizhawk from mgba for its scripting features, and being able to just run `nix run github:TASEmulators/BizHawk` would be great.

PR #4330, opened 2025-05-24
Respect the user's ancient

SuuperW:

(describe changeset here)

Check if completed:
- [ ] I, the commit author, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2024-06-22) and am compliant

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
@@ -93,22 +93,22 @@ namespace BizHawk.Client.Common
 
 			if (keepOldStates)
 			{
-				// For ancients ... lets just make sure we aren't keeping states with a gap below the new interval
+				// For ancients, let's throw out states if doing so still satisfies the ancient state interval.
 				if (settings.AncientStateInterval > _ancientInterval)
 				{
-					int lastReserved = 0;
-					List<int> framesToRemove = new List<int>();
-					foreach (int f in _reserved.Keys)
+					List<int> reservedFrames = _reserved.Keys.ToList();
+					reservedFrames.Sort();
+					for (int i = 1; i < reservedFrames.Count - 1; i++)
 					{
-						if (!_reserveCallback(f) && f - lastReserved < settings.AncientStateInterval)
-							framesToRemove.Add(f);
-						else
-							lastReserved = f;
-					}
-					foreach (int f in framesToRemove)
-					{
-						if (f != 0)
-							EvictReserved(f);
+						if (_reserveCallback(reservedFrames[i]))
+							continue;
+
+						if (reservedFrames[i + 1] - reservedFrames[i - 1] <= settings.AncientStateInterval)
+						{
+							EvictReserved(reservedFrames[i]);
+							reservedFrames.RemoveAt(i);
+							i--;
+						}
 					}
 				}
 			}
@@ -297,7 +297,7 @@ namespace BizHawk.Client.Common
 			}
 		}
 
-		public void EvictReserved(int frame)
+		private void EvictReserved(int frame)
 		{
 			if (frame == 0)
 			{
@@ -309,6 +309,16 @@ namespace BizHawk.Client.Common
 				_reserved.Remove(frame);
 				StateCache.Remove(frame);
 			}
+		}
+
+		public void Unreserve(int frame)
+		{
+			// Before removing the state, check if it should be still be reserved.
+			// For now, this just means checking if we need to keep this state to satisfy the ancient interval.
+			if (ShouldKeepForAncient(frame))
+				return;
+
+			EvictReserved(frame);
 		}
 
 		public void Capture(int frame, IStatable source, bool force = false)
@@ -366,41 +376,39 @@ namespace BizHawk.Client.Common
 						index2 =>
 						{
 							var state2 = _recent.GetState(index2);
-							StateCache.Remove(state2.Frame);
-
 							var isReserved = _reserveCallback(state2.Frame);
 
 							// Add to reserved if reserved, or if it matches an "ancient" state consideration
-							if (isReserved || !HasNearByReserved(state2.Frame))
+							if (isReserved || ShouldKeepForAncient(state2.Frame))
 							{
 								AddToReserved(state2);
 							}
+							else
+								StateCache.Remove(state2.Frame);
 						});
 				});
 		}
 
-		// Returns whether or not a frame has a reserved state within the frame interval on either side of it
-		private bool HasNearByReserved(int frame)
+		/// <summary>
+		/// Will removing the state on this leave us with a gap larger than the ancient interval?
+		/// </summary>
+		private bool ShouldKeepForAncient(int frame)
 		{
-			// An easy optimization, we know frame 0 always exists
-			if (frame < _ancientInterval)
+			int index = StateCache.BinarySearch(frame + 1);
+			if (index < 0)
+				index = ~index;
+			if (index == StateCache.Count)
 			{
-				return true;
+				// There is no future state... so why are we wanting to evict this state?
+				// We aren't decaying from _recent, that's for sure. So,
+				return false;
 			}
+			if (index <= 1)
+				return true; // This should never happen.
 
-			// Has nearby before
-			if (_reserved.Any(kvp => kvp.Key < frame && kvp.Key > frame - _ancientInterval))
-			{
-				return true;
-			}
-
-			// Has nearby after
-			if (_reserved.Any(kvp => kvp.Key > frame && kvp.Key < frame + _ancientInterval))
-			{
-				return true;
-			}
-
-			return false;
+			int nextState = StateCache[index];
+			int previousState = StateCache[index - 2]; // assume StateCache[index - 1] == frame
+			return nextState - previousState > _ancientInterval;
 		}
 
 		private bool NeedsGap(int frame)
--- a/src/BizHawk.Tests/Client.Common/Movie/ZwinderStateManagerTests.cs
+++ b/src/BizHawk.Tests/Client.Common/Movie/ZwinderStateManagerTests.cs
@@ -606,6 +606,45 @@ namespace BizHawk.Tests.Client.Common.Movie
 			}
 		}
 
+		[TestMethod]
+		public void TestKeepsAtLeastAncientInterval()
+		{
+			IStatable ss = CreateStateSource();
+			// For this test to work, AncientStateInterval must be greater than CurrentTargetFrameLength + RecentTargetFrameLength
+			// We also require the current and recent buffers fill up, so that states are evicted by the time we reach the ancient interval.
+			ZwinderStateManager manager = new ZwinderStateManager(new ZwinderStateManagerSettings
+			{
+				CurrentBufferSize = 1,
+				CurrentTargetFrameLength = 2000,
+
+				RecentBufferSize = 1,
+				RecentTargetFrameLength = 2000,
+
+				AncientStateInterval = 10000,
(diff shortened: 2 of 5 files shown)


src/BizHawk.Tests/Client.Common/Movie/ZwinderStateManagerTests.cs
```diff
@@ -413,7 +413,7 @@ public void StateCache()
 
 			for (int i = 400; i < 1000; i += 400)
 			{
-				zw.EvictReserved(i);
+				zw.Unreserve(i);
```
YoshiRulz, line 416: Would this pass if it was still calling `EvictReserved`?

SuuperW: Aside from the fact that it would no longer compile (since EvictReserved is and should be private (why doesn't C# let unit tests call privates?)), yes it would.

Morilli:
> Do not allow a gap (range of states with no states) to be larger than the ancient state interval.

Am I reading the code right in that previously this would treat the ancient state interval as a "minimum" distance, aka reserved states would always have a distance `>= interval`, and now it's `<= interval`?

I guess this fits the intended purpose better, although I don't think the other behavior was necessarily wrong.

SuuperW:
You are mostly correct. What I think you are missing is that the old code was checking if there was a state within `interval` on EITHER side, which would lead to distances of nearly `interval * 2`. Example:
Say you have ancient interval 1,000 (and the current+recent buffers fill fewer frames than that). You have two branches that diverge on frame 0 and were made on frame 2,000. When you load one, the entire "greenzone" is invalidated, but you have a state on frame 0 and on 2,000. Now, there is a state "within 1,000 frames" of frame 1,000. So it would never keep a state there.

I suppose that issue could have been solved well enough by just removing the "Has nearby after" check.

SuuperW:
Updated to include reserving states from gap buffer. This is necessary because of branches including a savestate.

src/BizHawk.Tests/Client.Common/Movie/ZwinderStateManagerTests.cs
YoshiRulz: Since this PR makes `Unreserve` the intended API, I think this is fine.

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -311,6 +311,49 @@ public void EvictReserved(int frame)
...
+		private bool ShouldKeepForAncient(int frame)
+		{
+			int index = StateCache.BinarySearch(frame + 1);
+			if (index < 0)
+				index = ~index;
+			if (index == StateCache.Count)
+			{
+				// There is no future state... so why are we wanting to evict this state?
+				// We aren't decaying from _recent, that's for sure. So,
+				return false;
```
Morilli: I don't really understand this comment and code. Assuming there is no next state, shouldn't the return value be `true`, aka always keep the most current state unconditionally?

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -311,6 +311,49 @@ public void EvictReserved(int frame)
...
+			}
+			if (index <= 1)
+				return true; // This should never happen.
```
Morilli: While `index < 0` cannot happen, this isn't true for the other possible cases.

`index == 0` will only happen when `frame < 0`, which is arguably nonsensical, so the return value can be anything. 
However `index == 1` can very much happen when `frame = 0`, and returning `true` is correct here. So while the code is fine, the comment is confusing.

Morilli (commented):
Have checked the code and I think this looks good. I did however encounter a scenario in my testing where I got a large section of missing greenzone >> ancient interval. I'm not sure yet how to reproduce it or where the issue could be, Not approving it until I've investigated further.

Morilli:
> I did however encounter a scenario in my testing where I got a large section of missing greenzone >> ancient interval.

I found the issue: This happens when a buffer is shrunk and the discarded states generate a large gap. This isn't caused by this PR but can now be fixed nicely with:
```diff
diff --git a/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs b/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
index 42e26994ec..e25734140b 100644
--- a/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
@@ -170,15 +170,11 @@ namespace BizHawk.Client.Common
 					for (int i = 0; i < old.Count; i++)
 					{
 						ZwinderBuffer.StateInformation si = old.GetState(i);
-						// don't allow states that should be reserved to decay here, where we don't attempt re-capture
-						if (_reserveCallback(si.Frame))
-							AddToReserved(si);
-						else
-							buffer.Capture(si.Frame, s =>
-							{
-								using var rs = si.GetReadStream();
-								rs.CopyTo(s);
-							}, null, true);
+						buffer.Capture(si.Frame, s =>
+						{
+							using var rs = si.GetReadStream();
+							rs.CopyTo(s);
+						}, index => PossiblyMoveToReserved(buffer, index), true);
 					}
 					old.Dispose();
 				}
```

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
SuuperW: There should be no situation in which that line is executed. Maybe in some weird situation if you really try you could make it happen, I'm not sure. But normally, you'd be calling this on the oldest state of "recent" or "gaps" and there'd be later ones.
As for what should happen if you ever did get there, I don't think there's a good answer, especially without knowing the context. I figured false was safer, but again that code should never run so it shouldn't matter.

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
SuuperW: It would not happen with the code I wrote for this branch, since calling with frame 0 would never happen.
However I think that does happen in the `small_tasproj` branch when saving with reserved only, and arguably we should also use this function on state history settings change too. But really both of those could just skip the call for frame 0 and then the comment is correct again.

SuuperW:
> I found the issue

Wouldn't it be nicer to run the states through the main capture logic when shrinking buffers? That way `recent` stays immediately before `current` instead of potentially having ancient interval sized gap between them.
Then again, that wouldn't be a common occurrence and there is no way to make it perfect (without re-emulating to get new states), so maybe it's not worth doing more than your suggested changes.

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
Morilli: Both this and the below case can absolutely be hit because `ShouldKeepForAncient` is called from `Unreserve` which is the public api used to unreserve states and could be called with any frame really.

> I figured false was safer

How so? false means allowing the state to be removed. I would imagine keeping it should be the safe option, especially when this is the state on the highest frame.

Morilli:
> Wouldn't it be nicer to run the states through the main capture logic when shrinking buffers? 

Theoretically yes, but I'm not sure it's necessary. Just using the `PossiblyMoveToReserved` function should be fine here.

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
SuuperW: Because if we return true, then if the user gets into a situation where this is happening they might end up putting lots of states into reserved.
I was going to provide an example of how this could happen, considering marker removal calls `Unreserve`, but it turns out auto-adjust input according to lag was recently removed and my example depended on that. Still, I think it is better to actually remove the state in the case of user removing a marker who's state is the last state, to avoid polluting reserved with unnecessary states. It seems less likely to me that the state really should be kept.

Morilli: Looking at it again I guess it doesn't really matter. I'd still like different comments to clarify that calls from `Unreserve` could hit this code path and why the return value is what it is (same for the `index <= 1` below).

Morilli:
Something I'm not sure about: How does this interact with cores with Encore that try to minimize savestates and don't greenzone at all? Now that reserved states can be kept due to ancient interval considerations this might be problematic for that core. Tagging @CasualPokePlayer for thoughts.

SuuperW:
> Something I'm not sure about: How does this interact with cores with Encore that try to minimize savestates and don't greenzone at all? Now that reserved states can be kept due to ancient interval considerations this might be problematic for that core. Tagging @CasualPokePlayer for thoughts.

I disagree with how that has been implemented, but I have never actually used it so I can't actually say what I think it should do nor am I familiar with how well it works in practice.

Morilli (approved):
Looks good to me now

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
```diff
@@ -311,6 +307,56 @@ public void EvictReserved(int frame)
...
+			{
+				// There is no future state, so there is no gap between states for us to measure.
+				// We're probably unreserving for a marker removal. Allow it to be removed, so we don't pollute _reserved.
+				return true;
+			}
+			if (index <= 1)
+				// index == 0 should not be possible. (It's the index of the state after the given frame.)
+				// index == 1 would mean we are considering removing the state on frame 0.
+				// We must always have a state on frame 0.
+				return true;
```
YoshiRulz, line 353: If I understand this method correctly, isn't this branch reachable by `Unreserve(1);` --> `[ 0, 2, 3 ].BinarySearch(1 + 1)`? Since `Unreserve` doesn't check that the frame is in the list.

YoshiRulz (commented):
Brace style needs fixing, but other than that I have no style nitpicks.

src/BizHawk.Client.Common/movie/tasproj/ZwinderStateManager.cs
Morilli: Technically true, but it doesn't really matter. This implies `Unreserve(1)` is called while there isn't actually a state on frame 1, in which case there is nothing to do.

Morilli:
> Brace style needs fixing, but other than that I have no style nitpicks.

Disagree, at least not in this PR. If we want consistent brace style we should make a rule for it instead of arbitrarily enforcing something that isn't already being done and agreed upon.

YoshiRulz:
https://github.com/TASEmulators/BizHawk/blob/1a438c6a8cef99b03f6731157898d170042c4af0/contributing.md?plain=1#L125
^ [added 2019-01](https://github.com/TASEmulators/BizHawk/commit/08fdb101ddd782e709a099f60e6f1f0d75ee2f7e%5E)
(There is of course a rule for it, in StyleCop IIRC, but if you were to enable it I expect there would be thousands of new warnings.)

SuuperW:
I just skimmed the description for Almman braces, but I didn't see anything about them being required. It seems to just be a style for how braces are used for situations where they *are* used, and not about when they are used.

As for StyleCop, I don't even know what that is. Visual Studio never made complained about or made any attempt to "fix" missing braces.
Imo, anything that's a required style rule should be a compile warning (or maybe error) locally and a build error in the automatic checks for PRs. (Unused `using`s at the top of files are errors in one of the PR checks. But not a compile error or warning locally.)

Issue #1399, opened 2018-12-16
tastudio: make input keys do something outside rec mode


vadosnaprimer:
taseditor has some logic, and АКЛ asked for some stuff in that regard:
http://tasvideos.org/forum/viewtopic.php?p=450129#450129

vadosnaprimer:
> Basically, I would like to tap the related control binding to toggle it on the current frame.

vadosnaprimer:
From Discord.

> [1:50 AM] suqd: in TAStudio, is there a way to bind an input (for example, [Right]) to keyboard? so I don't have to click every time?
> [1:56 AM] suqd: (only to the currently selected frame)

warmCabin:
Would be really good if it could toggle everything in the selection, too. Basically do the same as when the button at the top of the column is clicked.

vadosnaprimer:
When you have to do a lot of analog editing it'd help a lot to be able to do it all just with the keyboard for every individual cell, and selection cursor seems to be the best way to handle this without having to emulate the the frame you're editing and to enable rec mode.

Which ties back into selecting individual cells, maybe it'd make sense to use alt+arrows to select them?

Example. You use arrows to move selection up and down, and you can hold Shift to add selection instead of moving it. What if we use some modified combined with an arrow to switch from selecting full row to selecting just one cell? Like you'd select a row, then hold Alt and then Right or Left, first cell gets selected. Keep holding Alt and hit Right again, the next cell gets selected instead. And if you hit Alt+Up/Down then the corresponding cell is selected on the other frame.

Space for binary buttons sounds good. The initial idea was to just replicate taseditor behavior and use bound inputs to toggle corresponding keys on the selected frame, and that still sounds very useful, but when only 1 cell is selected I think space is a good universal way to toggle it.

And for analogs, we know that Space can't be a part of the numeric value so it might work as a way to actually enter the value you've typed. Tho what happens next is unclear: should you go to the next cell on the same row, or to the same cell on the next row?

SuuperW:
I feel like alt+arrow would be an annoying way to select cells. I think it'd be better to have an option for full row select vs single cell selection, with the single cell option being required for this way of toggling buttons as well as free-typing* axis values or entering axis editing mode on a particular column via keyboard. Is there a reason users would be regularly switching between full row and single cell?

[...]

PR #4332, opened 2025-05-24
Fix clicking TAStudio column header with non-contiguous rows selected

YoshiRulz:
...and prepare for #1399. I just need to figure out where TAStudio would get called into when buttons are pressed, since that's normally handled via `InputManager` I believe.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
@@ -390,18 +390,69 @@ namespace BizHawk.Client.EmuHawk
 						if (ModifierKeys != Keys.Alt)
 						{
 							// nifty taseditor logic
+							// your TAS Editor logic failed us (it didn't account for non-contiguous `SelectedRows`) --yoshi
 							var selection = TasView.SelectedRows.ToArray(); // sorted asc, length >= 1
 							bool allPressed = true;
-							foreach (var index in selection)
+							var contiguous = true;
+							var iSelection = 0;
+							var lastFrameIndexSeen = selection[iSelection] - 1;
+							while (iSelection < selection.Length)
 							{
+								var index = selection[iSelection];
+								if (index - lastFrameIndexSeen is not 1)
+								{
+									contiguous = false;
+									break;
+								}
+								lastFrameIndexSeen = index;
 								if (index == CurrentTasMovie.FrameCount // last movie frame can't have input, but can be selected
 									|| !CurrentTasMovie.BoolIsPressed(index, buttonName))
 								{
 									allPressed = false;
 									break;
 								}
+								iSelection++;
 							}
-							CurrentTasMovie.SetBoolStates(frame: selection[0], count: selection.Length, buttonName, val: !allPressed);
+							if (!allPressed)
+							{
+								// still need to check for discontinuity
+								while (iSelection < selection.Length)
+								{
+									var index = selection[iSelection];
+									if (index - lastFrameIndexSeen is not 1)
+									{
+										contiguous = false;
+										break;
+									}
+									lastFrameIndexSeen = index;
+									iSelection++;
+								}
+							}
+							else if (!contiguous)
+							{
+								// still need to check for `false` values
+								allPressed = allPressed && selection.Skip(iSelection - 1) // break'd before checking the next one, so less 1 (and can't get here if `iSelection` is 0 so that's safe)
+									.All(index => index < CurrentTasMovie.FrameCount // last movie frame can't have input, but can be selected
+										&& CurrentTasMovie.BoolIsPressed(index, buttonName));
+							}
+							// else all contiguous and all `true`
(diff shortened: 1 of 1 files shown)


Morilli:
That's a lot of code with logic I don't want to read to fix this niche thing...

How about making a `ToggleBoolStates` method on `ITasMovie` instead (ToggleBoolState already exists on there)?

YoshiRulz:
~~`ToggleBoolStates` would just be this same logic but in a different file.~~ Logic has been safely contained in an extension method. And this has different semantics to toggling each selected row anyway.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -383,25 +455,13 @@ private void TasView_ColumnClick(object sender, InputRoll.ColumnClickEventArgs e
...
 				else if (columnName != CursorColumnName)
 				{
-					var frame = TasView.AnyRowsSelected ? TasView.FirstSelectedRowIndex : 0;
 					var buttonName = TasView.CurrentCell.Column!.Name;
 
 					if (ControllerType.BoolButtons.Contains(buttonName))
 					{
 						if (ModifierKeys != Keys.Alt)
 						{
-							// nifty taseditor logic
```
vadosnaprimer, line 393: I remember writing that line but I don't remember what was nifty about taseditor logic. Can you explain in text what the problem was and how you fixed it?

vadosnaprimer: Oh I see. Yeah that's a bug then. TASEditor handles this correctly. Tho I still can't tell if the code is good :D

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -342,25 +342,21 @@ private void TasView_ColumnClick(object sender, InputRoll.ColumnClickEventArgs e
...
 					if (ControllerType.BoolButtons.Contains(buttonName))
 					{
 						if (ModifierKeys != Keys.Alt)
 						{
 							// nifty taseditor logic
-							bool allPressed = true;
-							foreach (var index in TasView.SelectedRows)
+							// your TAS Editor logic failed us (it didn't account for non-contiguous `SelectedRows`) --yoshi
+							var selection = TasView.SelectedRows.ToArray(); // sorted asc, length >= 1
+							var allPressed = selection.All(index => CurrentTasMovie.BoolIsPressed(index, buttonName))
```
SuuperW: This throws if the last row is selected. The bounds check that comes after the `&&` here should be first.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -342,25 +342,21 @@ private void TasView_ColumnClick(object sender, InputRoll.ColumnClickEventArgs e
...
+							foreach (var (start, count) in selection.ChunkConsecutive())
 							{
-								if (index == CurrentTasMovie.FrameCount // last movie frame can't have input, but can be selected
-									|| !CurrentTasMovie.BoolIsPressed(index, buttonName))
-								{
-									allPressed = false;
-									break;
-								}
+								CurrentTasMovie.SetBoolStates(frame: start, count: count, buttonName, val: !allPressed);
 							}
```
SuuperW: Make this produce only one undo action. Also best if it only does one invalidation since that's possible now. (I'm noticing now I didn't do changelog batching for alt column clicks...)

```cs
string set = allPressed ? "Unset" : "Set";
string batchName = $"{set} {selection.Length} frames of {buttonName} starting at {TasView.FirstSelectedRowIndex}";
CurrentTasMovie.ChangeLog.BeginNewBatch(batchName);
CurrentTasMovie.SingleInvalidation(() =>
{
	foreach (var (start, count) in selection.ChunkConsecutive())
	{
		CurrentTasMovie.SetBoolStates(frame: start, count: count, buttonName, val: !allPressed);
	}
});
CurrentTasMovie.ChangeLog.EndBatch();
```

You could even remove `ChunkConsecutive` and use the singular `SetBoolState` on each individual frame.

SuuperW (changes requested):

SuuperW:
> Reduce enumerations of selection in `TAStudio.TasView_ColumnClick`

I do not think you actually did this. There seem to be two immediately before this commit and two immediately after. Before: `foreach` and then `Count`. After: `ToArray` and then `foreach`. (And 1 enumeration in master, 3 in this branch) But I guess it's still an improvement since iterating over the array should be faster than the InputRoll re-constructing the rows based on individual cells?

But, this seems fine anyway. Toggling a button on 1,000,000 rows is faster than actually selecting them in the first place.

SuuperW (approved):

vadosnaprimer:
Please resolve conflicts.

PR #4335, opened 2025-05-27
Add more Lua functions for TAStudio markers

YoshiRulz:
Based on #4272 by @c7fab (thank you, and please let us know if this is insufficient for your use case).

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -416,6 +416,23 @@ namespace BizHawk.Client.EmuHawk
 				indexFrom: 0);
 		}
 
+		[LuaMethodExample("""
+			tastudio.setbranchtext("New label", tastudio.get_branch_index_by_id(branch_id));
+		""")]
+		[LuaMethod(
+			name: "get_branch_index_by_id",
+			description: "Finds the branch with the given UUID (0-indexed). Returns nil if not found.")]
+		public int? GetBranchIndexByID(string id)
+		{
+			if (!Guid.TryParseExact(id, format: "D", out var parsed))
+			{
+				Log($"not a valid UUID: {id}");
+				return null;
+			}
+			return Tastudio.CurrentTasMovie.Branches.Index()
+				.FirstOrNull(tuple => tuple.Item.Uuid == parsed)?.Index;
+		}
+
 		[LuaMethodExample("local nltasget = tastudio.getbranchinput( \"97021544-2454-4483-824f-47f75e7fcb6a\", 500 );")]
 		[LuaMethod("getbranchinput", "Gets the controller state of the given frame with the given branch identifier")]
 		public LuaTable GetBranchInput(string branchId, int frame)
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -459,8 +459,8 @@ namespace BizHawk.Client.EmuHawk
 		}
 
 		[LuaMethodExample("local sttasget = tastudio.getmarker( 500 );")]
-		[LuaMethod("getmarker", "returns the marker text at the given frame, or an empty string if there is no marker for the given frame")]
-		public string GetMarker(int frame)
+		[LuaMethod("getmarker", "returns the marker text at the given frame, or nil if there is no marker for the given frame")]
+		public string/*?*/ GetMarker(int frame)
 		{
 			if (Engaged())
 			{
@@ -471,7 +471,7 @@ namespace BizHawk.Client.EmuHawk
 				}
 			}
 
-			return "";
+			return null;
 		}
 
 		[LuaMethodExample("tastudio.removemarker( 500 );")]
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -459,6 +463,37 @@ public string GetMarker(int frame)
 			return "";
```
vadosnaprimer, line 463: > `tastudio.getmarker` would need to return nil whenever the corresponding frame isn't marked.

c7fab:
At first glance it looks those functions can do everything that was possible with my PR.

SuuperW:
I think it would be good to handle the case where the specified branch doesn't exist. So that `tastudio.getmarker(0, 500)` doesn't throw an exception. Maybe print a message saying no such branch exists, or maybe just return `nil` like when there is no marker on the given frame.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -444,21 +444,56 @@ public void LoadBranch(int index)
 		}
 
 		[LuaMethodExample("local sttasget = tastudio.getmarker( 500 );")]
-		[LuaMethod("getmarker", "returns the marker text at the given frame, or an empty string if there is no marker for the given frame")]
-		public string GetMarker(int frame)
+		[LuaMethod(
+			name: "getmarker",
+			description: "Returns the label of the marker on the given frame."
+				+ " If that frame doesn't have a marker (or TAStudio isn't running), returns nil."
+				+ " If branchIndex is specified, searches the markers in that branch instead.")]
```
YoshiRulz:
```suggestion
				+ " If branchIndex is specified, searches the markers in that branch instead (0-indexed, as with tastudio.getbranches).")]
```

SuuperW:
Actually, branches have an `Id` that can be obtained with `tastudio.getbranches` and this is used instead of an index in `tastudio.getbranchinput`. This seems like it could be useful as branch indexes can change by deleting one. Should this new function use the `Id` too?

Existing function `tastudio.setbranchtext` uses an index, not an id. So it's already not consistent in the API.

vadosnaprimer:
Please provide a summary for dummmies, I don't have the capacity to remember how this is different from all the previous variants and I can't review it by looking at the technical side.

YoshiRulz:
[diff between force push](https://github.com/TASEmulators/BizHawk/compare/2f85f211fd9e542b090cd28917b60b194aa56c69..929098e45d82a1d059c87a31cb9b31ea85d4921b)
[current API proposal](https://github.com/TASEmulators/BizHawk/pull/4335/files)
I implemented Suuper's suggestions of taking branch UUIDs as params and throwing less.
And to encourage people to use UUIDs more, I've also added `get_branch_index_by_id` so they can be used with existing index-based functions.

SuuperW:
> throwing less

I don't really like the idea of using the current marker set when the given branch ID doesn't exist.

I also think that the documentation given to users should specify what happens when the input is invalid.

SuuperW:
I played around with this, and I was surprised to find that `tastudio.find_marker_on_or_before(5, "x")` returned `0` instead of `nil`. And so does `tastudio.find_marker_on_or_before(5, 1)`.
I also found that `tastudio.find_marker_on_or_before(-1)` throws an exception.

It's not clear if either of these are intentional.

Everything else appears to behave as expected.

YoshiRulz:
> I played around with this, and I was surprised to find that `tastudio.find_marker_on_or_before(5, "x")` returned `0` instead of `nil`. And so does `tastudio.find_marker_on_or_before(5, 1)`.

Yeah it was `default(int)`. I think it should be `-1` rather than `nil` though.

> I also found that `tastudio.find_marker_on_or_before(-1)` throws an exception.

Oops.

kalimag:
> I think it should be `-1` rather than `nil` though.

IMO `nil` is more idiomatic in Lua. It's how the Lua standard library operates (e.g. `string.find`, `io.open`), and it ties in better with Lua patterns like `assert` and `foo or default_value`

YoshiRulz:
[Force push diff](https://github.com/TASEmulators/BizHawk/compare/635602005eaf248a85dc47612350e98a303d3cdb..00c2a7b016d7b0804ca16c66a59f9e07e60e5f64)—just changed `-1` to `nil` per above.
edit: feos pointed out I forgot to do the same for `get_branch_index_by_id`, fixed in the next force push

SuuperW:
I would like this to be merged before release.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -403,6 +403,19 @@ public LuaTable GetBranches()
 				indexFrom: 0);
 		}
 
+		[LuaMethodExample("""
+			tastudio.setbranchtext("New label", tastudio.get_branch_index_by_id(branch_id));
+		""")]
+		[LuaMethod(
+			name: "get_branch_index_by_id",
+			description: "Finds the branch with the given UUID (0-indexed). Returns -1 if not found.")]
```
vadosnaprimer: why not nil

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -446,21 +459,56 @@ public void LoadBranch(int index)
 		}
 
 		[LuaMethodExample("local sttasget = tastudio.getmarker( 500 );")]
-		[LuaMethod("getmarker", "returns the marker text at the given frame, or an empty string if there is no marker for the given frame")]
-		public string GetMarker(int frame)
+		[LuaMethod(
+			name: "getmarker",
+			description: "Returns the label of the marker on the given frame."
```
vadosnaprimer: Please add a not what happens if the marker exists but thee label is empty.

Issue #3734, opened 2023-08-13
Savestate/bk2 `Framebuffer` isn't easily user-accessible


Morilli:
Ever since the zstd compression changes, the `Framebuffer*.bmp` files are zstd-compressed, which makes them unviewable unless they are manually extracted and decompressed with zstd beforehand.

There is another, technically unrelated problem where the `Framebuffer.bmp` in a `.bk2` isn't actually in .bmp format, it's just the raw framebuffer. However the `Branches/FrameBuffer*.bmp` of a `.tasproj` are in bmp format, as well as the `FrameBuffer.bmp` file saved in a `.State` file.

So I wanna propose two changes (that would require bumping the `BizState` version):
1. Make the `Framebufer.bmp` always contain valid bmp data as the file extension suggests instead of just saving the raw frame buffer in the case of a `.bk2` movie.
2. Don't zstd-compress `.bmp` files or at the very least not the main one inside of a `.bk2` or `.State` file, as it prevents users from viewing them normally and based on some short testing, the difference in compression strength and speed is negligible between zstd and zip.

YoshiRulz:
So this would necessitate bumping the format version? We should add sync settings (#2090) at the same time then.

Issue #2090, opened 2020-06-03
BizHawk crashes when trying to load a savestate made with different sync settings


ThunderAxe31:
Create a savestate with GB or PSX, then change some core settings and reboot the core. Then, if you try to load that savestate, you get this for Gambatte:
```
System.InvalidOperationException: Save buffer size mismatch!
   at BizHawk.Emulation.Cores.Sony.PSX.Octoshock.LoadStateBinary(BinaryReader reader)
   at BizHawk.Client.Common.BinaryStateLoader.GetLump(BinaryStateLump lump, Boolean abort, Action`2 callback)
   at BizHawk.Client.Common.BinaryStateLoader.GetCoreState(Action`1 callbackBinary, Action`1 callbackText)
   at BizHawk.Client.Common.SavestateManager.LoadStateFile(IEmulator emulator, String path)
   at BizHawk.Client.EmuHawk.MainForm.LoadState(String path, String userFriendlyStateName, Boolean fromLua, Boolean suppressOSD)
   at BizHawk.Client.EmuHawk.MainForm.LoadQuickSave(String quickSlotName, Boolean fromLua, Boolean suppressOSD)
   at BizHawk.Client.EmuHawk.MainForm.CheckHotkey(String trigger)
   at BizHawk.Client.EmuHawk.MainForm.<ProcessInput>b__101_0(Boolean current, String trigger)
   at System.Linq.Enumerable.Aggregate[TSource,TAccumulate](IEnumerable`1 source, TAccumulate seed, Func`3 func)
   at BizHawk.Client.EmuHawk.MainForm.ProcessInput()
   at BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop()
   at BizHawk.Client.EmuHawk.Program.SubMain(String[] args)
```
And this for Octoshock:
```
System.InvalidOperationException: Savestate buffer size mismatch!
   at BizHawk.Emulation.Cores.Nintendo.Gameboy.Gameboy.LoadStateBinary(BinaryReader reader)
   at BizHawk.Client.Common.BinaryStateLoader.GetLump(BinaryStateLump lump, Boolean abort, Action`2 callback)
   at BizHawk.Client.Common.BinaryStateLoader.GetCoreState(Action`1 callbackBinary, Action`1 callbackText)
   at BizHawk.Client.Common.SavestateManager.LoadStateFile(IEmulator emulator, String path)
[...]

PR #4337, opened 2025-05-27
dont zstd compress sram

SuuperW:
Users should be able to extract the sram from a movie file and use it, and be able to replace the sram in a movie.

Check if completed:


Morilli:
I'm personally indifferent on whether or not sram should be compressed, but won't this require a version bump?
And if we do format changes, we should ideally bundle as many of them into one version change as possible, like #3734 or #2090 and perhaps renaming the text files that are actually json to `.json` already.

YoshiRulz:
Was the filename changed when Zstd compression was added?

And keep in mind that migrating movies from older versions isn't something we claim to support.

SuuperW:
I changed my approach for making movie SRAM usable/editable. to also address #3734. All zstd compressed files are now given an additional .zst extension. So MovieSaveRam.bin is now MovieSaveRam.bin.zst. These can be easily decompressed with either Windows' File Explorer or 7-Zip File Manager, which makes the uncompressed content readily available.
When loading, if a .zst version is not found it will look for a non-zst version and use that if found. This makes replacing the MovieSaveRam file easy.

This does not address #2090 or update JSON files to actually have .json extension. Do we want to do both of those along with this version bump?

YoshiRulz:
IMO the `.bin` should take precedence over `.bin.zst`

SuuperW:
> IMO the `.bin` should take precedence over `.bin.zst`

I just made the most likely to exist one be the first it checked, but I think your suggestion makes sense. Although it might also make sense to fail if they both exist and tell the user they should delete one, so that there's no ambiguity.

SuuperW:
I realized that the `ZipStateLoader` in master doesn't care at all what the file extensions are, which means we could change the extension of JSON files to .json with no real consequences. So I modified the commit to keep that behavior, and then changed file extensions to .json.

It also now throws if there are two files with the same name minus extension, except in the case that one is .zst in which case uncompressed takes precedence.

And why not go ahead and zstd compress the text/json files too? Since users can still access them, edit them, and replace them.

SuuperW:
Can this be merged? Tagging @CasualPokePlayer since he's been requested as a reviewer. (I did fix the thing you already commented on.)

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
```diff
@@ -59,14 +59,15 @@ private void PopulateEntries()
...
 			{
-				string name = z.FullName;
-				int i;
-				if ((i = name.LastIndexOf('.')) != -1)
-				{
-					name = name.Substring(0, i);
-				}
+				string name = z.FullName.Replace('\\', '/');
+				int i = name.IndexOf('.');
+				if (i != -1) name = name.Substring(0, i);
```
YoshiRulz: `var name = z.FullName.SubstringBefore('.').Replace('\\', '/');`

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
```diff
@@ -115,16 +116,30 @@ public static ZipStateLoader LoadAndDetect(string filename, bool isMovieLoad = f
...
 			{
+				isZstdCompressed = fileFound = _entriesByName.TryGetValue(lump.Name + ".zst", out e);
+			}
+
+			if (fileFound)
+			{
+				// In version 1.0.2, we did not use .zst to mark zstd compression (earlier versions had no zstd)
+				// These particular files/exensions were zstd compressed:
+				if (_ver == new Version(1, 0, 2) &&
+					(lump.Ext == "bin" || lump.Ext == "bmp" || lump.Name == "Greenzone"))
```
YoshiRulz:
```suggestion
					(lump.Ext is "bin" or "bmp" || lump.Name is "Greenzone"))
```
also put `&&` at start of this line

YoshiRulz (commented):
Does this resolve both of the issues we discussed earlier?

SuuperW:
Which two issues would that be?

Version bump: This does make a version bump.
.bin take precedence over .bin.zst: Yes, it does that now.

Mentioned GitHub issue "framebuffer isn't easily accessible": Resolved
Mentioned GitHub issue "crashes when loading savestate with different sync settings": I think this is unrelated and doesn't need a version bump, but I can look into it if desired.

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
SuuperW: Didn't know we had a `SubstringBefore`. Will use that.

src/BizHawk.Client.Common/savestates/ZipStateLoader.cs
SuuperW: I'm not a fan of `is` for checking values (we already have a dedicated operator for that, and `is` already means something else), but the `is or` syntax is kinda nice actually.

YoshiRulz:
re: #2090, the state just needs to include `SyncSettings.json`, which is then checked before loading a state. Whether that's done by comparing sync settings objects or comparing JSON strings doesn't really matter for UX. It can be in a separate PR.

PR #4341, opened 2025-05-31
Don't save when the user says no.

SuuperW:
If the user closed a movie in TAStudio while in recording mode and clicked "no" when asked if they want to save, the movie would get saved anyway. The recording mode check doesn't seem to do any good; if the user has recorded anything (or truncated the movie by loading a savestate to enter recording mode) then `Changes` will be set. From looking at the commit history, the check for recording mode was implemented before recording itself was implemented. So it probably should never have been there to begin with.
Edit: To clarify, this bug only actually happens if you open a (new or existing) movie while in recording mode, and not if you close TAStudio.

It's unclear why there is also a check for `IsActive`. It probably isn't doing any harm, but it is confusing. Commit 06605346e0eaf3352cd9b562bd6968e130aac14c says
> Fix a nasty bug in bk2 and tasmovie that was saving a movie even if it was inactive

I have no idea how this bug would have been triggered, since a movie shouldn't be able to be changed while inactive. The only way to make a movie inactive seems to be to stop it, and stopping it (used to) clears the `Changes` flag. So it would have had to be modified after stopping, then stopped again... even though ever call is made only on the condition that the movie is active. (Similarly, it is unclear why the movie stop method cleared the changes flag.)

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/movie/MovieSession.cs
+++ b/src/BizHawk.Client.Common/movie/MovieSession.cs
@@ -270,11 +270,12 @@ namespace BizHawk.Client.Common
 
 				message += "stopped.";
 
-				var result = Movie.Stop(saveChanges);
-				if (result)
+				if (saveChanges && Movie.Changes)
 				{
+					Movie.Save();
 					Output($"{Path.GetFileName(Movie.Filename)} written to disk.");
 				}
+				Movie.Stop();
 
 				Output(message);
 				ReadOnly = true;
@@ -390,7 +391,7 @@ namespace BizHawk.Client.Common
 			switch (Settings.MovieEndAction)
 			{
 				case MovieEndAction.Stop:
-					Movie.Stop();
+					StopMovie();
 					break;
 				case MovieEndAction.Record:
 					Movie.SwitchToRecord();
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.ModeApi.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.ModeApi.cs
@@ -20,23 +20,6 @@
 		public void SwitchToRecord() => Mode = MovieMode.Record;
 		public void SwitchToPlay() => Mode = MovieMode.Play;
 		public void FinishedMode() => Mode = MovieMode.Finished;
-
-		public virtual bool Stop(bool saveChanges = true)
-		{
-			bool saved = false;
-			if (saveChanges)
-			{
-				if (Mode == MovieMode.Record || (this.IsActive() && Changes))
-				{
-					Save();
-					saved = true;
-				}
-			}
-
-			Changes = false;
-			Mode = MovieMode.Inactive;
-
-			return saved;
-		}
+		public void Stop() =>  Mode = MovieMode.Inactive;
 	}
 }
--- a/src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
+++ b/src/BizHawk.Client.Common/movie/interfaces/IMovie.cs
@@ -121,11 +121,8 @@ namespace BizHawk.Client.Common
 
 		/// <summary>
 		/// Sets the movie to inactive (note that it will still be in memory)
-		/// The saveChanges flag will tell the movie to save its contents to disk
 		/// </summary>
-		/// <param name="saveChanges">if true, will save to disk</param>
-		/// <returns>Whether or not the movie was saved</returns>
-		bool Stop(bool saveChanges = true);
+		void Stop();
 
 		/// <summary>
 		/// Switches to record mode


Morilli:
Can you clarify how to reproduce this issue? I feel like I fixed this in b5893c0a0b860b0d3a16e7d3bca2087a588ec39c already.

SuuperW:
Yes, it happens whenever you open a different movie while in recording mode. But not when you close TAStudio.
Here's exact steps I just tested with a newly generated config file on the current master branch:
1) Open a ROM (I did a NDS one, but I don't think the core matters)
2) Open TAStudio
3) Enable recording mode in TAStudio
4) Select the File -> New menu item
5) Choose no
There should now be a file named default.tasproj in the default movies directory. If you'd saved the file with a name before selecting File -> New, then it would save it with your chosen file name after you said no.

Morilli:
Hm I see, so it happens when TAStudio itself is loading a new movie and the movie escapes to normal movie handling. The explicit `Mode == MovieMode.Record` check seems weird, but perhaps it's TAStudio's fault for even letting the movie escape in an active state, even though it clears the `Changes` flag.

src/BizHawk.Client.Common/movie/bk2/Bk2Movie.ModeApi.cs
```diff
@@ -26,14 +26,14 @@ public virtual bool Stop(bool saveChanges = true)
 			bool saved = false;
 			if (saveChanges)
 			{
-				if (Mode == MovieMode.Record || (this.IsActive() && Changes))
+				// How would a movie ever have changes while inactive?
```
Morilli, line 29: Adding this comment while simultaneously making this exact situation possible is probably not useful.

SuuperW:
> Hm I see, so it happens when TAStudio itself is loading a new movie and the movie escapes to normal movie handling. The explicit `Mode == MovieMode.Record` check seems weird, but perhaps it's TAStudio's fault for even letting the movie escape in an active state, even though it clears the `Changes` flag.

So, you suggest doing what my newly pushed commit does?

src/BizHawk.Client.Common/movie/bk2/Bk2Movie.ModeApi.cs
Morilli: You mean the check in `MovieSession`? That's true, so the active check should probably not exist here in the first place... to be honest I don't even know whether you can even get a movie into the `MovieSession` in an inactive state with the current code.

SuuperW:
Force pushed doing things a bit better. Is this good to merge?

src/BizHawk.Client.Common/movie/MovieSession.cs
```diff
@@ -270,11 +270,12 @@ public void StopMovie(bool saveChanges = true)
 
 				message += "stopped.";
 
-				var result = Movie.Stop(saveChanges);
-				if (result)
-				{
-					Output($"{Path.GetFileName(Movie.Filename)} written to disk.");
+				if (saveChanges && Movie.Changes)
+				{
```
YoshiRulz: Looks like you're mixing line endings (this file uses CRLF). Check your IDE settings.

SuuperW: Fixed. Do you know if there's a Visual Studio setting to auto-detect line endings? I am getting inconsistent behavior, with it using LF line endings in some places with no apparent reason. (updating BizHawk's .editorconfig might help... if line endings are consistent per-directory in BizHawk)

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -554,6 +554,8 @@ private bool StartNewMovieWrapper(ITasMovie movie, bool isNew)
 
 			SuspendLayout();
 			WantsToControlStopMovie = false;
+			// Stop the movie here, so that when main form calls movie stop, it won't save. (we've already saved, if needed)
+			MovieSession.StopMovie(false);
```
Morilli: I don't really like this. The problem here is that you're throwing out the current movie, but `StartNewMovie` could still fail to load the new movie (via the active cheat check for example, or just straight up unexpected error on reboot or something), in which case you would want the current movie to remain.

Perhaps using `CurrentTasMovie.ClearChanges()` here is actually the cleanest method. That way the movie stays until the new is loaded but won't be saved.

src/BizHawk.Client.Common/movie/MovieSession.cs
YoshiRulz: Apart from Designer files, I don't think there's any pattern to which files use CRLF. I do plan to nuke them all when we have fewer active branches.
I have no idea about VS.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
SuuperW: That's a good point. However, failure can still happen after MainForm's call to MovieSession.StopMovie so it would be unrecoverable anyway. MainForm stops the current movie on CloseGame (which happens during TAStudio loading of a movie), rather than wait until it is time to load the new one. (I'm not sure what the cause is, but I have had TAStudio and even EmuHawk crash/close themselves on movie load failures. I typically just re-open EmuHawk and try again and that usually works.)

I think the proper solution to that problem is not anything in TAStudio, but would require MovieSession.StopMovie to not throw away the movie, and make corresponding changes elsewhere. I think that's outside the scope of this PR.

Morilli: > failure can still happen after MainForm's call to MovieSession.StopMovie so it would be unrecoverable anyway

That's true, but it doesn't mean we need to make it worse by unconditionally stopping the movie here. If the logic in mainform was changed, this code would need to get changed too.

I agree that considering what `MovieSession.StopMovie` should or shouldn't is not part of this PR, but that doesn't mean using it here is good when it could be problematic.

Is there a problem with using `CurrentTasMovie.ClearChanges()` instead?

SuuperW: > Is there a problem with using CurrentTasMovie.ClearChanges() instead?

I don't think so. (putting it back in the ask save method; `CurrentTasMovie` might be `null` here because it could be a regular movie)

> If the logic in mainform was changed, this code would need to get changed too.

I don't see why. If stopping the movie doesn't throw it away, then it can be restarted without any fuss by calling `SwitchToPlay`. Although I suppose it may simplify the needed changes in MainForm if we don't stop the movie here.
So I think yeah it probably is better to use ClearChanges.

Morilli (approved):
This fixes the issue and seems like a safe change. Trying to ensure everything is correct is hard when considering what `MovieMode.Inactive` means (I don't know what it means) but this is a good step forward.

vadosnaprimer:
https://tasvideos.org/LawsOfTAS/OnSavestates#ModeInactive

Issue #2329, opened 2020-08-24
Clicking the left column in TAStudio should pause when it reaches the target frame


RetroEdit:
When you seek in TAStudio, it's supposed to pause when it reaches the goal frame. For some reason, that doesn't happen when emulation is unpaused before seeking.

~~This is annoying and I'm pretty sure this is not the intended behavior, but maybe it is? I vaguely get the impression previous versions didn't do this, but I need to do more testing.~~ Nope, this is an ever-present "feature".

I would classify this behavior as "runaway seek". I've noticed this before but I wasn't able to pin it down. Maybe it only occurs under more specific circumstances than simply being unpaused? I deleted my config before testing, so hopefully it's straightforward to reproduce.

RetroEdit:
This is probably the code for pausing after seeking:

https://github.com/TASVideos/BizHawk/blob/956b71060e5dbbc9d045cc7d20973a5e3bf24c59/src/BizHawk.Client.EmuHawk/MainForm.cs#L3005-L3013

But it doesn't pause when seeking while unpaused; maybe it did previously. In fact, `IsSeeking`, which is defined as `PauseOnFrame.HasValue` isn't true because `PauseOnFrame` is apparently never set when using the left cursor column to seek when emulation is unpaused, though it is set when clicking/drawing inputs. I'll test previous versions soon to find out where this regressed (if this is a regression).

RetroEdit:
Okay, I've tested several BizHawk versions (2.4, 2.2, and 2.0), and FCEUX 2.2.3, and this doesn't appear to be a regression I guess?

The reason why I don't think the current behavior is "correct" is because I interpret clicking the left column to be seeking to a specific frame. If emulation doesn't pause after reaching the frame, then it's not really seeking the frame, it's a form of resuming playback from that frame. Boiled down, the current behavior makes clicking the left column inconsistent in its purpose.

I am willing to consider the possibility that using cursor targeting to resume playback from a specific frame is desirable behavior and may be preferred by some. However, I would argue that unpausing after the frame is reached is straightforward and quick, while re-seeking the frame again requires scrolling back up to the frame to click on it again. Therefore, making it pause when the frame is reached is probably fine because unpausing after the frame has been reached is not that inconvenient.

Also, this reminds me that adelikat and I were discussing this in `#bizhawk` IRC a few weeks ago:

```
[...]

Issue #2081, opened 2020-05-31
Turbo seek doesn't render the proper frame


RetroEdit:
More specifically, turbo seek (sometimes? ~~always?~~) renders the frame that immediately precedes the frame that *should* be rendered. To verify this, one can turbo seek to a particular frame in a movie and examine the visual appearance, then seek to the frame that precedes it and frame advance, and compare the visual appearance to the visual appearance from turbo seeking directly. In my testing, doing these two operations that should appear identical resulted in two different rendered frames.

I discovered this while resyncing an mGBA BizHawk 2.4 movie to BizHawk 2.4.2. One instance of the emulator had turbo seek enabled, while the other didn't. I noticed that the frames appeared different when seeking to equivalent frames. After experimenting with settings, I discovered that the issue is present in both BizHawk 2.4 and BizHawk 2.4.2.

nattthebear:
Is this just framedrop from rendering being skipped?  I suppose turbo seek could seek to f - 1, then turn off turbo, then advance one.

adelikat:
The intent was in fact to do exactly that, f - 1, maybe that got broke at some point

RetroEdit:
I mean, I don't think it's frame drop per se, since it does render a frame *earlier* than what it should; unless the final frame is "dropped" for some weird reason? Maybe rendering happens too early in the turbo seek code, but should happen later? Seems less hacky than manually advancing one after seeking, especially if a fix could just address the underlying problem (unless it's a complicated fix, I suppose).

RetroEdit:
Still present in BizHawk 2.8; still seems to render at least one frame behind (maybe multiple?).

This makes `Turbo seek` annoying to use in conjunction with `Auto restore last position`, because sometimes the displayed results will be misleading. At first, I thought it was a desync issue.

Sophira:
I can confirm this still affects 2.9.1.

I did a bit of research on this bug and it looks like the last released version of BizHawk where this worked properly was 2.3.2; releases 2.3.3 onward contain this bug.

PR #4347, opened 2025-06-05
Fixes to seeking and auto restore

SuuperW:
The current implementation of TAStudio is a gigantic mess and has various pesky bugs. This is an attempt to resolve some of these issues.

Changes that are not bug fixes:
1) Show seek progress bar even for short seeks.
2) Include hotkey to go to green arrow, default is `R`.
3) Middle-click resumes a seek if a seek was in progress. (was this a bug? not sure)
4) Changed Lua API method `tastudio.setplayback` to return as soon as the seek begins. This is basically a side-effect of fixing bugs. See comments below for reasoning. (https://github.com/TASEmulators/BizHawk/pull/4347#issuecomment-2964930812)
5) Removed Lua API method `client.seekframe`. It would freeze the user interface for the duration of the seek, same as setplayback.

List of bugs fixed (also in commit message):
1) Auto-restore did not work when painting "axis" inputs.
2) If unpaused in recording mode, manual seek to frame A (click cursor column) then before that seek finishes seek to frame B. It would not unpause after reaching frame B to resume recording.
3) TAStudio would fail to pause with auto-restore if a second edit was made to a non-greenzoned frame while auto-restore seek was in progress.
4) Canceling seek would not remove the seek progress bar.
5) Mouse drag seeking was broken while seeking. (It would not change the target seek frame until prior seek had completed.)
6) Multiple issues with seeking to frames past the current movie length.
7) Modifier key + right click would jump to the last edited frame (even if this right click made no edits).
8) Clicking on an axis value while in axis editing mode would disable recording mode, regardless of whether an edit was made.
9) Clicking on an axis value while in axis editing mode would trigger auto-restore, even if no edit was made (seek to last edit frame).
EDIT: I'm not listing them all here anymore, see commit comments for more details. But, somewhere along the line I accidentally fixed these:
1) It used to be possible to invalidate greenzone without rewinding by pressing escape after mouse editing an axis value.
2) Right-click + scroll up (to seek backwards) immediately after opening TAStudio used to break the emulator pause functionality; it would be stuck unpaused.

I've done my best to avoid changing any behavior that isn't a bug. If you see any, please let me know.

Check if completed:


SuuperW:
added more fixes
plenty more still to come

SuuperW:
I'm not going to continue listing all the bugs. There are just too many, and I keep finding more as I go.

vadosnaprimer:
What matters is to mention them in commits, otherwise it'd be a pain to review.

Morilli:
> 2. If unpaused in recording mode, manual seek to frame A (click cursor column) then before that seek finishes seek to frame B. It would not unpause after reaching frame B to resume recording.

Does that even make sense though? Maybe seeking while unpaused *and* in recording mode actually *should* pause once seeking finishes? I generally do not really see a situation in which you would want to seek somewhere while unpaused and in recording mode, that sounds like a recipe to lose inputs.

SuuperW:
If we don't know what workflow would do that, then I think the best thing to do is provide consistent behavior. Making an exception for a special case we don't understand seems like a bad idea.
Anyway, I do it sometimes when I am messing around/exploring in-game. Never when I care about my precise inputs, but still I'd personally be annoyed if it paused unexpectedly or failed to re-enable recording mode.

As for losing inputs, the undo feature exists so I don't see this as being an issue. (It's currently totally broken in recording mode, but that's another issue and a bug I intend to fix.)

SuuperW:
Most recent commit (fix for Lua) changes the behavior of the Lua API. Previously, `tastudio.setplayback` would not return until seeking had completed and no Lua events/callbacks would happen during the seek. Now it returns when seeking begins (after load state, if that happened).
Note that seeks caused by `tastudio.applyinputchanges` were not broken in this way.

Although the old behavior is very bad, some Lua scripts could potentially no longer work without modifications. Is there significant value in making the old behavior available?

vadosnaprimer:
> Most recent commit (fix for Lua) changes the behavior of the Lua API. Previously, `tastudio.setplayback` would not return until seeking had completed and no Lua events/callbacks would happen during the seek. Now it returns when seeking begins (after load state, if that happened).

Sounds kinda weird. Callbacks being ignore while it's seeking may or may not be desirable (I never used lua to control tastudio/taseditor behavior in real tasing), but "set playback" implies a concrete event which is defined in that function's description:

> Seeks the given frame (a number) or marker (a string)

The goal is getting to that frame before we do something. After it finishes, it is expected that it's now safe to do whatever we wanted to do on that frame (or marker). Especially if there has to be a lot of seeking due to lack of states on those frames: if we say "Hey tastudio, please put playback on frame 9000!" we won't be satisfied if it puts it on frame 1000 and says "Done!!!".

> the old behavior is very bad

Why?

SuuperW:
> The goal is getting to that frame before we do something. After it finishes, it is expected that it's now safe to do whatever we wanted to do on that frame (or marker). Especially if there has to be a lot of seeking due to lack of states on those frames: if we say "Hey tastudio, please put playback on frame 9000!" we won't be satisfied if it puts it on frame 1000 and says "Done!!!".

This is at least trivial to fix in Lua: `while emu.framecount() < targetFrame do emu.frameadvance() end`

> Why?

1) It freezes the user interface. If TAStudio has to seek 8000 frames and the user can only emulate 200fps, it will be 40 seconds until ANYTHING is displayed and the user can take any further action. This would appear to be BizHawk crashing. This also means that it is impossible to cancel the seek, for example if the user didn't realize they were about to re-emulate thousands of frames.
2) During this time no Lua events can run. This means that any Lua functionality that depends on monitoring playback or keeping track of values over time can never work in a script that uses `tastudio.setplayback`. For example a script that I regularly use requires data from the previous frame (sometimes two) in order to display correct information on the current frame, and that is impossible if I use `tastudio.setplayback(frame_i_want_to_go_to)`.

vadosnaprimer:
> This is at least trivial to fix in Lua: `while emu.framecount() < targetFrame do emu.frameadvance() end`

It's not something that needs fixing because it's not a bug but an explicit function. We told it to get there, it got there, we're done.

> It freezes the user interface. If TAStudio has to seek 8000 frames and the user can only emulate 200fps, it will be 40 seconds until ANYTHING is displayed and the user can take any further action. This would appear to be BizHawk crashing. This also means that it is impossible to cancel the seek, for example if the user didn't realize they were about to re-emulate thousands of frames.

Didn't know about that. Does it depend on any option or is the freeze forced no matter what?

> During this time no Lua events can run. This means that any Lua functionality that depends on monitoring playback or keeping track of values over time can never work in a script that uses `tastudio.setplayback`. For example a script that I regularly use requires data from the previous frame (sometimes two) in order to display correct information on the current frame, and that is impossible if I use `tastudio.setplayback(frame_i_want_to_go_to)`.

This sounds like you have turbo seeking enabled and scripts during turbo disabled. If that's not the case, then I agree that it's also a problem.

However it doesn't mean that when we tell it to frame 1000 it should start seeking from frame 10 and immediately say "done!" There needs to be some proper solution to it doing this deep freeze.

SuuperW:
The freeze is forced no matter what.

> This sounds like you have turbo seeking enabled and scripts during turbo disabled. If that's not the case, then I agree that it's also a problem.

It freezes without turbo seek. I rarely have turbo seek enabled, and never turn off scripts during turbo.

I don't know what we could do to allow the script to run during the seek, without returning from `tastudio.setplayback`. It would not be too hard if the script uses the after-frame event, but since the "default" way to run some Lua code every frame is to run an infinite loop with `emu.frameadvance()` in it, ... well, I don't know how to make that work. The script will obviously be at the call to `tastudio.setplayback`, and I don't think there's any sensible way to move execution to wherever in the script `emu.frameadvance()` is.

vadosnaprimer:
Maybe I'm seeing it now. So what actually happens when I do `setplayback(9000)` from frame 10, given the most recent changes? Will it return from that function but actually keep seeking to 9000 while running all the tools normally?

What about `client.seekframe()`?

SuuperW:
Also, if the UI did not freeze then the user would be able to pause emulation and interrupt the seek. This will be much easier to handle if `tastudio.setplayback` has already returned. If it hasn't, and Lua is still waiting for the seek to complete, what happens if the seek never completes? We cannot possibly guarantee that emulation will ever reach the frame specified in the call to `tastudio.setplayback`, and suspending the Lua indefinitely would be very confusing behavior ... as would *sometimes* returning on the wrong frame but not always.

SuuperW:
> Maybe I'm seeing it now. So what actually happens when I do `setplayback(9000)` from frame 10, given the most recent changes? Will it return from that function but actually keep seeking to 9000 while running all the tools normally?
> 
> What about `client.seekframe()`?

There are two possibilities here.
First, if the movie does not have 9000 frames, setplayback will return immediately, doing nothing. (I do not know why this check exists.)
If the movie does have 9000 frames, the new setplayback behavior is to return right after the seek begins. It will keep seeking while all the tools run normally including the Lua script, which must explicitly check the current frame to know when frame 9000 is reached.

In either situation, seekframe will freeze the UI and seek, the same as old setplayback behavior (for when the movie did have 9000 frames) and just as it would in the current master branch since I haven't modified that code

vadosnaprimer:
Yeah okay. As long as it keeps seeking (until you disable it) returning right away is better.

SuuperW:
My list of things to change in this branch is now empty. Please review.

src/BizHawk.Client.Common/config/Binding.cs
```diff
@@ -133,6 +133,7 @@ void Bind(string tabGroup, string displayName, string defaultBinding = "", strin
 			Bind("TAStudio", "Show Cursor");
 			Bind("TAStudio", "Toggle Follow Cursor", "Shift+F");
 			Bind("TAStudio", "Toggle Auto-Restore", "Shift+R");
+			Bind("TAStudio", "Go To Green Arrow", "R");
```
YoshiRulz: Ambiguous; call it "Seek [...]" or "Scroll [...]"

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -102,39 +102,6 @@ public string GetLuaEngine()
 		public void InvisibleEmulation(bool invisible)
 			=> APIs.EmuClient.InvisibleEmulation(invisible);
 
-		[LuaMethodExample("client.seekframe( 100 );")]
-		[LuaMethod("seekframe", "Makes the emulator seek to the frame specified")]
-		public void SeekFrame(int frame)
```
YoshiRulz, line 107: This needs to still exist in some form. If the old behaviour is no longer possible, then it should just `Log()`.

src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
```diff
@@ -234,7 +240,9 @@ public int CopyOverInput(int frame, IEnumerable<IController> inputStates)
 
 			ChangeLog.EndBatch();
 			Changes = true;
-			InvalidateAfter(frame);
+			if (firstChangedFrame != -1)
+				// TOOD: Throw out the undo action if there are no changes.
+				InvalidateAfter(firstChangedFrame);
```
YoshiRulz, line 245: Missing braces (and typo)

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -549,7 +553,7 @@ private bool StartNewMovieWrapper(ITasMovie movie, bool isNew)
 
 			movie.InputRollSettingsForSave = () => TasView.UserSettingsSerialized();
 			movie.BindMarkersToInput = Settings.BindMarkersToInput;
-			movie.GreenzoneInvalidated = GreenzoneInvalidated;
+			movie.GreenzoneInvalidated = (f) => FrameEdited(f);
```
YoshiRulz:
```suggestion
			movie.GreenzoneInvalidated = f => _ = FrameEdited(f);
```
or propagate the new return type and
```suggestion
			movie.GreenzoneInvalidated = FrameEdited;
```

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
```diff
@@ -5,73 +5,44 @@ namespace BizHawk.Client.EmuHawk
...
+			StartSeeking(frame);
 
-					StartSeeking(frame);
-				}
-			}
+			if (!OnLeftMouseDown) { MaybeFollowCursor(); }
 		}
 
-		public void GoToPreviousFrame()
+		public void RestorePosition()
```
YoshiRulz: If this method were after `GoToNextMarker`, not only would this diff be clearer but it would be closer to alphabetical order :)

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
```diff
@@ -5,73 +5,44 @@ namespace BizHawk.Client.EmuHawk
...
 		/// <summary>
-		/// Only goes to go to the frame if it is an event before current emulation, otherwise it is just a future event that can freely be edited
+		/// Seek to the given frame, past or future, and load a state first if doing so gets us there faster.
+		/// Does nothing if we are already on the given frame.
 		/// </summary>
-		private void GoToLastEmulatedFrameIfNecessary(int frame, bool OnLeftMouseDown = false)
+		public void GoToFrame(int frame, bool OnLeftMouseDown = false)
 		{
-			if (frame != Emulator.Frame) // Don't go to a frame if you are already on it!
+			if (frame == Emulator.Frame)
```
YoshiRulz, line 13: I suspect the diff would be clearer if you wrote this as `if (frame == Emulator.Frame) return;`

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -1388,17 +1435,10 @@ public void EditAnalogProgrammatically(KeyEventArgs e)
...
-					_triggerAutoRestore = Emulator.Frame > _axisEditRow;
-					TastudioPlayMode(true);
-					JumpToGreenzone();
-					DoTriggeredAutoRestoreIfNeeded();
-				}
 			}
 
-			RefreshDialog();
+			if (!EndBatchEdit() && (prevTyped != _axisTypedValue || !AxisEditingMode))
+				RefreshDialog();
```
YoshiRulz, line 1463: Missing braces, though this could be 1 line.
~~Also, consider storing the result of `EndBatchEdit()` in a local variable, so the side-effects would be the same if the condition were to be rearranged in the future.~~

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +764,114 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+			_editIsFromLua = fromLua;
+		}
+		/// <returns>Returns true if the input list was redrawn.</returns>
+		public bool EndBatchEdit()
+		{
+			_batchEditing = false;
+
+			bool ret = false;
+			if (_batchEditMinFrame != -1)
+				ret = FrameEdited(_batchEditMinFrame);
```
YoshiRulz: Missing braces in this and the following method

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +764,114 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
 			}
 		}
 
+		/// <summary>
+		/// Begins a batch of edits, for auto-restore purposes. Auto-restore will be delayed until EndBatchEdit is called.
+		/// </summary>
+		public void BeginBatchEdit(bool fromLua = false)
```
YoshiRulz: What happens if there are mismatched begin and end calls?

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
SuuperW: OK.  ...but why? Is this one of those silly style guidelines? `_ = FrameEdited(f)` looks much uglier to me.
I'd have done the latter but that would require changing the type of GreenzoneInvalidated which seems a bit silly since the return value won't be used there either.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: An extra call to end would do nothing. A second call to begin before calling end would also do nothing. A call to begin without a call to end would result in not rewinding/auto-restoring at the right time, possibly emulating outside the greenzone (since no rewind), and maybe refreshes not happening at the right times.

Nested calls would just be an end at the first call to end, all future edits would be non-batched (no harm except doing needless work including repeated refreshing... which actually can be a bit of harm if there are tons of edits, for example in current master doing "Clone # Times" and entering something in the thousands will refresh that many times which takes a while during which the UI is unresponsive).

If you are worried that future edits might accidentally result in missing a `EndBatchEdit` call, there's already an unconditional call in list view MouseUp, so there's that at least. All calls currently should be matched except the one in list view MouseUp because it doesn't need a begin call as-is.

In current master it is possible to avoid the MouseUp event handler in Windows by pressing Alt+Tab to switch to another window while the mouse button is held. This is fixed in this branch by calling MouseUp when the window is "deactivated" (read: loses focus). If you know of any other ways to skip the MouseUp handler let me know how.

They are public because they are called from the Lua API, via `tastudio.applyinputchanges`.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
SuuperW: See my branch SuuperW:test and see if you like it.

SuuperW:
Fixed issues brought up in @YoshiRulz review along with one or two other cleanups. Also fixed regression and accidental behavior change: Lua's GreenzoneInvalidatedCallback was getting called at a different point and sometimes not at all.
And a few more fixes to tastudio.setplayback.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
YoshiRulz: So it won't be put in an unexpected state, and any weird behaviour would only last until you click again? That's good enough for me.

src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
```diff
@@ -82,13 +83,17 @@ public void SetFrame(int frame, string source)
 
 		public void ClearFrame(int frame)
 		{
-			ChangeLog.AddGeneralUndo(frame, frame, $"Clear Frame: {frame}");
+			string empty = Bk2LogEntryGenerator.EmptyEntry(Session.MovieController);
+			if (GetInputLogEntry(frame) != empty)
```
YoshiRulz: `if (GetInputLogEntry(frame) == empty) return;`

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -143,30 +143,32 @@ public void SetPlayback(object frame)
...
+				else if (frame is double frameNumber2)
+				{
+					f = (int)frameNumber2;
+				}
 				else
 				{
-					f = Tastudio.CurrentTasMovie.Markers.FindIndex((string)frame);
-					if (f == -1)
+					f = Tastudio.CurrentTasMovie.Markers.FindIndex((string) frame);
+					if (f != -1)
```
YoshiRulz: `>= 0`, and I think this should be using a separate temp variable rather than reassigning `f`.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs
```diff
@@ -94,18 +94,25 @@ protected override void UpdateAfter()
...
 
-			if (Settings.AutoPause)
+			if (Settings.AutoPause && _seekingTo == -1)
 			{
 				if (_doPause && CurrentTasMovie.IsAtEnd()) MainForm.PauseEmulator();
 				_doPause = !CurrentTasMovie.IsAtEnd();
 			}
 
+			if (!_seekingByEdit)
+				_shouldMoveGreenArrow = true;
```
YoshiRulz, line 105: missed one

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -523,10 +518,16 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
-					if (!record.Lagged.HasValue && LastPositionFrame > Emulator.Frame)
+					if (_seekingTo != -1)
 					{
-						StartSeeking(LastPositionFrame, true);
+						MainForm.UnpauseEmulator(); // resume seek
+						return;
+					}
+
+					var record = CurrentTasMovie[RestorePositionFrame];
+					if (!record.Lagged.HasValue)
```
YoshiRulz: `record.Lagged is null`

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +763,127 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+			{
+				return FrameEdited(_batchEditMinFrame);
+			}
+
+			return false;
+		}
+
+		public void BatchEdit(Action action)
+		{
+			// This is only caled from Lua.
```
YoshiRulz, line 787: If it's only called from Lua, put "Lua" or "ApiHawk" in the method name.
(Per the other thread, I think it's fine to keep the separate begin/end calls everywhere else.)

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +763,127 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+		/// If a mouse button is down, only tracks the edit so we can do this stuff on mouse up.
+		/// </summary>
+		/// <param name="frame">The frame that was just edited, or the earliest one if multiple were edited.</param>
+		/// <returns>Returns true if the input list was redrawn.</returns>
+		public bool FrameEdited(int frame)
+		{
+			GreenzoneInvalidatedCallback?.Invoke(frame); // lua callback
+
+			if (CurrentTasMovie.LastEditWasRecording)
+				return false;
```
YoshiRulz, line 813: missed one

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +763,127 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+					_seekingByEdit = true; // must be after GoToFrame & RestorePosition too (they'll set _seekingByEdit to false)
+
+					needsRefresh = false; // Refresh will happen via GoToFrame.
+				}
+				_batchEditMinFrame = -1;
+			}
+
+			if (needsRefresh)
+			{
+				if (TasView.IsPartiallyVisible(frame) || frame < TasView.FirstVisibleRow)
```
YoshiRulz, line 862:
```suggestion
				if (frame <= TasView.LastVisibleRow || TasView.IsPartiallyVisible(frame))
```
May be faster?

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +763,127 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+				{
+					// frame < FirstVisibleRow: Greenzone in visible rows has been invalidated
+					RefreshDialog();
+					return true;
+				}
+				else if (TasView.RowCount != CurrentTasMovie.FrameCount + 1)
+				{
+					// Row count must always be kept up to date even if last row is not directly visible.
+					TasView.RowCount = CurrentTasMovie.InputLogLength + 1;
```
YoshiRulz, line 871: Is this meant to be a different prop to the one in the condition?

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
YoshiRulz: Ehh it's marginally better. You can leave it like this or try moving this method after `GoToPreviousMarker`, idrc.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
YoshiRulz: Uglier, sure, but it communicates that the LHS('s caller) doesn't use the implicit lambda return.
And yeah it doesn't look like changing it would be useful.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.Navigation.cs
YoshiRulz: Ehh it's marginally better. You can leave it like this or try moving this method after `GoToPreviousMarker`, idrc.

YoshiRulz:
To fix the label in the Hotkeys dialog, increase this to 120 dp:
https://github.com/TASEmulators/BizHawk/blob/a030b16916015a86fe0b12028c67f8f07460673f/src/BizHawk.Client.EmuHawk/config/HotkeyConfig.cs#L98

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: silly me

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: Probably, but there are more significant performance optimizations in the implementation of ListView if you want to go there. In this particular case it's insignificant since (1) this isn't performance-sensitive logic and (2) the vast majority of edits will be in visible rows anyway.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: No, I didn't even realize there were two such properties. What is the theoretical difference? It's not explained in the property comments and their implementations are identical.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
SuuperW: I disagree with `>= 0`. `FindIndex` specifies that it returns -1 if no match is found. Checking for that specific value makes it clear what the code intends to do. And if we don't want to trust `FindIndex` to work correctly then we should be checking for the upper bound of valid indexes too.

YoshiRulz: I'm thinking we could later swap out the implementation with a binary search, but whoever does that would be wise to check every callsite anyway.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
YoshiRulz: From the docs I got the impression that the log length could include blank rows at the end. It was probably changed without updating the docs at some point.

src/BizHawk.Client.EmuHawk/config/HotkeyConfig.Designer.cs
```diff
@@ -198,7 +198,7 @@ private void InitializeComponent()
 			this.AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
 			this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
 			this.CancelButton = this.IDB_CANCEL;
-			this.ClientSize = new System.Drawing.Size(753, 463);
+			this.ClientSize = new System.Drawing.Size(783, 463);
```
YoshiRulz, line 201: Are any of these changes necessary? Just changing `iwOffsetX` was sufficient for me with Mono.

SuuperW: The window was not wide enough to see all the controls on the General tab when I only changed `iwOffsetX`. The button for "special tricky bindings" was cut off on the third column.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +763,129 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+		{
+			_batchEditing = true;
+		}
+		/// <returns>Returns true if the input list was redrawn.</returns>
+		private bool EndBatchEdit()
+		{
+			_batchEditing = false;
+			if (_batchEditMinFrame != -1)
+			{
+				return FrameEdited(_batchEditMinFrame);
```
YoshiRulz, line 779: I know I said you shouldn't worry about mismatched calls, but `_batchEditMinFrame = -1;` here could prevent some bugs.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -773,29 +763,129 @@ private void TasView_MouseDown(object sender, MouseEventArgs e)
...
+
+			return false;
+		}
+
+		public void ApiHawkBatchEdit(Action action)
+		{
+			// This is only caled from Lua.
+			_editIsFromLua = true;
+			BeginBatchEdit();
+			action();
```
YoshiRulz: ...and `try`/`finally` here

YoshiRulz (approved):

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: It's already inside `FrameEdited`. The only way it doesn't happen is if there's an exception during rewind. In which case the rewind didn't happen and setting that probably isn't a good idea.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: The Lua call shouldn't be able to throw, but I suppose this could be good if custom tools use it?

YoshiRulz:
Is bug fix 2) in OP the same as #2329?

SuuperW:
Probably not, since he doesn't mention recording mode. It's more likely he was experiencing bug 3, but none of his posts indicate the steps taken to reproduce the bug so I can't tell.

EDIT: No, not 3, since he said he was clicking on the left column not doing edits.

vadosnaprimer:
I missed that this breaks a script we ship (and [one we want to start shipping](https://tasvideos.org/Forum/Posts/520280)). Please fix.

https://github.com/TASEmulators/BizHawk/blob/e5817d9c024be188b0e127f062c4e8a1a7b69ed8/Assets/Lua/GBA/SonicAdvance_CamHack.lua#L24-L26

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -69,38 +68,21 @@ private int AxisEditRow
...
-
-			MainForm.PauseOnFrame = frame.Value;
-			int? diff = MainForm.PauseOnFrame - _seekStartFrame;
+			_seekStartFrame = Emulator.Frame;
+			_seekingByEdit = false;
 
-			WasRecording = CurrentTasMovie.IsRecording() || WasRecording;
-			TastudioPlayMode(); // suspend rec mode until seek ends, to allow mouse editing
+			_seekingTo = frame;
+			MainForm.PauseOnFrame = int.MaxValue; // This being set is how MainForm knows we are seeking, and controls TurboSeek.
```
Morilli, line 82: Is there any reason you changed this to be `int.MaxValue` instead of `frame`? This breaks numerous (reasonable) assumptions in the code that `PauseOnFrame` actually contains the correct frame number value and leads to incorrect behavior with turbo seeking because the frame advance code does not know when the turbo seek ends anymore.

SuuperW: What specific incorrect behavior are you seeing? The only reference to `PauseOnFrame` that care about the value other than checking if it is `null` or `<= Emulator.Frame` is MainForm line 3018. That might be an oversight by me (looks like it might cause the core to not render the last frame, for cores that support skipping rendering). But it's not related to frame advance.

For everything else, TAStudio handles things by setting `PauseOnFrame` to `null` at the correct time.

I would guess that the reason I set it to `MaxValue` was to avoid potential issues with having it get out of sync with `_seekingTo`, e.g. when using mouse wheel to change the seek frame. But the correct solution to that is probably to ~~change `_seekingTo` to a property and use `MainForm.PauseOnFrame` as a backing "field"~~. EDIT: Actually that would make it so that MainForm's pause-after-seek logic applies for turbo seeks, which is different from TAStudio's.

Morilli: Well it kinda breaks the tooltips in `SetPauseStatusBarIcon` although that's a really minor issue.

But yes, it also directly regresses #2081. Additionally this will now call `UpdateAfter` twice for lua scripts, once in `StepRunLoop_Core` because `atTurboSeekEnd` isn't set, and once in the `StopSeeking` call, which causes everything to be rendered twice which is very noticable with transparent drawings.

SuuperW: See #4743

PR #4349, opened 2025-06-08
Working examples for Lua functions emu.getregister and emu.setregister

SandTAS:
The old examples were trying to use `emu.getregisters()[0]` to get an arbitrary register name for the current core. But `emu.getregisters()` returns a key–value table, not a sequential table, so it's not correct to index it with an integer. Even if it were a sequential table, the first element would be at index 1, not 0.

The new examples use the [Lua function `next`](https://www.lua.org/manual/5.4/manual.html#pdf-next) to get an arbitrary register name. They also replace the apparently autogenerated variable name `inemuget` with more meaningful ones (`regname` and `value`).

Check if completed:


YoshiRulz:
The replacement is just as unhelpful.

SandTAS:
That's all right. I withdraw the suggestion.

PR #4355, opened 2025-06-14
64bit peek

### LINKED ISSUE #4353: Add 64-bit reads/writes to tools that manipulate memory
With lua this was impossible on 5.1 but 5.3 has 64-bit integers. Other tools like mem watch/search/hexeditor seem to always have been able to do this, but I guess low demand made it not happen.

In Doom a bunch of variables are 64-bit, and obviously in some other cores too, so it looks useful to extend the API.

YoshiRulz:
Untested. I just wanted to know precisely what changes would be required for this feature.

resolves #4353

### DIFF
--- a/src/BizHawk.Client.Common/tools/RamSearchEngine/RamSearchEngine.cs
+++ b/src/BizHawk.Client.Common/tools/RamSearchEngine/RamSearchEngine.cs
@@ -109,6 +109,22 @@ namespace BizHawk.Client.Common.RamSearchEngine
 						}
 					}
 					break;
+				case WatchSize.QWord:
+					if (_settings.IsDetailed())
+					{
+						for (var i = 0; i < _watchList.Length; i++)
+						{
+							_watchList[i] = new MiniQWordWatchDetailed(domain, i * stepSize, _settings.BigEndian);
+						}
+					}
+					else
+					{
+						for (var i = 0; i < _watchList.Length; i++)
+						{
+							_watchList[i] = new MiniQWordWatch(domain, i * stepSize, _settings.BigEndian);
+						}
+					}
+					break;
 			}
 		}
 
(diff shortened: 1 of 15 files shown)


src/BizHawk.Client.EmuHawk/tools/Watch/WatchValueBox.cs
```diff
@@ -31,18 +32,16 @@ public WatchSize ByteSize
...
+						WatchSize.Word => WordWatch.ValidTypes,
+						WatchSize.DWord => DWordWatch.ValidTypes,
+						_ => [ ],
 					};
-
-					if (!isTypeCompatible)
+					if (!validTypes.Contains(_type))
 					{
-						_type = WatchDisplayType.Unsigned;
+						_type = validTypes.Count is 0 ? WatchDisplayType.Unsigned : validTypes[1];
```
Morilli: Typo? Should be `validTypes[0]` instead? Also the above switch should probably include `WatchSize.QWord` now.

src/BizHawk.Client.Common/tools/RamSearchEngine/RamSearchEngine.cs
```diff
@@ -301,28 +307,19 @@ private IEnumerable<long> AllAddresses()
 		{
 			foreach (var watch in _watchList)
 			{
-				if (_settings.CheckMisAligned)
+				var addr = watch.Address;
+				yield return addr;
+				if (_settings.CheckMisAligned) continue; // I guess we're assuming the other parts of the address are also in `_watchList`? --yoshi
```
Morilli, line 336: Right... if you have a better idea for how to implement this in a simple way without getting duplicates feel free to do that here.

src/BizHawk.Client.EmuHawk/tools/Watch/WatchValueBox.cs
YoshiRulz: gdi Lua rewiring my brain

src/BizHawk.Client.Common/tools/RamSearchEngine/RamSearchEngine.cs
YoshiRulz: Caller can `Distinct()`, but I didn't bother looking for where the caller(s) are.

vadosnaprimer:
Still draft?

PR #4373, opened 2025-06-25
Implement a new TAS state manager

SuuperW:
This PR writes a new state manager. It takes some inspiration from the current Zwinder manager, but has a few differences. The primary goal of this state manager was to enable capturing states out of order. This provides two primary benefits:
1) Users who load a branch, emulate forward a little, then rewind can still use the full buffer space.
2) It becomes much easier to "thin out" states (keeping only a subset of them) for saving and to then use them again on load. With the default settings, auto-save is now much less annoying.

In my very limited testing, performance seems virtually identical, except when re-playing an old part of a movie where it will be capturing states (what Zwinder calls "gap states"), for movies with a large number of states. That's slower due to a stupid of .NET's SortedSet which I haven't figured out a nice workaround for.

There is no support for loading states from or automatically migrating settings from the old manager. That doesn't seem super important, and anyway BizHawk already doesn't claim any backward compatibility in this area.

There is also no support for compressing states with ZStandard. This can be added, but is in my opinion very much not worth the performance loss.
There is also no support for TempFile states. I do plan on implementing this, but I currently don't see that as a high priority.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/IStateManagerSettings.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/IStateManagerSettings.cs
@@ -1,0 +1,63 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace BizHawk.Client.Common
+{
+	[JsonConverter(typeof(StateManagerSettingsConverter))]
+	public interface IStateManagerSettings
+	{
+		IStateManager CreateManager(Func<int, bool> reserveCallback);
+
+		IStateManagerSettings Clone();
+	}
+
+	// Legacy deserialization support. It's ugly, but Newtonsoft is dumb like that.
+	// All classes that implement IStateManager MUST use NoConverter as their converter or face infinite recursion.
+	public class StateManagerSettingsConverter : JsonConverter<IStateManagerSettings>
+	{
+		public override IStateManagerSettings ReadJson(JsonReader reader, Type objectType, IStateManagerSettings existingValue, bool hasExistingValue, JsonSerializer serializer)
+		{
+			var jsonObject = Newtonsoft.Json.Linq.JObject.Load(reader);
+			if (jsonObject.TryGetValue("$type", out JToken value))
+			{
+				string type = (string)value;
+				if (type.Contains(nameof(PagedStateManager.PagedSettings)))
+					return jsonObject.ToObject<PagedStateManager.PagedSettings>(serializer);
+				else if (type.Contains(nameof(ZwinderStateManagerSettings)))
+					return jsonObject.ToObject<ZwinderStateManagerSettings>(serializer);
+				else
+					throw new Exception("Invalid state manager type."); // This will get eaten.
+			}
+			else
+			{
+				if (reader.Path.Contains("DefaultTasStateManagerSettings"))
+				{
+					// We want to default to the new manager.
+					// Users will only keep the zwinder manager as their default if they specify it after we've added the paged manager.
+					return new PagedStateManager.PagedSettings();
+				}
+				else
+				{
+					// But inside a movie file, keep existing settings.
+					return jsonObject.ToObject<ZwinderStateManagerSettings>(serializer);
+				}
+			}
+		}
+
+		public override bool CanWrite => false;
+
+		public override void WriteJson(JsonWriter writer, IStateManagerSettings value, JsonSerializer serializer) => throw new NotImplementedException();
+	}
+
+	// Newtonsoft makes classes inherit converters from implemented interfaces. Using this converter undoes that.
+	public class NoConverter : JsonConverter
+	{
+		public override bool CanConvert(Type objectType) => false;
+		public override bool CanRead => false;
+		public override bool CanWrite => false;
+
+		public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer) => throw new NotImplementedException();
+
+		public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) => throw new NotImplementedException();
+	}
+}
--- a/src/BizHawk.Client.Common/tools/TAStudio/PagedStateManager.cs
+++ b/src/BizHawk.Client.Common/tools/TAStudio/PagedStateManager.cs
@@ -122,6 +122,7 @@ namespace BizHawk.Client.Common
 			New,
 			Mid,
 			Old,
+			None,
 		}
 
 		private struct StateInfo: IComparable<StateInfo>
@@ -130,6 +131,7 @@ namespace BizHawk.Client.Common
 			public int LastPage;
 			public int Size;
 			public int Frame = -1;
+			public StateGroup Group;
 
 			public StateInfo() { }
 
@@ -159,6 +161,8 @@ namespace BizHawk.Client.Common
 		private bool _bufferIsFull = false;
 		private int _newPagesUsed = 0;
 		private int _midPagesUsed = 0;
+
+		private int _lastForceCapture = -1;
 
 		private class PagedStream : Stream
 		{
@@ -417,6 +421,7 @@ namespace BizHawk.Client.Common
 			{
 				Frame = frame,
 				FirstPage = _firstFree,
+				Group = destinationGroup,
 			};
 
 			using PagedStream stream = new(newState, this, false);
@@ -450,24 +455,32 @@ namespace BizHawk.Client.Common
 
 			if (HasState(frame)) return;
 
+			if (force)
+			{
+				if (HasState(_lastForceCapture))
+				{
+					StateInfo state = _states.GetViewBetween(new(_lastForceCapture), new(_lastForceCapture)).Min;
+					if (state.Group == StateGroup.Old && !_reserveCallback(_lastForceCapture) && !ShouldKeepForOld(_lastForceCapture))
+						RemoveState(state);
+				}
+				_lastForceCapture = frame;
+			}
+
+			StateGroup group = StateGroup.None;
 			if (_reserveCallback(frame))
 			{
-				InternalCapture(frame, source, StateGroup.Old);
+				group = StateGroup.Old;
 			}
 			else if (source.AvoidRewind)
 			{
 				// Zwinder did this, so I will too. I'm not sure it's a good idea but maybe it is.
-				return;
-			}
-			else if (force)
-			{
-				InternalCapture(frame, source, StateGroup.Mid);
+				// (do nothing)
 			}
 			else if (!_bufferIsFull || _newStates.Count == 0 || frame > _newStates.Min.Frame)
 			{
 				int max = frame - 1;
 				int min = frame - Settings.FramesBetweenNewStates;
-				int newestOlderState = _states.GetViewBetween(new(min), new(max)).Max.Frame;
+				int newestOlderState;
 
 				// Special case: If the buffer is entirely full of old states, we don't attempt to capture into new.
 				if (_bufferIsFull && _midStates.Count == 0 && _newStates.Count < 2)
@@ -477,19 +490,27 @@ namespace BizHawk.Client.Common
 					// In this case, we kinda want to ignore that new state.
 					if (_newStates.Count == 1)
 					{
-						newestOlderState = _states.GetViewBetween(new(min), new(max)).Max.Frame;
-						max = newestOlderState - 1;
+						int stateToIgnore = _states.GetViewBetween(new(min), new(max)).Max.Frame;
+						max = stateToIgnore - 1;
 					}
 					if (min <= max) newestOlderState = _states.GetViewBetween(new(min), new(max)).Max.Frame;
 					else newestOlderState = 0;
 
 					if (frame - newestOlderState >= Settings.FramesBetweenOldStates)
-						InternalCapture(frame, source, StateGroup.Old);
-					return;
+						group = StateGroup.Old;
 				}
+				else
+				{
+					// Don't consider non-new states when looking at the gap. (unless we have no new states)
+					// This keeps the gaps regular even when some frames are temporarily force captured.
+					if (_newStates.Count != 0)
+						newestOlderState = _newStates.GetViewBetween(new(min), new(max)).Max.Frame;
+					else
+						newestOlderState = _states.GetViewBetween(new(min), new(max)).Max.Frame;
 
-				if (frame - newestOlderState >= Settings.FramesBetweenNewStates)
-					InternalCapture(frame, source, StateGroup.New);
+					if (frame - newestOlderState >= Settings.FramesBetweenNewStates)
+						group = StateGroup.New;
+				}
 			}
 			else
 			{
@@ -497,8 +518,13 @@ namespace BizHawk.Client.Common
 				int min = frame - Settings.FramesBetweenMidStates;
 				StateInfo newestOlderState = _states.GetViewBetween(new(min), new(max)).Max;
 				if (frame - newestOlderState.Frame >= Settings.FramesBetweenMidStates)
-					InternalCapture(frame, source, StateGroup.Mid);
+					group = StateGroup.Mid;
 			}
+
+			if (group != StateGroup.None)
+				InternalCapture(frame, source, group);
+			else if (force)
+				InternalCapture(frame, source, StateGroup.Old);
 		}
 
 		public void Clear() => InvalidateAfter(0);
@@ -558,7 +584,7 @@ namespace BizHawk.Client.Common
 			if (state.Frame == 0) return;
 
 			// Remove the state if it's an old state we don't need.
-			if (!_newStates.Contains(state) && !_midStates.Contains(state) && !ShouldKeepForOld(frame))
+			if (state.Group == StateGroup.Old && !ShouldKeepForOld(frame))
 			{
 				RemoveState(state);
 			}
@@ -587,8 +613,7 @@ namespace BizHawk.Client.Common
 				{
 					foreach (StateInfo state in _states)
 					{
-						if (_newStates.Contains(state)) continue;
-						if (_midStates.Contains(state)) continue;
+						if (state.Group != StateGroup.Old) continue;
 
 						if (!_reserveCallback(state.Frame) && !ShouldKeepForOld(state.Frame))
 							RemoveState(state);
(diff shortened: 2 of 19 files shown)


SuuperW:
It seems the code analyzer use in the automatic build check is broken. It's complaining that there's no access modifier on a field marked `private`. What should I do to resolve this?

YoshiRulz:
Those `const` fields are not `private`. Maybe you forgot to push?

src/BizHawk.Tests/Client.Common/Movie/PagedStateManagerTests.cs
```diff
@@ -0,0 +1,503 @@
...
+using BizHawk.Emulation.Common;
+
+namespace BizHawk.Tests.Client.Common.Movie
+{
+	[TestClass]
+	public class PagedStateManagerTests
+	{
+		// Current PagedStateManager uses 4KB pages (4092 bytes of state data per page)
+		private const int PAGE_COUNT = 256;
+		private const int STATE_BYTES_PER_PAGE = 4092;
```
Morilli: 4KiB is 4096 bytes, is this a typo or intentionally -4?

SuuperW: Intentional. 4 bytes are for page metadata.

nattthebear:
The zwinder was designed for the requirements and limitations of linear rewind, and has worked out well there.  But it always seemed to be a bit of a mismatch for tasstudio, and I never quite understood the tasstudio requirements.  So I'm not surprised you're able to come up with something better, and it will be good to replace this.

SuuperW:
I don't think anybody quite understands the TAStudio requirements, because different users have different workflows, the same user may have different workflows for different games or TASes, and different consoles have very different limitations arising from state size and the time it takes to make states.

Although I believe this implementation will fit my use cases very well (and I've been using it a little since creating it), I would not be surprised if someone else wants some different behavior or options. (or even if I do if/when I TAS a new game/console) We'd need feedback from actual users to know that.

vadosnaprimer:
One of my biggest problems with the current state manager (that it introduced compared to the [decay algo](https://github.com/TASEmulators/BizHawk/pull/1128) I added years ago) is that if you go back and reemulate past states without changing any input, the very end of the greenzone (the future that has already been emulated) will not release states. That means if I watch some movie I'm judging and I notice a problem at frame 10k, and it's a system with big states, trying to go back to, say, frame 5k, I will only have A FEW states at my disposal. I may emulate frame 6k, then 6100, and trying to go back to frame 6k will launch the long seeking process from somewhere like frame 1k. Because the entire buffer is already fully occupied with all the states that are currently in the future compared to my playback position. To regain those states I have to change input, so those future states are erased and I can finally navigate comfortably.

The decay I added was not only thining out states from the far past but also from the far future. If there's little to no probability that I'll suddenly jump back to the future, I don't need all of those future states at all times. And when I need them, I'll reemulate them from some remaining keypoint states that are, say, 500 frames apart.

tl;dr: the area we're currently in needs the highest amount of states while we're navigating it, and farther away from it we should be gradually dropping more and more states.

I haven't checked the code for this PR, but how does it handle this problem?

SuuperW:
First, your use case is not one I had considered. When I rewind to watch old sections it is while I am working on a TAS, and I absolutely will want to jump back to the present and keep states there. It's usually just to review how I did a certain trick at some earlier point or what my strategy was.

Now for how this state manager handles rewind and states in "gaps". The manager keeps three categories of states which are similar to the old manager: old (ancient), mid (recent), and new (current). It does not have a separate "gaps" category. When you rewind to watch an old section, it pulls states out of the "mid" category. If there are any mid states past the currently emulated frame it will take the oldest one of those. If not, it will take the oldest mid state behind current emulation. This enables you to rewind and keep a section of states that spans as many frames as what the "mid" category can handle.

This should in general be much more than the old manager's "gap" category but I am not sure it would suit your use case.
Is it be okay for your use case that it uses the interval for mid states (now called "Frames Between Mid States", user-set rather than computed based on state size and buffer size) rather than the interval for new states? You said the currently emulated region should have the highest density. I think this is at odds with what I want from rewinding (at least, removing 90% of "new" states while rewinding is), but maybe we could have more frequent states while watching old sections and thin out the "new" states to whatever the "mid" interval is? That would require significant changes to how it currently tracks states in order to do it efficiently.
Or, you might be able to make the current one work for you by setting the "new to mid ratio" to something very low. This will place most of the states in the "mid" category, leaving you with just a few "new" states at the end of the greenzone if you re-emulate a bunch of past states. But there still won't be any "thinning out" of those future states.

vadosnaprimer:
I'll need some time on testing this with several different cores so see how it feels.

> You said the currently emulated region should have the highest density. I think this is at odds with what I want from rewinding (at least, removing 90% of "new" states while rewinding is), but maybe we could have more frequent states while watching old sections and thin out the "new" states to whatever the "mid" interval is? That would require significant changes to how it currently tracks states in order to do it efficiently.

The most recent example of this wasn't even to just watch a movie in tastudio and try improvements that come to mind. With huge states, every time you have to go back to check or redo something that affects the future, you can find yourself in that situation I'm describing. Huge states saturate the buffer very quickly, so how distant that past you need to check is depends on state size and the amount of gigabytes you can afford to give them. If future states persist, precise navigation to find the right point when something in memory changes becomes much harder. You've just emulated a few frames so it's natural to think that it means they're still in the buffer, but in fact only one of them is, and to navigate in there you have to either rewind from a secondary class state every time, or change input every time to future states don't pile up.

With tiny states it's not a problem because you can afford having hundreds or thousands of them at all times.

> Or, you might be able to make the current one work for you by setting the "new to mid ratio" to something very low. This will place most of the states in the "mid" category, leaving you with just a few "new" states at the end of the greenzone if you re-emulate a bunch of past states. But there still won't be any "thinning out" of those future states.

Removing first class states form a far future region and only leaving secondary class states there for backup would probably be ideal. Current TSM seems to only work in one direction, so it feels super weird to be unable to navigate around the place you've just been to and emulated it all. Maybe my whole vision of TSM is built around how states are captured in real time as opposed to emulated time. States that were captured an hour ago should be the first ones to go if we're out of capacity, regardless of their position. Then states captured 50 minutes ago, etc. If I need to really investigate some area, I only need to emulate it all from some keypoint (secondary or tertiary class states) and it's again all mine for a while, until I move to work on another section.

State decay was one solution to this but now thinking about it, I can say that it wasn't an exact implementation of this vision (because I never directly formulated it) but an incredibly close approximation.

SuuperW:
> I'll need some time on testing this with several different cores so see how it feels.

You posted this then edited your post. Is anything you say in that edit based on testing?

> You've just emulated a few frames so it's natural to think that it means they're still in the buffer, but in fact only one of them is

Are there any existing cores with such huge states, that don't have `AvoidRewind` set? NDS is relatively large but still you get a decent number of states with the default buffer size. I am getting nearly 200 frames with "new" states (so nearly 200/4 = 50 states) with the default settings. Granted 200 is still not a lot, but probably enough to go back after accidentally overshooting the intended frame. Edit: For rewinding it'd use "mid" states. With default settings, that'd be about 24 states = 24 * 20 = 480 frames.

> first class states ... secondary or tertiary class states

You are suggesting a new way of categorizing states, then? Based on "how states are captured in real time"?

> State decay was one solution to this but now thinking about it, I can say that it wasn't an exact implementation

What do you mean by "state decay"? There is a concept of "state decay" in the Zwinder state manager, but it was highly similar to what this one does, not based on states' real creation time like you are talking about.

SuuperW:
What do you think about always trying to keep enough states to satisfy "ancient interval" (from Zwinder) or "Frames Between Old States" (this manager)? Would you want to keep states such that you can always load one that is within X frames of where you want to go?

Perhaps the best solution to support watching movies like you want would be to implement another state manager that's based on state real creation time, and let the user choose which one they want?

> States that were captured an hour ago should be the first ones to go if we're out of capacity, regardless of their position. Then states captured 50 minutes ago, etc.

Would you expect the oldest (by real time) state to be the first to go even if it's the only state within 1,000 frames behind the currently emulated frame?
If what you want truly is as simple as always removing the oldest state, by real time, that would be easy to implement as a new state manager.

vadosnaprimer:
State decay is https://github.com/TASEmulators/BizHawk/pull/1128

Will reply to the rest tomorrow.

> Is anything you say in that edit based on testing?

No.

vadosnaprimer:
Getting around to test this is taking a long time...

SuuperW:
New commit makes it possible to have different movies with different state manager implementations.
There is no UI, ~~and old movie files do not keep the old state manager without modification. You need to add `"$type": "BizHawk.Client.Common.ZwinderStateManagerSettings, BizHawk.Client.Common"` in the `GreenZoneSettings.txt` file~~.

src/BizHawk.Client.Common/tools/TAStudio/PagedStateManager.cs
```diff
@@ -0,0 +1,665 @@
...
+	/// 2) It becomes much easier to "thin out" states (keeping only a subset of them) for saving and to then use them again on load.
+	///
+	/// Out of order states also means we do not need separately allocated buffers for "current", "recent", etc.
+	/// This gives us some more flexibility, but still there won't be any one-size-fits-all strategy for state managment.
+	/// For the initial implementation I will be using similar settings as Zwinder has but this may change in the future.
+	///
+	/// Additionally, this approach allows us to take the main goals of ZwinderBuffer into a state manager:
+	/// 1. No copies, ever.  States are deposited directly to, and read directly from, one giant buffer. (assuming no TempFile storage which is not currently implemented)
+	/// 2. Support for arbitrary and changeable state sizes.
+	/// </summary>
```
YoshiRulz, line 29:
```csharp
	/// <summary>
	/// This class will manage savestates for TAStudio, with a very similar API as <see cref="ZwinderStateManager"/>.
	/// <br/>This manager intends to address a shortcoming of Zwinder, which is that it could not accept states out of order.
	/// Allowing states to be added out of order has two primary benefits:
	/// <list type="number">
	/// <item><description>Users who load a branch, emulate forward a little, then rewind can still use the full buffer space.</description></item>
	/// <item><description>It becomes much easier to "thin out" states (keeping only a subset of them) for saving and to then use them again on load.</description></item>
	/// </list>
	/// Out of order states also means we do not need separately allocated buffers for "current", "recent", etc.
	/// This gives us some more flexibility, but still there won't be any one-size-fits-all strategy for state management.
	/// <br/>For the initial implementation I will be using similar settings as Zwinder has, but this may change in the future.
	/// <br/>Additionally, this approach allows us to take the main goals of <see cref="ZwinderBuffer"/> into a state manager:
	/// <list type="number">
	/// <item><description>No copies, ever. States are deposited directly to, and read directly from, one giant buffer (assuming no TempFile storage, which is not currently implemented).</description></item>
	/// <item><description>Support for arbitrary and changeable state sizes.</description></item>
	/// </list>
	/// </summary>
```

SuuperW:
Force push fixes a bug from my prior fix to force captures.

src/BizHawk.Client.Common/movie/tasproj/IStateManagerSettings.cs
```diff
@@ -0,0 +1,63 @@
...
+				}
+			}
+		}
+
+		public override bool CanWrite => false;
+
+		public override void WriteJson(JsonWriter writer, IStateManagerSettings value, JsonSerializer serializer) => throw new NotImplementedException();
+	}
+
+	// Newtonsoft makes classes inherit converters from implemented interfaces. Using this converter undoes that.
```
YoshiRulz, line 52: Are you able to set `[JsonConverter]` on `IStateManager.Settings` directly?

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
```diff
@@ -887,8 +895,27 @@ private void StateHistorySettingsMenuItem_Click(object sender, EventArgs e)
 		{
 			using GreenzoneSettings form = new(
 				DialogController,
-				new ZwinderStateManagerSettings(CurrentTasMovie.TasStateManager.Settings),
-				(s, k) => { CurrentTasMovie.TasStateManager.UpdateSettings(s, k); },
+				CurrentTasMovie.TasStateManager.Settings.Clone(),
+				(s, k) => { CurrentTasMovie.TasStateManager = CurrentTasMovie.TasStateManager.UpdateSettings(s, k); },
```
YoshiRulz, line 899: This weird construct suggests to me that you're doing something wrong, but I couldn't say what.

src/BizHawk.Tests.Client.Common/config/SerializationStabilityTests.cs
```diff
@@ -45,14 +49,15 @@ public sealed class SerializationStabilityTests
...
 			[typeof(PathEntry)] = PATHENTRY_SER,
 			[typeof(PathEntryCollection)] = $@"{{""Paths"":[{PATHENTRY_SER}],""UseRecentForRoms"":false,""LastRomPath"":"".""}}",
 			[typeof(RecentFiles)] = RECENT_SER,
 			[typeof(RewindConfig)] = @"{""UseCompression"":false,""UseDelta"":false,""Enabled"":true,""AllowSlowStates"":false,""BufferSize"":512,""UseFixedRewindInterval"":false,""TargetFrameLength"":600,""TargetRewindInterval"":5,""AllowOutOfOrderStates"":true,""BackingStore"":0}",
 			[typeof(SaveStateConfig)] = @"{""Type"":0,""CompressionLevelNormal"":1,""CompressionLevelRewind"":0,""MakeBackups"":true,""SaveScreenshot"":true,""BigScreenshotSize"":131072,""NoLowResLargeScreenshots"":false}",
 			[typeof(ToolDialogSettings)] = @"{""_wndx"":52,""_wndy"":44,""Width"":796,""Height"":455,""SaveWindowPosition"":true,""TopMost"":false,""FloatingWindow"":true,""AutoLoad"":false}",
-			[typeof(ZwinderStateManagerSettings)] = ZWINDER_SER,
+			//[typeof(ZwinderStateManagerSettings)] = ZWINDER_SER,
```
YoshiRulz: If this type isn't serialised anymore then remove the test case.

src/BizHawk.Client.Common/movie/tasproj/IStateManagerSettings.cs
```diff
@@ -0,0 +1,63 @@
...
+	public class StateManagerSettingsConverter : JsonConverter<IStateManagerSettings>
+	{
+		public override IStateManagerSettings ReadJson(JsonReader reader, Type objectType, IStateManagerSettings existingValue, bool hasExistingValue, JsonSerializer serializer)
+		{
+			var jsonObject = Newtonsoft.Json.Linq.JObject.Load(reader);
+			if (jsonObject.TryGetValue("$type", out JToken value))
+			{
+				string type = (string)value;
+				if (type.Contains(nameof(PagedStateManager.PagedSettings)))
+					return jsonObject.ToObject<PagedStateManager.PagedSettings>(serializer);
```
YoshiRulz, line 25: braces

src/BizHawk.Client.Common/movie/tasproj/IStateManagerSettings.cs
SuuperW: That would do nothing even if it does work, because we never serialize state managers.
It would be possible to get the same results by modifying the code that does the serialization (with some serializer settings), but I thought doing it here would be better. That maybe is not actually the case given how convoluted it makes things.
But I was also thinking we'd probably eventually remove the old one, or at least drop support for deserializing things that don't contain type info. In that case, we would just delete the converter and referencing attributes when that happens.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
SuuperW: I think the way the greenzone settings form is set up is bad, but I didn't want to change it right now.

src/BizHawk.Tests.Client.Common/config/SerializationStabilityTests.cs
SuuperW: It will get serialized if the user chooses to use the old state manager. I have not reviewed these tests though, perhaps there is something to update.

src/BizHawk.Client.Common/movie/tasproj/IStateManagerSettings.cs
SuuperW: There's lots of no-braces if statements inside PagedStateManager too.

SuuperW: You might be the only contributor who cares about braces. Multiple other recent committers do not use them, so I don't think it makes sense for me to change what I've written here.

YoshiRulz: Multiple other recent committers have been yelled at by yours truly for not following the documented brace style. Just because the warning has been muted doesn't mean you should create more future work. Your IDE should be able to add braces automatically as you type.

vadosnaprimer:
Finally testing! I think it's fine that going back to old frames and rerunning them without editing the movie doesn't fill them with newest and closest-together states. There seems to plenty of room for second-level states there and their gap is quite reasonable. And even that gap only exists until we start editing.

I really love how fast projects save now. My only wish would be to have some variety in configs depending on state size. For systems with giant states current defaults will have to be changed all the time. Ideally we'd have a way for user to have several presets, ideally self-created too. Ideally also in a form that'd allow sharing those presets across different hawk releases.

SuuperW:
I have thought about changing the current config's "default state manager settings" thing to a per-core thing. I kinda want to re-organize all the existing options into a new TAStudio settings dialog first though, since it's already a mess with how many options there are spread across various tool strip menus.

vadosnaprimer (approved):
Approving. We can sort out the nuances later.

vadosnaprimer:
What changed this time? Or was it just a rebase?

SuuperW:
Just a rebase. I was waiting for the build checks to pass before merging, but then I started doing something else and forgot about it.

PR #4377, opened 2025-06-28
Add some hotkeys for markers, and display them in tool tips.

SuuperW:
Hotkeys:
Add marker (default M): Adds a marker to the emulated frame, same as the add marker button under the markers list.
Delete marker (default Ctrl+M): Deletes a marker on the emulated frame
Seek to prev/next marker (default Shift+PageUp/Down): Already existed, they are just now configurable.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
@@ -999,24 +999,40 @@ namespace BizHawk.Client.EmuHawk
 			}
 		}
 
+		public void SetMarker() => SetMarker(Emulator.Frame);
+
+		private void SetMarker(int frame)
+		{
+			TasMovieMarker/*?*/ existingMarker = CurrentTasMovie.Markers.FirstOrDefault(m => m.Frame == frame);
+
+			if (existingMarker != null)
+			{
+				MarkerControl.EditMarkerPopUp(existingMarker);
+			}
+			else
+			{
+				MarkerControl.AddMarker(frame);
+			}
+		}
+
+		public void RemoveMarker()
+		{
+			TasMovieMarker/*?*/ existingMarker = CurrentTasMovie.Markers.FirstOrDefault(m => m.Frame == Emulator.Frame);
+			if (existingMarker == null) return;
+
+			CurrentTasMovie. Markers.Remove(existingMarker);
+			MarkerControl.UpdateMarkerCount();
+		}
+
 		private void TasView_MouseDoubleClick(object sender, MouseEventArgs e)
 		{
 			if (TasView.CurrentCell?.Column is not { Name: var columnName }) return;
 
 			if (e.Button == MouseButtons.Left)
 			{
-				if (!AxisEditingMode && TasView.CurrentCell.RowIndex is not null && columnName is FrameColumnName)
+				if (!AxisEditingMode && columnName is FrameColumnName)
 				{
-					var existingMarker = CurrentTasMovie.Markers.FirstOrDefault(m => m.Frame == TasView.CurrentCell.RowIndex.Value);
-
-					if (existingMarker != null)
-					{
-						MarkerControl.EditMarkerPopUp(existingMarker);
-					}
-					else
-					{
-						MarkerControl.AddMarker(TasView.CurrentCell.RowIndex.Value);
-					}
+					SetMarker(TasView.CurrentCell.RowIndex.Value);
 				}
 			}
 		}
@@ -1469,15 +1485,7 @@ namespace BizHawk.Client.EmuHawk
 		private void TasView_KeyDown(object sender, KeyEventArgs e)
 		{
 			// taseditor uses Ctrl for selection and Shift for frame cursor
-			if (e.IsShift(Keys.PageUp))
-			{
-				GoToPreviousMarker();
-			}
-			else if (e.IsShift(Keys.PageDown))
-			{
-				GoToNextMarker();
-			}
-			else if (e.IsShift(Keys.Home))
+			if (e.IsShift(Keys.Home))
 			{
 				GoToFrame(0);
 			}
--- a/src/BizHawk.Client.Common/config/Binding.cs
+++ b/src/BizHawk.Client.Common/config/Binding.cs
@@ -137,6 +137,10 @@ namespace BizHawk.Client.Common
 			Bind("TAStudio", "Toggle Turbo Seek", "Shift+S");
 			Bind("TAStudio", "Undo", "Ctrl+Z"); // TODO: these are getting not unique enough
 			Bind("TAStudio", "Redo", "Ctrl+Y");
+			Bind("TAStudio", "Seek To Prev Marker", "Shift+PageUp");
+			Bind("TAStudio", "Seek To Next Marker", "Shift+PageDown");
+			Bind("TAStudio", "Set Marker", "M");
+			Bind("TAStudio", "Delete Marker", "Ctrl+M");
 			Bind("TAStudio", "Sel. bet. Markers", "Ctrl+A");
 			Bind("TAStudio", "Select All", "Ctrl+Shift+A");
 			Bind("TAStudio", "Reselect Clip.", "Ctrl+B");
--- a/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Events.cs
@@ -755,6 +755,8 @@ namespace BizHawk.Client.EmuHawk
 
 			InitControls();
 			InputManager.SyncControls(Emulator, MovieSession, Config);
+
+			Tools.HandleHotkeyUpdate();
 		}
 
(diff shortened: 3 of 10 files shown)


YoshiRulz (dismissed):
I don't like adding an event to `Config`. What if `HandleHotkeyUpdate` was a virtual method on `ToolFormBase` and you called it on all open tools at the end of this method?
https://github.com/TASEmulators/BizHawk/blob/adf8e38c7e41554af2e5a1d55879258573935946/src/BizHawk.Client.EmuHawk/MainForm.Events.cs#L750-L758

SuuperW:
I think that would work, but what is wrong with using an event? Personally I think more things in BizHawk should be events. I don't like adding methods to base classes or interfaces just for them to be implemented in only a single class.

YoshiRulz:
It's applicable to other tools too. (If it was just for TAStudio, why set up multicast?)
Events add overhead and are prone to creating subtle bugs.
To guarantee the serialisation remains good, you could use an `Action` property, but again it really needs to be multicast because other tools could use it.

SuuperW:
A virtual method on all tools that most tools don't use (though I don't actually know how many would actually benefit from this particular one) also adds overhead.
Events are somewhat more prone to bugs though. (I really wish handlers would be automatically removed from the event when the owning object is disposed.)

I'll switch the implementation.

vadosnaprimer:
I haven't checked the code but conceptually this PR makes sense. The only difference from the current workflow is that we're adding markers on selected frames using the menu, and to arbitrary frames using double-click. So these hotkeys are kinda unique in functionality by adding it on emulated frame instead. But I don't mind having options for all 3 methods of adding them.

SuuperW:
There is already a button for adding a marker on the emulated frame, but not for removing on the emulated frame.
I could add one for adding to the selected frame too... but I feel like that's not as useful. It's also a bit ambiguous since you can have multiple frames selected. The context menu "Set Markers" option adds a marker to each selected frame which seems pointless but someone probably had a reason for implementing it.

vadosnaprimer:
Yeah no need for extra stuff for now.

SuuperW:
Now uses a virtual method instead of an event.

src/BizHawk.Client.Common/tools/Interfaces/IToolForm.cs
```diff
@@ -58,5 +58,7 @@ public interface IToolForm
 		bool ContainsFocus { get; }
 		void Show();
 		void Close();
+
+		void HandleHotkeyUpdate();
```
YoshiRulz: How should this be implemented, either with or without WinForms?

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -999,24 +999,42 @@ private void TasView_MouseWheel(object sender, MouseEventArgs e)
 			}
 		}
 
+		public void SetMarker(int frame = -1)
+		{
+			if (frame == -1) frame = Emulator.Frame;
```
YoshiRulz: Is `-1` expected via `TasView.CurrentCell.RowIndex`, or is this just to handle the default argument? If it's the latter, you should use `(int?) null`, or have a separate overload `public void SetMarker() => SetMarker(Emulator.Frame);`

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -999,24 +999,42 @@ private void TasView_MouseWheel(object sender, MouseEventArgs e)
 			}
 		}
 
+		public void SetMarker(int frame = -1)
+		{
+			if (frame == -1) frame = Emulator.Frame;
+
+			TasMovieMarker existingMarker = CurrentTasMovie.Markers.FirstOrDefault(m => m.Frame == frame);
```
YoshiRulz:
```suggestion
			var existingMarker = CurrentTasMovie.Markers.FirstOrDefault(m => m.Frame == frame);
```
(I've been writing e.g. `TasMovieMarker/*?*/` where NRTs are disabled, but this doesn't need an explicit type IMO.)

src/BizHawk.Tests/Client.Common/config/SerializationStabilityTests.cs
```diff
@@ -76,8 +77,13 @@ static void CheckAll<T>(string? groupDesc = null)
 				foreach (var mi in t.GetMembers(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic))
 				{
 					if (mi.GetCustomAttribute<JsonIgnoreAttribute>() is not null) continue;
+					if (mi.MemberType == MemberTypes.Event) continue;
```
YoshiRulz: I believe you no longer need to change this test after the refactor, but consider
```suggestion
					if (mi.MemberType is MemberTypes.Event) Assert.Fail("cannot serialise event {t.FullName}.{mi.Name} in {groupDesc}");
```

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: `var` is stupid imo. It is less readable in almost all situations and removes the ability to quickly navigate to definition or find other references since I can't put the cursor over the type name. In this particular case that's not such a big issue but I think it's better to always be explicit unless there's a special reason to not be.
I wouldn't mind using `/*?*/` though.

src/BizHawk.Client.Common/tools/Interfaces/IToolForm.cs
SuuperW: I don't care how one implements it, but I can add a doc comment explaining why it exists. I don't know why WinForms would matter.

YoshiRulz: For ext. tools, `IToolForm` is not necessarily `Form`, hence why I suggested the method be added to `ToolFormBase`. That also guarantees that method overrides can read global `Config`.

SuuperW: I don't follow why the `IToolForm` not being a `Form` matters. It might still care about hotkey updates.
The method is also a virtual method on `ToolFormBase`. (Although I would have instead used a default interface implementation if we could use those in this project.)
Any tool that cares what the hotkeys are I expect to have already obtained a reference to the global `Config` when it first loads so providing it here doesn't seem necessary.

YoshiRulz: Pretty sure only a `Form` can receive keystrokes in our WinForms hierarchy. So no, there would be no need for updating hotkey hints.

It now occurs to me that `ToolManager` should pass in `Config.HotkeyBindings` to the method. We're trying to move away from tools holding references to the global `Config`.

SuuperW: Tools do not receive the keystrokes associated with hotkeys. All hotkeys are handled by `MainForm.CheckHotkey`, which checks if the relevant tool is loaded (if it is a tool hotkey) and directly calls the relevant method. So it does not matter if a tool can receive keystrokes. Besides, a tool could want to display a hotkey that it doesn't provide functionality for. (TAStudio has a tool tip for `Pause` but uses its own special mouse binding instead of the global hotkey. It could have a tooltip for toggle record mode, but that checkbox currently has no tooltip.)

I think passing in the hotkey bindings should wait until the refactor that removes references to the config object. Having it in now would be somewhat confusing, and there are decisions about how to structure the code in said refactor that need to be made and might affect how this is done. I don't know right now what other option might be used, but things seem too interdependent right now to be making assumptions about that future.

YoshiRulz: The fact that ext. tools can't add new hotkeys (which had slipped my mind) is yet another reason why this method should not be added to `IToolForm`.

Displaying hotkeys which aren't directly related to the tool is an interesting use case, but I think it's reasonable for those to inherit `ToolFormBase` or otherwise get a reference to `Config`.

SuuperW: Okay, it's not in `IToolForm` anymore.
I didn't know switching away from an event would bring up so many questions. Maybe an event really was the best way to handle it.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
SuuperW: Is now `/*?*/` here and in `RemoveMarker` too.

src/BizHawk.Tests/Client.Common/config/SerializationStabilityTests.cs
SuuperW: I left this commit out in the new force push, since you're right it isn't needed so not relevant to this PR.
I don't see the point of explicitly failing when events are present, though. Yeah they can't be serialized but that's because there is nothing to be serialized, just like methods.

PR #4380, opened 2025-06-28
Put alt handling code in FormBase. Fixes #3584.

### LINKED ISSUE #3584: When TAStudio is focused, keybinds that include Alt don't work
The default keybind for "Delete Branch" is Alt+Delete, causing some confusion.

Cause appears to be this:
https://github.com/TASEmulators/BizHawk/blob/44bf67e8a3cbdca597ccf07b9a5ce4d54b0e1867/src/BizHawk.Client.EmuHawk/Input/Input.cs#L254-L260
https://github.com/TASEmulators/BizHawk/blob/44bf67e8a3cbdca597ccf07b9a5ce4d54b0e1867/src/BizHawk.Client.EmuHawk/MainForm.cs#L528-L535

SuuperW:
This also allows for the use of alt keys to use the menu items in Virtual Pad. Although for some reason I can't figure out, pressing alt does not underline the relevant letters in the tool bar like it does for other windows.

Note: If a hotkey is set to a key combination that normally would open a menu (e.g. Alt+F, which normally opens the File menu) then the hotkey is triggered and the menu is not opened. This is the same as prior behavior in MainForm, just now applied to tools as well.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/FormBase.cs
+++ b/src/BizHawk.Client.EmuHawk/FormBase.cs
@@ -94,5 +94,40 @@ namespace BizHawk.Client.EmuHawk
 			=> base.Text = Config?.UseStaticWindowTitles == true
 				? (_windowTitleStatic ??= WindowTitleStatic)
 				: WindowTitle;
+
+		// Alt key hacks. We need this in order for hotkey bindings with alt to work.
+		/// <summary>sends a simulation of a plain alt key keystroke</summary>
+		internal void SendPlainAltKey(int lparam)
+		{
+			var m = new Message { WParam = new IntPtr(0xF100), LParam = new IntPtr(lparam), Msg = 0x0112, HWnd = Handle };
+			base.WndProc(ref m);
+		}
+
+		/// <summary>HACK to send an alt+mnemonic combination</summary>
+		internal void SendAltKeyChar(char c) => ProcessMnemonic(c);
+
+		protected override void WndProc(ref Message m)
+		{
+			if (!BlocksInputWhenFocused)
+			{
+				// this is necessary to trap plain alt keypresses so that only our hotkey system gets them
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/FormBase.cs
```diff
@@ -94,5 +94,40 @@ public void UpdateWindowTitle()
 			=> base.Text = Config?.UseStaticWindowTitles == true
 				? (_windowTitleStatic ??= WindowTitleStatic)
 				: WindowTitle;
+
+		// Alt key hacks. We need this in order for hotkey bindings with alt to work.
+		/// <summary>sends a simulation of a plain alt key keystroke</summary>
+		internal void SendPlainAltKey(int lparam)
```
YoshiRulz, line 103: These should be `public`.

YoshiRulz (commented):
I can't really review this since I don't use Windows.

src/BizHawk.Client.EmuHawk/FormBase.cs
SuuperW: Why should they be more visible than they need to be?

YoshiRulz: ~~`FormBase` and the caller, `Input`, aren't semantically related.~~ I misread the diff viewer. My point still stands: They may end up in separate projects in the future.

SuuperW: The caller is `MainForm`.

SuuperW: > My point still stands: They may end up in separate projects in the future.

If these are ever separated, the visibility can be increased at that time. Making everything public just in case of potential future refactoring doesn't make sense.

Issue #4375, opened 2025-06-27
[🐧] TAStudio "Clone # Times" doesn't work


YoshiRulz:
`TasView.Focused` is `false` here under Mono:
https://github.com/TASEmulators/BizHawk/blob/5e7fb1c42fef27b77240d496abac61266f4559c3/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs#L549-L555
Single "Clone" *does* work, which makes me think the modal dialog is what's causing this.

SuuperW:
What could be the reason for checking `TasView.Focused`? I would think the solution to this is to remove that.

SuuperW:
Does #4381 resolve this issue?

YoshiRulz:
No

Morilli:
This is due to the `FramesPrompt` dialog being shown. After `ShowDialog()` returns, the `TasView` is immediately re-focused on windows, but not under mono. I don't wanna debug mono or dig into the source code to try and figure out why so I'm just gonna assume it's another intricate small mono behavior bug.

SuuperW:
@YoshiRulz Please verify that updated #4381 fixes this.

edit: merged https://github.com/TASEmulators/BizHawk/commit/87b268af646acb64f65e1472de0f3e527577ef7f

PR #4381, opened 2025-06-29
Fix a situation where copy/paste hotkeys don't work.

SuuperW:
I sometimes find that I am unable to copy/paste inputs in TAStudio. Maybe this fixes it. It does fix this particular situation:
1) Click the marker list or another control to ensure it has focus
2) Do not move the mouse over the input list
3) Alt+tab to another window
4) Alt+tab back
5) Try to copy with Ctrl+C.

~~This might not fix it in every case. There have been several times when I am unable to copy/paste with the keyboard even after clicking the input list, moving the mouse around, clicking another window then going back to TAStudio... nothing seems to work. I have no idea how to reproduce that, but I do know by using the debugger that when it happens pressing Ctrl+C does not enter `CopyMenuItem_Click`.~~ That issue has been fixed.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
@@ -143,7 +143,7 @@ namespace BizHawk.Client.EmuHawk
 
 		private void SaveSelectionToMacroMenuItem_Click(object sender, EventArgs e)
 		{
-			if (!TasView.Focused && TasView.AnyRowsSelected)
+			if (!TasView.AnyRowsSelected)
 			{
 				return;
 			}
@@ -177,7 +177,7 @@ namespace BizHawk.Client.EmuHawk
 
 		private void PlaceMacroAtSelectionMenuItem_Click(object sender, EventArgs e)
 		{
-			if (!TasView.Focused && TasView.AnyRowsSelected)
+			if (!TasView.AnyRowsSelected)
 			{
 				return;
 			}
@@ -321,7 +321,7 @@ namespace BizHawk.Client.EmuHawk
 
 		private void SelectBetweenMarkersMenuItem_Click(object sender, EventArgs e)
 		{
-			if (TasView.Focused && TasView.AnyRowsSelected)
+			if (TasView.AnyRowsSelected)
 			{
 				var selectionEnd = TasView.SelectionEndIndex ?? 0;
 				var prevMarker = CurrentTasMovie.Markers.PreviousOrCurrent(selectionEnd);
@@ -355,7 +355,7 @@ namespace BizHawk.Client.EmuHawk
 
 		private void CopyMenuItem_Click(object sender, EventArgs e)
 		{
-			if (TasView.Focused && TasView.AnyRowsSelected)
+			if (TasView.AnyRowsSelected)
 			{
 				_tasClipboard.Clear();
 				var list = TasView.SelectedRows.ToArray();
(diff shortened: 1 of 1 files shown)


Morilli:
No idea what steps 3 and 4 are for, they are not required. With that in mind this isn't a "fix", just a behavior change that allows hotkeys for the input roll to work everywhere in the tastudio window, not just when the input roll is focused.

Considering mono does not handle focus correctly (#4375) and I don't see a situation in which not focusing the input roll should do anything meaningful I think the best course of action here is not to try and select the `TasView` on any random user action but rather allow it to not be selected and still handle the corresponding hotkeys.

There's already a random `TasView.Select()` on `TasView_MouseEnter` (the whole reason your step 2 is even necessary), let's not add another. It seems like the best solution is to remove all explicit `Select()` calls and all `.Focused` checks.

SuuperW:
> I think the best course of action here is not to try and select the TasView on any random user action but rather allow it to not be selected and still handle the corresponding hotkeys. It seems like the best solution is to remove all explicit Select() calls and all .Focused checks.

I think that would be good. Probably the way to make copy/paste work would be making them BizHawk hotkeys (instead of relying on the tool strip menu item to handle it via shortcut keys). Like everything else is.

SuuperW:
Removed the focus checks.

The seemingly random `TasView.Focus()` is left. Commit 4d97d062e7591b0dad54774ae4f9513cff968030 says it's about RMB+wheel scroll. It's still a bad solution but simply removing it will cause issues.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
```diff
@@ -143,7 +143,7 @@ private void SaveBk2BackupMenuItem_Click(object sender, EventArgs e)
 
 		private void SaveSelectionToMacroMenuItem_Click(object sender, EventArgs e)
 		{
-			if (!TasView.Focused && TasView.AnyRowsSelected)
+			if (!TasView.AnyRowsSelected)
```
YoshiRulz, line 146: This changes the semantics—but I think it was a typo before and should have read `!TasView.Focused || !TasView.AnyRowsSelected` like elsewhere in this file, so it would now work as intended.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.MenuItems.cs
```diff
@@ -177,7 +177,7 @@ private void SaveSelectionToMacroMenuItem_Click(object sender, EventArgs e)
 
 		private void PlaceMacroAtSelectionMenuItem_Click(object sender, EventArgs e)
 		{
-			if (!TasView.Focused && TasView.AnyRowsSelected)
+			if (!TasView.AnyRowsSelected)
```
YoshiRulz, line 180: ditto

YoshiRulz (approved):
Yes, "Clone # Times" works on my machine with this patch.

Morilli:
> The seemingly random TasView.Focus() is left. [...] simply removing it will cause issues.

Have you tested this? I tested briefly with it removed and didn't notice any difference in behavior.

SuuperW:
Yes, RMB+wheel scroll doesn't work if you click outside the TasView, until it is re-focused.

Morilli:
> Yes, RMB+wheel scroll doesn't work if you click outside the TasView, until it is re-focused.

I tested exactly that and it does work just fine. Did you test this before removing all the `Focused` checks? In general I don't see how you would do a rightclick without also focusing the tasview at the same time.

SuuperW:
I tested it after removing the focused checks.

> In general I don't see how you would do a rightclick without also focusing the tasview at the same time.

Don't right click on the tasview.

Morilli:
> Don't right click on the tasview.

... but then the scroll won't move frames in first place? Are you sure that whatever you tested confirmed that this Select check is necessary?

SuuperW:
Oh, yes you are right. (silly me)
I think the select on mouse enter should be removed.

SuuperW:
Upon further examination, there are still some actions that require TasView focus. In particular, anything handled by `TasView_KeyDown`. The seeks probably should be turned into hotkeys. As for `EditAnalogProgrammatically`, I think it would make sense to end axis editing mode if the TasView loses focus. Once those are done, then the TasView focus on mouse enter will no longer do anything.

Issue #2301, opened 2020-08-13
TAStudio's undo/redo code probably needs a rewrite


RetroEdit:
There are many problems, and I seek to enumerate them in this post:

- It's inefficient, which leads to situations like #1988. There's a fair bit of one-by-one element copying going on. I still find plain copy-pasting of inputs to be pretty slow for large sets of inputs, and I'm pretty sure slow code is responsible.
- It's confusing and hard to understand. I looked into #2293 and found the code to be confusing and slight tweaks resulting in wildly varying outcomes. Because of this inclarity, it probably has bugs.
   - For instance, recently (2020-09-11) I've realized the greenzone invalidation for undoing things is questionable, particularly for undoing multiple things. The greenzone isn't invalidated properly, which could lead to desyncs if one isn't careful.
  - It resets last cursor position, which is inideal. Even undoing a single change can alter the auto-restore cursor's last position.
- It has repetitive code (note the `if (IsRecording || force)` checks):

https://github.com/TASEmulators/BizHawk/blob/0818e40149049939e505abf73167073f58cde477/src/BizHawk.Client.Common/movie/tasproj/TasMovie.History.cs#L326-L409

I think the current code is generally going to be pretty difficult to salvage, and a full rewrite is probably in order. Ideally, we'd write testcases along the way to try and prevent regressions as much as possible. Also, good documentation (and code following that documentation) would help a great deal.

----

I have no plans to do such a rewrite immediately (I view MovieSession cleanup to be of higher priority), but I have thought a bit about what I want.

Presently, I think the actions themselves and the ChangeLog management code should be completely decoupled. I think other TAStudio code (and Lua functions) should call through the ChangeLog interface, which will then call methods that directly carry out the actions. Most likely, no other code should be utilizing the action methods directly (even if they didn't want ChangeLog to be logged, they could just use some bool in the ChangeLog object to turn off the logging).

I think doing this decoupling would clean up recursive loops around undoing/redoing. Currently, every check is parameterized for determining whether it should create a HistoryLog entry, which makes it messier for the HistoryLog itself to apply those operations (e.g. in the process of redoing the same change again, or using inverse operations to carry out the change).

Eventually, being able to serialize the ChangeLog might be nice, for replicating TAStudio operations. This would be useful for bug reports, to demo things, or just plain for internal project testing itself.

My thoughts may change; this problem seems complicated, and I may learn new things as I begin to approach the problem.

----

EDIT 2020-08-27:

[...]

PR #4382, opened 2025-06-29
Undo redo

SuuperW:
Fixes various issues with undo and redo.

There are new tests. I put together a `FakeEmulator` and `FakeMovieSession` to make these tests possible. I just got the tests to run, this may not be the best way to set things up.

Questions:
1) What should happen if the user presses Ctrl+Z in the middle of painting inputs (while the mouse button is down)? The current behavior appears to be that it undoes whatever paint has already been done but the user can still continue painting. In this state, each new painted frame will be its own undo action. Should we disallow undo during paint? Stop the paint? Keep painting but start a new batch?
2) It's been mentioned in #2301 that using undo can move the green arrow. I am not sure what the exact behavior was when that was written, but current behavior will move the green arrow if (and only if) making that same edit manually (e.g. paint an input with the mouse) would move the green arrow. (That is, if the user has manually changed the current frame [seeking by doing anything other than editing inputs], is not currently seeking to the green arrow, and the modified frame is in the past.) This seems reasonable to me, but @RetroEdit might disagree.

Also I am aware of one issue that persists: If the movie has no inputs and you record a frame (which will be frame 0), then undo that frame record, TAStudio will automatically re-record a blank frame on frame 0 which deletes the redo history. This seems like a minor edge case bug and I don't see an easy fix so it may not be worth fixing.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/movie/bk2/StringLogs.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/StringLogs.cs
@@ -37,6 +37,9 @@ namespace BizHawk.Client.Common
 				}
 			}
 
+			if (newLog.Count != currentLog.Count)
+				return Math.Min(newLog.Count, currentLog.Count);
+
 			return null;
 		}
 
--- a/src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
@@ -11,13 +11,7 @@ namespace BizHawk.Client.Common
 
 		public override void RecordFrame(int frame, IController source)
 		{
-			// RetroEdit: This check is questionable; recording at frame 0 is valid and should be reversible.
-			// Also, frame - 1, why?
-			// Is the precondition compensating for frame - 1 reindexing?
-			if (frame != 0)
-			{
-				ChangeLog.AddGeneralUndo(frame - 1, frame - 1, $"Record Frame: {frame}");
-			}
+			ChangeLog.AddGeneralUndo(frame, frame, $"Record Frame: {frame}");
 
 			SetFrameAt(frame, Bk2LogEntryGenerator.GenerateLogEntry(source));
 
@@ -31,10 +25,7 @@ namespace BizHawk.Client.Common
 				InvalidateAfter(frame);
 			}
 
-			if (frame != 0)
-			{
-				ChangeLog.SetGeneralRedo();
-			}
+			ChangeLog.SetGeneralRedo();
 		}
 
 		public override void Truncate(int frame)
--- a/src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/TasMovie.Editing.cs
@@ -164,8 +164,6 @@ namespace BizHawk.Client.Common
 			ShiftBindedMarkers(removeUpTo, removeStart - removeUpTo);
 
 			Changes = true;
-			InvalidateAfter(removeStart);
-
 			ChangeLog.AddRemoveFrames(
 				removeStart,
 				removeUpTo,
@@ -173,6 +171,8 @@ namespace BizHawk.Client.Common
 				removedMarkers,
 				$"Remove frames {removeStart}-{removeUpTo - 1}"
 			);
+
+			InvalidateAfter(removeStart);
 		}
 
 		public void InsertInput(int frame, string inputState)
@@ -188,9 +188,9 @@ namespace BizHawk.Client.Common
 			ShiftBindedMarkers(frame, inputLog.Count());
 
 			Changes = true;
-			InvalidateAfter(frame);
-
 			ChangeLog.AddInsertInput(frame, inputLog.ToList(), $"Insert {inputLog.Count()} frame(s) at {frame}");
+
+			InvalidateAfter(frame);
 		}
 
 		public void InsertInput(int frame, IEnumerable<IController> inputStates)
@@ -258,9 +258,9 @@ namespace BizHawk.Client.Common
 			ShiftBindedMarkers(frame, count);
 
 			Changes = true;
-			InvalidateAfter(frame);
-
 			ChangeLog.AddInsertFrames(frame, count, $"Insert {count} empty frame(s) at {frame}");
+
+			InvalidateAfter(frame);
 		}
 
 		private void ExtendMovieForEdit(int numFrames)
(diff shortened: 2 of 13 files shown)


YoshiRulz:
Testing more systems is definitely welcome! Are you able to add the tests first, or would it not compile?

SuuperW:
I could probably make the tests be the first commit. I only added them when I was trying to figure out why there was a check for frame 0 and why there was a minus 1, but I never did actually figure that out. Both were added with commits that say they fix things without details so I assume they were not super obscure bugs, and probably related to some other code that has since changed.

I also don't think the tests are really relevant to the commits that are before the testing commit. I could try to add some more, but most of the changes are in TAStudio logic which isn't easily testable.

vadosnaprimer:
> Keep painting but start a new batch?

This.

SuuperW:
So it turns out that tying undo batching to TAStudio batching isn't a great idea, because they are two different things and handling the edge cases correctly is confusing. I've re-done this PR (see updated OP for new approach and implications), including more bug fixes and tests.

vadosnaprimer:
> Is this actually any better?

Probably, tho can't tell if it has a functional difference too.

SuuperW:
Force push added a new commit batching recording mode edits. #2301 should be fully resolved with this.

SuuperW:
Can we go ahead and merge this?

Morilli:
I don't feel comfortable reviewing >1k LOC changes, especially when I'm not even familiar with the original logic.

If you think this is a clear improvement over the previous implementation then I'm personally fine with merging and dealing with the inevitable fallout later in fixup patches. Especially the test infrastructure is probably going to help, there's lots of things that should be verified via tests but building the required test structure around it is hard, so having something to work off of is great.

YoshiRulz:
I'd like to start with the unit tests. I had a look at them and they're not structured in the usual way, with a `[TestInitialize]` and `[DataRow]` (or separate methods where they pass delegates).
edit: snipe'd

SuuperW:
> I don't feel comfortable reviewing >1k LOC changes, especially when I'm not even familiar with the original logic.
 
 That's fair. Especially since the original logic was difficult to understand, not well-organized, and inconsistent. I do think this is a clear improvement.

> I'd like to start with the unit tests. I had a look at them and they're not structured in the usual way, with a `[TestInitialize]` and `[DataRow]` (or separate methods where they pass delegates). edit: snipe'd

The tests don't all initialize the same way, so I'm not sure `[TestInitialize]` is relevant. I'm also not sure where you'd want to use `[DataRow]`. But the primary reason I didn't use them is really just that I don't like declarative/attribute-based coding here so I didn't try.

Since you said "snipe'd" I will take that to mean you agree that it's fine to merge.

vadosnaprimer:
I will test it in a couple days. Have the branch checked out, just distracted by something else.

SuuperW:
rebased onto master, will wait for vadosnaprimer's test before merge

YoshiRulz:
> The tests don't all initialize the same way, so I'm not sure `[TestInitialize]` is relevant. I'm also not sure where you'd want to use `[DataRow]`. But the primary reason I didn't use them is really just that I don't like declarative/attribute-based coding here so I didn't try.

I imagined [something like this](https://github.com/YoshiRulz/BizHawk_BranchBackup/blob/undo_redo/src/BizHawk.Tests.Client.Common/Movie/MovieUndoTests.cs).
https://github.com/TASEmulators/BizHawk/pull/4382/commits/8a654545d527e541512cb46a93aaae864f765b9c adds 25 tests but only 11 of them pass, and I was hoping that such a refactor would isolate the failures. With 94 of the now 121 tests passing it sort of did.

> Since you said "snipe'd" I will take that to mean you agree that it's fine to merge.

No? It simply means that Morilli coincidentally posted while I was typing.
I've barely looked at anything apart from the tests.

SuuperW:
> I imagined [something like this](https://github.com/YoshiRulz/BizHawk_BranchBackup/blob/undo_redo/src/BizHawk.Tests.Client.Common/Movie/MovieUndoTests.cs).

That looks absolutely horrid. This is so much more difficult to understand than what I wrote and looks way more error prone and difficult to maintain, since they share code in non-obvious ways. You've also converted some similar tests to a single method with multiple `DataRow` but this just removes the helpful test name (e.g. "RecordFrameZero" and "RecordFrameAtEnd" are now not as obviously being intentionally tested).
And I don't know what you mean by isolate the failures. Are they not already mostly isolated? A few tests are initially affected by two bugs but trying to avoid that would make the tests more convoluted and difficult to understand than they already are.

vadosnaprimer:
~~I'm pressing Ctrl+Z and nothing is happening. Only works if I open the Undo dialog and press its buttons or doubleclick its rows.~~ Shouldn't have mixed up configs, it does work from clean one.

Actual bug:
- Create new project
- Advance one frame.
- Undo
- Turn on Recording mode
- Unpause by hitting `>||`

> System.ArgumentOutOfRangeException: Индекс за пределами диапазона. Индекс должен быть положительным числом, а его размер не должен превышать размер коллекции.
Имя параметра: index
   в System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   в BizHawk.Client.Common.TasMovieChangeLog.MergeLastActions() в /src/BizHawk.Client.Common/movie/tasproj/TasMovie.History.cs:строка 196
   в BizHawk.Client.EmuHawk.TAStudio.FastUpdateAfter() в /src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs:строка 119
   в BizHawk.Client.EmuHawk.TAStudio.UpdateAfter() в /src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs:строка 109
   в BizHawk.Client.EmuHawk.ToolManager.UpdateToolsAfter() в /src/BizHawk.Client.EmuHawk/tools/ToolManager.cs:строка 769
   в BizHawk.Client.EmuHawk.MainForm.StepRunLoop_Core(Boolean force) в /src/BizHawk.Client.EmuHawk/MainForm.cs:строка 3265
   в BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop() в /src/BizHawk.Client.EmuHawk/MainForm.cs:строка 960
   в BizHawk.Client.EmuHawk.Program.SubMain(String[] args) в /src/BizHawk.Client.EmuHawk/Program.cs:строка 377

Otherwise things I tried work ok.

SuuperW:
Thanks, good find. I have fixed that bug.

> Otherwise things I tried work ok.

OK to merge?

vadosnaprimer:
If nobody minds.

SuuperW:
@YoshiRulz do you mind?

YoshiRulz:
I think merging and putting out a release candidate for testing would be a good idea, certainly more useful than a code review from me.

PR #4393, opened 2025-07-07
Make .editorconfig not root, and remove all rules less than a warning.

SuuperW:
This allows users to configure their own style rules for things we allow both ways. It also allows users to make Visual Studio not fade out code that triggers a "silent" style rule.

It's very annoying having Visual Studio reduce contrast of code that violates "silent" rules (for example a new private method that isn't yet called, or using `this` when not necessary), and this is the only way I know how to disable that. (Really if Visual Studio weren't stupid and let the user disable that anti-feature I wouldn't care about this .editorconfig change).

Check if completed:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,3 @@
-root = true
-
 [*]
 indent_style = tab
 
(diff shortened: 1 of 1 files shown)


YoshiRulz:
The intent is for all devs to see the same warnings and suggestions on the same code. (Desregarding the fact that Kate isn't currently showing these `IDExxxx` rules, and that Rider adds extra rules on top of any Analyzers.)

> It's very annoying having Visual Studio reduce contrast of code that violates "silent" rules

The way you've described it, that does sound like a bug. Per [docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#severity-level), `silent` indicates that the diagnostic is not shown, but may appear as a quick fix option and be applied to code generated by the IDE.

I'd be open to further lowering the severity of some rules.

Morilli:
I disagree with most of this. The editorconfig *should* be root to prevent parsing arbitrary editorconfig files from above this repository's directory, and you should be able to ignore slightly faded code (or configure it, idk how visual studio works).

I do believe that if visual studio is highlighting silent rules in the IDE that that constitutes a bug (they should be silenced, subtly showing something should be the suggestion level).

That said, many of the rules that are currently marked as `silent` shouldn't be. If we care slightly about how code should look there's `suggestion`, and if we really care we can use `warning` or `error`. Otherwise, they should be marked as `none` if they are otherwise visible.

I have used `silent` in the past to mark rules that I think should be enabled but cannot be without changing many lines of code. Randomly removing `silent` and `suggestion` rules is not the way (especially without checking what their default are).

SuuperW:
> The editorconfig should be root to prevent parsing arbitrary editorconfig files from above this repository's directory

Why? Any "arbitrary files" it would be parsing were put there by the user. Anything that conflicts with what's in ours will be ignored.

> slightly faded code

It's not slightly faded.

> or configure it
> I do believe that if visual studio is highlighting silent rules in the IDE that that constitutes a bug (they should be silenced, subtly showing something should be the suggestion level).

I agree. I reported the bug months ago. Ideally VS would put the user in control of how code appears and what shows up in the editor.

> That said, many of the rules that are currently marked as silent shouldn't be. If we care slightly about how code should look there's suggestion, and if we really care we can use warning or error. Otherwise, they should be marked as none if they are otherwise visible.

That would actually work for me. "suggestion" is something I can tolerate.


I think the best solution would be to remove .editorconfig entirely (because it doesn't work well) and have a script that flags code style issues. We could configure the build to run this script and fail if there are errors. Maybe use the same scripts that run automatically for Github PRs?

YoshiRulz:
>> The editorconfig should be root to prevent parsing arbitrary editorconfig files from above this repository's directory
>
> Why? Any "arbitrary files" it would be parsing were put there by the user. Anything that conflicts with what's in ours will be ignored.

Imagine someone has included BizHawk as a submodule in [their project](https://gitlab.com/tasbot/movie-manip), but they use a different brace style and want to enforce it with EditorConfig. That might cause BizHawk submodule builds to fail.

> [...] and have a script that flags code style issues. We could configure the build to run this script and fail if there are errors. Maybe use the same scripts that run automatically for Github PRs?

This exists, it's called Roslyn, and we're already using it.

SuuperW:
> Imagine someone has included BizHawk as a submodule in [their project](https://gitlab.com/tasbot/movie-manip), but they use a different brace style and want to enforce it with EditorConfig. That might cause BizHawk submodule builds to fail.

I don't see the problem. They'd put BizHawk's code at `bizhawk/bizhawk` and create a `bizhawk/.editorconfig` file that says it is root.

> This exists, it's called Roslyn, and we're already using it.

Visual Studio won't tell me anything about suggestions or even warnings on files that aren't currently open. ...but I just tested and `dotnet build` does bring up a warning for IDE0005 at least if I add a violation, so.... maybe it's a bug in VS? I don't know why it would not show warnings that `dotnet build` does. It is set to "entire solution". But yeah it seems we do already have such a script.

SuuperW:
It seems I can get Visual Studio to not fade out my text by overriding the relevant rules in `src/.editorconfig` and adding that to my local exclude file.

Issue #3337, opened 2022-07-24
API Hawk: Cannot load configuration file when attempting to load persisted configurations for external tools.


tommai78101:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"
[//]: # "One more thing: if you're on Linux, please comment on #1430 instead of opening an issue so we don't annoy the other devs."

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."

If the external tool is opened at least once, and the external tool uses the following C# codes below, in theory, the configuration file `config.ini` will include information of the external tool.

```csharp
[ConfigPersist]
public GeneticAlgorithmBotSettings Settings { get; set; }
```

```csharp
public class GeneticAlgorithmBotSettings {
	public RecentFiles RecentBotFiles { get; set; } = new RecentFiles();
	public bool TurboWhenBotting { get; set; } = true;
	public bool InvisibleEmulation { get; set; }
}
```

Below is the snippet of the `config.ini`:

```json
  ...

  "CommonToolSettings": {
    "BizHawk.Client.EmuHawk.TAStudio": {
      "_wndx": 543,
      "_wndy": 55,
      "Width": 525,
      "Height": 615,
      "SaveWindowPosition": true,
      "TopMost": false,
      "FloatingWindow": true,
      "AutoLoad": false
    },
    "GeneticAlgorithmBot.GeneticAlgorithmBot": {
      "_wndx": 1091,
      "_wndy": 149,
      "Width": 723,
      "Height": 626,
      "SaveWindowPosition": true,
      "TopMost": false,
      "FloatingWindow": true,
      "AutoLoad": false
    }
  },
  "CustomToolSettings": {
    "BizHawk.Client.EmuHawk.TAStudio": {
      "Settings": {
        "$type": "BizHawk.Client.EmuHawk.TAStudio+TAStudioSettings, EmuHawk",
        "RecentTas": {
          "recentlist": [],
          "MAX_RECENT_FILES": 8,
          "AutoLoad": false,
          "Frozen": false
        },
        "AutoPause": true,
[...]

PR #4395, opened 2025-07-09
Some style rule updates with fixes to existing violations

SuuperW:
Interestingly, IDE0038 `Use pattern matching to avoid is check followed by a cast (without variable)` doesn't work with `dotnet build` but does work inside Visual Studio.

There are quite a few other rules I find stupid (like preferring `var` over explicit types; explicit types are superior) but other people probably disagree so I just did the ones I think pretty much anyone would agree with.

While fixing mixed use of tabs and spaces, a few files had majority spaces but a few lines with tabs. They're all tabs now. There are still files that use spaces exclusively.

Check if completed:


SuuperW:
Okay so apparently the build checks for PRs will flag private unused members that are used via attributes+reflection, but `dotnet build` will not. So this needs more work.

EDIT: Okay but what about BizHawk.Emulation.Cores/Consoles/Nintendo/GBHawk/Audio.cs(1235,16): 'Audio.GetSamples'? Why isn't that flagged when I run `dotnet build`? EDIT: Oh it was probably because I made a local editorconfig that set its severity to none as a workaround for a VS bug. ...I thought I did a working build without that though. My bad.

YoshiRulz:
Oh, I have a local branch which cleans up mixed tabs+spaces, let me dust that off.

You shouldn't try to remove unused fields in at least the Cores project. Assume the core maintainer is in the middle of a refactoring.

SuuperW:
There were some violations (including some with explicit suppression) of unused/unread members that were actually used but only via reflection. I refactored some stuff to fix these, using less reflection. Old code is deleted, but maybe it shouldn't be as that breaks external tools (until they refactor themselves, which isn't hard).

Morilli:
428d1feae143d876e61c910392c96fc892c5214f is a breaking change and should not be in this PR imo as it's less a style change and more an actual functional change.

e7264721cbdc25781f0d6e9a3f938832bf0933de looks fine and while technically breaking is very easy to fix. I just wonder: Will this break existing external tools silently or does it error on load? In any case this change should be communicated clearly in changelog and whatever.

95b70f2538fcca81841ba6875b4a062aa53cd4aa looks fine (i've definitely looked at every single line change) but should be added to `.git-blame-ignore-revs` after. 

Rest is fine.

SuuperW:
> [428d1fe](https://github.com/TASEmulators/BizHawk/commit/428d1feae143d876e61c910392c96fc892c5214f) is a breaking change and should not be in this PR imo as it's less a style change and more an actual functional change.

> [e726472](https://github.com/TASEmulators/BizHawk/commit/e7264721cbdc25781f0d6e9a3f938832bf0933de) looks fine and while technically breaking is very easy to fix. I just wonder: Will this break existing external tools silently or does it error on load? In any case this change should be communicated clearly in changelog and whatever.

Yeah I wasn't sure if either of these belonged here. They're both easy updates for external tools to make. If this PR is otherwise good to merge I'll remove one or both and do in-source suppression.

Is there a place I can put a note about these changes for changelog?

src/BizHawk.Client.Common/cheats/GbaGameSharkDecoder.cs
```diff
@@ -28,21 +28,21 @@ private static string Decrypt(string code)
...
-		private static string DecryptPro(string code)
-		{
-			var sum = 0xC6EF3720;
-			var op1 = uint.Parse(code.Remove(8, 9), NumberStyles.HexNumber);
-			var op2 = uint.Parse(code.Remove(0, 9), NumberStyles.HexNumber);
+		//private static string DecryptPro(string code)
+		//{
+		//	var sum = 0xC6EF3720;
+		//	var op1 = uint.Parse(code.Remove(8, 9), NumberStyles.HexNumber);
+		//	var op2 = uint.Parse(code.Remove(0, 9), NumberStyles.HexNumber);
```
YoshiRulz: Rather than commenting out lines, move the `static` field beside this method and wrap the whole thing in `#if false`.

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -17,9 +17,6 @@ namespace BizHawk.Client.Common
 	[Description("A library for manipulating the EmuHawk client UI")]
 	public sealed class ClientLuaLibrary : LuaLibraryBase
 	{
-		[RequiredService]
-		private IEmulator Emulator { get; set; }
```
YoshiRulz: Looks like the core is accessed via `this.MainForm` instead for some reason?

src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
```diff
@@ -10,8 +9,6 @@ namespace BizHawk.Client.Common
 	[Description("A library for communicating with other programs")]
 	public sealed class CommLuaLibrary : LuaLibraryBase
 	{
-		private readonly IDictionary<Guid, ClientWebSocketWrapper> _websockets = new Dictionary<Guid, ClientWebSocketWrapper>();
```
YoshiRulz, line 13: Keep this, but in `#if ENABLE_WEBSOCKETS`

src/BizHawk.Client.EmuHawk/AVOut/SynclessRecorder.cs
```diff
@@ -113,33 +112,6 @@ public void SetMetaData(string gameName, string authors, ulong lengthMs, ulong r
 
 		public string DesiredExtension() => "syncless.txt";
 
-		/// <summary>
-		/// splits the string into chunks of length s
-		/// </summary>
-		private static List<string> StringChunkSplit(string s, int len)
```
YoshiRulz: Keep this in `#if false`.

src/BizHawk.Client.EmuHawk/tools/Debugger/AddBreakpointDialog.cs
```diff
@@ -23,12 +23,8 @@ public AddBreakpointDialog(BreakpointOperation op, uint address, uint mask, Memo
 			BreakType = type;
 		}
 
-		private BreakpointOperation _operation;
-
 		private BreakpointOperation Operation
 		{
-			get => _operation;
-
 			set
```
YoshiRulz: `set`-only prop? This can be a local method in the constructor.

src/BizHawk.Client.Common/config/IConfigPersist.cs
```diff
@@ -0,0 +1,56 @@
...
+		public class Provider
+		{
+			private Dictionary<string, object> _data;
+
+			public Provider(Dictionary<string, object> data)
+			{
+				_data = data;
+			}
+
+			public bool Get<T>(string name, ref T value)
```
YoshiRulz, line 18: Shouldn't `value` be `out`? It seems that way from the call-sites.

Also call this `TryGet`.

src/BizHawk.Client.Common/config/ConfigPersistAttribute.cs
```diff
@@ -1,8 +0,0 @@
-namespace BizHawk.Client.Common
-{
-	/// <summary>Indicates that a property is to be saved to config for persistence.</summary>
-	[AttributeUsage(AttributeTargets.Property)]
-	public sealed class ConfigPersistAttribute : Attribute
```
YoshiRulz, line 5: As Morilli said, this needs to stay to give ext. tools time to migrate.

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -963,6 +958,7 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
 			return base.ProcessCmdKey(ref msg, keyData);
 		}
 
+#pragma warning disable IDE0051 // Remove unread private members (We might wish to do something with this. Was removed due to being broken.)
```
YoshiRulz: `#if false`

src/BizHawk.Client.Common/config/IConfigPersist.cs
```diff
@@ -0,0 +1,56 @@
...
+
+		/// <summary>
+		/// Load the provided configuration.
+		/// </summary>
+		void LoadConfig(Provider provider);
+
+		/// <summary>
+		/// Return a dictionary representing the current configuration.
+		/// </summary>
+		Dictionary<string, object> SaveConfig();
```
YoshiRulz, line 54: Does the JSON serialiser require a `Dictionary<,>`?

src/BizHawk.Emulation.DiscSystem/DiscFormats/Blobs/Blob_ECM.cs
```diff
@@ -118,7 +118,7 @@ public void Load(string path)
 			//TODO - endian bug. need an endian-independent binary reader with good license (miscutils is apache license)
 			//extension methods on binary reader wont suffice, we need something that lets you control the endianness used for reading. a complete replacement.
 			var br = new BinaryReader(stream);
-			EDC = br.ReadInt32();
+			/*EDC =*/ br.ReadInt32();
```
YoshiRulz:
```suggestion
			/*EDC*/_ = br.ReadInt32();
```

src/BizHawk.Emulation.DiscSystem/DiscFormats/Blobs/Blob_ECM.cs
```diff
@@ -47,10 +47,10 @@ public IndexEntry(int type, uint number, long ecmOffset, long logicalOffset)
...
 		private readonly List<IndexEntry> Index = new();
 
-		/// <summary>
-		/// the ECMfile-provided EDC integrity checksum. not being used right now
-		/// </summary>
-		private int EDC;
+		///// <summary>
+		///// the ECMfile-provided EDC integrity checksum. not being used right now
+		///// </summary>
+		//private int EDC;
```
YoshiRulz: `#if false`

src/BizHawk.Emulation.DiscSystem/DiscUtils.cs
```diff
@@ -2,11 +2,11 @@ namespace BizHawk.Emulation.DiscSystem
...
-		private static byte IntToBCD(int n)
-		{
-			var tens = Math.DivRem(n, 10, out var ones);
-			return (byte)((tens << 4) | ones);
-		}
+		//private static byte IntToBCD(int n)
+		//{
+		//	var tens = Math.DivRem(n, 10, out var ones);
+		//	return (byte)((tens << 4) | ones);
+		//}
```
YoshiRulz: Just remove this one since it's redundant with with `BCD2.IntToBCD` (`DiscTypes.cs`)

YoshiRulz:
I checked upstream and `IDE0038` not working in CLI was closed as wontfix -_- and it doesn't work in Rider either.

Which rule was flagging the blank lines at the end of scopes?

src/BizHawk.Client.Common/config/IConfigPersist.cs
SuuperW: Making it `ref` greatly simplifies most call sites as there's no need for an intermediate variable (with ref we can set the variable we want directly, instead of setting an out param then the caller having to assign that to the variable) and also because it makes type inference work much nicer.

src/BizHawk.Client.Common/config/IConfigPersist.cs
SuuperW: No, I just wanted to make this change as small as reasonably possible. The stuff was already getting shoved into a dictionary.

SuuperW:
> Which rule was flagging the blank lines at the end of scopes?

"Fix violations of SA1508 and SA1509 (braces should [not] be preceded by blank line). Don't enable them because they flag some silly things."

src/BizHawk.Client.Common/config/IConfigPersist.cs
YoshiRulz: You can use `out` without `out var` if that's what you mean.

src/BizHawk.Client.Common/config/IConfigPersist.cs
YoshiRulz: If a return type of `IEnumerable<KeyValuePair<string, object>>` serialises correctly, the implementations can use `yield`.

src/BizHawk.Client.Common/config/IConfigPersist.cs
SuuperW: I should have mentioned, using `out` introduces these problems because I wanted the uses of `Get` (or TryGet) method to not touch `value` if no value is found. Since an `out` param must be set within the method, that is why it would require an intermediate variable.

If we do `bool TryGet<T>(string name, out T value)` then `public string SomePersistentString;` and `TryGet("SomePersistentString", out SomePersistentString)` then the persisted value would be assigned `default` which likely isn't a sensible default. So `LoadConfig` would be required to check the return value and if false assign a sensible default. Keeping default definitions in just one place, outside of `LoadConfig`, seems cleaner.

src/BizHawk.Client.Common/config/IConfigPersist.cs
SuuperW: I do not know if it would serialise, but a quick attempt runs into this error at ToolManager.cs:141:

`IEnumerable<KeyValuePair<string, object>>' must be a non-abstract type with a public parameterless constructor in order to use it as parameter 'TValue' in the generic type or method 'CollectionExtensions.GetValueOrPutNew<TKey, TValue>(IDictionary<TKey, TValue>, TKey)`

So you'd need to resolve that somehow. I also don't see any benefit to changing an implementation such as
```cs
return new()
{
	[nameof(Settings)] = Settings,
	[nameof(TasViewFont)] = TasViewFont,
};
```
into
```cs
yield return new(nameof(Settings), Settings);
yield return new(nameof(TasViewFont), TasViewFont);
```

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
SuuperW: Using the `Emulator` property is better imo.
I'm not sure if `Emulator` ever was used, but the current uses of `MainForm.Emulator` were new methods added after the `Emulator` property, so that's weird.

I'll make it use the `Emulator` property instead of deleting it.

src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
SuuperW: Prefer
```cs
#if ENABLE_WEBSOCKETS
		private readonly System.Collections.Generic.IDictionary<Guid, ClientWebSocketWrapper> _websockets = new System.Collections.Generic.Dictionary<Guid, ClientWebSocketWrapper>();
#endif
```
or wrapping the using directive in its own #if directive?

src/BizHawk.Emulation.DiscSystem/DiscUtils.cs
SuuperW: Will also remove the unused public `BCD_Byte` below it, since it actually just does the same thing and has the same visibility as `BCD2`.

src/BizHawk.Client.Common/config/ConfigPersistAttribute.cs
SuuperW: Should I also leave in `RestoreDefaultsAttribute`?

src/BizHawk.Client.Common/lua/CommonLibs/CommLuaLibrary.cs
YoshiRulz: Yeah do the same for the `using` if it would become unused otherwise

src/BizHawk.Client.Common/config/ConfigPersistAttribute.cs
YoshiRulz: I think so.
The attributes can be `[Obsolete(IsError = true)]` if you can't keep them functional.

src/BizHawk.Client.Common/config/IConfigPersist.cs
YoshiRulz: `GetValueOrPutNew` would be something like `.FirstOrNull(kvp => kvp.Key == toolTypeName)?.Value ?? new()`, but you couldn't do that without enumerating the whole collection anyway, so nevermind.
The benefits would have been avoiding creating a hashmap and I think reducing allocation.

src/BizHawk.Client.Common/config/ConfigPersistAttribute.cs
SuuperW: My toy external tool with both obsoleted attributes works in both 2.10 and this branch now.

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
YoshiRulz: https://github.com/TASEmulators/BizHawk/commit/f1fe5a03eeff69e29eb5e13edb2d615209dc1c73

YoshiRulz:
I've pushed a bunch of code style commits to `master`, including https://github.com/TASEmulators/BizHawk/commit/71d3d659b3629a21a7bb2c417c24bd4d3b4aed5c which addresses `SA1508` (`SA1509` seemed unhelpful), and https://github.com/TASEmulators/BizHawk/commit/ef9d64eb8af139fffc6d3f2dcdb321699b760d51 which adds `.editorconfig`s for individual cores. Rebasing should make your diff much smaller now.

SuuperW:
> I've pushed a bunch of code style commits to `master`, including [71d3d65](https://github.com/TASEmulators/BizHawk/commit/71d3d659b3629a21a7bb2c417c24bd4d3b4aed5c) which addresses `SA1508` (`SA1509` seemed unhelpful), and [ef9d64e](https://github.com/TASEmulators/BizHawk/commit/ef9d64eb8af139fffc6d3f2dcdb321699b760d51) which adds `.editorconfig`s for individual cores. Rebasing should make your diff much smaller now.

There are a couple hundred fewer line removals now, but the vast majority of line changes are indentation changes which hasn't changed much.

src/BizHawk.Client.Common/config/IConfigPersist.cs
YoshiRulz: > [...] then the persisted value would be assigned default which likely isn't a sensible default. So LoadConfig would be required to check the return value and if false assign a sensible default.

That sounds reasonable to me. The `bool TryX(out T result)` pattern is well-established.

I think the proper way to do fire-and-forget would be a `void Get<T>(string name, ref T value)` for fields and a `void Get<T>(string name, Action<T> setValue)` for props.

src/BizHawk.Client.Common/config/IConfigPersist.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace BizHawk.Client.Common
+{
+	public interface IConfigPersist
```
YoshiRulz, line 7: Does this new approach also suffer from #3337?

SuuperW: Yes.

Issue #4258, opened 2025-03-04
(issue title -- summarise the summary)


ladybunne:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
When BizHawk attempts to save to a file that is being accessed by Dropbox, an error dialog will pop up, then BizHawk will crash entirely.

I'm on Windows, using a personally modified version of BizHawk forked from late 2.9, with some minor changes to file extensions - nothing that would affect this.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Run BizHawk while its save directory is pointed at a directory managed by Dropbox.
2. Play a game, optionally with autosaving enabled.
3. Wait for a clash between BizHawk autosaving and Dropbox syncing.

### Output
[//]: # "Paste the contents of the error dialog if there is one (try Ctrl+C, it usually works), or paste the output from the Lua Console, or delete this section."
```
System.IO.IOException: The process cannot access the file 'C:\Users\Ladybunne\Dropbox\Media\Games\Emulation\SaveRAM\GBA\Yu-Gi-Oh! - Ultimate Masters - World Championship Tournament 2006 (USA) (En,Ja,Fr,De,Es,It).Autosave.srm.bak' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileInfo.Delete()
   at BizHawk.Client.EmuHawk.MainForm.FlushSaveRAM(Boolean autosave) in C:\Users\Ladybunne\repos\BizHawk\src\BizHawk.Client.EmuHawk\MainForm.cs:line 2002
   at BizHawk.Client.EmuHawk.MainForm.StepRunLoop_Core(Boolean force) in C:\Users\Ladybunne\repos\BizHawk\src\BizHawk.Client.EmuHawk\MainForm.cs:line 3051
[...]

Issue #4329, opened 2025-05-23
Stop returning Lua tables with 0-indexing


YoshiRulz:
These are non-idiomatic and annoying to work with (though IMO the normal 1-indexed tables are also annoying).
Switching a function's indexing is a breaking change, so we should plan to change them all at once. The candidates are:
- TODO

SuuperW:
Will need to add functions that take a 0-based index as an argument too. Idk if any exist, but I saw the update in #4335.

Also if we do this, can we please also standardize the function names? E.g. make them all use underscores, `client.getavailabletools` -> `client.get_available_tools`.

Issue #4242, opened 2025-02-25
Stop hardcoding core SaveRAM metadata into frontend, use a service


YoshiRulz:
https://github.com/TASEmulators/BizHawk/blob/f6538289dc7adb9c76862bd426c17a57c1ed4dab/src/BizHawk.Client.EmuHawk/MainForm.cs#L2000-L2005

see also #2344

Issue #4173, opened 2025-01-07
Unify emu.setislagged, emu.setlagcount, and tastudio.setlag


warmCabin:
### Summary
There are three separate ways of looking at lag in the Hawk Lua API: the lag flag (`emu.setislagged`), lag counter (`emu.setlagcount`), and TAStudio red bars (`tastudio.setlag`). These three values should be the three different ways of looking at the same information, but are in fact completely unrelated. There should be one simple `setlag` call that keeps all these values in sync.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
```lua
while true do
    emu.setislagged(math.random() < 0.5)
    emu.setlagcount(math.random() * 1000)
    tastudio.setlag(emu.framecount(), math.random() < 0.5)

    emu.frameadvance()
end
```

### Host env.
[//]: # "List the computers you've found the bug with. If there's a version that doesn't have the bug, please put that in too. Here are some examples:"
- BizHawk 2.10.0 Windows 10

YoshiRulz:
See also #3447 and https://github.com/TASEmulators/BizHawk/issues/3447#issuecomment-1321218179.

SuuperW:
What is the purpose of setting the lag value?

Traditionally, a "lag frame" is one where no input is polled. If we are using this meaning, it makes no sense to allow Lua to set a frame's lag state. TAStudio has (used to have) a feature to automatically adjust input according to lag, and it relies on this particular meaning of lag. (It was removed due to "people not complaining about it being broken", but it was very useful to me the one single time I did a TAS of a game with lag mid-gameplay so... I rather think it's worth having.)

By default, melonDS uses an "alternate" meaning of lag that just ignores touch screen input polling. (which is on by default, but the setting calls the alternate meaning "False" for some reason) This is more accurate to what is generally considered lag, shows lag while the game is loading stuff, and is usually but not always indicative of the game actually using inputs on that frame. This is still close to the traditional meaning and still makes no sense to allow Lua to set the lag value. (An example of the game using input on a "lag frame" is in certain level loads in Mario Kart DS, one specific frame near the end of the "lag" area will change RNG if there is touch input.) But this being the default indicates that we have officially abandoned the traditional meaning.

[...]

PR #4400, opened 2025-07-13
Handle exceptions while writing to files

SuuperW:
Most file writes now use a new system that allows for error detection and handling. (includes movies, config file, and all tools)
When an exception occurs, a message box appears alerting the user. If there is an exception while doing a save-on-close, the user is given the option to try again, skip saving, or cancel closing. (Includes a bug fix where an existing dialog offering the cancel option did not actually cancel.)

Addresses #4258.

Fixes an issue where the CheatList was being forcibly autosaved when closing a game, ignoring the autosave config option.

Changes behavior when saving the config file fails. Previously it would silently fail (and report success if "saved" via the menu item), now it has the same error handling as other updated file writes. The silent fail seems to have been intended to allow using a read-only config file. If we want to bring this back, I suggest making a config option to not save on close instead. As-is, a read-only file will still work you just have to click "No" after closing EmuHawk when it tells you it can't save the config file.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/MovieApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/MovieApi.cs
@@ -52,6 +52,7 @@ namespace BizHawk.Client.Common
 			return Bk2LogEntryGenerator.GenerateLogEntry(_movieSession.Movie.GetInputState(frame));
 		}
 
+		// TODO: Change return type to FileWriteResult
 		public void Save(string filename)
 		{
 			if (_movieSession.Movie.NotActive())
@@ -69,7 +70,8 @@ namespace BizHawk.Client.Common
 				}
 				_movieSession.Movie.Filename = filename;
 			}
-			_movieSession.Movie.Save();
+			FileWriteResult result = _movieSession.Movie.Save();
+			if (result.Exception != null) throw result.Exception;
 		}
 
 		public IReadOnlyDictionary<string, string> GetHeader()
--- a/src/BizHawk.Client.Common/DialogControllerExtensions.cs
+++ b/src/BizHawk.Client.Common/DialogControllerExtensions.cs
@@ -1,6 +1,7 @@
 #nullable enable
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 
 namespace BizHawk.Client.Common
@@ -63,6 +64,20 @@ namespace BizHawk.Client.Common
 			string? caption = null,
 			EMsgBoxIcon? icon = null)
 				=> dialogParent.DialogController.ShowMessageBox3(owner: dialogParent, text: text, caption: caption, icon: icon);
+
+		public static void ErrorMessageBox(
+			this IDialogParent dialogParent,
+			FileWriteResult fileResult,
+			string? prefixMessage = null)
+		{
+			Debug.Assert(fileResult.IsError && fileResult.Exception != null, "Error box must have an error.");
+
+			string prefix = prefixMessage == null ? "" : prefixMessage + "\n";
+			dialogParent.ModalMessageBox(
+				text: $"{prefix}{fileResult.UserFriendlyErrorMessage()}\n{fileResult.Exception!.Message}",
+				caption: "Error",
+				icon: EMsgBoxIcon.Error);
+		}
 
 		/// <summary>Creates and shows a <c>System.Windows.Forms.OpenFileDialog</c> or equivalent with the receiver (<paramref name="dialogParent"/>) as its parent</summary>
 		/// <param name="discardCWDChange"><c>OpenFileDialog.RestoreDirectory</c> (isn't this useless when specifying <paramref name="initDir"/>? keeping it for backcompat)</param>
--- a/src/BizHawk.Client.Common/DialogControllerExtensions.cs
+++ b/src/BizHawk.Client.Common/DialogControllerExtensions.cs
@@ -72,11 +72,38 @@ namespace BizHawk.Client.Common
 		{
 			Debug.Assert(fileResult.IsError && fileResult.Exception != null, "Error box must have an error.");
 
-			string prefix = prefixMessage == null ? "" : prefixMessage + "\n";
+			string prefix = prefixMessage ?? "";
 			dialogParent.ModalMessageBox(
-				text: $"{prefix}{fileResult.UserFriendlyErrorMessage()}\n{fileResult.Exception!.Message}",
+				text: $"{prefix}\n{fileResult.UserFriendlyErrorMessage()}\n{fileResult.Exception!.Message}",
 				caption: "Error",
 				icon: EMsgBoxIcon.Error);
+		}
+
+		/// <summary>
+		/// If the action fails, asks the user if they want to try again.
+		/// The user will be repeatedly asked if they want to try again until either success or the user says no.
+		/// </summary>
+		/// <returns>Returns true on success or if the user said no. Returns false if the user said cancel.</returns>
+		public static bool DoWithTryAgainBox(
+			this IDialogParent dialogParent,
+			Func<FileWriteResult> action,
+			string message)
+		{
+			FileWriteResult fileResult = action();
+			while (fileResult.IsError)
+			{
+				string prefix = message ?? "";
+				bool? askResult = dialogParent.ModalMessageBox3(
+						text: $"{prefix} Do you want to try again?\n\nError details:" +
+							$"{fileResult.UserFriendlyErrorMessage()}\n{fileResult.Exception!.Message}",
+						caption: "Error",
+						icon: EMsgBoxIcon.Error);
+				if (askResult == null) return false;
+				if (askResult == false) return true;
+				if (askResult == true) fileResult = action();
+			}
+
+			return true;
 		}
 
 		/// <summary>Creates and shows a <c>System.Windows.Forms.OpenFileDialog</c> or equivalent with the receiver (<paramref name="dialogParent"/>) as its parent</summary>
--- a/src/BizHawk.Client.Common/DialogControllerExtensions.cs
+++ b/src/BizHawk.Client.Common/DialogControllerExtensions.cs
@@ -6,6 +6,13 @@ using System.Diagnostics.CodeAnalysis;
 
 namespace BizHawk.Client.Common
 {
+	public enum TryAgainResult
+	{
+		Saved,
+		IgnoredFailure,
+		Canceled,
+	}
+
 	public static class DialogControllerExtensions
 	{
 		public static void AddOnScreenMessage(
@@ -84,7 +91,7 @@ namespace BizHawk.Client.Common
 		/// The user will be repeatedly asked if they want to try again until either success or the user says no.
 		/// </summary>
 		/// <returns>Returns true on success or if the user said no. Returns false if the user said cancel.</returns>
-		public static bool DoWithTryAgainBox(
+		public static TryAgainResult DoWithTryAgainBox(
 			this IDialogParent dialogParent,
 			Func<FileWriteResult> action,
 			string message)
@@ -98,12 +105,12 @@ namespace BizHawk.Client.Common
 							$"{fileResult.UserFriendlyErrorMessage()}\n{fileResult.Exception!.Message}",
 						caption: "Error",
 						icon: EMsgBoxIcon.Error);
-				if (askResult == null) return false;
-				if (askResult == false) return true;
+				if (askResult == null) return TryAgainResult.Canceled;
+				if (askResult == false) return TryAgainResult.IgnoredFailure;
 				if (askResult == true) fileResult = action();
 			}
 
-			return true;
+			return TryAgainResult.Saved;
 		}
 
 		/// <summary>Creates and shows a <c>System.Windows.Forms.OpenFileDialog</c> or equivalent with the receiver (<paramref name="dialogParent"/>) as its parent</summary>
--- a/src/BizHawk.Client.Common/FileWriteResult.cs
+++ b/src/BizHawk.Client.Common/FileWriteResult.cs
@@ -1,7 +1,6 @@
 ﻿#nullable enable
 
 using System.Diagnostics;
-using System.IO;
 
 namespace BizHawk.Client.Common
 {
@@ -10,6 +9,8 @@ namespace BizHawk.Client.Common
 		Success,
 		FailedToOpen,
 		FailedDuringWrite,
+		FailedToDeleteOldBackup,
+		FailedToMakeBackup,
 		FailedToDeleteOldFile,
 		FailedToRename,
 	}
@@ -21,16 +22,18 @@ namespace BizHawk.Client.Common
 	{
 		public readonly FileWriteEnum Error = FileWriteEnum.Success;
 		public readonly Exception? Exception;
-		public readonly string WritePath;
+		internal readonly FileWritePaths Paths;
 
 		public bool IsError => Error != FileWriteEnum.Success;
 
-		public FileWriteResult(FileWriteEnum error, string path, Exception? exception)
+		internal FileWriteResult(FileWriteEnum error, FileWritePaths writer, Exception? exception)
 		{
 			Error = error;
 			Exception = exception;
-			WritePath = path;
+			Paths = writer;
 		}
+
+		public FileWriteResult() : this(FileWriteEnum.Success, new("", ""), null) { }
 
 		/// <summary>
 		/// Converts this instance to a different generic type.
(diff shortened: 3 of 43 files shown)


src/BizHawk.Client.Common/FrameworkZipWriter.cs
```diff
@@ -51,19 +72,17 @@ public void WriteItem(string name, Action<Stream> callback, bool zstdCompress)
...
-			{
-				_archive.Dispose();
-				_archive = null;
-			}
-			_fs?.Flush(flushToDisk: true);
-			_fs?.Dispose();
+			if (_disposed) return;
+			_disposed = true;
+
+			_archive!.Dispose();
```
YoshiRulz: [Don't throw in `Dispose`](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=netstandard-2.0#remarks) (this needs to be `?.` or it will throw when run a second time)

src/BizHawk.Client.Common/FrameworkZipWriter.cs
```diff
@@ -32,8 +33,28 @@ public FrameworkZipWriter(string path, int compressionLevel)
...
+			FrameworkZipWriter ret = new(compressionLevel);
+			ret._fs = fs.Value!;
+			ret._archive = new(ret._fs.Stream, ZipArchiveMode.Create, leaveOpen: true);
+
+			return fs.Convert(ret);
+		}
+
+		public FileWriteResult Finalize()
+		{
+			if (_fs == null) throw new InvalidOperationException("Cannot use disposed ZipWriter.");
```
YoshiRulz: Use `ObjectDisposedException`

src/BizHawk.Client.Common/IZipWriter.cs
```diff
@@ -5,5 +5,10 @@ namespace BizHawk.Client.Common
 	public interface IZipWriter : IDisposable
 	{
 		void WriteItem(string name, Action<Stream> callback, bool zstdCompress);
+
+		/// <summary>
+		/// This method must be called after writing has finished.
+		/// </summary>
+		FileWriteResult Finalize();
```
YoshiRulz: Not sold on this name, it could be confused for the finaliser (destructor).

Maybe this logic should just be in `Dispose`? Then we can use `using` at call-sites.

src/BizHawk.Client.Common/FileWriteResult.cs
```diff
@@ -0,0 +1,82 @@
...
+					return "unreachable";
+			}
+		}
+	}
+
+	/// <summary>
+	/// Provides information about the success or failure of an attempt to write to a file.
+	/// If successful, also provides a related object instance.
+	/// </summary>
+	public class FileWriteResult<T> : FileWriteResult
```
YoshiRulz: Is this set up like `Task`/`Task<T>`?

src/BizHawk.Client.Common/Api/Classes/MovieApi.cs
```diff
@@ -53,6 +53,7 @@ public string GetInputAsMnemonic(int frame)
 			return Bk2LogEntryGenerator.GenerateLogEntry(_movieSession.Movie.GetInputState(frame));
 		}
 
+		// TODO: Change return type to indicate success/failure? This method is part of BizHawk's API for external tools/Lua.
```
YoshiRulz: `bool` for success is fine, and we can always expose the `FileWriteResult` later.

src/BizHawk.Client.Common/FileStreamWithTemp.cs
```diff
@@ -0,0 +1,106 @@
...
+			FinalPath = final;
+			TempPath = temp;
+			Stream = new(temp, FileMode.Create, FileAccess.Write);
+		}
+
+		// There is no public constructor. This is the only way to create an instance.
+		public static FileWriteResult<FileStreamWithTemp> Create(string path)
+		{
+			string writePath = path;
+			if (File.Exists(path))
```
YoshiRulz: What is the intent for when the path doesn't exist? Should it be created?

src/BizHawk.Client.Common/FileStreamWithTemp.cs
```diff
@@ -0,0 +1,106 @@
...
+			_finalized = true;
+			return new(FileWriteEnum.Success, FinalPath, null);
+		}
+
+		public void Dispose()
+		{
+			Stream.Dispose();
+
+			// The caller should call Finalize and handle potential failure.
+			if (!_finalized) throw new Exception("FileStreamWithTemp was closed without finalizing.");
```
YoshiRulz: Don't throw in `Dispose`.
```suggestion
			if (_finalized) return;
			const string ERR_MSG_NEVER_FINISHED = $"{nameof(FileStreamWithTemp)} was closed without calling {nameof(Finalize)}!";
#if DEBUG
			throw new Exception(ERR_MSG_NEVER_FINISHED);
#else
			Console.WriteLine(ERR_MSG_NEVER_FINISHED);
#endif
```

src/BizHawk.Client.Common/movie/bk2/Bk2Movie.IO.cs
```diff
@@ -1,3 +1,5 @@
+#nullable enable
```
YoshiRulz, line 1: Did you check the whole file for null-correctness?

src/BizHawk.Client.EmuHawk/MainForm.Movie.cs
```diff
@@ -125,7 +125,14 @@ public void StopMovie(bool saveChanges = true)
 			}
 			else
 			{
-				MovieSession.StopMovie(saveChanges);
+				FileWriteResult saveResult = MovieSession.StopMovie(saveChanges);
+				if (saveResult.IsError)
+				{
+					this.ShowMessageBox(
+						$"Failed to save movie.\n{saveResult.UserFriendlyErrorMessage()}\n{saveResult.Exception.Message}",
+						"Error",
```
YoshiRulz, line 133:
```suggestion
						caption: "Error",
```
Also should this be `ModalMessageBox`? I forget what the difference is.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -4327,10 +4346,12 @@ public void SaveState(string path, string userFriendlyStateName, bool fromLua =
 				return;
 			}
 
-			try
+			if (new SavestateFile(Emulator, MovieSession, MovieSession.UserBag).Create(path, Config.Savestates).IsError)
```
YoshiRulz:
```suggestion
			var result = new SavestateFile(Emulator, MovieSession, MovieSession.UserBag).Create(path, Config.Savestates);
			if (result.IsError)
```
(since `Create` has side-effects)

src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.IToolForm.cs
```diff
@@ -155,12 +155,23 @@ public override bool AskSaveChanges()
...
-			if (result is null) return false;
-			if (result.Value) SaveTas();
+			if (shouldSaveResult == true)
+			{
+				FileWriteResult saveResult = SaveTas();
+				while (saveResult.IsError && shouldSaveResult != true)
+				{
+					shouldSaveResult = this.ModalMessageBox3(
+						$"Failed to save movie. {saveResult.UserFriendlyErrorMessage()}\n{saveResult.Exception.Message}\n\nTry again?",
+						"Error",
```
YoshiRulz:
```suggestion
						caption: "Error",
```

YoshiRulz (commented):
Better I/O handling has been sorely needed for a while. I suspect the `.tasproj` corruptions will end with this.

src/BizHawk.Client.Common/FrameworkZipWriter.cs
SuuperW: That's what `if (_disposed) return;` is for.

"the object must ignore all calls after the first one" and that is what it does.

src/BizHawk.Client.Common/IZipWriter.cs
SuuperW: You unfortunately can't get the return value if you don't explicitly call it, so `using` is useless. As for better names I'm not sure. `Commit`? `Finished`? `FinalWrite`? `Confirm`?

src/BizHawk.Client.Common/FileWriteResult.cs
SuuperW: I guess so? One version represents the result of an operation without a return value, one with a return value. I didn't think about `Task` while writing it.

src/BizHawk.Client.Common/Api/Classes/MovieApi.cs
SuuperW: That could work, but I could imagine some tools would be unhappy about it no longer throwing. Probably not a big concern, but I haven't even considered what a tool or Lua script using this method would be doing so I don't know.

src/BizHawk.Client.Common/FileStreamWithTemp.cs
SuuperW: Yup. The reason for using a temp file (with `.saving` added into the name) was to avoid destroying the old file until the new one is fully made, and so the file can still get saved somewhere if the file is already being used. Neither of those apply if the desired file name doesn't exist.

src/BizHawk.Client.Common/FileStreamWithTemp.cs
SuuperW: How about `Debug.Assert`? That's cleaner than #if DEBUG imo and I see no point in using `Console.WriteLine` for non-debug since that would never be seen by a typical user.

src/BizHawk.Client.Common/movie/bk2/Bk2Movie.IO.cs
SuuperW: With so much code changes I knew I'd forget something. Checking it now, compiler incorrectly infers `string?` for `var backupName` yet doesn't complain about passing potentially null to `Path.GetFileName` for some reason. There's also a `!` null-suppression operator in `Directory.CreateDirectory(Path.GetDirectoryName(fn)!);` which would be an issue if the user selected a file in the root directory but that's not one I put there. And there's a `?` that should be added later.

Aside from that, it is of course touching properties of the movie that don't use nullable but it looks like it is handling them correctly.

src/BizHawk.Client.EmuHawk/MainForm.Movie.cs
SuuperW: I never knew, so /shrug. By the method descriptions, one will have a parent and the other I guess won't. They appear to behave the same.
Why a named parameter there?

SuuperW:
Highlights of force push:
1) Added constraint to FileWriteResult<T> and .editorconfig rules to enforce compliance.
2) Re-did some closing and disposing stuff and order of things to make it work in all situations that I could think to test. Including a new "Abort" mechanism.

SuuperW:
Well I just re-read the linked issue and realized that this PR does nothing to address that. I had seen BizHawk and autosave and assumed it meant TAStudio auto-saving because I forget that there is another autosave. Handling of sram saving failure will come later.

src/BizHawk.Client.Common/Api/Classes/MovieApi.cs
YoshiRulz: Better idea:
```csharp
bool Save(string filename, out Exception e);
[Obsolete]
void Save(string filename); // throws
```

src/BizHawk.Client.Common/FileStreamWithTemp.cs
YoshiRulz: Rewritten `Dispose` LGTM

src/BizHawk.Client.EmuHawk/MainForm.Movie.cs
YoshiRulz: Dialogs without a parent window break keep-above, see #4103.

Using named arguments when there are two params of the same type prevents bugs. That's not just theoretical either, we have had a confirmation dialog with the title and message swapped.

src/BizHawk.Client.Common/Api/Classes/MovieApi.cs
SuuperW: If we are ever to remove the obsolete version, we'd be left with an API method with a signature that doesn't make sense. `FileWriteResult Save(string filename);` is clearly better.
I propose leaving this as a TODO that will be made along with other API changes at whatever future time we decide to update the API such as #4173 and #4329.

src/BizHawk.Client.EmuHawk/MainForm.Movie.cs
SuuperW: OK, so someone wrote incorrect code once. I don't think that's a reason to change coding style. Besides, there are many more instances of not using named parameters in such cases (including uses of the various message box methods here) than there are of using them.

As for keeping the window on top, I do not observe that issue with either method. But I do observe that issue with the current implementation of the state history form. Since all other message boxes in this file use the `ShowMessageBox` variants I think it's best to go with that.

SuuperW:
Now handles errors for SRAM, config file, savestates, and all tools' `AskSaveChanges` (and related save menu items).
Some file writes have not been updated. Are there any others that should be done? TAStudio macros probably should be, I guess.

SuuperW:
I believe this is ready for merge. See edited OP for notes on some behavior changes.

src/BizHawk.Emulation.Common/Interfaces/Services/ISaveRam.cs
```diff
@@ -28,5 +28,10 @@ public interface ISaveRam : IEmulatorService
 		/// This value should be considered a hint more than an absolute truth.
 		/// </summary>
 		bool SaveRamModified { get; }
+
+		/// <summary>
+		/// Certain cores may support SaveRam only in certain situations, for example only for certain games.
+		/// </summary>
+		bool SupportsSaveRam { get; }
```
YoshiRulz: Cores can conditionally implement services by mutating the service provider from in the ctor body; can that be done for `ISaveRam`?

YoshiRulz (commented):
> Changes SRAM behavior. It no longer silently modifies a loaded SRAM file when it is not the expected size. Instead cores will/should throw if the SRAM is invalid. This may "break" things since previously-working .SaveRam files (e.g. bad dumps maybe?) may no longer work. EDIT: I don't recall why I originally put this change in this PR. It seems unrelated, so I guess I could remove it.

Yes, please open a new PR for this, I'm eager to redesign that old interface. I have [a few vague notes](https://gitlab.com/YoshiRulz/yoshis-hawk-thoughts/-/issues/86) of which the main proposal was to have the core "declare" the expected size with a new prop, to be checked by the frontend. Making cores throw when the array is the wrong size has the same effect, and I think would have been necessary anyway, so I'm on board with that. It would also solve #4242.

src/BizHawk.Emulation.Common/Interfaces/Services/ISaveRam.cs
SuuperW: I think so. I don't know how much work it would be to make the necessary modifications.

SuuperW:
Removed ISaveRam refactor.

SuuperW:
Are there any reasons to not merge this?

YoshiRulz:
I'd like to get more eyes on this, especially the new API's design.

SuuperW:
For what it's worth, all my TASing since creating the PR has been done on a branch with these commits and I have not encountered any issues.

vadosnaprimer:
If there's been no eyes on this for half a year we shouldn't expect them to appear. Unless something looks broken it should be fine to just merge and see. Conflicts tho.

src/BizHawk.Client.Common/DialogControllerExtensions.cs
```diff
@@ -64,6 +72,47 @@ public static bool ModalMessageBox2(
...
+			Func<FileWriteResult> action,
+			string message)
+		{
+			FileWriteResult fileResult = action();
+			while (fileResult.IsError)
+			{
+				string prefix = message ?? "";
+				bool? askResult = dialogParent.ModalMessageBox3(
+						text: $"{prefix} Do you want to try again?\n\nError details:" +
+							$"{fileResult.UserFriendlyErrorMessage()}\n{fileResult.Exception!.Message}",
```
YoshiRulz, line 105: This "friendlier" message omits the important details of the exception type and stacktrace, making it useless for debugging. Fixed in https://github.com/TASEmulators/BizHawk/commit/773f300b32201c0629f02c0819b29a760e9b4637.
(You also missed a space after the colon, gave the dialog the nondescriptive title "Error", forgot to update this method's documentation to reflect the new return type, and haven't followed the code style conventions.)

YoshiRulz:
(Didn't mention it here, but for first regression, see 3ec7c2a96a62829af4f38b876cdb12e4cbedd13c.)

A third regression. https://github.com/TASEmulators/BizHawk/commit/a2e8eed8b6aa50d2a9784c23540e0129e99cc210

I'd like to revert this wide-reaching and unreviewed changeset before we release 2.11.1.

PR #4403, opened 2025-07-24
Fix input/hotkey bugs

SuuperW:
Fix: Hotkey overrides input would cause an input to continue after a button was released if the input is bound to a combination such as Shift+A and an "internal hotkey" is bound to a single button (such as A). Also input bound to Alt would not have the alt ignored by menus.

Fix: Input overrides hotkeys did not work if both input and hotkey were bound to a combination such as Shift+A.

Fix: Controller inputs bound to a modifier key combination would not be triggered if additional modifier keys were also pressed.

Fix: autofire/autohold hotkeys would sometimes activate buttons that were already being held when the hotkey was pressed.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Controller.cs
+++ b/src/BizHawk.Client.Common/Controller.cs
@@ -59,11 +59,20 @@ namespace BizHawk.Client.Common
 		public List<string> SearchBindings(string button)
 			=> _bindings.Where(b => b.Value.Contains(button)).Select(static b => b.Key).ToList();
 
-		// Searches bindings for the controller and returns true if this binding is mapped somewhere in this controller
-		public bool HasBinding(string button) =>
-			_bindings
+		/// <summary>
+		/// Checks if the given button combination would be a controller input.
+		/// This means Shift+A will return true if an input is bound to either A or Shift+A.
+		/// </summary>
+		public bool HasBinding(string button)
+		{
+			string[] buttons = button.Split('+');
+			return _bindings
 				.SelectMany(kvp => kvp.Value)
-				.Any(boundButton => boundButton == button);
+				.Any((boundCombination) => {
+					string[] boundButtons = boundCombination.Split('+');
+					return boundButtons.All((b) => buttons.Contains(b));
+				});
+		}
 
 		/// <summary>
 		/// uses the bindings to latch our own logical button state from the source controller's button state (which are assumed to be the physical side of the binding).
--- a/src/BizHawk.Client.Common/input/InputCoalescerControllers.cs
+++ b/src/BizHawk.Client.Common/input/InputCoalescerControllers.cs
@@ -11,14 +11,16 @@ namespace BizHawk.Client.Common
 		public InputCoalescer()
 			: base(NullController.Instance.Definition) {} // is Definition ever read on these subclasses? --yoshi
 
-		protected virtual void ProcessSubsets(string button, bool state) {}
+		protected virtual void ProcessInput(string button, bool state)
+		{
+			Buttons[button] = state;
+		}
 
 		public void Receive(InputEvent ie)
 		{
 			var state = ie.EventType is InputEventType.Press;
 			var button = ie.LogicalButton.ToString();
-			Buttons[button] = state;
-			ProcessSubsets(button, state);
+			ProcessInput(button, state);
 			if (state) return;
 			// when a button or modifier key is released, all modified key variants with it are released as well
 			foreach (var k in Buttons.Keys.Where(k =>
@@ -30,10 +32,17 @@ namespace BizHawk.Client.Common
 
 	public sealed class ControllerInputCoalescer : InputCoalescer
 	{
-		protected override void ProcessSubsets(string button, bool state)
+		protected override void ProcessInput(string button, bool state)
 		{
 			// For controller input, we want Shift+X to register as both Shift and X (for Keyboard controllers)
 			foreach (var s in button.Split('+')) Buttons[s] = state;
+		}
+
+		public override bool IsPressed(string button)
+		{
+			// Since we split all inputs into their separate physical buttons, we need to check combinations here.
+			string[] buttons = button.Split('+');
+			return buttons.All(Buttons.GetValueOrDefault);
 		}
 	}
 }
--- a/src/BizHawk.Client.Common/inputAdapters/InputManager.cs
+++ b/src/BizHawk.Client.Common/inputAdapters/InputManager.cs
@@ -158,7 +158,7 @@ namespace BizHawk.Client.Common
 		/// <param name="processUnboundInput">Events that did not do anything are forwarded out here.
 		/// This allows things like Windows' standard alt hotkeys (for menu items) to be handled by the
 		/// caller if the input didn't alrady do something else.</param>
-		public void ProcessInput(IPhysicalInputSource source, Func<string, bool> processHotkey, Config config, Action<InputEvent> processUnboundInput, Func<string, bool> isInternalHotkey)
+		public void ProcessInput(IPhysicalInputSource source, Func<string, bool> processHotkey, Config config, Action<InputEvent> processUnboundInput)
 		{
 			// loop through all available events
 			InputEvent ie;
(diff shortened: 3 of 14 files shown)


src/BizHawk.Client.Common/Controller.cs
```diff
@@ -59,11 +59,20 @@ public bool AnyInputHeld
...
+		{
+			HashSet<string> buttons = button.Split('+').ToHashSet();
+			return _bindings
 				.SelectMany(kvp => kvp.Value)
-				.Any(boundButton => boundButton == button);
+				.Any((boundCombination) => {
+					HashSet<string> boundButtons = boundCombination.Split('+').ToHashSet();
+					return boundButtons.IsSubsetOf(buttons);
+				});
+		}
```
Morilli: I don't think this logic is fully correct. When an input is bound to e.g. `Ctrl+E` and a hotkey is bound to `Ctrl+Shift+E` and "Input overrides Hotkeys" is active, pressing `Ctrl+Shift+E` will do nothing.

SuuperW: In my testing, `Ctrl+Shift+E` will press the controller input and not trigger the hotkey, which is the expected behavior.
Inputs work even when pressing extra modifier keys because (from Discord)
> If we didn't consider Shift+A as also A then you could not press whatever is bound to Shift and then whatever is bound to A while still holding shift.

Morilli: Have you tested this with `Input overrides Hotkeys`? Because I get nothing.

SuuperW: Yes, but turns out I was messing up elsewhere. I have gotten the bug now. It is not a bug with this code, but another bug that exists in master branch. The hotkey should not be triggered because the input should work.

SuuperW: Fix with new commit.

Morilli:
It's pretty hard to verify things are correct now because nothing defines "correct". Let me recap a couple things that happen.

If a key combination like `Ctrl+Shift+A` is pressed, it will now activate controller bindings for `Ctrl`, `Shift`, `A`, `Ctrl+Shift`, `Ctrl+A`, `Shift+A` and `Ctrl+Shift+A`. Only a hotkey bound to `Ctrl+Shift+A` will be triggered though.

"Internal" hotkeys are a separate group from normal hotkeys. If any key press activates a normal hotkey then no internal hotkey will be triggered, regardless of priority setting, except for the `Frame Inch` hotkey binding. The `Frame Inch` exception also means that binding a key combination to `Frame Inch` and some controller input will still have it go through in "Hotkeys override Input" mode.

The "Hotkeys override Input" branch is especially difficult for me to understand. Am I correct that the following code is functionally equivalent to what the code is currently doing?
```csharp
if (ie.EventType is InputEventType.Press)
{
	var handled = triggers.Aggregate(false, (current, trigger) => current | CheckHotkey(trigger));
	if (!handled)
	{
		hotkeyCoalescer.Receive(ie);
		if (!triggers.Exists(IsInternalHotkey))
		{
			finalHostController.Receive(ie);
		}
	}
}
else
{
	hotkeyCoalescer.Receive(ie);
	finalHostController.Receive(ie);
}
```

SuuperW:
> It's pretty hard to verify things are correct now because nothing defines "correct".

Yeah, we really should have some tests for this.

> If a key combination like `Ctrl+Shift+A` is pressed, it will now activate controller bindings for `Ctrl`, `Shift`, `A`, `Ctrl+Shift`, `Ctrl+A`, `Shift+A` and `Ctrl+Shift+A`. Only a hotkey bound to `Ctrl+Shift+A` will be triggered though.

This is correct, with the caveat that you cannot press `Ctrl+Shift+A` without first pressing `Ctrl+Shift` and before that either `Ctrl` or `Shift`, so hotkeys bound to those may trigger in the process. `A` must be pressed last or `Ctrl+Shift+A` won't happen for a hotkey, but a controller binding would be activated when pressing the keys in any order.
 
> "Internal" hotkeys are a separate group from normal hotkeys. If any key press activates a normal hotkey then no internal hotkey will be triggered, regardless of priority setting, except for the `Frame Inch` hotkey binding. The `Frame Inch` exception also means that binding a key combination to `Frame Inch` and some controller input will still have it go through in "Hotkeys override Input" mode.

Interesting. I had not looked that much into the internal hotkeys.
I think that system should be re-worked. One reason for this is that if you set a very low emulation speed, the frame advance hotkey is unreliable. So is pausing. (because you're likely to press and release between loops, so the loop will never see it pressed)

> The "Hotkeys override Input" branch is especially difficult for me to understand. Am I correct that the following code is functionally equivalent to what the code is currently doing?

Yeah lots of this code is tough to read. I believe your code is equivalent.

YoshiRulz:
> [...] if you set a very low emulation speed, the frame advance hotkey is unreliable. So is pausing. (because you're likely to press and release between loops, so the loop will never see it pressed)

#2959?

SuuperW:
OP is unclear, but yeah this is the same issue as the comments there are discussing.

Morilli:
> if you set a very low emulation speed, the frame advance hotkey is unreliable. So is pausing.

I can't reproduce that with the pausing hotkey. However for frame advance you can use frame inch instead as that immediately pauses instead of waiting for the actual frame loop.

SuuperW:
Oh pause isn't an internal hotkey. Huh.
I must have been using the frame advance button to try to pause when I experienced that.

SuuperW:
Tests and additional fixes.

> "Internal" hotkeys are a separate group from normal hotkeys. If any key press activates a normal hotkey then no internal hotkey will be triggered

This is also fixed.

Rebased onto my undo_redo branch (has its own pending PR) in order to make use of test code there.

YoshiRulz:
> Rebased onto my undo_redo branch (has its own pending PR) in order to make use of test code there.

please don't

SuuperW:
Rebased onto master.
I'd like to get this branch merged too.

src/BizHawk.Client.Common/input/InputCoalescerControllers.cs
```diff
@@ -30,10 +32,21 @@ public void Receive(InputEvent ie)
...
+
+		public override bool IsPressed(string button)
+		{
+			// Since we split all inputs into their separate physical buttons, we need to check combinations here.
+			bool allPressed = true;
+			foreach (var individualButton in button.Split('+'))
+			{
+				allPressed &= Buttons.GetValueOrDefault(individualButton);
+			}
+			return allPressed;
```
YoshiRulz: `return button.Split('+').All(Buttons.GetValueOrDefault);`

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1860,8 +1717,6 @@ public bool RunLibretroCoreChooser()
 
 		// input state which has been destined for game controller inputs are coalesced here
 		// public static ControllerInputCoalescer ControllerInputCoalescer = new ControllerInputCoalescer();
```
YoshiRulz: This commented-out field should have been removed ages ago.

src/BizHawk.Tests.Client.Common/InputManagerTests.cs
```diff
@@ -0,0 +1,500 @@
...
+
+			return context;
+		}
+
+		[TestMethod]
+		public void BasicControllerInput()
+		{
+			Context c = MakeContext();
+			InputManager manager = c.manager;
+			FakeInputSource source = new FakeInputSource();
```
YoshiRulz: Unlike in the other PR, `[TestInitialize]` would be a clear improvement here.

src/BizHawk.Tests.Client.Common/InputManagerTests.cs
```diff
@@ -0,0 +1,500 @@
...
+			}
+		}
+
+		private string[] _hotkeys = [ "Hotkey 1", "Autofire", "Autohold" ];
+
+		private static readonly IReadOnlyList<string> _modifierKeys = new[] { "Super", "Ctrl", "Alt", "Shift" };
+
+		private static readonly int PRIORITY_BOTH = 0;
+		private static readonly int PRIORITY_INPUT = 1;
+		private static readonly int PRIORITY_HOTKEY = 2;
```
YoshiRulz: At this point you should declare an enum for it. Not sure why no-one had done that yet.

src/BizHawk.Tests.Client.Common/InputManagerTests.cs
```diff
@@ -0,0 +1,500 @@
...
+			{
+				emulator.FrameAdvance(manager.ControllerOutput, false);
+				manager.ProcessInput(new FakeInputSource(), processHotkey, config, (_) => { });
+
+				if (lag) manager.AutoFireController.IncrementStarts();
+				manager.StickyAutofireController.IncrementLoops(lag);
+			}
+		}
+
+		private string[] _hotkeys = [ "Hotkey 1", "Autofire", "Autohold" ];
```
YoshiRulz: `static readonly`

src/BizHawk.Client.Common/inputAdapters/InputManager.cs
```diff
@@ -147,5 +137,113 @@ private static AutofireController BindToDefinitionAF(
...
 			return ret;
 		}
+
+		// input state which has been destined for client hotkey consumption are colesced here
+		private readonly InputCoalescer _hotkeyCoalescer = new InputCoalescer();
+
+		/// <param name="processUnboundInput">Events that did not do anything are forwarded out here.
+		/// This allows things like Windows' standard alt hotkeys (for menu items) to be handled by the
+		/// caller if the input didn't alrady do something else.</param>
+		public void ProcessInput(IInputSource source, Func<string, bool> processHotkey, Config config, Action<InputEvent> processUnboundInput)
```
YoshiRulz: This move looks good to me, my only concern is whether RMouse still works.

src/BizHawk.Client.Common/input/IInputSource.cs
```diff
@@ -0,0 +1,13 @@
+﻿#nullable enable
+
+using System.Collections.Generic;
+
+namespace BizHawk.Client.Common
+{
+	public interface IInputSource
```
YoshiRulz: Name could be more specific

YoshiRulz (approved):

src/BizHawk.Tests.Client.Common/InputManagerTests.cs
SuuperW: Disagree. `[TestInitialize]` is never better, as it is more difficult to understand, especially if one doesn't know to look for that method. Imo, all the code necessary to understand a test should be visible when looking at the test method. ("visible" includes any methods you can see being used; you see it is used so you can navigate to it if needed)

src/BizHawk.Tests.Client.Common/InputManagerTests.cs
SuuperW: Idk either, but it didn't seem important to add it here. I'll add a commit for that since you suggested it though.

src/BizHawk.Client.Common/inputAdapters/InputManager.cs
SuuperW: Right-clicking for the context menu still works, and binding a controller input to right-click still works. Anything else with RMouse need testing?

SuuperW:
I realized that I messed up mouse coordinate transforming here. I've added a new commit 29ed1b3 that fixes that. (It's weird that we translate screen coordinates to client coordinates then to -10k to +10k range and back to client coordinates again but I assume there's a reason for that somewhere.)
I'll give you a chance to review the updates before I merge, @YoshiRulz.

src/BizHawk.Client.Common/inputAdapters/InputManager.cs
YoshiRulz: Not RMB. RMouse X/Y is the relative mouse mechanism introduced for DOOM/DOSBox. I just checked at https://github.com/TASEmulators/BizHawk/commit/38a69e89e7bfdb418a5304cbad3ed49defd36f17 and it seems to work the same.

YoshiRulz:
> (It's weird that we translate screen coordinates to client coordinates then to -10k to +10k range and back to client coordinates again but I assume there's a reason for that somewhere.)

Not a good reason. It goes back to before https://github.com/TASEmulators/BizHawk/commit/e12b5d8137805b74c17cfd91d00ef65de6380e6a when axes used `float`.

[diff](https://github.com/SuuperW/BizHawk/compare/a2223107bfc475441cfa620aad52a323e26b630c..38a69e89e7bfdb418a5304cbad3ed49defd36f17) still LGTM

Issue #64, opened 2015-06-22
GensHawk: Savestates


GoogleCodeExporter:
```
Implement savestates for GensHawk.
```

Original issue reported on code.google.com by `brandman...@gmail.com` on 12 May 2012 at 7:49

PR #4406, opened 2025-07-26
Experimenting with Fluent.Net (i18n) in DiscoHawk

YoshiRulz:
See [Fluent homepage](https://projectfluent.org) and [my notes #64](https://gitlab.com/YoshiRulz/yoshis-hawk-thoughts/-/issues/64).

If someone on Windows could check the Designer inlining matches the previous layout, those can be pushed to master immediately.


Morilli:
The DiscoHawk window looks exactly as it did with the Designer files. That said I really don't like "inlining" designer files as that completely defeats its purpose. It gets much harder to edit the layout and test things when changing anything (and if you're not changing anything, why bother converting it?).

vadosnaprimer:
> That said I really don't like "inlining" designer files as that completely defeats its purpose. It gets much harder to edit the layout and test things when changing anything (and if you're not changing anything, why bother converting it?).

Agreed.

YoshiRulz:
I will be making changes in this very PR, namely by replacing all the strings with translation key lookups.

At the risk of repeating myself, the drawbacks of the Designer far outweigh its benefits, and we'll have to stop using it anyway as part of leaving WinForms.

Morilli:
> we'll have to stop using it anyway as part of leaving WinForms.

That's true and fine, but it doesn't look like that's what you're doing in here. Is using fluent not possible while keeping the existing structure?

YoshiRulz:
To keep the Designer files, I would have to move every `.Text` assignment into the ctor, similar to this:
https://github.com/TASEmulators/BizHawk/blob/56f3a2d69200e8f5ca502b8bd8e0a0aa9d551053/src/BizHawk.Client.EmuHawk/tools/Watch/RamWatch.cs#L41-L88

PR #4413, opened 2025-08-10
[Fix] Adding a fix to QuickerNES to support certain games with submapper > 0

SergioMartin86:

This is a very small fix to the QuickerNES core which prevents it from rejecting roms based on mappers that are now supported. The rejecting legacy lines of code were relevant back then when the mappers weren't supported. 

The motivating test case for this fix is the following game:

Pac-Man - Championship Edition (USA, Europe) (Namco Museum Archives Vol 1).nes
SHA1: 4CBAD49930253086FBAF4D082288DF74C76D1ABC
MD5: EE8BC8BAED5B9C5299E84E80E6490DE6

Before:  "Unsupported Mapper" Error
Now:

image

Check if completed:

### DIFF
--- a/quicknes/msvc/libquicknes.vcxproj
+++ b/quicknes/msvc/libquicknes.vcxproj
@@ -204,3 +204,4 @@
     <Copy SourceFiles="$(OutDir)libquicknes.dll" DestinationFolder="$(ProjectDir)../../output/dll" SkipUnchangedFiles="true" />
   </Target>
 </Project>
+


CasualPokePlayer:
If a submapper is supported, it should be checked explicitly, not just always shoved through. Submappers indicate different behavior than iNES, which cannot be assumed to always work out of the box in quickerNES. Any NES 2.0 header with extended mapper high bits too cannot be assumed to be supported: those mappers are above the 255 mapper limit of iNES, which quickerNES does not support (and your changes don't add support for such high mapper codes at all, you're going to just end up shoving the ROM through with a completely different mapper that is very likely not going to work correctly).

That said, a quick look at No-Intro indicates the ROM should not have any submapper nor extended mapper high bits, so this change shouldn't affect anything here. What is the header present in your ROM dump?

EDIT: Was reading the wrong thing in No-Intro, it appears the game wants submapper 3 of mapper 19 here?

CasualPokePlayer:
Seems submappers for mapper 19 specifically are just these to explicitly indicate expansion sound being present or not, along with the expansion sound volume relative to the NES APU's audio. "Default" iNES behavior (submapper 0) just leaves the presence of such unspecified. quickerNES behavior just emulates such regardless, so it's probably fine to allow submappers through for mapper 19 specifically.

CasualPokePlayer:
https://github.com/TASEmulators/quickerNES/blob/03860f74711ae1d59a2736db4b742a5c4a6d08a1/source/quickerNES/core/cart.hpp#L83-L95

The change would better be:
```c
      if (h.ex_mapper != 0)
      {
        if ((h.ex_mapper & 0x0F) != 0)
          return "Unsupported mapper";

        bool has_compatible_submapper = false;

        // For some submappers, functionality doesn't actually change in an incompatible way
        // For these submappers, we can allow them through

        int mapper_code = (h.flags2 & 0xF0) | ((h.flags >> 4) & 0x0F);

        // https://www.nesdev.org/wiki/INES_Mapper_019
        // Mapper 19 submappers mostly just specify expansion sound volume
        // Submapper 1 and 2 indicate no expansion sound
        // However, having such doesn't add compatibility issues
        if (mapper_code == 19)
          has_compatible_submapper = true;

        if (!has_compatible_submapper)
          return "Unsupported mapper";
      }
```

Also, the printf should be masking out unrelated flags and properly combining the variables for a mapper + submapper code.
```c
int mapper_code = (h.flags2 & 0xF0) | ((h.flags >> 4) & 0x0F);
int submapper_code = 0;
if ((h.flags2 & 0x0C) == 0x08)
{
  mapper_code |= (h.ex_mapper << 8) & 0xF00;
  submapper_code = (h.ex_mapper >> 4) & 0xF;
}

printf("[QuickerNES] Cart Mapper Info: Mapper: %d - Submapper: %d\n", mapper_code, submapper_code);
```

(Can then remove the `mapper_code` being set in the `if (h.ex_mapper != 0)` branch then)

PR #4433, opened 2025-08-31
Seek fixes

SuuperW:
Several fixes to seeking behavior.

This PR introduces a change to how a return value from the API is used: The value of `Rewind` is no longer ignored. What it signifies is also slightly different. It used to say "if the rewind occurred" but now "if a frame advance is required". (In TAStudio, these were usually the same thing.) The ApiHawk [changelog](https://github.com/TASEmulators/BizHawk-ExternalTools/wiki/Partial-changelog-for-ApiHawk) should be updated.

Bugs that affect 2.10:
1. Wheel scroll up behavior was inconsistent, depending on if one or two events are raised for a two-notch scroll.
2. Green arrow was not reset when loading a new movie.
3. Seek would be canceled if user triggers a TAStudio close but cancels when asked if they want to save.
4. Starting a new movie then unpausing could then immediately auto-pause.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/IControlMainform.cs
+++ b/src/BizHawk.Client.EmuHawk/IControlMainform.cs
@@ -54,8 +54,8 @@
 		/// <summary>
 		/// Function that is called by Mainform instead of using its own code
 		/// when a Tool sets WantsToControlRewind
-		/// Returns whether or not the rewind action actually occured
 		/// </summary>
+		/// <returns>Returns true if a frame advance is required.</returns>
 		bool Rewind();
 
 		bool WantsToControlRestartMovie { get; }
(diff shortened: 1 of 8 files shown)


vadosnaprimer:
Works okay with and without state gaps.

SuuperW:
Ready for merge.

Morilli:
> Green arrow was not reset when loading a new movie.

I don't know what the green arrow is but at the very least there's a new bug here: Starting a new tasproj does not cancel a seek. I see what d9efa0536996d8d49af8123a8d4fece7d47c55ba is trying to do but there should be a `StopSeeking` call in tastudio startup then or something.

SuuperW:
> Starting a new tasproj does not cancel a seek.

Thanks, fixed.

PR #4444, opened 2025-09-14
Remove the button for loading macro tool.

SuuperW:
It is still possible to open via the toolbox but only when VersionInfo.DeveloperBuild is true.
It will eventually be removed entirely, since it's very broken and it seems nobody wants it. That will take a little more effort since some of its stuff is tied into TAStudio, but I figured it might be good to do this before release.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.Designer.cs
@@ -203,7 +203,6 @@ namespace BizHawk.Client.EmuHawk
 			this.TraceLoggerMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.DebuggerMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
 			this.CodeDataLoggerMenuItem = new BizHawk.WinForms.Controls.ToolStripMenuItemEx();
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
How did you suddenly realize that it's bad to have this tool? The tool is clunky, and the menu is disabled without tastudio, but it's weird to just remove something that's not completely broken and not obsolete by a better workflow.

SuuperW:
It's not a sudden realization. I just never bothered to remove it before. The UI is not very good, and I haven't used it since macros were possible to use from inside TAStudio (which is something I did not very long after I made the macro tool). I doubt other people are using it either.

> The tool is clunky, and the menu is disabled without tastudio

It's disabled when not recording a movie. TAStudio is not required. (However if TAStudio is not opened, some features are disabled and it's even less user-friendly.)

One of the features was broken at the time I made this PR, but I actually fixed it recently. At the time I made this PR I thought it was more broken than just that one feature, but I think I was confusing bad UX without TAStudio for actual brokenness. The macro tool is technically more capable than what you can do in TAStudio, but again I doubt anyone is using it.

vadosnaprimer:
That still doesn't explain why it's bad to have it.

SuuperW:
If it had been as broken as I thought, I would probably not have ever fixed it and having broken features that won't get fixed is bad imo. Since I did fix it, it probably isn't bad to have. I still might eventually remove it if I can come up with a nice UI for putting the extra features inside TAStudio but that might not happen any time soon.

Issue #4356, opened 2025-06-16
Lua memory.write_bytes_as_array may not preserve order, keeps non-integer keys


SandTAS:
### Summary

The Lua function [`memory.write_bytes_as_array`](https://github.com/TASEmulators/BizHawk/blob/2.10/src/BizHawk.Client.Common/lua/CommonLibs/MemoryLuaLibrary.cs#L103) works by first [converting its input table into a C# `Dictionary`](https://github.com/TASEmulators/BizHawk/blob/2.10/ExternalProjects/NLua/src/Lua.cs#L858), then [writing bytes in the order returned by the `Dictionary`'s `.Values`](https://github.com/TASEmulators/BizHawk/blob/2.10/ExternalProjects/NLua/src/LuaTable.cs#L61). Neither of these operations—iterating over a Lua table or iterating over a C# `Dictionary`—is guaranteed to preserve the order of sequential integer keys. What is more, if the input Lua table contains non-integer keys, the corresponding values are also written to memory, in some arbitrary place in the ordering.

It happens that, in many common cases, iterating over the keys of a sequential Lua table *does* preserve the order of integer keys. And it appears that, in a C# `Dictionary` whose values are written only once, `.Values` preserves insertion order. So often, in practice, `memory.write_bytes_as_array` works as desired. But correct behavior is dependent on Lua and C# implementation details, and even then, there are ways to subvert it so that bytes are written out of order, or extra bytes are written.

### Repro

This Lua script tries `memory.write_bytes_as_array` with 4 different sequential tables. For me, only test 1 works; tests 2, 3, 4 all show some problem. I ran this with NesHawk, but I expect it will work with most cores and ROMs.

```lua
local PAD_BYTE = 0xff

-- Put a small block of memory into a known state.
for i = 0, 15 do
	memory.write_u8(i, PAD_BYTE)
end

-- Return a new table with PAD_BYTE added to the end up to length at least 16.
function pad(t)
	local new = table.move(t, 1, #t, 1, {})
	while #new < 16 do
		table.insert(new, PAD_BYTE)
	end
	return new
end

-- Return true iff t1 and t2 have the same keys and the same values
-- (only looking at keys seen by ipairs).
function sequences_equal(t1, t2)
	for i in ipairs(t1) do
		if t1[i] ~= t2[i] then
			return false
		end
	end
	for i in ipairs(t2) do
		if t1[i] ~= t2[i] then
			return false
		end
	end
	return true
end

[...]

PR #4446, opened 2025-09-14
Fix some items not appearing in Lua form dropdowns

kalimag:
- Add `NLuaTableHelper.EnumerateAllValues` which works like the old `EnumerateValues` implementation before 7bd9e218
- Add doc comments to both methods to clarify the difference
- Use this new (old) method in `FormsLuaLibrary.Dropdown` and `SetDropdownItems` to restore old behavior

7bd9e218 changed the behavior of `NLuaTableHelper.EnumerateValues` to enumerate keys sequentially and ignore non-sequential/non-numeric keys, the way Lua's `ipairs` does it. This makes sense for most uses of `EnumerateValues`, but `forms.dropdown` is explicitly documented as not requiring numeric/sequential keys, and some scripts* rely on this behavior. `forms.setdropdownitems` doesn't document what the table should look like, but can be assumed to work the same.

\* I ran into this issue while testing the bizhawk-shuffler-2 script on a dev build, a quick [search](https://github.com/search?q=%22forms.dropdown%22+OR+%22forms.setdropdownitems%22+language%3ALua+&type=code) shows other scripts that are likely affected, but I didn't test those.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/lua/NLuaTableHelper.cs
+++ b/src/BizHawk.Client.Common/lua/NLuaTableHelper.cs
@@ -2,6 +2,7 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
+using System.Linq;
 
 using NLua;
 
@@ -28,6 +29,18 @@ namespace BizHawk.Client.Common
 			return table;
 		}
 
+		/// <summary>
+		/// Enumerates all values in <paramref name="table"/>, regardless of what their keys are.
+		/// </summary>
+		/// <returns>
+		/// An unordered sequence of values.
+		/// </returns>
+		public IEnumerable<T> EnumerateAllValues<T>(LuaTable table) => table.Values.Cast<T>();
+
+		/// <summary>
+		/// Enumerates values in a Lua-style array, starting with table[1], table[2], ... up to the first absent index.
+		/// Ignores all non-sequential and non-numeric keys.
+		/// </summary>
 		public ArraySegment<T> EnumerateValues<T>(LuaTable table)
 		{
 			var n = table.Count;
--- a/src/BizHawk.Client.Common/lua/NLuaTableHelper.cs
+++ b/src/BizHawk.Client.Common/lua/NLuaTableHelper.cs
@@ -2,7 +2,6 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
-using System.Linq;
 
 using NLua;
 
@@ -28,14 +27,6 @@ namespace BizHawk.Client.Common
 			foreach (var (k, v) in dictionary) table[k] = v;
 			return table;
 		}
-
-		/// <summary>
-		/// Enumerates all values in <paramref name="table"/>, regardless of what their keys are.
-		/// </summary>
-		/// <returns>
-		/// An unordered sequence of values.
-		/// </returns>
-		public IEnumerable<T> EnumerateAllValues<T>(LuaTable table) => table.Values.Cast<T>();
 
 		/// <summary>
 		/// Enumerates values in a Lua-style array, starting with table[1], table[2], ... up to the first absent index.
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/FormsLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/FormsLuaLibrary.cs
@@ -179,7 +179,7 @@ namespace BizHawk.Client.EmuHawk
 				return 0;
 			}
 
-			var dropdownItems = _th.EnumerateValues<string>(items).ToList();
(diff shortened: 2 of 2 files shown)


YoshiRulz:
> [...] `forms.dropdown` is explicitly documented as not requiring numeric/sequential keys, and some scripts rely on this behavior.

Only since https://github.com/TASEmulators/BizHawk/commit/e2fb1c7f083b228f89a0c8ca82af05dc70e6b335. That would have been written to reflect the current (at the time) behaviour of the function.
But I can see that https://github.com/TASEmulators/BizHawk/commit/7bd9e218ca4714e39269b10b70ead02b89c05689 was a breaking change and should have at least updated the function docs.

Still, I don't know why you would be using a dict-like table for a dropdown menu in the first place? The keys are literally not used, even for an array-like table they're meaningless since the values are sorted.
Looking through your search query, code like [this](https://github.com/TestRunnerSRL/bizhawk-co-op/blob/c82024650ce1fb1c98231cbd8b157a7a7fa70e6c/bizhawk-co-op.lua#L250) or [this](https://github.com/bryanveloso/ironmon-tracker/blob/0430396acf677c413cd80e830eadcaea8eb71cef/ironmon_tracker/Theme.lua#L266) is clearly wrong.
The standard `table.pack` can produce a dict-like table as mentioned in https://github.com/TASEmulators/BizHawk/issues/4356#issue-3148373444, but even then the non-numeric key is meant to be like metadata, separate from the contents. Any thoughts @SandTAS?

kalimag:
Of course, accepting the breaking change and updating the Lua docs to reflect the new behavior is perfectly reasonable. It's fairly easily fixable on the scripting end, I just figured maintaining compatibility would be preferable.

The only reason it's used in the shuffler script is because an associative table already exists for a different purpose and it's convenient to reuse it.

SandTAS:
I haven't used one of these forms dropdowns before. The documented and past behavior of `forms.dropdown` I would find surprising in both respects: that it uses values from non-sequential keys and that it unconditionally sorts the values.

I would tend to say that maintaining backward compatibility is more important, especially as the change in behavior was unintended in this case.

The proposed patch, however, has the same ordering problem as in #4356. `table.Values` (which converts a Lua table to a C# `Dictionary` and enumerates the values of the dictionary) is not guaranteed to return the values in any particular ordering. It happens to work mostly as expected, but only because of internal implementation details.

The ordering issue doesn't matter for `forms.dropdown`, because that function always unconditionally sorts the items anyway. But it matters for `forms.setdropdownitems` with `alphabetize = false`. I expect that, with the current pull request, you could get an unexpected ordering using one of the tricks from #4356, something like:
```lua
forms.setdropdownitems(handle, {[1]="A", [2]="B", [3]="C", [4]="D"}, false)
```
The trick here is to use a different table literal syntax that causes the integer keys to go into the [hash part](https://www.lua.org/gems/sample.pdf#page=5) of the Lua table rather than the array part, which causes them to be unordered when inserted into the C# `Dictionary`.

In short, if `NLuaTableHelper.EnumerateAllValues` needs to produce values in a certain order (e.g. all sequential integer keys first, then non-integer keys), then that ordering should be defined and then implemented specifically. If `NLuaTableHelper.EnumerateAllValues` is not supposed to have any guarantees on order, then functions like `forms.setdropdownitems` that case about sequential order need to use something other than `NLuaTableHelper.EnumerateAllValues`.

YoshiRulz:
(I suspect the unconditional sort was *because* `Values` didn't have a stable order, though [it's not mentioned](https://github.com/TASEmulators/BizHawk/commit/91063b7ffdfb99dfe6404870c44888f5990884c7) and it was before my time so I can't be sure.)

IMO, if someone has passed `false` for `alphabetize` but then also passed in anything besides a simple list, they have no right to complain when it fails.

> In short, if `NLuaTableHelper.EnumerateAllValues` needs to produce values in a certain order (e.g. all sequential integer keys first, then non-integer keys), then that ordering should be defined and then implemented specifically.

I think I agree with this. When `alphabetize` is `false`, use something like `dropdownItems.AddRange(t.Where(kvp => !(kvp.Key is long i && 1 <= i && i <= dropdownItems.Count)).Select(kvp => kvp.Value).Order())`, and flaky scripts will just be broken. And when it's `true` use `t.Values.Order()` (directly, no need to keep helpers around that shouldn't be used) to restore the old behaviour, rather than `EnumerateValues` which will definitely break scripts.
edit: And "breaking scripts" here only amounts to displaying items out of order; the script can only read back the selected item's value, not its index.

kalimag:
In the case of `SetDropdownItems(..., alphabetize: false)`, items are now sorted by integer keys first, so any "proper" sequentially indexed tables will maintain their intended order.
Non-integer keys come afterwards in an undetermined order (not necessarily the same order as in 2.10).
To keep the code simple it doesn't differentiate between sequential 1...n keys and negative/zero/non-sequential keys, but I think that shouldn't be a problem.

YoshiRulz:
Yeah that's probably more performant than what I suggested. I'll update the function docs. Thank you.

Issue #4449, opened 2025-09-16
Tool/Lua input API no longer returns key combinations


kalimag:
### Summary
Previously, `InputApi.Get()`/`GetPressedButtons()` and through it Lua's `input.get` would include key combination strings like `"Ctrl+Shift+A"` (no clue what the BizHawk terminology for this is). After 6eecdd2194d49493c10d02b0b651e3574e8f49fb, pressed keys are only returned individually like `"Ctrl", "Shift", "A"`.

The previous behavior was quite useful for implementing complex hotkeys in scripts and tools. In one of my external tools, I was able to completely reuse BizHawk's input widgets to do hotkey and input binding. It's possible to rewrite scripts to check modifiers and buttons individually, but much less convenient and there's no easy way to replicate the full logic of how BizHawk processes modifier key combinations.

This could be seemingly easily solved by decoupling the tool/Lua API from `InputManager.ControllerInputCoalescer` and having a separate API coalescer with the old behavior. For better or worse, this would also decouple scripts from the input priority setting.

### Repro
1. Load Lua script
2. Press key combination such as `Ctrl+Shift+A`

```Lua
local prev_inputs = {}
while true do
	local inputs = input.get()
	for key, pressed in pairs(inputs) do
		if pressed and not prev_inputs[key] then
			print("Press: "..key)
		end
		prev_inputs[key] = nil
	end
	for key, pressed in pairs(prev_inputs) do
		if pressed and not inputs[key] then
			print("Unpress: "..key)
		end
	end
	prev_inputs = inputs
	emu.yield();
end
```

### Output
On 2.10:
```
Press: Ctrl
Press: Ctrl+Shift
Press: Shift
Press: A
Press: Ctrl+Shift+A
Unpress: A
Unpress: Ctrl+Shift+A
Unpress: Ctrl+Shift
Unpress: Shift
Unpress: Ctrl
```

On 6eecdd2194d49493c10d02b0b651e3574e8f49fb:
```
Press: Ctrl
Press: Shift
Press: A
Unpress: A
Unpress: Shift
Unpress: Ctrl
```

### Host env.
- BizHawk dev builds after 6eecdd2194d49493c10d02b0b651e3574e8f49fb

kalimag:
[...]

PR #4451, opened 2025-09-18
Fix InputApi

SuuperW:
See issue #4449 

It is questionable if the accidental change to the API behavior is good. This PR restores the old behavior and updates the API documentation to clarify behavior. (with the caveat that hotkey priority can no longer hide inputs from the API)

Note that the new `ApiInputCoalescer` class would not be necessary if we do not want controllers to see holding A and pressing Shift as Shift+A.

If there is consensus that we want to do this, I will update the comments in the test code and merge.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/InputApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/InputApi.cs
@@ -18,7 +18,7 @@ namespace BizHawk.Client.Common
 		public Dictionary<string, bool> Get()
 		{
 			var buttons = new Dictionary<string, bool>();
-			foreach (var (button, _) in _inputManager.ControllerInputCoalescer.BoolButtons().Where(kvp => kvp.Value)) buttons[button] = true;
+			foreach (var (button, _) in _inputManager.HostInputCoalescer.BoolButtons().Where(kvp => kvp.Value)) buttons[button] = true;
 			return buttons;
 		}
 
@@ -44,7 +44,7 @@ namespace BizHawk.Client.Common
 			=> _inputManager.ControllerInputCoalescer.AxisValues().ToDictionary();
 
 		public IReadOnlyList<string> GetPressedButtons()
-			=> _inputManager.ControllerInputCoalescer.BoolButtons().Where(static kvp => kvp.Value)
+			=> _inputManager.HostInputCoalescer.BoolButtons().Where(static kvp => kvp.Value)
 				.Select(static kvp => kvp.Key).ToList();
 	}
 }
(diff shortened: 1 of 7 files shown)


YoshiRulz:
If having a subclass just for ApiHawk is the easiest way to do this then so be it.
> Note that the new `ApiInputCoalescer` class would not be necessary if we do not want controllers to see holding A and pressing Shift as Shift+A.

Typo? We don't want that, and you have `TestOutOfOrderModifierCombinationIsNotCombined` which explicitly tests against that.

Have some more tests:
```diff
From 7f89459854522b2174860db614e2ed21122421b0 Mon Sep 17 00:00:00 2001
From: YoshiRulz <OSSYoshiRulz+git@gmail.com>
Date: Thu, 18 Sep 2025 20:13:02 +1000
Subject: [PATCH] Add tests with multiple modifier keys

---
 .../Api/InputApiTests.cs                      | 27 +++++++++++++++++++
 1 file changed, 27 insertions(+)

diff --git a/src/BizHawk.Tests.Client.Common/Api/InputApiTests.cs b/src/BizHawk.Tests.Client.Common/Api/InputApiTests.cs
index e992821298..2e3ddab863 100644
--- a/src/BizHawk.Tests.Client.Common/Api/InputApiTests.cs
+++ b/src/BizHawk.Tests.Client.Common/Api/InputApiTests.cs
@@ -141,6 +141,33 @@ namespace BizHawk.Tests.Client.Common.Api
 			Assert.IsTrue(buttons.Contains("Shift+A"));
 		}
 
+		[TestMethod]
+		public void TestMultipleModifiersBeforeKey()
+		{
+			Context context = new();
+			context.MakePressEvent("Shift");
+			context.MakePressEvent("Ctrl");
+			context.MakePressEvent("Shift+Ctrl");
+			context.MakePressEvent("A");
+			context.MakePressEvent("Ctrl+Shift+A");
+			CollectionAssert.AreEquivalent(
+				new[] { "Shift", "Ctrl", "Shift+Ctrl", "A", "Ctrl+Shift+A" },
+				context.api.GetPressedButtons().ToArray());
+		}
+
+		[TestMethod]
+		public void TestMultipleModifiersAfterKey()
+		{
+			Context context = new();
+			context.MakePressEvent("A");
+			context.MakePressEvent("Shift");
+			context.MakePressEvent("Ctrl");
+			context.MakePressEvent("Shift+Ctrl");
+			CollectionAssert.AreEquivalent(
+				new[] { "A", "Shift", "Ctrl", "Shift+Ctrl" },
+				context.api.GetPressedButtons().ToArray());
+		}
+
 		[TestMethod]
 		public void TestReleaseRegularButtonReleasesCombination()
 		{
-- 
2.49.0
```

SuuperW:
> Typo? We don't want that, and you have `TestOutOfOrderModifierCombinationIsNotCombined` which explicitly tests against that.

*controllers* do that. You're right that the Api does not.

I will add your tests and another, but if you press a second button while holding a modifier there is no press event for the second key by itself.

SuuperW:
Merged so that we can keep the old API behavior in the upcoming release. If anyone has a concrete proposal for a better API, make a new issue/PR.

Issue #3224, opened 2022-04-22
EmuHawk on Win10 produces "Illegal characters in path" error when opening a recent ROM via Windows's "Recent" menu


webbedspace:
![image](https://user-images.githubusercontent.com/83939/164650301-5b7c2179-0304-422a-9548-7cb24432da23.png)
![image](https://user-images.githubusercontent.com/83939/164650800-d97ea68d-81ba-48df-a257-c26ae1a1519d.png)
Whenever I try to open the following ROM path `B:\ROMs\B3313 v0.7 by Christopher Rios Lillo.z64` using Windows 10's "Recent" menu, in either the Start menu or by right-clicking it on the Taskbar, the following error occurs:
```
System.ArgumentException: Illegal characters in path.
   at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
   at System.IO.FileInfo.Init(String fileName, Boolean checkHost)
   at BizHawk.Common.PathExtensions.PathExtensions.MakeAbsolute(String path, String cwd)
   at BizHawk.Client.EmuHawk.MainForm..ctor(ParsedCLIFlags cliFlags, IGL gl, Func`1 getConfigPath, Func`1 getGlobalConfig, Action`1 updateGlobalSound, String[] args, IMovieSession& movieSession, Boolean& exitEarly)
   at BizHawk.Client.EmuHawk.Program.SubMain(String[] args)
```
This error does NOT occur if opening it using the **internal** Recent ROM menu inside EmuHawk.

YoshiRulz:
Interesting. Stacktrace suggests it's a filesystem issue.

Which version are you using?

webbedspace:
This is EmuHawk 2.8 on Windows 10 Pro 19044

Meerkov:
Didn't this happen with the Lua system too? It's probably because it's taking an Absolute path here, and trying to append it to the Working Directory.

(Completely blind guess, but it sounds right to me).

YoshiRulz:
It's writing e.g. `.\EmuHawk.exe "*OpenRom*C:\path\to.rom"` but only expecting `C:\path\to.rom`.

kalimag:
This still happens with roms inside archives, due to the pipe in the path:

```
[...]

PR #4452, opened 2025-09-18
Ensure absolute paths in HawkFile

kalimag:
### Changes

- Use `Path.GetFullPath()` to turn the path passed into `HawkFile` into an absolute path (if it isn't already)
- Use that path in the rest of `RomLoader.LoadRom`

---

Generally, the input paths that reach `HawkFile` are already absolute. However, relative paths can enter through the tool/Lua `openrom` API. This is particularly a problem for Lua because 1) the working directory is changed to the script directory during execution and 2) Lua doesn't have easy access to path manipulation/working directory APIs to generate absolute paths.

Calling `client.openrom`  with a path relative to the script works for the most part, but causes various problems when BizHawk tries to access that same relative path outside of script execution, with a different working directory:
- "Restart Core" doesn't work (this just tries to re-open the current rom, but now doesn't find it)
- The "Recent ROM" menu entries are useless
- CCD reader sometimes can't open the associated `.sub` file (https://github.com/authorblues/bizhawk-shuffler-2/issues/80)
- Hex editor file domain doesn't work

This PR fixes those issues.

Note: the issue could also be handled in the Lua API where it is most pertinent, but I figured doing it centrally would be preferable. It also needs to do the `"foo.zip|rom.nes"` parsing first, and I didn't want to duplicate that logic.

Not sure about just catching all exceptions, but previously invalid paths passed to `HawkFile` would not raise exceptions and I didn't want to change that behavior. There are places without exception handling that just rely on `HawkFile.Exists`, e.g. the NesHawk palette config.

Check if completed:

### DIFF
--- a/src/BizHawk.Common/HawkFile/HawkFile.cs
+++ b/src/BizHawk.Common/HawkFile/HawkFile.cs
@@ -81,16 +81,9 @@ namespace BizHawk.Common
 			string? autobind = null;
 			var split1 = SplitArchiveMemberPath(path);
 			if (split1 != null) (path, autobind) = split1.Value;
-			try
-			{
-				if (!string.IsNullOrEmpty(path)) path = Path.GetFullPath(path);
-			}
-			catch
-			{
-				// let invalid paths be handled by File.Exists check instead of exception
-			}
-			FullPathWithoutMember = path;
 			Exists = _rootExists = File.Exists(path);
+			if (_rootExists) path = Path.GetFullPath(path);
+			FullPathWithoutMember = path;
 			if (!_rootExists) return;
 
 			if (DearchivalMethod != null && allowArchives)


kalimag:
While checking how other places in BizHawk handle this, I just noticed that #3224 still happens for zipped roms, because `MakeAbsolute()` can't handle the `|` in the path. If this PR is merged, that could also be solved by removing this line altogether:

https://github.com/TASEmulators/BizHawk/blob/b98968c4b802551e9fbadc5f981dc9905450358b/src/BizHawk.Client.EmuHawk/MainForm.cs#L695

YoshiRulz:
Hmm was `HawkFile` meant to support relative paths? The ctor argument is written into `FullPathWithoutMember` which does seem like it should be absolute. But then it was written in a lax way:
https://github.com/TASEmulators/BizHawk/blob/5f5bf639c2509c606871faceb06760ebf302496a/BizHawk.Common/HawkFile.cs#L211-L223

I'd like to let users use relative paths if they want, but it's hard to get right and bugs often slip through, like with the recents menu. Lua scripts having a different working directory makes it worse. So maybe converting inputs to absolute right away is for the best.

kalimag:
>Hmm was HawkFile meant to support relative paths?

There's no issue with HawkFile, it doesn't care about relative vs absolute paths and reads from them just fine. It opens the file in the constructor, so it's not affected by working directory changes once created.

This PR doesn't really affect how `HawkFile` works at all, it mainly affects places that rely on `HawkFile.FullPathWithoutMember`/`CanonicalFullPath` properties to reopen the file or open associated files in different contexts, because they're not actually guaranteed to be full paths.

My thinking was:
1. The relative-to-absolute conversion needs to happen somewhere
2. It'd be better to do it in one central place rather than doing it at every possible source of paths, because that's a source of bugs
3. It needs to happen in conjunction with the archive member splitting because `Path.GetFullPath("foo.zip|bar.nes")` doesn't work, and that happens in `HawkFile` (though it is duplicated in other places)
4. Other APIs like `FileInfo` and `FileStream` work the same way, in that they accept relative paths but immediately convert them to absolute paths

src/BizHawk.Common/HawkFile/HawkFile.cs
```diff
@@ -79,22 +79,28 @@ public static bool PathContainsPipe(string s)
 		/// <param name="delayIOAndDearchive">Pass <see langword="true"/> to only populate a few fields (those that can be computed from the string <paramref name="path"/>), which is less computationally expensive.</param>
 		public HawkFile([HawkFilePath] string path, bool delayIOAndDearchive = false, bool allowArchives = true)
 		{
+			string? autobind = null;
+			if (SplitArchiveMemberPath(path) is { } split) (path, autobind) = split;
+			try
+			{
+				if (!string.IsNullOrEmpty(path)) path = Path.GetFullPath(path);
```
YoshiRulz: I don't think there should be any I/O calls when `delayIOAndDearchive` is `true`. Maybe the path should be stored in a new field, with `FullPathWithoutMember` implemented as `=> Path.GetFullPath(RelPathWithoutMember);`?

edit: Looks like there's only 1 callsite which passes `delayIOAndDearchive: true`, and guess what the line immediately after it does?
https://github.com/TASEmulators/BizHawk/blob/4ca33d9765f630986888d93cf8d61f7acbabb13d/src/BizHawk.Client.EmuHawk/Extensions/ToolExtensions.cs#L62-L63
This needs refactoring, please hold...

kalimag: That wouldn't work if you do
```cs
var file = new HawkFile("foo.bin");
Environment.CurrentDirectory = "...";
Console.WriteLine(file.FullPathWithoutMember);
```
Though I think no code currently does anything like this.

But I also don't think `GetFullPath` actually does any IO, except for weird cases like short names? It certainly works on non-existing paths.

And the only place I can find where `delayIOAndDearchive` is even used immediately accesses `FullPathWithoutMember` anyway:
https://github.com/TASEmulators/BizHawk/blob/4ca33d9765f630986888d93cf8d61f7acbabb13d/src/BizHawk.Client.EmuHawk/Extensions/ToolExtensions.cs#L62-L63

YoshiRulz: Yep, removed it, please rebase.

kalimag: Done

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3612,8 +3611,9 @@ public bool LoadRom(string path, LoadRomArgs args, out bool failureIsFromAskSave
 			// what's the meaning of the last rom path when opening an archive? based on the archive file location
 			if (args.OpenAdvanced is OpenAdvanced_OpenRom)
 			{
-				var leftPart = path.Split('|')[0];
-				Config.PathEntries.LastRomPath = Path.GetFullPath(Path.GetDirectoryName(leftPart) ?? "");
+				var physicalPath = path;
+				if (HawkFile.SplitArchiveMemberPath(path) is { } split) physicalPath = split.ArchivePath;
+				Config.PathEntries.LastRomPath = Path.GetDirectoryName(Path.GetFullPath(physicalPath)) ?? "";
```
YoshiRulz, line 3616: This whole thing is just
```csharp
using (HawkFile hf = new(path)) Config.PathEntries.LastRomPath = hf.Directory ?? string.Empty;
```
(not sure how it could be `null` but want to be safe there)

src/BizHawk.Common/HawkFile/HawkFile.cs
```diff
@@ -81,6 +81,14 @@ public HawkFile([HawkFilePath] string path, bool allowArchives = true)
 			string? autobind = null;
 			var split1 = SplitArchiveMemberPath(path);
 			if (split1 != null) (path, autobind) = split1.Value;
+			try
+			{
+				if (!string.IsNullOrEmpty(path)) path = Path.GetFullPath(path);
+			}
+			catch
+			{
+				// let invalid paths be handled by File.Exists check instead of exception
```
YoshiRulz: Would this be better after the `File.Exists` call then?

src/BizHawk.Client.EmuHawk/MainForm.cs
kalimag: Well now that `delayIOAndDearchive` flag would be handy

kalimag: But yeah I think potentially parsing a big archive file twice file just to construct a path should be avoided?

The call sites could be a bit more succinct with this change:

```cs
public static (string ArchivePath, string? MemberPath) SplitArchiveMemberPath([HawkFilePath] string path)
{
	// ...
	return i == -1 ? (path, null) : (path.Substring(0, i), path.Substring(i + 1));
}

(path, var autobind) = SplitArchiveMemberPath(path);

(var physicalPath, _) = HawkFile.SplitArchiveMemberPath(path);
```

src/BizHawk.Common/HawkFile/HawkFile.cs
kalimag: Uh, probably? `GetFullPath` came first because of the `delayIOAndDearchive` path, but now that's gone.

I don't believe there's any code that cares about the `FullPath` properties if `Exists` is false, but technically the interface doesn't prevent it.

YoshiRulz: Well if the constructor throws it doesn't matter, the object isn't usable.

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: Oh right. Nevermind.

src/BizHawk.Common/HawkFile/HawkFile.cs
kalimag: Just to make sure I'm understanding you correctly:

In the current PR and on master, the `HawkFile` constructor does not throw even if you pass `""` or `"*?:"` as the path, `File.Exists` simply returns false. I added the try-catch because `Path.GetFullPath` *would* throw in cases that previously did not cause an exception.

Which of these options do you want:
1. [Current] Try-catch, no exceptions thrown, `FullPathWithoutMember` is absolute when `path` is valid but file does not exist
2. Only call `GetFullPath` if `File.Exists` is true, no exceptions,  `FullPathWithoutMember` may be relative when `path` is valid but file does not exist
3. Always call `GetFullPath`, exception will be thrown when `path` is invalid

YoshiRulz: I'm thinking 2). When `$PWD/$path` exists then that's the only reasonable inference for the user's intent (but for the Lua thing...). But when it doesn't exist, it's not obvious why and we shouldn't make any assumptions, certainly not without prompting the user. Maybe the relative path will end up in the config file and then get resolved next time EmuHawk is launched.

YoshiRulz:
To confirm: #3224 wasn't fully resolved, but now is? And the bug you linked in the shuffler repo?

kalimag:
Yes to both

PR #4453, opened 2025-09-19
Fix Windows jump list entries for libretro

kalimag:
- Escape double quotes in command line argument stored in the Windows jump list entry, to let serialized `OpenAdvanced_Libretro` values roundtrip properly

---

`OpenAdvanced_Libretro` objects are serialized like this:

    *Libretro*{"Path":"F:\\Emulation\\GBC\\rom.gb","CorePath":"F:\\Emulation\\BizHawk\\Libretro\\Cores\\sameboy_libretro.dll"}

.NET command line parsing strips out the double quotes, which either causes JSON deserialization errors or breaks the argument into multiple parts if the paths contain spaces. This is unrelated to BizHawk code or `System.CommandLine`, the arguments are already mangled in `Main(string args[])`.

Double quotes need to be escaped in [a particular way](https://learn.microsoft.com/en-us/dotnet/api/system.environment.getcommandlineargs#remarks). I think this is a MS/Windows thing and I'm not sure if .NET parses arguments the same on Linux, but the jump list stuff only works on Windows anyway.

(Realistically speaking a simple `string.Replace("\"", "\\\"")` probably would have worked fine in this case, since none of the json values should ever end in a backslash that could cause incorrect escape sequences)

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/JumpLists.cs
+++ b/src/BizHawk.Client.EmuHawk/JumpLists.cs
@@ -1,4 +1,5 @@
 using System.Reflection;
+using System.Text.RegularExpressions;
 using System.Windows;
 using System.Windows.Shell;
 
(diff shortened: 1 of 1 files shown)


kalimag:
Now that all these bugs are fixed, we can finally go back to not remembering the jump list is a thing that exists

PR #4455, opened 2025-09-21
Fixed DMC DMA bus conflicts with controllers

100thCoin:
In a previous PR I modified how controller strobing was done, and also made changes to controller clocking. After those changes, I added more tests to my test ROM, which (unbeknownst to me) was now failing due to my aforementioned changes. Here is what broke:

When the DMC DMA has a bus conflict with the controller ports, the controller does not get clocked an additional time, unlike the situation where the dummy read cycles of a DMC DMA read from the controller port.

Apparently, the TAS of Tiny Toons desynced after my previous changes. After fixing it, it once again completes the game:

TinyToonTAS

In addition to fixing how the DMC DMA bus conflict does or does not clock the controller, I also implemented the way the DMC DMA Bus conflict affects the data bus:

Conflcit

Keep in mind, this bus conflict can read the controllers, and read from $4015, but reading from address $4015 does not update the data bus, so the conflict really only affects the data bus if reading from the controller ports.

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/APU.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/APU.cs
@@ -950,9 +950,22 @@ namespace BizHawk.Emulation.Cores.Nintendo.NES
 			{
 				if (sample_length != 0)
 				{
+					
 					sample_buffer = apu.nes.ReadMemory((ushort)sample_address);
 					sample_buffer_filled = true;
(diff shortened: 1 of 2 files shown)


100thCoin:
Okay, so I re-did how the bus conflicts are handled.

Just to confirm, both of these tests are passing:

controller1

And now, so too are these:

controller2

Here's the full table of my test ROM:

table

vadosnaprimer:
For the reference: https://tasvideos.org/Forum/Posts/537942

PR #4458, opened 2025-09-21
Add player data to `Doom/things-lines.lua`

kalimag:
- Add offset data for the new Players memory domain/`player_t` to script
- Add an example function for using this data
- Add helper for defining struct offsets (and use this for mobj offsets as well)
- Some padding adjustments of the `mobj_t` offsets
- Update the mobj offset comment (this didn't match the actual offsets)

cc @vadosnaprimer 

Check if completed:
- [x] I have run any relevant test suites
- [x] I, the commit author, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2024-06-22) and am compliant

PR #4464, opened 2025-09-24
DSDA Lua script

kalimag:
(continued from #4458)

- Split definitions out into reusable module
- Add sector and line definitions

TODO:
- [x] Complete player and mobj structs
- [x] Heretic and Hexen enums
- [x] ~~Misc domain?~~
- [x] Make structs more convenient to use
  Ideally you can write code like `for mobj in pairs(dsda.mobjs) do print(mobj.type) end` and all the memory reading is abstracted away
  - [x] Better array handling
- [x] Remove `Objects` cache in `things-lines.lua` (doesn't work with spawned objects)
- [x] Cache lines instead?
  Seems to be a big performance cost on fancy Hexen maps
- [ ] Documentation
- [ ] Example code

If time permits:
- [x] Center zoom
- [x] Filter more mobj types (blood, smoke, decorations, etc.)
- [x] Stop shooting while clicking the map buttons
  - ~~Annoyingly, moving the map while paused isn't possible because of the way `gui.text` works~~

Check if completed:


vadosnaprimer:
Misc domain may have to wait until I gather feedback from Doom people.

Assets/Lua/Doom/dsda-data.lua
```diff
@@ -156,6 +157,126 @@ dsda.mobj = dsda.struct_layout()
...
+	.s32  ("floorlightsec")
+	.s32  ("ceilinglightsec")
+	.s16  ("floorpic")
+	.s16  ("ceilingpic")
+	.s16  ("lightlevel")
+	.s16  ("special")
+	.s16  ("tag")
+	.s32  ("cachedheight")
+	.s32  ("scaleindex")
+  	-- e6y
```
vadosnaprimer: That's just signature of who added it to the code.
https://doomwiki.org/wiki/Andrey_Budko_(entryway)

Assets/Lua/Doom/dsda-data.lua
```diff
@@ -156,6 +157,126 @@ dsda.mobj = dsda.struct_layout()
...
+	.s32  ("prevsec")
+	.s32  ("nextsec")
+	.s32  ("heightsec")
+	.s16  ("bottommap")
+	.s16  ("midmapmap")
+	.s16  ("topmap")
+	.s16  ("colormap")
+	.ptr  ("touching_thinglist")
+	.s32  ("linecount")
+	.ptr  ("lines")
```
vadosnaprimer: This pointer is to array of lines this sector consists of, but it's unclear how to expose those. Normally when TASing people just look up sector ID in level editor and track it by that ID in XDRE.

kalimag: Yeah, variable sized arrays don't really fit into the current system. I assume you could inline `line->frontsector/backsector->iSectorID` to the line domain, like the `v1`/`v2` coords. 

But I'm not familiar with Doom formats and have no idea how difficult it actually is to turn a list of lines into a polygon to display on the map, for example. What's the intended use of the sector domain?

vadosnaprimer: XDRE shows these when you enter sector ID
изображение

vadosnaprimer:
I think center zoom is done. Not 100% perfect but combined with mouse drag feels sufficient.

Assets/Lua/Doom/things-lines.lua
```diff
@@ -1,35 +1,54 @@
...
+-- Map colors (0xAARRGGBB or "name" or false)
+local ENEMY_COLOR       = 0xFFFF0000
+local ENEMY_WAIT_COLOR  = 0xFFCC0000
+local CORPSE_COLOR      = false -- 0x80CC0000
+local MISSILE_COLOR     = 0xFFFF8000
+local PLAYER_COLOR      = 0xFF0080FF
+local COUNTITEM_COLOR   = false -- 0xA0FFFFFF
+local ITEM_COLOR        = false -- 0xA0FFFFFF
+local MISC_COLOR        = false -- 0xFFFFFFFF
+local INERT_COLOR       = false -- 0x80808080
```
vadosnaprimer: I recommend a `Colors` table.

Assets/Lua/Doom/dsda-data.lua
```diff
@@ -0,0 +1,1864 @@
...
+local OUT_OF_BOUNDS <const>  = 0xFFFFFFFF -- no such index
+local WBX_POINTER_HI <const> = 0x36F
+local BusDomain <const>      = "System Bus"
+
+dsda.MAX_PLAYERS = 4
+-- sizes in bytes
+dsda.LINE_SIZE   = 256  -- sizeof(line_t) is 232, but we padded it for niceness
+dsda.MOBJ_SIZE   = 512  -- sizeof(mobj_t) is 464, but we padded it for niceness
+dsda.PLAYER_SIZE = 1024 -- sizeof(player_t) is 729, but we padded it for niceness
+dsda.SECTOR_SIZE = 512  -- sizeof(sector_t) is 344, but we padded it for niceness
```
vadosnaprimer: Maybe a `Sizes` table for these too?

Assets/Lua/Doom/dsda-data.lua
```diff
@@ -0,0 +1,1864 @@
...
+
+
+
+-- Locals
+local read_u8   = memory.read_u8
+local read_u24  = memory.read_u24_le
+local read_u32  = memory.read_u32_le
+local readfloat = memory.readfloat
```
vadosnaprimer: Similar shortcuts exist in the main script, looks like it makes sense to declare them once and then use everywhere?

Assets/Lua/Doom/things-lines.lua
```diff
@@ -471,24 +489,39 @@ event.onexit(function()
...
 
 while true do
-	if Init then init_objects() end
-	
+	local framecount = emu.framecount()
+	local paused = client.ispaused()
+
+	if Init then init_mobj_bounds() end
+
+	-- re-init cache after state load, rewind, etc.
```
vadosnaprimer: `event.onloadstate`

Assets/Lua/Doom/dsda-data.lua
```diff
@@ -494,6 +502,12 @@ dsda.line = dsda.struct_layout(nil, dsda.LINE_SIZE, "Lines")
 	.s32  ("healthgroup")
 	.ptr  ("tranmap")
 	.float("alpha")
+	.done ()
+
+assert(dsda.line.size   == 232, "line.size does not match sizeof(line_t)")
+assert(dsda.mobj.size   == 464, "mobj.size does not match sizeof(mobj_t)")
+assert(dsda.player.size == 792, "player.size does not match sizeof(player_t)")
+assert(dsda.sector.size == 344, "sector.size does not match sizeof(sector_t)")
```
vadosnaprimer: Having these as consts at the top is better because we can be descriptive with their names and reuse them if needed. Especially sensible since we have 2 types of sizes: padded and internal.

Assets/Lua/Doom/dsda-data.lua
```diff
@@ -20,11 +20,6 @@ local WBX_POINTER_HI <const> = 0x36F
 local BusDomain <const>      = "System Bus"
 
 dsda.MAX_PLAYERS = 4
--- sizes in bytes
-dsda.LINE_SIZE   = 256  -- sizeof(line_t) is 232, but we padded it for niceness
-dsda.MOBJ_SIZE   = 512  -- sizeof(mobj_t) is 464, but we padded it for niceness
-dsda.PLAYER_SIZE = 1024 -- sizeof(player_t) is 729, but we padded it for niceness
-dsda.SECTOR_SIZE = 512  -- sizeof(sector_t) is 344, but we padded it for niceness
```
vadosnaprimer: Having these as consts at the top is better because we can be descriptive with their names and reuse them if needed. Especially sensible since we have 2 types of sizes: padded and internal.

Assets/Lua/Doom/things-lines.lua
kalimag: That isn't raised on rewind

Although I guess it's possible to load a state with the same framecount as the current state, so doing both might be good

Assets/Lua/Doom/dsda-data.lua
kalimag: All of them can be accessed like `dsda.line.size` and `dsda.line.padded_size`, which seems descriptive enough. The asserts are just there to make sure the script calculated them correctly

But sure, I'll put them them back into tables

kalimag:
Should the script pause during TAStudio seek?

vadosnaprimer:
Scripts only pause by default if you seek with turbo, and there's an option to not pause them even during turbo. Outside of turbo we don't normally make them pause, because sometimes it's needed to keep running, and switching that on and off in the script itself would be annoying. Turbo-seek checkbox is faster to use and more organized.

vadosnaprimer:
Found the address of `rndindex` in system bus - 0x003489C0. Seems consistent across IWADs and PWADs. That's the value xdre shows, so we probably should too, as a start.

kalimag:
Stupid question: can we get symbols from the core build?

Edit: turns out the wbx ELF already contains symbols:

```
  1754: 0000036f0039cd40  2048 OBJECT  LOCAL  DEFAULT   15 channelinfo
  1755: 0000036f003488bc   268 OBJECT  LOCAL  DEFAULT   15 rng
  1756: 0000036f00345468     4 OBJECT  LOCAL  DEFAULT   15 demo_insurance
  1757: 0000036f0039ad88     4 OBJECT  LOCAL  DEFAULT   15 compatibility_level
  1758: 0000036f0031e498     8 OBJECT  LOCAL  DEFAULT   14 rndtable
```
`0x003488bc + NUMPRCLASS*sizeof(int) == 0x003489C0`

Only issue here is that they'll presumably change with every core update. But we could probably automate extract wbx > objdump > lua table with a script

vadosnaprimer:
I have a feeling we don't need the Misc domain anymore :D

Assets/Lua/Doom/things-lines.lua
```diff
@@ -267,7 +267,7 @@ end
 local function iterate()
 	if Init then return end
 
-	for addr, mobj in pairs(structs.mobj.items) do
+	for _, mobj in pairs(structs.global.mobjs) do
```
vadosnaprimer: Did you notice any speed difference?

kalimag: Benchmarking reads from Lines and System Bus shows small improvements for u32 reads and huge improvements for bulk reads (which the script currently doesn't use), but I didn't notice a significant FPS increase from this change in practice.

```
             Lines read_u32_le took 0.00495250
            SysBus read_u32_le took 0.00378650
      Lines read_binary_string took 0.00148250
     SysBus read_binary_string took 0.00009050
```

<details>
<summary>Code</summary>

```lua
local clock = os.clock
local print = print
local function bench(name, iterations, func, ...)
	local start = clock() -- this is wall time on Windows, CPU time on Linux
	local result
	for i = 1, iterations do
		result = func(i, ...)
	end
	local diff = clock() - start
	print(string.format("%30s took %.8f", name, diff / iterations, result ~= nil and ", result: "..result or ""))
end

console.clear()

local iterations = 2000
local page_size = 4096
local word_count = page_size * 2 / 4
local byte_count = page_size * 2

local read_u32_le = memory.read_u32_le
local read_bytes_as_binary_string = memory.read_bytes_as_binary_string

local function bench_read_u32_le_aligned(_, domain)
	for i = 0, word_count - 1 do
		read_u32_le(i * 4, domain)
	end
end

local function bench_read_binary_string(_, domain)
	read_bytes_as_binary_string(0, byte_count, domain)
end

bench("Lines read_u32_le", iterations, bench_read_u32_le_aligned, "Lines")
bench("SysBus read_u32_le", iterations, bench_read_u32_le_aligned, "System Bus")
bench("Lines read_binary_string", iterations, bench_read_binary_string, "Lines")
bench("SysBus read_binary_string", iterations, bench_read_binary_string, "System Bus")
```

</details>

vadosnaprimer: I should implement reading other things than 1 byte...

kalimag: To be honest, at this point I'm not sure if the special domains are even necessary anymore. With the symbols and source code being available, you can get everything from the system bus without much trouble.

I suppose some people who don't want to use the dsda script for whatever reason may still find them easier to work with

vadosnaprimer: Fair! And with array sizes changing, only Players can be turned into a pointer based domain, but players are far from majority of this script so I don't think it'd matter.

Assets/Lua/Doom/dsda-data.lua
vadosnaprimer: What about these tho?

kalimag: I don't think it makes sense to share them across scripts, but in any case they're not used anymore in `things-lines.lua` because all the memory reads happen through the dsda script now. 

Unless you would like to keep them for something, I'll just remove them from `things-lines.lua`

vadosnaprimer: Okay, please remove them then!

PR #4466, opened 2025-09-25
Fix printing Lua tables that contain mixed or non-comparable keys

### LINKED ISSUE #4543: Lua Console stack trace in `print` when a table has keys of different types
### Summary

Calling `print` on a table that contains keys of more than one type (e.g. string, boolean, integer) results in a C# exception and stack trace ("A .NET exception occured in user-code").

I suspect [`SerializeTable`](https://github.com/TASEmulators/BizHawk/blob/2.11/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/ConsoleLuaLibrary.cs#L62), which has a special-case check for all integer keys, and then tries to sort the keys generically.

### Repro

Try this, a table where one key is an integer and one is a string:

```
print({[1] = "a", x = "b"})
```

It works with other combinations, such as:

```
print({[true] = "a", x      = "b"})
print({[1]    = "a", [true] = "b"})
print({[1.5]  = "a", x      = "b"})
print({[1.5]  = "a", [2]    = "b"})
```

### Output

```
NLua.Exceptions.LuaScriptException: [string "main"]:1: A .NET exception occured in user-code
System.ArgumentException: Object must be of type Int64.
  at System.Int64.CompareTo (System.Object value) [0x00030] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Collections.Comparer.Compare (System.Object a, System.Object b) [0x0004e] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Collections.Generic.ObjectComparer`1[T].Compare (T x, T y) [0x00000] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Linq.EnumerableSorter`2[TElement,TKey].CompareAnyKeys (System.Int32 index1, System.Int32 index2) [0x00000] in <d22af090bceb4be792f53595cf074724>:0 
  at System.Collections.Generic.ComparisonComparer`1[T].Compare (T x, T y) [0x00000] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Collections.Generic.ArraySortHelper`1[T].SwapIfGreater (T[] keys, System.Comparison`1[T] comparer, System.Int32 a, System.Int32 b) [0x00007] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Collections.Generic.ArraySortHelper`1[T].IntroSort (T[] keys, System.Int32 lo, System.Int32 hi, System.Int32 depthLimit, System.Comparison`1[T] comparer) [0x00022] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Collections.Generic.ArraySortHelper`1[T].IntrospectiveSort (T[] keys, System.Int32 left, System.Int32 length, System.Comparison`1[T] comparer) [0x00016] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
  at System.Collections.Generic.ArraySortHelper`1[T].Sort (T[] keys, System.Int32 index, System.Int32 length, System.Collections.Generic.IComparer`1[T] comparer) [0x0000e] in <12b418a7818c4ca0893feeaaf67f1e7f>:0 
```

Depending on the table, different type are possible in the `ArgumentExecption`:

```
System.ArgumentException: Object must be of type String.
System.ArgumentException: Object must be of type Int64.
System.ArgumentException: Object must be of type Double.
```

### Host env.

- BizHawk 2.11; Debian 12

kalimag:
- Fix `ArgumentException` when printing a table that contains keys of mixed types `{1, a=1}` or non-`IComparable` types `{[assert]="function", [{}]="table"}` (regression in 631e5c37b4df48bd68f9c4e1b8ace033aca05b5a)
- Sort keys by most types, even in mixed tables
- Rewrite printing code with a stringbuilder
- Only surround strings in quotes, both for keys and values

### Repro

```lua
print({
	1,
	[56789.56789] = 56789.56789,
	p = "p",
	[true] = true,
	2,
	[1234.1234] = 1234.1234,
	[forms.createcolor(255, 0, 0, 255)] = "red",
	x = "x",
	3,
	c = "c",
	[coroutine.create(function() end)] = "coroutine",
	[false] = false,
	["1"] = "one",
	[{}] = {},
	[assert] = assert,
	[-1] = -1,
	[forms.createcolor(0, 0, 0, 255)] = "black",
	[0xFFFFFFFFFFFFFFF] = "longmax",
	[math.huge]= "infinity",
})
```

#### 2.10
```
"1": "1"
"-1": "-1"
"1": "one"
"1152921504606846975": "longmax"
"1234,1234": "1234,1234"
"2": "2"
"3": "3"
"56789,56789": "56789,56789"
"∞": "infinity"
"c": "c"
"Color [A=255, R=0, G=0, B=0]": "black"
"Color [A=255, R=255, G=0, B=0]": "red"
"False": "False"
"function": "function"
"p": "p"
"table": "table"
"thread": "coroutine"
"True": "True"
"x": "x"
```

#### 2.11
```
NLua.Exceptions.LuaScriptException: [string "main"]:1: A .NET exception occured in user-code
System.ArgumentException: Object must be of type "Int64".
```
with a simple table like `{1, 2, 3}`:
```
1: "1"
2: "2"
3: "3"
```

#### PR
```
-1: -1
1: 1
2: 2
3: 3
1152921504606846975: "longmax"
"1": "one"
"c": "c"
"p": "p"
"x": "x"
1234,1234: 1234,1234
56789,56789: 56789,56789
∞: "infinity"
False: False
True: True
Color [A=255, R=0, G=0, B=0]: "black"
Color [A=255, R=255, G=0, B=0]: "red"
function: function
table: table
thread: "coroutine"
```

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/lua/LuaHelper.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelper.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 
@@ -39,7 +40,7 @@ namespace BizHawk.Client.Common
 			static void Append(StringBuilder sb, object value)
 			{
 				if (value is string str) sb.Append('"').Append(str).Append('"');
-				else sb.Append(value);
+				else sb.AppendFormat(CultureInfo.InvariantCulture, "{0}", value);
 			}
 		}
 	}
--- a/src/BizHawk.Client.Common/lua/LuaHelper.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelper.cs
@@ -12,7 +12,8 @@ namespace BizHawk.Client.Common
 		public static LuaTable EnumerateToLuaTable<T>(this NLuaTableHelper tableHelper, IEnumerable<T> list, int indexFrom = 1)
 			=> tableHelper.ListToTable(list as IReadOnlyList<T> ?? list.ToList(), indexFrom);
 
-		public static string Serialize(this LuaTable lti)
+		/// <remarks>Intended for printing tables to the console</remarks>
+		public static string PrettyPrintShallow(this LuaTable lti)
 		{
 			var sorted = lti
 				.OrderBy(static item => item.Key switch
--- a/src/BizHawk.Client.Common/lua/LuaHelper.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelper.cs
@@ -28,7 +28,7 @@ namespace BizHawk.Client.Common
 				{
 					double d => d,
 					long i => i,
-					_ => (double?)null
+					_ => (double?)null,
 				})
 				.ThenBy(static item => item.Key as long?) // sort large integers that double can't represent
 				.ThenBy(static item => item.Key is not (long or double) ? item.Key.ToString() : null, StringComparer.InvariantCulture);
--- a/src/BizHawk.Client.Common/lua/LuaHelper.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelper.cs
@@ -18,13 +18,18 @@ namespace BizHawk.Client.Common
 				.OrderBy(static item => item.Key switch
 				{
 					long => 0,
+					double => 0,
 					string => 1,
-					double => 2,
-					bool => 3,
-					_ => 4, // tables, functions, ...
+					bool => 2,
+					_ => 3, // tables, functions, ...
 				})
-				.ThenBy(static item => item.Key as long?)
-				.ThenBy(static item => item.Key as double?)
+				.ThenBy(static item => item.Key switch
+				{
+					double d => d,
+					long i => i,
+					_ => (double?)null
+				})
+				.ThenBy(static item => item.Key as long?) // sort large integers that double can't represent
 				.ThenBy(static item => item.Key is not (long or double) ? item.Key.ToString() : null, StringComparer.InvariantCulture);
 
 			var sb = new StringBuilder();
--- a/src/BizHawk.Client.Common/lua/LuaHelper.cs
+++ b/src/BizHawk.Client.Common/lua/LuaHelper.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 
 using NLua;
 
@@ -9,5 +10,37 @@ namespace BizHawk.Client.Common
 	{
 		public static LuaTable EnumerateToLuaTable<T>(this NLuaTableHelper tableHelper, IEnumerable<T> list, int indexFrom = 1)
 			=> tableHelper.ListToTable(list as IReadOnlyList<T> ?? list.ToList(), indexFrom);
+
+		public static string Serialize(this LuaTable lti)
+		{
+			var sorted = lti
+				.OrderBy(static item => item.Key switch
+				{
+					long => 0,
+					string => 1,
+					double => 2,
+					bool => 3,
+					_ => 4, // tables, functions, ...
+				})
+				.ThenBy(static item => item.Key as long?)
+				.ThenBy(static item => item.Key as double?)
+				.ThenBy(static item => item.Key is not (long or double) ? item.Key.ToString() : null, StringComparer.InvariantCulture);
+
+			var sb = new StringBuilder();
+			foreach (var item in sorted)
+			{
+				Append(sb, item.Key);
+				sb.Append(": ");
+				Append(sb, item.Value);
+				sb.Append('\n');
+			}
+			return sb.ToString();
+
+			static void Append(StringBuilder sb, object value)
+			{
+				if (value is string str) sb.Append('"').Append(str).Append('"');
+				else sb.Append(value);
+			}
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/ConsoleLuaLibrary.cs
```diff
@@ -61,12 +60,34 @@ private void LogWithSeparator(string separator, string terminator, params object
...
+					{
+						long => 0,
+						string => 1,
+						double => 2,
+						bool => 3,
+						_ => 4, // tables, functions, ...
+					})
+					.ThenBy(static item => item.Key as long?)
+					.ThenBy(static item => item.Key as double?)
+					.ThenBy(static item => item.ToString());
```
YoshiRulz: Can you pass `StringComparer.Invariant`?

YoshiRulz (changes requested):
Add a unit test please.

I think you'll now be calling `ToString` twice on each table/other. Maybe do `.Select(o => o is not (long or string or double or bool) ? o.ToString() : o)` first.

kalimag:
> Add a unit test please.

I'm not sure how, since it's so tightly integrated into printing directly to LuaConsole. I guess I could pull out the local `SerializeTable` method and make it public?

> I think you'll now be calling `ToString` twice on each table/other. Maybe do `.Select(o => o is not (long or string or double or bool) ? o.ToString() : o)` first.

Ugh, I was quite sure that the `ThenBy` clauses would only be evaluated for items that are equal in previous steps, but surprisingly that's not the case. That's very annoying.

I was under the impression that `long` and `double` values wouldn't end up in that code path, and most other types don't allocate in `ToString()`, with the exception of `userdata` types like `Color` that can have a custom `ToString` implementation. But I also unintentionally used `item.ToString()` rather than `item.Key.ToString()`

YoshiRulz:
>> Add a unit test please.
>
> I'm not sure how, since it's so tightly integrated into printing directly to LuaConsole. I guess I could pull out the local `SerializeTable` method and make it public?

It's already a `static` local method, if you ~~lift it to a member~~ move it to a member of the base class then you should be able to call it from `LuaTests`.

>> I think you'll now be calling `ToString` twice on each table/other. Maybe do `.Select(o => o is not (long or string or double or bool) ? o.ToString() : o)` first.
>
> Ugh, I was quite sure that the `ThenBy` clauses would only be evaluated for items that are equal in previous steps, but surprisingly that's not the case. That's very annoying.

edit: I was unaware of that, actually I was referring to the implicit `ToString` as part of the `StringBuilder.Append(object)` call.

kalimag:
> I was unaware of that, actually I was referring to the implicit `ToString` as part of the `StringBuilder.Append(object)` call.

Yea, I got it. `ToString` is still called twice for misc values, but for most values that can end up here `ToString()` is a trivial `return` (`string`, `boolean`, `table`, `function`, `thread`). The only exception are `userdata` values (e.g. `forms.createcolor()``, but printing tables with keys like that is unusual enough not to be a concern, IMO.

src/BizHawk.Client.Common/lua/LuaHelper.cs
```diff
@@ -9,5 +10,37 @@ public static class LuaExtensions
 	{
 		public static LuaTable EnumerateToLuaTable<T>(this NLuaTableHelper tableHelper, IEnumerable<T> list, int indexFrom = 1)
 			=> tableHelper.ListToTable(list as IReadOnlyList<T> ?? list.ToList(), indexFrom);
+
+		public static string Serialize(this LuaTable lti)
```
YoshiRulz: Maybe call it `PrettyPrintShallow` or something to indicate it's not for general use

src/BizHawk.Client.Common/lua/LuaHelper.cs
```diff
@@ -9,5 +10,37 @@ public static class LuaExtensions
...
 			=> tableHelper.ListToTable(list as IReadOnlyList<T> ?? list.ToList(), indexFrom);
+
+		public static string Serialize(this LuaTable lti)
+		{
+			var sorted = lti
+				.OrderBy(static item => item.Key switch
+				{
+					long => 0,
+					string => 1,
+					double => 2,
```
YoshiRulz: Can you add another test case with a mix of integers and floats? Because I have a feeling this will give an unintuitive ordering (which is part of why I only did `long`, that approximated Lua's sequence rules).

kalimag: Yeah, mixed integer and decimal numbers will not be sorted together. I don't think there's any simple solution to this without making the sorting much more complex, but maybe `double` should come directly after `long`.

I didn't include them in the test because their formatting is locale dependent (`.` vs `,`). Maybe that should change? FWIW, Lua's `string.format` seems to always use `.` as the decimal separator.

YoshiRulz: `AppendFormat(CultureInfo.InvariantCulture, "{0}", value)` should fix the separator.

YoshiRulz (approved):

PR #4471, opened 2025-09-25
Create a `MemoryDomain` subclass which wraps `IDebuggable` reg get/set

YoshiRulz:
[dev build for branch | TASEmulators:regs-mem-domain](https://nightly.link/TASEmulators/BizHawk/workflows/ci/regs-mem-domain?preview)

Screenshot_20250926_042924

### DIFF
--- a/src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
+++ b/src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
@@ -1,0 +1,115 @@
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+using BizHawk.Common.StringExtensions;
+
+using LookupEntry = (string RegName, byte RegWidthBytes, byte ByteOffset);
+
+namespace BizHawk.Emulation.Common
+{
+	public sealed class RegistersMemoryDomain : MemoryDomain
+	{
+		private const string ERR_FMT_STR_INVALID_ADDR = "invalid address, must be in 0..<{0}";
+
+		public static ImmutableArray<LookupEntry> GenLookup(IDictionary<string, RegisterValue> regs)
+		{
+			List<LookupEntry> entries = new();
+			foreach (var (name, val) in regs)
+			{
+				var widthBytes = unchecked((byte) (val.BitSize / 8));
+				if (val.BitSize % 8 is not 0) widthBytes++;
+				byte i = widthBytes;
+				while (i > 0) entries.Add((RegName: name, RegWidthBytes: widthBytes, ByteOffset: --i));
+			}
+			while (entries.Count % sizeof(uint) is not 0) entries.Add((string.Empty, 0, 0)); // padding for Hex Editor
+			return entries.ToImmutableArray();
+		}
+
+		private readonly IDebuggable _debuggableCore;
+
+		private readonly ImmutableArray<LookupEntry> _lookup;
+
+		public RegistersMemoryDomain(IDebuggable debuggableCore)
+		{
+			_debuggableCore = debuggableCore;
+			var regs = _debuggableCore.GetCpuFlagsAndRegisters();
+			_lookup = GenLookup(regs);
+			EndianType = Endian.Big;
+			var cpuName = regs.Keys.CommonPrefix();
+			Name = cpuName.Length is 0 ? "CPU registers" : $"{new string(cpuName)/*trailing space*/}registers";
+			Size = _lookup.Sum(static entry => entry.RegWidthBytes);
+			WordSize = sizeof(byte);
+			Writable = !debuggableCore.GetType().GetMethod("SetCpuRegister").CustomAttributes
+				.Any(ad => ad.AttributeType == typeof(FeatureNotImplementedAttribute));
+		}
+
+		public override byte PeekByte(long addr)
+		{
+			if (addr < 0 || _lookup.Length <= addr) throw new ArgumentOutOfRangeException(paramName: nameof(addr), addr, message: string.Format(ERR_FMT_STR_INVALID_ADDR, _lookup.Length));
+			var (name, width, offset) = _lookup[unchecked((int) addr)];
+			if (width is 0) return default; // padding for Hex Editor
+			var toReturn = unchecked((uint) _debuggableCore.GetCpuFlagsAndRegisters()[name].Value);
+			toReturn >>= 8 * offset;
+			return unchecked((byte) toReturn);
+		}
+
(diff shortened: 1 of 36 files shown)


vadosnaprimer (approved):

vadosnaprimer:
I like the concept. No idea about the code tho.

src/BizHawk.Emulation.Common/Interfaces/Services/IMemoryDomains.cs
```diff
@@ -13,7 +13,7 @@ namespace BizHawk.Emulation.Common
 	/// If this service is available the client will expose many RAM related tools such as the Hex Editor, RAM Search/Watch, and Cheats
 	/// In addition, this is an essential service for effective LUA scripting, and many other tools
 	/// </summary>
-	public interface IMemoryDomains : IEnumerable<MemoryDomain>, IEmulatorService
+	public interface IMemoryDomains : IReadOnlyList<MemoryDomain>, IEmulatorService
```
SuuperW, line 16: Is this change related to the new memory domain? I don't like seeing random unrelated diffs.

src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
```diff
@@ -0,0 +1,115 @@
...
+
+using LookupEntry = (string RegName, byte RegWidthBytes, byte ByteOffset);
+
+namespace BizHawk.Emulation.Common
+{
+	public sealed class RegistersMemoryDomain : MemoryDomain
+	{
+		private const string ERR_FMT_STR_INVALID_ADDR = "invalid address, must be in 0..<{0}";
+
+		public static ImmutableArray<LookupEntry> GenLookup(IDictionary<string, RegisterValue> regs)
```
SuuperW: Should be private.

src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
```diff
@@ -0,0 +1,115 @@
...
+		private readonly ImmutableArray<LookupEntry> _lookup;
+
+		public RegistersMemoryDomain(IDebuggable debuggableCore)
+		{
+			_debuggableCore = debuggableCore;
+			var regs = _debuggableCore.GetCpuFlagsAndRegisters();
+			_lookup = GenLookup(regs);
+			EndianType = Endian.Big;
+			var cpuName = regs.Keys.CommonPrefix();
+			Name = cpuName.Length is 0 ? "CPU registers" : $"{new string(cpuName)/*trailing space*/}registers";
```
SuuperW, line 40: Not sure it makes sense to assume all registers are "CPU" registers when there is no common prefix.

SuuperW (commented):
1) I think it would be better to not modify cores' code. It should be possible to conditionally add the registers domain outside of the core, depending on if it provides the required service.

2) I don't exactly like the inefficiencies here with reading and writing more than necessary, but I suppose there's not a lot you could realistically do about that.

3) I also see `unchecked` in a lot of places. Does this provide any benefit? Since unchecked is default, this seems like unneeded fluff.

src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
YoshiRulz: What else would they be? It's called `GetCpuFlagsAndRegisters` in the service.

src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
YoshiRulz:
```suggestion
		private static ImmutableArray<LookupEntry> GenLookup(IDictionary<string, RegisterValue> regs)
```

src/BizHawk.Emulation.Common/Interfaces/Services/IMemoryDomains.cs
YoshiRulz: Yes, I'm indexing into it here:
https://github.com/TASEmulators/BizHawk/blob/e4a0490cf06eda7a2596d32c230f07633d4ae399/src/BizHawk.Client.EmuHawk/tools/HexEditor/HexEditor.cs#L1510
And there was an existing `.First()` call:
https://github.com/TASEmulators/BizHawk/blob/e4a0490cf06eda7a2596d32c230f07633d4ae399/src/BizHawk.Client.EmuHawk/tools/Watch/WatchEditor.cs#L196
The implementation class was subclassing `List<>` so no change was needed there.

YoshiRulz:
> 1. I think it would be better to not modify cores' code. It should be possible to conditionally add the registers domain outside of the core, depending on if it provides the required service.

In `BasicServiceProvider` maybe? It can't be something the Hex Editor does internally if we want it to be usable from the APIs.

> 3. I also see `unchecked` in a lot of places. Does this provide any benefit? Since unchecked is default, this seems like unneeded fluff.

[Enabling `$(CheckForOverflowUnderflow)`](https://gitlab.com/YoshiRulz/yoshis-hawk-thoughts/-/issues/37) is one of my long-term cleanup goals.
Some casts in `unchecked` are equivalent to the same cast without it, not sure which, but if someone cared enough to make a list we could flag those with an Analyzer. I just have muscle memory to write `unchecked((uint) n)` instead of something clearer like `Unsafe.As<int, uint>(ref n)`.

src/BizHawk.Emulation.Common/Base Implementations/RegistersMemoryDomain.cs
SuuperW: Ah. Well. The linked issue sounds like they're talking about non-CPU registers (VDP registers), although I'm not familiar with what that actually is. Perhaps the method name is wrong; we don't have a separate method for getting non-CPU registers. Or the Cpu in the method name only refers to the flags, Idk.

SuuperW:
> [Enabling `$(CheckForOverflowUnderflow)`](https://gitlab.com/YoshiRulz/yoshis-hawk-thoughts/-/issues/37) is one of my long-term cleanup goals.

Okay, but why? Why would you want to change the language's behavior?
Surely it would be better to instead use `checked` where casts should be checked.

YoshiRulz:
Because *most* operations should be `checked`. Developers who write, for example, a simple addition don't (and shouldn't have to) think over the minutiae of how machines approximate the infinite space of addition over the integers. Only in edge cases does it cause problems, and I'd like the computer to raise those problems to the dev explicitly. It's just like being warned of a possible `NullReferenceException` or of a naming collision.

Issue #4469, opened 2025-09-25
[🐧] I can't see anything! (Regression)


cD1rtX3:
### Summary
Since 2.11, all normal text in dialogues is unreadable.

### Repro
1. Open a dialogue.

### Host env.
- BizHawk 2.11; Ubuntu 24.04.2 LTS; Intel/Intel

### Screenshots:
<img width="906" height="531" alt="Image" src="https://github.com/user-attachments/assets/74c97f0d-b476-42ac-b7dd-a17d884c69e3" />

cD1rtX3:
Dialogue in 2.10:  
<img width="876" height="538" alt="Image" src="https://github.com/user-attachments/assets/4d4d89cc-48d2-4029-884e-e185c144f845" />

YoshiRulz:
Can you bisect? Even if I had an idea of what commit could have done this, I can't repro; theming is broken for me in both releases (and both are working on my other machine).

YoshiRulz:
From linked thread, if your issue is the same then https://github.com/TASEmulators/BizHawk/commit/a64d7627ce9348e5dcade6941787d59ebc286cae may be the cause?
@Morilli How can we conditionally use the old scaling mode for Linux only?

Morilli:
*sigh*... previously for windows, `AutoScaleFactorX` was 1 and `AutoScaleFactorY` was 1. After that commit, `AutoScaleFactorX` is 1 and `AutoScaleFactorY` is 1. Naively I didn't expect that to differ for linux. Perhaps using `AutoScaleMode.Font` is actually the correct thing to do if we want to ensure things are sized properly for contained text.

cD1rtX3:
My internet is not good enough to download and test that PR 😔

vadosnaprimer:
It's just 80MBs, same as release...
https://github.com/TASEmulators/BizHawk/actions/runs/18020954575/artifacts/4110492392

YoshiRulz:
Future humans: Please try launching as `GTK_DATA_PREFIX='' ./EmuHawkMono.sh` and report whether it worked and what distro+DE you use.

---

Report from Discord of a user who switched to a dev build in order to fix the binds dialog. They found that, on this other dialog:
https://github.com/TASEmulators/BizHawk/blob/d2a604954be95d3a2006cae6d24ad0f5ceaaaf5c/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs#L713-L716
[...]

Issue #4456, opened 2025-09-21
[🐧] Hex editor memory highlighter shows incorrectly on Linux


NimbusFox:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
When taking a look at the memory using the Hex Editor it does not line the highlight up with where memory is and sometimes can appear in the wrong offset 

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Launch BizHawk and a game
2. Highlight a bit of memory
3. Scroll down a little bit

### Host env.
[//]: # "List the computers you've found the bug with. If there's a version that doesn't have the bug, please put that in too. Here are some examples:"
- BizHawk 2.10; Manjaro; AMD/AMD
- BizHawk 2.11; Manjaro; AMD/AMD

[//]: # "(screenshots, if applicable)"

Selected 0x1B8 says I'm looking at 0x158

<img width="669" height="634" alt="Image" src="https://github.com/user-attachments/assets/b65b02f7-bf47-45c1-b02a-415249bf1eef" />

Selected 0x64 says I'm looking between 0x44 and 0x54

<img width="674" height="630" alt="Image" src="https://github.com/user-attachments/assets/79d6782e-24bb-4a79-8a18-55f23425a03e" />

[//]: # "That's it! If you'd like to help more, you could try a dev build (see Testing in the readme) or an older release. Click submit now and you can edit it later."

This effects all cores as far as I can tell. As I have seen this when using mGBA, GameBatte and MelonDS as well

YoshiRulz:
I think this has been reported before, but it must have been on Discord because I can only find #3925.
Anyway I still can't reproduce on NixOS (Plasma 6) nor Ubuntu (GNOME 42). It might be related to the GTK theme being picked up, though I have that on NixOS, and it might be related to system UI scale, though I have that on Ubuntu.

[...]

Issue #4246, opened 2025-02-27
QuickNES crashed when I tried to switch games (Bizhawk 2.10)


Grandmaster9918:
[//]: # "This description supports Markdown syntax. There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "These lines are comments, for letting you know what you should be writing. You can delete them or leave them in."
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
Loaded ROM: Akumajou Densetsu Japan
Loaded Core: QuickNES
Open Tools: PPU viewer.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Paused the game
2. Opened the PPU viewer
3. Tried to switch game
4. Emulator crashed
### Output
[//]: # "Paste the contents of the error dialog if there is one (try Ctrl+C, it usually works), or paste the output from the Lua Console, or delete this section."
```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at BizHawk.Client.EmuHawk.NesPPU.HandlePaletteViewMouseMove(Point e) in /_/src/BizHawk.Client.EmuHawk/tools/NES/NESPPU.cs:line 669
   at BizHawk.Client.EmuHawk.NesPPU.Generate(Boolean now) in /_/src/BizHawk.Client.EmuHawk/tools/NES/NESPPU.cs:line 279
   at BizHawk.Emulation.Cores.Consoles.Nintendo.QuickNES.QuickNES.FrameAdvance(IController controller, Boolean render, Boolean rendersound) in /_/src/BizHawk.Emulation.Cores/Consoles/Nintendo/QuickNES/QuickNES.cs:line 352
   at BizHawk.Client.EmuHawk.MainForm.StepRunLoop_Core(Boolean force) in /_/src/BizHawk.Client.EmuHawk/MainForm.cs:line 3063
   at BizHawk.Client.EmuHawk.MainForm.ProgramRunLoop() in /_/src/BizHawk.Client.EmuHawk/MainForm.cs:line 898
[...]

PR #4472, opened 2025-09-25
fix text not appearing on Linux

rjgabel:

[dev build for branch | USERNAME:BRANCHNAME](https://nightly.link/USERNAME/BizHawk/workflows/ci/BRANCHNAME?preview)

Fix description text not appearing in the Controller/Hotkey configuration menu on some Linux systems. https://github.com/TASEmulators/BizHawk/issues/1430#issuecomment-3326319711

Check if completed:
- [X] I, the commit author, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2024-06-22) and am compliant

### DIFF
--- a/src/BizHawk.Client.EmuHawk/UIHelper.cs
+++ b/src/BizHawk.Client.EmuHawk/UIHelper.cs
@@ -1,7 +1,6 @@
 using System.Drawing;
 using System.Windows.Forms;
 using BizHawk.Common.NumberExtensions;
-using static BizHawk.Client.Common.DisplayManagerBase;
 
 namespace BizHawk.Client.EmuHawk
 {
@@ -9,9 +8,11 @@ namespace BizHawk.Client.EmuHawk
 	{
 		private static SizeF AutoScaleCurrentSize { get; } = GetCurrentAutoScaleSize();
 
-		public static float AutoScaleFactorX { get; } = AutoScaleCurrentSize.Width / DEFAULT_DPI;
+        public static SizeF AutoScaleBaseSize { get; } = new SizeF(6F, 13F);
 
-		public static float AutoScaleFactorY { get; } = AutoScaleCurrentSize.Height / DEFAULT_DPI;
+		public static float AutoScaleFactorX { get; } = AutoScaleCurrentSize.Width / AutoScaleBaseSize.Width;
+
+		public static float AutoScaleFactorY { get; } = AutoScaleCurrentSize.Height / AutoScaleBaseSize.Height;
 
 		public static SizeF AutoScaleFactor { get; } = new SizeF(AutoScaleFactorX, AutoScaleFactorY);
 
@@ -37,7 +38,7 @@ namespace BizHawk.Client.EmuHawk
 
 		private static SizeF GetCurrentAutoScaleSize()
 		{
-			using var form = new Form { AutoScaleMode = AutoScaleMode.Dpi };
+			using var form = new Form { AutoScaleMode = AutoScaleMode.Font };
 			return form.CurrentAutoScaleDimensions;
 		}
 


Morilli:
Yeah I don't really want to straight up revert this change because it will regress #4246, but we probably do need to change the autoscale mode in these functions back to `Font`. 

The reason I wanted it to use `Dpi` is because font scaling is not consistent and usually just depends on the dpi setting again, but it makes sense that things that contain text need to be sized so that the text fits, even if that changes the layout.

YoshiRulz:
CI failure is about whitespace, but as written `AutoScaleBaseSize` is only used in those 2 other props and can be inlined.

rjgabel:
> CI failure is about whitespace, but as written `AutoScaleBaseSize` is only used in those 2 other props and can be inlined.

AutoScaleBaseSize is inline and whitespace should be fixed now

Morilli:
Here's what I would do: https://github.com/Morilli/BizHawk/commit/0726c2d379118302cf498b47307b5d67d2644837

edit: auto-link assumes this repo, fixed --yoshi

rjgabel:
> Here's what I would do: [0726c2d](https://github.com/TASEmulators/BizHawk/commit/0726c2d379118302cf498b47307b5d67d2644837)

Having trouble testing/building this commit, it says that it's not part of any brach?

YoshiRulz:
https://github.com/Morilli/BizHawk/actions/runs/18021719539

rjgabel:
> https://github.com/Morilli/BizHawk/actions/runs/18021719539

Seems like this build fixes the issue.

YoshiRulz:
I've tested both that build and this PR, and saw no difference on Ubuntu, while both are still broken on NixOS&dagger;.
So @Morilli it's your call on which approach to take.

&dagger; So no difference there either.
I did have an idea: You said you use Mint 22? Which would default to an X11 session. My machines run GNOME Wayland and Plasma Wayland respectively, and I suspect the Bazzite user who CNR would also be on Wayland. But then in #4469 OP said she uses Ubuntu 24.04, which I thought would default to Wayland?

Morilli:
> while both are still broken on NixOS

So you can't see any labels currently on either 2.10, 2.11 or this PR? If we're gonna fix those labels might as well do it properly so it works anywhere. Can you test if changing this to something like `Size = new(iwOffsetX - UIHelper.ScaleX(2), UIHelper.ScaleY(25))` fixes it?
https://github.com/TASEmulators/BizHawk/blob/4a34d6a2abf2d221cd92c27bb96f3969ffb77331/src/BizHawk.Client.EmuHawk/config/HotkeyConfig.cs#L110

YoshiRulz:
Not exactly, the keybinds dialogs are fine:
Screenshot_20250926_233928

This is the theming issue I have:
Screenshot_20250926_233920
And as you can see in the first screenshot, opening a modal dialog makes the text in other windows visible... for some reason. (RAM Watch is fine, forgot to actually add a watch.)

rjgabel:
> † So no difference there either. I did have an idea: You said you use Mint 22? Which would default to an X11 session. My machines run GNOME Wayland and Plasma Wayland respectively, and I suspect the Bazzite user who CNR would also be on Wayland. But then in #4469 OP said she uses Ubuntu 24.04, which I thought would default to Wayland?

That's correct, I am using Mint 22. I don't think there's a way for me to switch to Wayland on Mint (at least from the 60 seconds of searching I've done, I'm new to Linux), but if there is I'd be happy to test on Wayland.

> Not exactly, the keybinds dialogs are fine

Does this issue happen on 2.11, or is it a regression caused by the fixes? Interestingly, after doing some testing the fix I made seems to actually fix the Memory Editor window as well on my machine (it's broken on 2.10 and 2.11).

Let me know if you want me to close this PR in favor of Morilli's, or integrate their changes into this PR.

YoshiRulz:
That's 2.11 via Nix. 2.10 is the same.

Someone else reported a visual bug in the Hex Editor: #4456 (fixed in dev builds).

rjgabel:
> That's 2.11 via Nix. 2.10 is the same.
> 
> Someone else reported a visual bug in the Hex Editor: #4456 (fixed in dev builds).

Okay, I think the Nix issue might be different then, I was just worried my changes somehow broke Nix. As for the hex editor, I was comparing my PR (based on master) to 2.11, which is why I thought my changes fixed it for some reason.

Issue #4058, opened 2024-09-28
TAStudio Recording mode gets extremely slow for longer movies


RetroEdit:
To be more precise, the later in the movie you are, the slower recording mode is. So if you seek to an early frame in a long TAStudio movie and turn on recording mode, recording will often be fast.

It does also seem to be affected by how long the current movie has been open in TAStudio, but later frames are slower regardless.

This doesn't affect non-recording mode playback in TAStudio or read+write mode in traditional re-recording. It applies whether or not display is enabled.

I observed with both Gambatte and mGBA, and assume it affects all other cores.

### Repro
I find it pretty easy to just take a longer movie and observe how slow recording mode is later in the movie.

I went a bit overkill and created a script to measure it more rigorously under different conditions: [measure_playback_speed.zip](https://github.com/user-attachments/files/17171423/measure_playback_speed.zip)

And from that data, I created this graph:

(for the below test, display was "absolute zero", clock unthrottled, "follow cursor" off, TAStudio autosave interval set to 0, and TAStudio minimized.)

![2 10-rc1_GB_speeds](https://github.com/user-attachments/assets/ce480caf-0e16-4a72-9931-e7b0cf93b3b8)
(edit: An alternate view of the same data reveals that the curve for recording mode is hyperbolic—that is, frame times increase linearly.)

### Host env.
- Does *not* affect BizHawk 2.3.1
- (Affects BizHawk 2.3.2; Win11, BizHawk 2.3.3, BizHawk 2.4.2; Win11)
- BizHawk 2.9.1; Win11
- BizHawk 2.10-rc1; Win11
- BizHawk 2.10 dev build at 15ff0ec; Win11
  - (Latest dev build at time of writing)

Morilli:
Can you clarify what state the tasmovie was in when you did these tests? Mainly I'm interesting in the length of the tasmovie and whether or not states were recorded during these runs.

If you have a tasproj with 300k frames pre-recorded and are playing through them I'd expect a different result compared to recording 300k frames from start. I don't see how either of those would produce an ever-increasing frametime based on movie length, but I would still like to know for reproducing purposes.

Morilli:
Okay I can easily reproduce this and did some basic profiling:
Playback mode:
![image](https://github.com/user-attachments/assets/c82ce32a-cb06-4977-b7f6-597b25513b41)

Recording mode:
[...]

Issue #2641, opened 2021-02-24
Recording mode keeps getting toggled off


lunjesse:
[//]: # "This issue description supports Markdown syntax (this is what comments look like). There's a cheatsheet here: https://guides.github.com/features/mastering-markdown/"
[//]: # "You can leave these comments here or delete them. Also, please remember to search for similar issues before writing anything, including in closed issues!"
[//]: # "One more thing: if you're on Linux, please comment on #1430 instead of opening an issue so we don't annoy the other devs."

### Summary
[//]: # "Briefly describe what's broken. Include relevant details: loaded core, loaded rom's hash, open tools, running scripts... You can embed a screenshot if it's easier to show the bug, but if you need more than one please put them at the end."
Clicking on a previous frame in the piano roll can turn off recording mode. Given that's reloading a state, that makes it hard to use recording mode unless I never reload a state.

### Repro
[//]: # "If you can't figure out the list of steps, delete this section and put 'heisenbug' in the summary somewhere. If a Lua script can cause the bug, you can embed that instead (as simple as possible please)."
1. Open up a ROM (I used GBA)
2. Open TAStudios
3. Toggle "Recording Mode"
4. Let the piano roll run a bit
5. Click an earlier part of the piano roll
There's now a chance the Recording Mode ends. Can there please be a way to prevent this?

### Host env.
[//]: # "List the computers you've found the bug with. If there's a version that doesn't have the bug, please put that in too. Here are some examples:"
- BizHawk 2.6.1; Win10 Home; Intel

[//]: # "(screenshots, if applicable)"

[//]: # "That's it! If you'd like to help more, you could try a dev build (see Testing in the readme) or an older release. Click submit now and you can edit it later."

CasualPokePlayer:
I've seen this before, although it was limited to Recording Mode being turned off while TAStudio was seeking to the frame I selected, and Recording Mode was turned back on when it finished seeking. I think this is intentional?

YoshiRulz:
May be related to #2619?

lunjesse:
> I've seen this before, although it was limited to Recording Mode being turned off while TAStudio was seeking to the frame I selected, and Recording Mode was automatically turned back on when it finished seeking. I think this is intentional?

If it's true, is there a way to toggle this behavior off? I want to explore a crash prone area in game while visualizing the input so far and recording mode is the only way so far.

lunjesse:
> May be related to #2619?

[...]

Issue #2143, opened 2020-06-16
In TAStudio recording mode, input can't be easily undone


RetroEdit:
When entering recording mode in TAStudio, the previous input will be overwritten. This is sometimes accidental, and it would be desirable to be able to undo the current session of recorded input, rather than undoing each input one-by-one.

Since I've already looked at the undo handling code, I might have a look at this; I already plan to do some more undo/redo revamping, since undo/redo insert inefficiency (#1988) wasn't the only slow thing.

vadosnaprimer:
FCEUX handles this by combining consecutively recorded frames into a single Undo step.

RetroEdit:
Yeah, that's exactly what I was thinking should happen. Architecturally speaking, it might require some modifications to BizHawk's undo/redo API, since I'm not sure it can handle actions occurring for multiple consecutive frames (doing stuff like deleting multiple frames, etc. is different because it can be handled as a batch action).

SuuperW:
Fixed with #4382

PR #4475, opened 2025-09-28
Do not disable recording mode when loading a branch.

SuuperW:
Multiple people (at least 2) in Discord have brought up a recent change, thinking it may be a bug: Loading a branch disables recording mode, but did not in 2.10. This indicates that users find the change surprising and prefer the old behavior. This PR will restore that old behavior.

The change was originally done as part of making TAStudio more consistent. Editing inputs in a way that is not recording a frame should disable recording mode, but it used to be that not all edits would do this. Loading a branch is one way of changing inputs, and so it now (in master) disables recording mode.

Although disabling recording mode on branch load makes sense to me, I can see how it would be a nuisance to people who actually regularly use recording mode as part of TASing.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
@@ -757,6 +757,8 @@ namespace BizHawk.Client.EmuHawk
 		{
 			GreenzoneInvalidatedCallback?.Invoke(frame); // lua callback
 
+			if (_suspendEditLogic) return false;
+
 			// Recording multiple frames, or auto-extending the movie, while unpaused should count as a single undo action.
 			if (CurrentTasMovie.LastEditWasRecording && !MainForm.EmulatorPaused)
 			{
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
@@ -1179,6 +1179,7 @@ namespace BizHawk.Client.EmuHawk
 			TasView.AllColumns.ColumnsChanged();
 		}
 
+		private bool _suspendEditLogic;
 		public void LoadBranch(TasBranch branch)
 		{
 			if (Settings.OldControlSchemeForBranches && !TasPlaybackBox.RecordingMode)
@@ -1187,7 +1188,9 @@ namespace BizHawk.Client.EmuHawk
 				return;
 			}
 
+			_suspendEditLogic = true;
 			CurrentTasMovie.LoadBranch(branch);
+			_suspendEditLogic = false;
 			LoadState(new(branch.Frame, new MemoryStream(branch.CoreData, false)));
 
 			CurrentTasMovie.TasStateManager.Capture(Emulator.Frame, Emulator.AsStatable());


vadosnaprimer:
Loading a branch in rec mode is the same as loading a savestate without tastudio and then continuing to record. Pretty standard workflow. Obviously loading a branch changes inputs, but it's not a part of manual editing.

YoshiRulz:
Does loading a branch always pause?

SuuperW:
> Does loading a branch always pause?

No.

YoshiRulz:
Maybe it should when recording mode is active.

vadosnaprimer:
Is that suggestion based on actual TASing workflow that someone uses all the time?

YoshiRulz:
This fallacy again? I'll bite: https://github.com/TASEmulators/BizHawk/issues/2144#issuecomment-645515184, and I'm sure I've also seen users confused by recording mode clearing inputs.

But to actually reason through my suggestion, it's weighing the "false positives"—pausing when the user intended to keep recording immediately after loading the branch, requiring the user unpause—against the "false negatives"—clearing inputs in the new branch, requiring the user pause, undo (see #2143), possibly disable recording mode, and unpause. We may not know which is more common, but the former is obviously easier for the user to remedy.

Another possibility would be to disable *recording mode* if loadstating while unpaused, but re-enabling that is no easier than unpausing, and based on #2641 would probably be the more annoying requirement. If the user was using recording mode on the old branch then they're probably aware that it's enabled, and they probably intend to continue recording once they've reoriented themself in the new branch.

vadosnaprimer:
You're making it sound like asking for a proper explanation and practical reasoning is something that should be avoided. It's not fallacy. It's how this entire tool was initially built in fceux. Workflow is everything.

SuuperW:
> But to actually reason through my suggestion, it's weighing the "false positives"—pausing when the user intended to keep recording immediately after loading the branch, requiring the user unpause—against the "false negatives"—clearing inputs in the new branch, requiring the user pause, undo (see #2143), possibly disable recording mode, and unpause. We may not know which is more common, but the former is obviously easier for the user to remedy.

The later is not any harder to do than the former. The issue you linked is irrelevant since the bug has been fixed (oops, that should be closed now) and also because undo is simply not required: the user can just load the branch again. I do not think keeping it on will be a significant annoyance even if one wants it off.

Unless people have complained about recording mode being kept on when loading branches in the past, or someone has a real workflow where keeping it on is a significant hindrance, I think we should keep it on. Because the recent complaints in Discord show there are real workflows where users want it kept on.

Edit: and regarding pausing on branch load: please no. I would be annoyed by that. It's not uncommon for me to load branch A while unpaused, watch a few seconds, then load branch B, watch a few seconds, and maybe go back and forth a couple times to visually compare two strategies.

YoshiRulz:
>> [...] it's weighing the "false positives" [...] against the "false negatives" [...] We may not know which is more common, but the former is obviously easier for the user to remedy.
>
> The later is not any harder to do than the former. [...] undo is simply not required: the user can just load the branch again. [...] Because the recent complaints in Discord show there are real workflows where users want it kept on.

That's the conclusion I came to as well. I hadn't thought of loading the branch again, but unpausing is still less effort IMO.

> Edit: and regarding pausing on branch load: please no. I would be annoyed by that. It's not uncommon for me to load branch A while unpaused, watch a few seconds, then load branch B, watch a few seconds, and maybe go back and forth a couple times to visually compare two strategies.

While recording though?

YoshiRulz:
> You're making it sound like asking for a proper explanation and practical reasoning is something that should be avoided. It's not fallacy.

Asking for more reasoning: good, expected
Dismissing my suggestions because I'm not a TASer: fallacious<sup>[1](https://en.wikipedia.org/wiki/Courtier's_reply),[2](https://en.wikipedia.org/wiki/Appeal_to_accomplishment)</sup> (and now I am guilty of the fallacy of linking to Wikipedia articles about fallacies)

vadosnaprimer:
> Dismissing

Where?

RetroEdit:
> Maybe it should [pause] when recording mode is active.

Automatic pausing upon branch load is more likely to be perceived by end-users as an unexplained inconsistency rather than a protective measure.

> This PR will restore that old behavior.

I'm of two minds. As someone who uses recording mode rarely and with great caution, I'd somewhat inclined to suggest this PR's changes should be optional, with a TAStudio config option `Disable recording mode on branch load`.

On the other hand, ideally TAStudio should mostly be an enhancement to traditional movie rerecording. Under that view, loading states traditionally preserves recording mode, so loading branches definitely should.

And with the new undo/redo changes, unfixable recording mode blunders should be less common. (As a side note, it's hard to properly test this branch while this new regression remains: #4477 )

SuuperW:
> On the other hand, ideally TAStudio should mostly be an enhancement to traditional movie rerecording.

I see TAStudio as a complete replacement for traditional movie recording.

RetroEdit:
Sort of off-topic, but I don't think TAStudio can or should *replace* traditional rerecording. TAStudio cannot currently fulfill the role of being lightweight, reliable, and consistent with traditional rerecording workflows that many existing users have been using for years. Major bugs like #4058 went unnoticed for years, and I still run into TAStudio-specific bugs often and I imagine there are many undiscovered edge cases.

(Getting even more off-topic, I've personally thought for years it'd be better to just rewrite TAStudio from scratch with robust testcases and special attention to workflow efficiency. But perhaps it's too ambitious for me as one person, and I haven't written much actual code -- I have been slowly building up a library of personal notes about implementation details, abstract testcases, and backwards compatibility notes though.)

vadosnaprimer:
> > On the other hand, ideally TAStudio should mostly be an enhancement to traditional movie rerecording.
> 
> I see TAStudio as a complete replacement for traditional movie recording.

It's both an upgrade and a replacement, depending on one's preferences. It being an upgrate helps people get used to it and potentially migrate over. But if not, it should at least not be annoying as an enhancement.

SuuperW:
> undo is simply not required: the user can just load the branch again

And with 1c0c915, it won't even mess up any undo history.

> it should at least not be annoying as an enhancement.

If this is to be the case, I think it's pretty clear that loading a branch should not turn off recording mode.

PR #4480, opened 2025-10-02
Improved menu keyboard navigation

SuuperW:
Several bug fixes.

New behavior existing in `master` and this PR but not in 2.11, arguably a bug:
- Pressing Alt+F does not work when F is bound to a controller input. You can at least in this PR use Alt, then F.

Bugs that exist in 2.11 and `master`, fixed in this PR:
- Pressing space while the tool strip menu is focused does not work. (Main form only)
- Alt does not properly focus the menu. (Main form only in 2.11, main form + TAStudio in `master`)
- Pressing a hotkey while the main menu was open activated the hotkey. (Main form only) (the system menu [alt+space] still has this issue)

Regression from 2.11 in `master`, fixed in this PR:
- Pressing Alt combinations did not work in input-blocking tools.

Bug existing in all three:
- Alt+Space -> Escape leaves the program is suspended as if menu selected, but pressing a menu key does not open a menu and instead resumes the main program loop. (All forms)
- Pressing context menu key in TAStudio does not work.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/inputAdapters/InputManager.cs
+++ b/src/BizHawk.Client.Common/inputAdapters/InputManager.cs
@@ -149,10 +149,11 @@ namespace BizHawk.Client.Common
 		/// <summary>
 		/// Processes queued inputs and triggers input evets (i.e. hotkeys), but does not update output controllers.<br/>
 		/// </summary>
-		/// <param name="processUnboundInput">Events that did not do anything are forwarded out here.
+		/// <param name="processSpecialInput">All input events are forwarded out here.
 		/// This allows things like Windows' standard alt hotkeys (for menu items) to be handled by the
-		/// caller if the input didn't alrady do something else.</param>
-		public void ProcessInput(IPhysicalInputSource source, Func<string, bool> processHotkey, Config config, Action<InputEvent> processUnboundInput)
+		/// caller if the input didn't alrady do something else.
+		/// <br/>The second parameter is true if the input already did something (hotkey or controller input).</param>
+		public void ProcessInput(IPhysicalInputSource source, Func<string, bool> processHotkey, Config config, Action<InputEvent, bool> processSpecialInput)
 		{
 			// loop through all available events
 			InputEvent ie;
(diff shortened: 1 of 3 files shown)


vadosnaprimer:
I don't understand what these mean:
- Pressing space while the tool strip menu is focused does not work. (Main form only)
- Pressing Alt combinations did not work in input-blocking tools.

Also did you mean 2.11 instead of 2.10?

SuuperW:
> * Pressing space while the tool strip menu is focused does not work. (Main form only)

1) Click on the File menu.
2) Press Esc key
3) Observe that the menu strip is focused: arrow keys move the selection
4) Press space. This should open the menu that Alt+Space opens, but it doesn't.

> * Pressing Alt combinations did not work in input-blocking tools.

1) Open Lua Console
2) Press Alt+F. This should open the File menu, but it doesn't.

> Also did you mean 2.11 instead of 2.10?

Yes.

vadosnaprimer:
While comparing the first case to other apps I found that as you switch the menus using arrows, before the first menu and after the last one there's a hidden menu that represents that Alt+Space thing, and if you hit Enter while that hidden menu is selected, Alt+Space thing will show up. And yeah simply pressing Space while some other menu is selected, shows Alt+Space thing too.

That functionality available through Enter is still not present in this PR, because you can't select that hidden menu, tho I dunno if that matters.

~~> Press Alt+F. This should open the File menu, but it doesn't.~~
~~Only Alt+S fails to work because lua console has 2 menus using the S key. File and Help menus open fine from Alt+F/Alt+H.~~
Nevermind you said regression from 2.11.

Morilli:
I don't like that this focuses the top menu now when I alt-tab, this was not the case in any previous version and a good thing imo. I also don't see this behavior in any other app.

SuuperW:
I missed that. Apparently it's not an issue if you release alt before releasing tab, which is my habit.
I will fix it. edit: https://github.com/TASEmulators/BizHawk/commit/4fc9e207892278b7713aae53f458f2d38243ab8c

PR #4485, opened 2025-10-04
Do not default bind two hotkeys to the same key combination.

SuuperW:
Title.

### DIFF
--- a/src/BizHawk.Client.Common/config/Binding.cs
+++ b/src/BizHawk.Client.Common/config/Binding.cs
@@ -1,4 +1,7 @@
 ﻿using System.Collections.Generic;
+#if DEBUG
+using System.Diagnostics;
+#endif
 using System.Linq;
 
 namespace BizHawk.Client.Common
@@ -132,7 +135,7 @@ namespace BizHawk.Client.Common
 			Bind("TAStudio", "Delete Branch", "Alt+Delete");
 			Bind("TAStudio", "Show Cursor");
 			Bind("TAStudio", "Toggle Follow Cursor", "Shift+F");
-			Bind("TAStudio", "Toggle Auto-Restore", "Shift+R");
+			Bind("TAStudio", "Toggle Auto-Restore", "Alt+R");
 			Bind("TAStudio", "Seek To Green Arrow", "R");
 			Bind("TAStudio", "Toggle Turbo Seek", "Shift+S");
 			Bind("TAStudio", "Undo", "Ctrl+Z"); // TODO: these are getting not unique enough
@@ -198,6 +201,15 @@ namespace BizHawk.Client.Common
 
 			AllHotkeys = dict;
 			Groupings = dict.Values.Select(static info => info.TabGroup).Distinct().ToList();
+
+#if DEBUG
+			var bindings = dict.Values
+				.Where(static info => !info.DisplayName.StartsWith("RA ") && !string.IsNullOrEmpty(info.DefaultBinding))
+				.Select(static info => info.DefaultBinding);
+#pragma warning disable MA0031 // Optimize Enumerable.Count() usage
+			Debug.Assert(bindings.Count() == bindings.Distinct().Count(), "Do not default bind multiple hotkeys to the same button combination.");
+#pragma warning restore MA0031 // Optimize Enumerable.Count() usage
+#endif
 		}
 
 		public static void ResolveWithDefaults(IDictionary<string, string> dict)


src/BizHawk.Client.Common/config/Binding.cs
```diff
@@ -198,6 +201,15 @@ void Bind(string tabGroup, string displayName, string defaultBinding = "", strin
 
 			AllHotkeys = dict;
 			Groupings = dict.Values.Select(static info => info.TabGroup).Distinct().ToList();
+
+#if DEBUG
+			var bindings = dict.Values
+				.Where(static info => !info.DisplayName.StartsWith("RA ") && !string.IsNullOrEmpty(info.DefaultBinding))
+				.Select(static info => info.DefaultBinding);
+#pragma warning disable MA0031 // Optimize Enumerable.Count() usage
+			Debug.Assert(bindings.Count() == bindings.Distinct().Count(), "Do not default bind multiple hotkeys to the same button combination.");
```
YoshiRulz: The actual problem here is that this is a double enumeration. Enumerate `.ToArray()` first.

YoshiRulz (approved):
Otherwise LGTM.

src/BizHawk.Client.Common/config/Binding.cs
SuuperW: What's the double enumeration? I enumerate `bindings` twice, but doing a `ToArray` won't change that.

YoshiRulz: [double/multiple enumeration](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/CA1851)

~~Looks like the built-in diagnostic for catching those wasn't actually enabled...~~ pushed https://github.com/TASEmulators/BizHawk/commit/997cc598abc22a80b3c0eace705f83c27993a952

vadosnaprimer: Any changes still required here?

YoshiRulz: There haven't been any new commits, so yes. I would have just fixed it but I can't push to this branch. https://github.com/TASEmulators/BizHawk/commit/983947c53d9fd806ad9d8e83392bc0b9f29ea422 <-- untested since exceptions in static ctors are silent under Mono

vadosnaprimer: @SuuperW

SuuperW: I don't think any change is required. The impact of unoptimized `Count` usage is minimal and only exists in debug mode.

If the reason for wanting a change is doing this meaningless optimization, sure, I'll force-push an update. But if the reason is avoiding the in-line warning suppression, using `ToArray` won't do that anyway. It still doesn't like `bindings.Length == bindings.Distinct().Count()` and the recommended fix is less readable (as is Yoshi's version). Readability and clear intent > tiny debug startup performance.

YoshiRulz: In general, a double enumeration is a correctness issue, not a perf issue.
If you want to put readability first, make an `IsDuplicateFree` extension.

SuuperW: In what way is it incorrect? The warning message itself indicates it's a optimization (aka performance) issue.
Either way, an extension method is not a bad idea.

YoshiRulz: MA0031 is suggesting `.CountIsExactly(n)` instead of `.Count() == n`, which would be faster in the worst case so that message is apt.

I have no idea why CA1851 is in the Performance category. The docs literally say "If the enumeration operation has side effects, multiple enumerations might result in bugs," which isn't a lot of detail but it at least hints at how you could build an example.
```csharp
static void Example(IEnumerable<int> l) {
	Console.WriteLine("running incorrect duplicate-free check...");
	var count = l.Count();
	Console.WriteLine($"contains {count} total");
	var uniqCount = l.Distinct().Count();
	Console.WriteLine($"contains {uniqCount} distinct ({(uniqCount == count ? "dupe-free" : "not dupe-free")})");
}
Random rng = new(1234);
Example(Enumerable.Range(1, 5).Select(_ => {
	var n = rng.Next(0, 10);
	Console.WriteLine($"\t{n}");
	return n;
}));
/*
running incorrect duplicate-free check...
	3
	8
	3
	9
	3
contains 5 total
	9
	8
	5
	6
	3
contains 5 distinct (dupe-free)
*/
```
(I suspect there's a second way involving shared instance state, rather than this kind of side-effect, but I couldn't figure that out quickly.)

SuuperW: I did not get that suggestion (of `CountIsExactly`). I get "Replace 'Count() == n' with 'Take(n + 1).Count() == n'" Looking at `CountIsExactly`, it's a BizHawk extension method that uses `Take`. Using it removes the warning.

I also did not get CA1851. And these enumerations do not have side effects, so I don't see why you would be getting that message.

Issue #2694, opened 2021-04-01
Log Window is not thread safe


dgp1130:
### Summary

Using `Console.WriteLine("...");` from a backend thread and viewing it in Log Window causes the thread to crash. For example:

### Repro

1. Make a new external tool with at least this implementation:

```cs
public sealed class MyTool : Form, IExternalToolForm {
    // ...
    public void Restart() {
        Task.Run(() => {
            Console.WriteLine("Hello");
            Console.WriteLine("World");
        });
    }
}
```

2. Open BizHawk and click `View > Open Log Window...`.
3. Click `Tools > MyTool`.
4. Observe log window reads simply: "Hello"

You can also put a breakpoint on the two `Console.WriteLine("...")` statements. The first will trigger, but the second will not. If you view the thread in the Visual Studio threads window, it is listed as `<not available>`.

If you do the same thing with the Log Window **closed**, both statements will execute just fine.

### Host env.

- BizHawk 2.5.2; Win10

### Additional Info

I believe the root cause here is that Log Window [does not use a thread safe implementation for `Console.SetOut()`](https://github.com/TASVideos/BizHawk/blob/2.5.2/src/BizHawk.Client.EmuHawk/LogWindow.cs#L49). According the C# docs, the [`Console` class is thread-safe](https://docs.microsoft.com/en-us/dotnet/api/system.console?redirectedfrom=MSDN&view=net-5.0#thread-safety), so I think it's reasonable for devs to expect this to work. Unfortunately I'm not well-versed enough in C# threads to know the best fix, but I think we just need to marshal the input buffer to the main thread.

Aside from just breaking the code, it also is difficult and unintuitive to debug. I was expecting an exception of some kind but just never got one. The thread silently died without explanation.

nattthebear:
I chewed someone out about this a while back, but there was no interest in fixing it.  The new console was so utterly worthless for other reasons that I certainly had no interest in fixing it.

When I'm working I usually build Bizhawk in debug mode, which has a different console that actually functions.  If I need to use release mode, I make a quick code mod to turn on the same console that debug mode uses.  Is that awful?  Of course.  Oh well.

vadosnaprimer:
[...]

PR #4490, opened 2025-10-05
Small fix for crashing log window when written from other thread

Skarsnik:
For some reason, changing this code to this makes it stop crashing when writing to the console from another thread. The value from `_lines.Count` was not the same in `virtualListView1.VirtualListSize`, maybe write to VirtuListSize are delayed?

Prevent crashs when writing in the console from multiple threads.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/LogWindow.cs
+++ b/src/BizHawk.Client.EmuHawk/LogWindow.cs
@@ -103,7 +103,7 @@ namespace BizHawk.Client.EmuHawk
 		private void doUpdateListSize()
 		{
 			virtualListView1.VirtualListSize = _lines.Count;
-			virtualListView1.EnsureVisible(_lines.Count - 1);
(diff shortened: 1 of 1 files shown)


YoshiRulz:
What pattern were you using that wasn't working with the log window? Something like `new Thread(() => Console.WriteLine(...)).Start();`? edit: Well CNR with that, I'll just take your word for it.

Skarsnik:
I have an external tool that does network stuff in its own thread. The main thread and the network thread both do some Console.Writeln(). 
I think my issue compared to #2694 is that I do a lot of multiline output (for example, in https://github.com/Skarsnik/Bizhawk-nwa-tool/blob/master/src/NWAServer.cs#L118 ) and if I remember correctly (sorry, I fixed this like over 6 months ago?) both `_lines.Count` and `virtualListView1.VirtualListSize` were not the same despiste the asignation.

vadosnaprimer:
I don't mind merging this without testing.

Issue #4479, opened 2025-10-02
Refactor RCheevosAchievementListForm and RCheevosLeaderboardListForm to not use a separate form for each respective item


CasualPokePlayer:
These forms currently are just wrappers against FlowLayoutPanels that take in an array of RCheevosAchievementForm's/RCheevosLeaderboardForm's. This is grossly inefficient and can crash due to window handle exhaustion, and as such they're currently disabled (see #4478). They need to be refactored to not taken in a list of forms which all each end up creating a window handle.

YoshiRulz:
I didn't even know you could put a `Form` in a controls collection. I guess it makes sense since it's a subclass of `Control`.

Is
```diff
-public partial class RCheevosAchievementForm : Form
+public partial class RCheevosAchievementForm : UserControl
```
sufficient?

CasualPokePlayer:
> Is
> 
> -public partial class RCheevosAchievementForm : Form
> +public partial class RCheevosAchievementForm : UserControl
> 
> sufficient?

This doesn't seem to be sufficient, this rather just makes it so window handle creation for each control is deferred until showing the list form (which proceeds to crash if there's too many controls).

CasualPokePlayer:
Note that it's not just the window handle of the RCheevosAchievementForm itself being a problem, but also all its components (Label/TextBox/CheckBox/PictureBox) which also each create a window handle, resulting in 14 window handles per achievement (1000 window handles would be fine technically, but not 14000).

YoshiRulz:
Is this limit per window or per process? Could you, say, `.Take(1000)` and be sure it would never hit the limit? That would be much better than disabling the feature outright.

CasualPokePlayer:
The limit is per process.

I wouldn't want to just arbitrary take some amount of achievements are there's handling to re-order achievements listed based on levels of importance (see OrderByKey).

PR #4508, opened 2025-10-12
Use a virtualized FlowLayoutPanel for the rcheevos achievments form

Morilli:
See #4479. Currently WIP and kinda garbage code, not yet sure how to structure this. But it should generally work.

todo:
- [x] handle form resizing
- [ ] make this implementation less manual and hacky

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/RetroAchievements/RCheevosLeaderboardListForm.cs
+++ b/src/BizHawk.Client.EmuHawk/RetroAchievements/RCheevosLeaderboardListForm.cs
@@ -27,7 +27,6 @@ namespace BizHawk.Client.EmuHawk
 			flowLayoutPanel1.BoundScrollBar = vScrollBar1;
 			if (OSTailoredCode.IsUnixHost)
 				vScrollBar1.Location += new Size(7, 0);
-
 		}
 
 		private void DisposeLboardForms()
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.EmuHawk/Extensions/ControlExtensions.cs
```diff
@@ -199,6 +201,16 @@ public static CheckState ToCheckState(this bool? tristate)
 				false => CheckState.Unchecked,
 				null => CheckState.Indeterminate,
 			};
+
+		public static void SuspendDrawing(this Control control)
+		{
+			WmImports.SendMessageW(control.Handle, WM_SETREDRAW, (IntPtr) 0, IntPtr.Zero);
```
CasualPokePlayer: Be mindful that this would not work on Linux (should probably just no-op this call if it's not Windows)

Morilli:
By the way this is about ready from my side. I'm still not too happy with the code structure but it works and fixes the issue, so 🤷.

Unless there's complaints I'd like to merge this soon so we have a working achievements and leaderboards form for the next release.

Edit: Seems to crash in my wsl ubuntu because `OnSizeChanged` is called during `InitializeComponent`... sigh

PR #4512, opened 2025-10-18
Fix video filename with ffmpeg writer

SuuperW:
Issue fixed:
1) Set window size to 2x.
2) Enable A/V capture OSD
3) Record video with ffmpeg writer, give the name "test.mp4"
4) Frame advance to actually capture a frame
Result: The video is written to "test_1.mp4". File "test.mp4" was created but has no video.

This PR fixes the issue by pulling out the logic for acquiring a `IVideoProvider` that was being used in `MainForm.AvFrameAdvance` to its own method, and using that method to obtain the initial video resolution.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -3210,14 +3210,9 @@ namespace BizHawk.Client.EmuHawk
 
 				aw = Config.VideoWriterAudioSyncEffective ? new VideoStretcher(aw) : new AudioStretcher(aw);
 				aw.SetMovieParameters(Emulator.VsyncNumerator(), Emulator.VsyncDenominator());
-				if (Config.AVWriterResizeWidth > 0 && Config.AVWriterResizeHeight > 0)
-				{
-					aw.SetVideoParameters(Config.AVWriterResizeWidth, Config.AVWriterResizeHeight);
-				}
-				else
-				{
-					aw.SetVideoParameters(_currentVideoProvider.BufferWidth, _currentVideoProvider.BufferHeight);
-				}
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3451,6 +3448,10 @@ private void AvFrameAdvance()
 					ShowMessageBox(owner: null, $"Video dumping died:\n\n{e}");
 					AbortAv();
 				}
+				finally
+				{
+					(output as IDisposable)?.Dispose();
```
Morilli: This will dispose the current core if `output = _currentVideoProvider` ran previously.

Morilli (changes requested):

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: Fixed.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3322,6 +3315,70 @@ private void RecordAvBase(string videoWriterName, string filename, bool unattend
 			RewireSound();
 		}
 
+		private (IVideoProvider, Action/*?*/ Dispose) GetCaptureProvider()
```
YoshiRulz:
```suggestion
		private (IVideoProvider Output, Action/*?*/ Dispose) GetCaptureProvider()
```
(And the other build error was from master, fixed now if you would rebase again.)

SuuperW: Done

PR #4516, opened 2025-10-21
Un-hardcode some hotkeys

RetroEdit:
Still a work-in-progress.

Check if completed:


SuuperW:
> per-hotkey overrides for hotkey overlap check

It would be nice actually to have some sort of general context check, where certain hotkeys are only active in a specific context.

And then TAStudio axis edit mode would be a context, that also needs to disable everything outside that context when active.

PR #4552, opened 2025-11-21
Controller backend improvements

### LINKED ISSUE #4550: New extra trailing `|` in input LogEntry for handheld consoles.
## Preface/TLDR
This issue itself isn't significant, but I think it shows some other issues that could be addressed.

But basically, it at least gives me peace of mind that this schema change is known.

## Details
I recently noticed for certain cores like mGBA and Gambatte, there's an extra trailing `|` in each input string. For example in Gambatte, before a blank entry was `|.........|`, and after, it was: `|.........||`

Probably only single-player handheld cores are affected; i.e., any core where this code is run:

https://github.com/TASEmulators/BizHawk/blob/4fc9e207892278b7713aae53f458f2d38243ab8c/src/BizHawk.Emulation.Common/Base%20Implementations/ControllerDefinition.cs#L162-L163

## Cause
I bisected to: a0800862b3678cbfeaad1290418a17c2b3e51fea (2.10 dev)

[`CreateLogEntry`](https://github.com/TASEmulators/BizHawk/blob/4fc9e207892278b7713aae53f458f2d38243ab8c/src/BizHawk.Client.Common/movie/bk2/Bk2LogEntryGenerator.cs#L42) is the proximate cause, because the second group in `source.Definition.ControlsOrdered` is empty. Specifically, prior to the above commit, `Bk2LogEntryGenerator` has an internal field `_controlsOrdered` that was initialized to `_source.Definition.ControlsOrdered.Where(static c => c.Count is not 0).ToList();` and reused; afterwards, the filter `c => c.Count is not 0` was no longer applied.

Arguably the root cause here is that [`GenOrderedControls`](https://github.com/TASEmulators/BizHawk/blob/4fc9e207892278b7713aae53f458f2d38243ab8c/src/BizHawk.Emulation.Common/Base%20Implementations/ControllerDefinition.cs#L117) inits with a list of length `PlayerCount + 1`.

## Resolution
A few things could be improved here:

- [ ] **Code**: `ControllerDefinition` could probably have an boolean (or maybe enum?) internal field that identifies consoles like GB/GBA/TI-83, etc.
  - `IsHandheld` maybe?
  - [x] It's also worth considering whether the LogEntry schema should be reverted/altered for these cases.
    - Note though the specific change in schema here isn't necessarily a huge issue; based on some recent TASVideos submissions I've seen, old movies still can be imported sensibly.
    - Answer (2026-04-07): [it should](https://github.com/TASEmulators/BizHawk/issues/4550#issuecomment-4201153345)
- [x] **Documentation**: the remark before `GenOrderedControls` mentions a "player 0" without further context. Above, [`ControlsOrdered`](https://github.com/TASEmulators/BizHawk/blob/4fc9e207892278b7713aae53f458f2d38243ab8c/src/BizHawk.Emulation.Common/Base%20Implementations/ControllerDefinition.cs#L23-L24) *does* mention the list has sublists of console buttons followed by player buttons. I'd like to see consistent verbiage around console buttons, and clear specification of how cases like GB are encoded.
- [ ] **Testing**: should catch schema changes like this. [`LogGeneratorTests`](https://github.com/TASEmulators/BizHawk/blob/master/src/BizHawk.Tests.Client.Common/Movie/LogGeneratorTests.cs) does test generically test logs, but there doesn't appear to be core-specific tests.
  - [ ] At the very least, there should be a consistency check for each core (with default sync settings) comparing the blank frame to a hardcoded value. This is not to say the schema should never change, but it should be intentional, to consider the effects on importing movies from prior versions of BizHawk.
  - [ ] Maybe somehow test whether trailing or extraneous `|` or lack thereof are loaded in a backwards compatible manner?
  - [ ] Maybe somehow test per-button whether the button mnemonic/ordering corresponds to what it was previously?

## See also
A semi-related commit and issue I found: b0913448966d65cc5c19a7b72168cea0d540fa5e (commit comment: "fixes a8368849a9bb03da0ed2183b4fb68caabb53af49 and 3384ce8629f1ef03b4b9296c311b33d0ea3d685c"), #4132

RetroEdit:
Fixes #4550

Check if completed:
   - Also, spot-tested on GB and NES

### DIFF
--- a/src/BizHawk.Tests.Client.Common/Movie/LogGeneratorTests.cs
+++ b/src/BizHawk.Tests.Client.Common/Movie/LogGeneratorTests.cs
@@ -89,22 +89,25 @@ namespace BizHawk.Tests.Client.Common.Movie
 		[TestMethod]
 		public void GenerateLogEntry_Bk2Controller()
 		{
-			var upController = new SimpleController(new ControllerDefinition("Dummy Gamepad") { BoolButtons = { "P1 Up", "P3 A" } }.MakeImmutable());
-			upController.Definition.BuildMnemonicsCache(VSystemID.Raw.NES);
+			var simpleController = new SimpleController(new ControllerDefinition("Dummy Gamepad") { BoolButtons = { "P1 Up", "P3 A" } }.MakeImmutable());
+			simpleController.Definition.BuildMnemonicsCache(VSystemID.Raw.NES);
 
-			var logEntry = Bk2LogEntryGenerator.GenerateLogEntry(upController);
-			Assert.AreEqual("||.||.|", logEntry);
+			var originalLogEntry = Bk2LogEntryGenerator.GenerateLogEntry(simpleController);
+			var originalLogKey = Bk2LogEntryGenerator.GenerateLogKey(simpleController.Definition);
 
-			var logKey = Bk2LogEntryGenerator.GenerateLogKey(upController.Definition);
-			Assert.AreEqual("##P1 Up|##P3 A|", logKey);
+			// just for safety, should be covered by the above tests already
+			Assert.AreEqual("||.||.|", originalLogEntry);
+			Assert.AreEqual("##P1 Up|##P3 A|", originalLogKey);
 
-			Bk2Controller bk2Controller = new Bk2Controller(upController.Definition, logKey);
+			// ensure a Bk2Controller constructed with ControllerDefinition and LogKey
+			// generates the exact same outputs as the original SimpleController
+			Bk2Controller bk2Controller = new Bk2Controller(simpleController.Definition, originalLogKey);
 
 			var newLogEntry = Bk2LogEntryGenerator.GenerateLogEntry(bk2Controller);
-			Assert.AreEqual(logEntry, newLogEntry);
+			Assert.AreEqual(originalLogEntry, newLogEntry);
 
 			var newLogKey = Bk2LogEntryGenerator.GenerateLogKey(bk2Controller.Definition);
-			Assert.AreEqual(logKey, newLogKey);
+			Assert.AreEqual(originalLogKey, newLogKey);
 		}
 #pragma warning restore BHI1600
 	}


src/BizHawk.Emulation.Common/Base Implementations/ControllerDefinition.cs
```diff
@@ -144,6 +155,11 @@ public static int PlayerNumber(string buttonName)
 
 		private static readonly Regex PlayerRegex = new Regex("^P(\\d+) ");
 
+		/// <summary>
+		/// Returns the number of players.
+		/// Currently only returns 0 for consoles where all control ports are empty,
+		/// so returns 1 for (non-linked) Game Boy and similar cases.
+		/// </summary>
 		public int PlayerCount
```
YoshiRulz: Where is this used? Would it not be better to change the semantics such that player 0 counts as a player (that is, `PlayerCount` is usually 1 more than the number of plugged-in gamepads)?

RetroEdit: I'll look and consider it. I'm not sure I entirely like the semantics of treating the console itself as a "player" though. At that point, I would also question why one couldn't just get the length of `ControlsOrdered` directly.

Morilli: Well `ControlsOrdered` relies on `PlayerCount` to know how many control groups there are. I do think documenting the special behavior is not useful when it should be changed instead.

src/BizHawk.Client.EmuHawk/tools/Macros/MovieZone.cs
```diff
@@ -241,7 +241,7 @@ public MovieZone(string fileName, IDialogController dialogController, IEmulator
 			{
 				if (!emuKeys.Contains(macro))
 				{
-					dialogController.ShowMessageBox($"The selected macro is not compatible with the current emulator core.\nMacro controller: {readText[1]}\nMacro player count: {readText[2]}", "Error");
+					dialogController.ShowMessageBox($"The selected macro is not compatible with the current emulator core.\nMacro controller: {readText[1]}\nMacro controller groups (\"players\" - 1) count: {readText[2]}", "Error");
```
SuuperW: Shouldn't this be `(\"players\" + 1)`? (plus)

RetroEdit: Yes, good catch!

SuuperW:
`Bk2Controller.Bk2ControllerDefinition.GenOrderedControls` is also incorrect. The base implementation remarks that other implementers should include empty lists for groups with no buttons. It seems like the only reason that empty lists should be included is that there's special logic expecting console buttons to be in the first group. (e.g. for NES, the power button is in group 0, player 1's buttons are in group 1) But this implementation removes empty lists.

Related, `Bk2LogEntryGenerator.GenerateLogKey` also removes empty lists.

Removing empty lists in these places probably solved some issues that were created by reporting `PlayerCount` of 1 for handhelds and then assuming there are `PlayerCount + 1` groups, which this PR fixes.

This removal of empty lists should only be a problem if there are any cores that have no console buttons and have multiple controllers. In that case, input display will show player 1's inputs last in the list of inputs instead of first. I do not know if there are any cores like this, but I think these implementations should be fixed regardless.

And it might be a good idea for this PR to change the public `NumControllerGroups` to return `ControlsOrdered.Count` instead of re-calculating it. It seems like these two values should always match. (The use of `NumControllerGroups` inside `GenOrderedControls` would need to be updated.)

RetroEdit:
One wonders if `NumControllerGroups` should even exist if it's just wrapping the public `ControlsOrdered.Count`...

While I agree `Bk2Controller` doesn't match the base implementation's specification in that comment, and I think you're probably right it would be okay and more correct to remove the empty entry filtering, I don't want to change more than necessary.

But if you feel strongly, feel free to add it to this PR (or make your own commit).

SuuperW:
I think it's OK to leave it, since I doubt it's causing any issues.

I pushed some updates for macros. Merge conflict here should just be because `MovieZone.cs` was renamed.

src/BizHawk.Emulation.Common/Base Implementations/ControllerDefinition.cs
```diff
@@ -144,24 +154,25 @@ public static int PlayerNumber(string buttonName)
...
-				{
-					return player;
-				}
+			get => ControlsOrdered.Count;
+		}
 
-				// Hack for things like gameboy/ti-83 as opposed to genesis with no controllers plugged in
-				return allNames.Exists(static b => b.StartsWithOrdinal("Up")) ? 1 : 0;
-			}
+		private int _calculateNumControllerGroups() {
```
Morilli: This cannot exist if other implementations of `GenOrderedControls` cannot use it. In general I don't think it's a good idea to allow infinite recursion when `NumControllerGroups` is called from within `GenOrderedControls` (as is the case in `BkmControllerDefinition.GenOrderedControls` currently). You're right that `NumControllerGroups` doesn't need to exist when it's just a thin wrapper over an otherwise public property, so why replace the previous public property that calculated the count with this?

If I'm seeing it correctly all places that need access to the player count could either use `ControlsOrdered` directly (like how LsmvImport is using it) or need the number because `ControlsOrdered` doesn't exist yet (GenOrderedControls).

I would prefer you revert this to a public function that gets player count / controller group count and possibly alter the usages that should just be using `ControlsOrdered` directly.

RetroEdit: I'll revert and fix that -- it feels convoluted even inside the base `ControllerDefinition` implementation, let alone in the recursion mishap over in `BkmControllerDefinition.GenOrderedControls` (whoops).

src/BizHawk.Emulation.Common/Base Implementations/ControllerDefinition.cs
```diff
@@ -153,14 +170,7 @@ public int PlayerCount
...
 					.Max();
-
-				if (player > 0)
-				{
-					return player;
-				}
-
-				// Hack for things like gameboy/ti-83 as opposed to genesis with no controllers plugged in
-				return allNames.Exists(static b => b.StartsWithOrdinal("Up")) ? 1 : 0;
+				return player + 1;
```
YoshiRulz, line 173: Is this the only behavioural change in this set?

RetroEdit: Yes, for removing the special-case; callers have been adjusted appropriately to account for the `+ 1`

src/BizHawk.Tests.Client.Common/Movie/LogGeneratorTests.cs
```diff
@@ -55,6 +55,57 @@ public void GenerateLogEntry_Floats()
...
+		{
+			var upControllerDefinition = new ControllerDefinition("Dummy Gamepad") { BoolButtons = { "P2 Up" } }.MakeImmutable();
+			upControllerDefinition.BuildMnemonicsCache(VSystemID.Raw.NES);
+
+			var logKey = Bk2LogEntryGenerator.GenerateLogKey(upControllerDefinition);
+			Assert.AreEqual("###P2 Up|", logKey);
+		}
+
+		[TestMethod]
+		public void GenerateLogEntry_Bk2Controller()
```
SuuperW, line 90: What is this one testing that the others don't? Also it seems to be testing multiple things, which is not ideal.

Morilli: This one specifically tests that a `Bk2Controller`  constructed from a `ControllerDefinition` and `LogKey` generates the exact same input log and logkey as the original controller does (basically testing the loading movie from file case).

And yes it's kind of also testing the same thing as the two tests before, but if anything in the initial log entry/logkey generation fails that should get caught individually by the other tests which makes it easier to identify the specific problem.

SuuperW: OK, I see it now. Comments would be helpful.
And are the first two asserts necessary? Seems like they should be removed (or given a comment / assertion message saying that this isn't what this test is really about, it's really about the last two).

The tests I've made recently have used arrange, act, assert comments. The "act" part is what is being tested, and having these simple comments can help clarify non-trivial (or highish line count?) tests like this.

Example for how I might reformat this one:
```cs
public void GenerateLogEntry_Bk2ControllerFromLogKey()
{
	// arrange
	var upController = new SimpleController(new ControllerDefinition("Dummy Gamepad") { BoolButtons = { "P1 Up", "P3 A" } }.MakeImmutable());
	upController.Definition.BuildMnemonicsCache(VSystemID.Raw.NES);

	var logEntry = Bk2LogEntryGenerator.GenerateLogEntry(upController); // should be "||.||.|"
	var logKey = Bk2LogEntryGenerator.GenerateLogKey(upController.Definition); // should be "##P1 Up|##P3 A|"

	// act
	Bk2Controller bk2Controller = new Bk2Controller(upController.Definition, logKey);

	// assert
	var newLogEntry = Bk2LogEntryGenerator.GenerateLogEntry(bk2Controller);
	Assert.AreEqual(logEntry, newLogEntry);

	var newLogKey = Bk2LogEntryGenerator.GenerateLogKey(bk2Controller.Definition);
	Assert.AreEqual(logKey, newLogKey);
}
```

Although I might hard-code the expected log key and entry, to get a minimal "arrange" section.

Morilli: I don't think such a structure is necessary here but I've added some additional comments to explain what exactly this is testing.

PR #4562, opened 2025-11-28
update melonDS to release 1.1

SuuperW:
Updates melonDS to release 1.1. Some info about changes in 1.1 can be found at https://melonds.kuribo64.net/comments.php?id=240.

The particular change I'm interested in right now is the graphics fixes to the OpenGL compute shader renderer. The current version of melonDS in BizHawk does not have a good option for high-resolution rendering as both OpenGL rendering modes contain significant graphical issues. With 1.1, the OpenGL compute looks correct.

The only non-trivial conflicts with the merge were in SPU.cpp and SPU.h. Since upstream now also uses blip buf our changes there shouldn't be necessary anymore. I made those files exactly match the upstream versions. Sound appears to work correctly and does not desync from audio in a 2 minute A/V recording.

The referenced commit for the melonDS submodule is at https://github.com/SuuperW/melonDS/tree/release1.1. I can make a PR for TASEmulator's melonDS fork if desired.


YoshiRulz:
> I can make a PR for TASEmulator's melonDS fork if desired.

Yes, the submodule will need to reference a commit that exists in
https://github.com/TASEmulators/BizHawk/blob/acfc1f878144ab0048e58f746c7e4a81e6c5adfd/.gitmodules#L20
otherwise it can't be cloned.

CasualPokePlayer:
There are more changes that are needed to properly merge 1.1, I'll do the merge myself later.

PR #4576, opened 2025-12-06
[WIP] `libarchive`-based dearchiving

YoshiRulz:
tests fail for ¯\\\_(ツ)\_/¯ reasons

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,6 +6,7 @@
 		<PackageVersion Include="Google.FlatBuffers" Version="23.5.26" /> <!-- last version with .NET Standard 2.0 support -->
 		<PackageVersion Include="ImGui.NET" Version="1.90.6.1" />
 		<PackageVersion Include="JunitXml.TestLogger" Version="3.1.12" />
+		<PackageVersion Include="LibArchive.Net" Version="0.2.0-alpha.0.82" />
 		<PackageVersion Include="Magick.NET-Q8-AnyCPU" Version="14.8.2" />
 		<PackageVersion Include="Menees.Analyzers" Version="3.2.2" />
 		<PackageVersion Include="Meziantou.Analyzer" Version="2.0.206" />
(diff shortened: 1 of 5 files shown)


Morilli:
Dare I ask what this is for?

YoshiRulz:
It [supports more formats](https://github.com/libarchive/libarchive/wiki/LibarchiveFormats) than SharpCompress.

SuuperW:
I assume the particular format you're interested in is zstd, and you want to simplify our zstd use. Including making it so that zstd compressed files in our archives won't need the .zst extension anymore and can be edited via text editor more easily.

YoshiRulz:
I don't even know what you're suggesting.
My goal here is to use the de facto standard library for working with archives, and in doing so add support for `.tar.gz`, `.tar.bz2`, `.tar.lz`, and sure why not `.tar.zst`.
https://github.com/TASEmulators/BizHawk/blob/db070b93bbe23a9ed0bb9c10f77958361863e9f7/src/BizHawk.Client.Common/SharpCompressDearchivalMethod.cs#L70-L80

SuuperW:
> My goal here is to use the de facto standard

Your goal is to "be like everybody else", then? Or is there any use case/feature that this enables, or issue that it fixes?

It doesn't look like you're actually using this new library anyway. I have no idea where or how you might intend to use it.

RetroEdit:
I would imagine opening games inside more types of archives would be the intended use-case. That *is* a useful feature.

YoshiRulz:
Currently the dearchival method is set globally on program startup, so it would be a 1-line change to use only libarchive:
https://github.com/TASEmulators/BizHawk/blob/db070b93bbe23a9ed0bb9c10f77958361863e9f7/src/BizHawk.Client.EmuHawk/Program.cs#L181

CasualPokePlayer:
> and in doing so add support for .tar.gz, .tar.bz2, .tar.lz, and sure why not .tar.zst

All of those are *already* [supported](https://github.com/mitchcapper/sharpcompress/blob/master/FORMATS.md) by SharpCompress. At most they are just not supported within BizHawk's usage of SharpCompress.

PR #4581, opened 2025-12-10
Improve handling of Lua script callbacks

### LINKED ISSUE #4442: Lua scripts' callbacks aren't always cleared
> &lt;CasualPokePlayer> in particular, callbacks which are set in a callback
https://github.com/TASEmulators/BizHawk/blob/ed497a76f6fc3e30c4078dc89c85e3ee64a7a03c/src/BizHawk.Client.Common/lua/NamedLuaFunction.cs#L42-L47

```lua
--TODO write repro script
```

see also #3036

SuuperW:
Lua scripts that show as not running should not do anything. This means any events it registers should be removed.

Changes:
1. Lua scripts that register one or more callbacks (other than onexit) will show as still running once the script completes. (So if there's no "frame loop", but a `onframeend` callback, it will show as running.) Removing the callback(s) will stop the script.
2. Lua scripts that create a `LuaWinform` (via `forms.newform`) will show as still running once the script completes. Closing the form(s) will stop the script.
3. When a script is stopped, all registered callbacks for that script will be removed and all its forms will be closed.
4. As a consequence of the above, an existing Lua script will no longer work if it (a) has a frame loop, an exit event, plus one or more events and (b) relies on exiting the frame loop to stop the script + trigger the exit event. (It's really easy to update the script though, assuming its exit event properly cleans things up: manually call the exit event handler after the frame loop.)
5. Fix various bugs that caused Lua execution to run outside of its sandbox and/or without setting the `CurrentFile`. This fixes #4442.
6. Make Lua exception handling more consistent.
7. Make TAStudio's Lua events work like the other Lua events. For events that return a value (the various OnQuery ones), TAStudio will call the events in order until one returns something other than `nil`.
8. Much refactoring to support the above changes.

There are tests for most of the changed behavior. A significant amount of refactoring was specifically to support testing without depending on BizHawk.Client.EmuHawk, which creates problems of its own.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
@@ -524,14 +524,10 @@ namespace BizHawk.Client.EmuHawk
 				OutputBox.ScrollToCaret();
 			});
 			if (!Settings.DuplicateToStdout) return;
-#if true
-			Console.Write("[Lua script] {0}", message);
-#else //TODO keep track of origin script and include that as well
 			Console.Write(
-				senderFilenameNoExt is null ? "[Lua REPL] {1}" : "[Lua script: {0}] {1}",
-				senderFilenameNoExt,
+				LuaImp.CurrentFile == _nonFile ? "[Lua REPL] {1}" : "[Lua script: {0}] {1}",
+				Path.GetFileNameWithoutExtension(LuaImp.CurrentFile.Path),
 				message);
-#endif
 			if (!message.EndsWith('\n')) Console.WriteLine('\\'); // indicate continued lines; still won't make sense w.r.t. backslashes as an escape character, but it's good enough for humans
 		}
 


src/BizHawk.Client.Common/DependencyInjection/DependencyProvider.cs
```diff
@@ -0,0 +1,25 @@
+﻿#nullable enable
+
+using System.Collections.Generic;
+
+namespace BizHawk.Client.Common
+{
+	public class DependencyProvider : IDependencyProvider
```
YoshiRulz: A third bespoke DI system is the last thing we need. Why is it necessary, what led you to add it?

src/BizHawk.Tests.Client.Common/Display/FakeDisplayManager.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Drawing;
+using System.Drawing.Text;
+
+using BizHawk.Bizware.Graphics;
+using BizHawk.Client.Common;
+
+namespace BizHawk.Tests.Client.Common.Display
+{
+	internal class FakeDisplayManager : IDisplayManager
```
YoshiRulz: Are you not able to use `SimpleGDIPDisplayManager`? I moved it out of the testroms project so it could be re-used.

src/BizHawk.Tests.Client.Common/BizHawk.Tests.Client.Common.csproj
```diff
@@ -6,6 +6,8 @@
 	<ItemGroup>
 		<ProjectReference Include="$(ProjectDir)../BizHawk.Client.Common/BizHawk.Client.Common.csproj" />
 		<ProjectReference Include="$(ProjectDir)../BizHawk.Tests.Common/BizHawk.Tests.Common.csproj" />
+		<PackageReference Include="System.Drawing.Common" />
 		<EmbeddedResource Include="data/**/*" />
+		<InternalsVisibleTo Include="BizHawk.Tests.Client.EmuHawk" />
```
YoshiRulz: Make the type(s) you need `public` instead.
There's no reason to manage the public API surface of a unit test project.

src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
```diff
@@ -0,0 +1,69 @@
...
+		public void SaveState(string path, string userFriendlyStateName, bool fromLua = false, bool suppressOSD = false) => throw new NotImplementedException();
+		public void SeekFrameAdvance() => throw new NotImplementedException();
+		public void StepRunLoop_Throttle() => throw new NotImplementedException();
+		public void StopMovie(bool saveChanges = true) => throw new NotImplementedException();
+		public void TakeScreenshot() => throw new NotImplementedException();
+		public void TakeScreenshot(string path) => throw new NotImplementedException();
+		public void TakeScreenshotToClipboard() => throw new NotImplementedException();
+		public void TogglePause() => throw new NotImplementedException();
+		public void ToggleSound() => throw new NotImplementedException();
+		public void UnpauseEmulator() => throw new NotImplementedException();
```
YoshiRulz, line 67: Everything throws? What is this for then?

src/BizHawk.Tests.Client.EmuHawk/BizHawk.Tests.Client.EmuHawk.csproj
```diff
@@ -0,0 +1,30 @@
...
+		<LangVersion>latest</LangVersion>
+		<Nullable>enable</Nullable>
+		<UseWindowsForms>true</UseWindowsForms>
+	</PropertyGroup>
+	<Import Project="../MainSlnTest.props" />
+	<ItemGroup>
+		<!-- Note that the Properties field here is ignored by Visual Studio. Run tests with `dotnet run` or:
+		Add <TESTING>true</TESTING> at line 4 of MainSlnExecutable.props for tests to work in Visual Studio.
+		This will break regular builds, so be sure to remove it after running the tests. -->
+		<ProjectReference Include="$(ProjectDir)../BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj" Properties="TESTING=true" />
```
YoshiRulz: Per our Discord conversation, this isn't a suitable solution.

src/BizHawk.Tests.Client.Common/FakeMainFormTools.cs
```diff
@@ -0,0 +1,75 @@
...
+		public bool StartNewMovie(IMovie movie, bool newMovie) => throw new NotImplementedException();
+		public void StartSound() => throw new NotImplementedException();
+		public void StopSound() => throw new NotImplementedException();
+		public void Throttle() => throw new NotImplementedException();
+		public void TogglePause() => throw new NotImplementedException();
+		public void UnpauseEmulator() => throw new NotImplementedException();
+		public void Unthrottle() => throw new NotImplementedException();
+		public void UpdateDumpInfo(RomStatus? newStatus = null) => throw new NotImplementedException();
+		public void UpdateStatusSlots() => throw new NotImplementedException();
+		public void UpdateWindowTitle() => throw new NotImplementedException();
```
YoshiRulz: As with the other fake, why is this necessary if all the implementations would throw?

src/BizHawk.Tests.Client.EmuHawk/AssemblyInfo.cs
```diff
@@ -0,0 +1 @@
+﻿[assembly: Parallelize(Scope = ExecutionScope.MethodLevel)]
```
YoshiRulz: Call this file `AssemblyInfo.cs` per convention

src/BizHawk.Client.Common/DependencyInjection/DependencyProvider.cs
SuuperW: Really just a quick way to get tests running. I thought I might end up needing to use it more than I did.
Since I don't believe there's any good way to make these tests run (aside from massive refactoring of the code under test) I didn't spend much time thinking about this.

src/BizHawk.Tests.Client.Common/Display/FakeDisplayManager.cs
SuuperW: Will try that.

src/BizHawk.Tests.Client.Common/BizHawk.Tests.Client.Common.csproj
SuuperW: Yes, there is. I get compiler errors/warnings (I forgot which) when I made a class that isn't a test class public. If you change that analyzer, I'll be happy to make the needed classes public instead or using `InternalsVisibleTo`.

src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
SuuperW: Not everything throws. `NetworkingHelpers` doesn't, and it is accessed when instantiating `CommApi`. Obviously I don't need the comm api, but given the mess of reflection that handles its instantiation I wasn't about to change it for this.

src/BizHawk.Tests.Client.EmuHawk/BizHawk.Tests.Client.EmuHawk.csproj
SuuperW: We never agreed on anything there. I agree it isn't good, but I do not know of a better way. If you have any suggestions, please share.
Removing the tests entirely (or perhaps leaving them in but removing the TestClass/TestMethod attributes) might be the best option until the code under test is refactored such that it can move out of EmuHawk.

src/BizHawk.Tests.Client.Common/FakeMainFormTools.cs
SuuperW: See property `MainForApi`.
The code under test obtained it's `IMainFormForApi` by casting the `IMainFormForTools` reference, assuming they are both the same instance. If you want to refactor such that it doesn't need the `IMainFormForTools` reference, then this class can be deleted.

SuuperW:
Updated to make the tests not depend on BizHawk.Client.EmuHawk, so all GitHub's checks pass now.

src/BizHawk.Client.Common/DisplayManager/SimpleGDIPDisplayManager.cs
```diff
@@ -20,7 +20,7 @@ private SimpleGDIPDisplayManager(Config config, IEmulator emuCore, IGL_GDIPlus g
 				new GDIPlusGuiRenderer(glImpl))
 			{}
 
-		public SimpleGDIPDisplayManager(Config config, IEmulator emuCore, Func<(int Width, int Height)> getVirtualSize)
+		public SimpleGDIPDisplayManager(Config config, IEmulator emuCore)
```
YoshiRulz: Don't know why this wasn't removed earlier, but this signature change will need to be propagated to testroms:
https://github.com/TASEmulators/BizHawk/blob/7acd38983ad3e64935e31daa02943fe1558e3921/src/BizHawk.Tests.Testroms/DummyFrontend.cs#L156

src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
YoshiRulz: Adding an indirection is a 5 LOC change:
```diff
-private readonly (HttpCommunication? HTTP, MemoryMappedFiles MMF, SocketServer? Sockets) _networkingHelpers;
+private readonly IMainFormForApi _mainForm;
+private (HttpCommunication? HTTP, MemoryMappedFiles MMF, SocketServer? Sockets) _networkingHelpers => _mainForm.NetworkingHelpers;
-public CommApi(IMainFormForApi mainForm) => _networkingHelpers = mainForm.NetworkingHelpers;
+public CommApi(IMainFormForApi mainForm) => _mainForm = mainForm;
```
Then you should be able to pass `(IMainFormForApi) null!`.

YoshiRulz: Adding an indirection is a 5 LOC change:
```diff
-private readonly (HttpCommunication? HTTP, MemoryMappedFiles MMF, SocketServer? Sockets) _networkingHelpers;
+private readonly IMainFormForApi _mainForm;
+private (HttpCommunication? HTTP, MemoryMappedFiles MMF, SocketServer? Sockets) _networkingHelpers => _mainForm.NetworkingHelpers;
-public CommApi(IMainFormForApi mainForm) => _networkingHelpers = mainForm.NetworkingHelpers;
+public CommApi(IMainFormForApi mainForm) => _mainForm = mainForm;
```
Then you should be able to pass `(IMainFormForApi) null!`.

src/BizHawk.Tests.Client.Common/FakeMainFormTools.cs
YoshiRulz: They're casts to `MainForm`, not `IMainFormForApi`. I'll look into resolving the hack.

src/BizHawk.Client.Common/DisplayManager/SimpleGDIPDisplayManager.cs
YoshiRulz, line 23: Don't know why this wasn't removed earlier, but this signature change will need to be propagated to testroms:
https://github.com/TASEmulators/BizHawk/blob/7acd38983ad3e64935e31daa02943fe1558e3921/src/BizHawk.Tests.Testroms/DummyFrontend.cs#L156

src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
SuuperW: What's the value of removing `FakeMainFormForApi`, though? Is it worth making this extra other change?
The fake will be necessary for testing other things eventually, so the class would come back, even if not for Lua stuff.

src/BizHawk.Client.Common/DisplayManager/SimpleGDIPDisplayManager.cs
SuuperW: Why did this not come up as a compilation error?

src/BizHawk.Tests.Client.Common/FakeMainFormTools.cs
YoshiRulz: Pushed https://github.com/TASEmulators/BizHawk/commit/f902c10b378c6d9f9e42f6d366517f1de9b65516.

src/BizHawk.Client.Common/DisplayManager/SimpleGDIPDisplayManager.cs
YoshiRulz: The testroms projects aren't included in the main solution.
Pushed https://github.com/TASEmulators/BizHawk/commit/3b19e4ece52a4f654d277ad492eef7baab453ab3.

src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
YoshiRulz: It's a fake that doesn't fake anything, it's bloat. When the need for a fake arises then one can be written for that. (Though I'm convinced that testing the frontend will require a model-view split, at which point the real `MainFormModel` or w/e can be used.)

src/BizHawk.Tests.Client.Common/FakeMainFormTools.cs
SuuperW: That's so much worse than having a fake implementation of an interface, and hardly better than assuming our main form reference is actually the MainForm class imo.

src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
SuuperW: Yeah, it's bloat. But straightforward, harmless bloat. Since it's only in test code, I don't see it as much of an issue.
And `IMainFormForApi` would be needed to test API things, not frontend things. Using the UI model for that would be a bad idea.

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
```diff
@@ -125,26 +54,22 @@ public string GuidStr
...
 
 		public string Event { get; }
 
-		private Func<object[], object[]> Callback { get; }
-
-		public Action InputCallback { get; }
-
-		public MemoryCallbackDelegate MemCallback { get; }
-
-		public Action<int> RandomCallback { get; }
```
YoshiRulz: These can stay:
```csharp
public void CallDoomRandomCallback(int pr_class) => _ = Call([ pr_class ]);
```
then in `DoomLuaLibrary`:
```csharp
callbacks.Add(nlf.CallDoomRandomCallback);
nlf.OnRemove += () => callbacks.Remove(nlf.CallDoomRandomCallback);
```
Probably less relevant for this since one there's only 1 callsite.

SuuperW: What can stay? The code above or under "These can stay"?
And what would be the benefit for any of them staying as members of `NamedLuaFunction`?

SuuperW:
Rebased onto master.

SuuperW:
Anyone who's willing to take a look at or test this PR, feedback is appreciated.
I'd like to merge this in the near future.

SuuperW:
Force push because rebase
Two new commits are added to fix issues, otherwise identical.

src/BizHawk.Client.Common/Api/ApiManager.cs
```diff
@@ -4,27 +4,29 @@
...
-				var interfaceType = implType.GetInterfaces().FirstOrDefault(t => typeof(IExternalApi).IsAssignableFrom(t) && t != typeof(IExternalApi));
-				if (interfaceType == null) continue; // if we couldn't determine what it's implementing, then it's not an api impl. type
-				var ctor = implType.GetConstructors().Single();
-				list.Add((implType, interfaceType, ctor, ctor.GetParameters().Select(pi => pi.ParameterType).ToArray()));
+				AddApiType(implType);
 			}
-			_apiTypes = list.ToArray();
+		}
+
+		public static void AddApiType(Type type)
```
YoshiRulz, line 24: Don't like this (used in `EmuHawk.ToolApi..cctor`). `public static` state should be immutable, and certainly not affected by call order.

src/BizHawk.Client.Common/Api/ApiManager.cs
```diff
@@ -4,27 +4,29 @@
...
 	public static class ApiManager
 	{
-		private static readonly IReadOnlyList<(Type ImplType, Type InterfaceType, ConstructorInfo Ctor, Type[] CtorTypes)> _apiTypes;
+		private static readonly List<(Type ImplType, Type InterfaceType, ConstructorInfo Ctor, Type[] CtorTypes)> _apiTypes = new();
 
 		static ApiManager()
 		{
-			var list = new List<(Type, Type, ConstructorInfo, Type[])>();
-			foreach (var implType in ReflectionCache_Biz_Cli_Com.Types.Concat(ReflectionCache.Types)
+			foreach (var implType in ReflectionCache_Biz_Cli_Com.Types
```
YoshiRulz, line 17:
```suggestion
			foreach (var implType in ReflectionCache.Types
```

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -111,6 +113,11 @@ public void SeekFrame(int frame)
 		public int GetApproxFramerate()
 			=> APIs.EmuClient.GetApproxFramerate();
 
+		[LuaMethodExample("local stconget = console.getluafunctionslist( );")]
+		[LuaMethod("getluafunctionslist", "returns a list of implemented functions")]
+		public string GetLuaFunctionsList()
```
YoshiRulz, line 118: Why is this moving?
You forgot to update the example anyway.

src/BizHawk.Client.Common/lua/NamedLuaFunction.cs
YoshiRulz: These specialised `*Callback` properties can stay. My suggestion is to change them to methods (like you've done with `Callback.Invoke(...)` --> `Call(...)`) but keep the logic centralised in this file.

src/BizHawk.Client.Common/lua/ILuaLibraries.cs
```diff
@@ -2,6 +2,8 @@ namespace BizHawk.Client.Common
 {
 	public interface ILuaLibraries
```
YoshiRulz, line 3: Why does this interface exist, again? https://github.com/TASEmulators/BizHawk/commit/b3c7f0fa48ab583060c508f95b2d1d4d3533ab69 eliminated the multiple-implementations aspect, so is this just an abstraction to make it usable from Client.Common? I feel like we should be able to get rid of it.

src/BizHawk.Client.Common/lua/IRegisterFunctions.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace BizHawk.Client.Common
+{
+	public interface IRegisterFunctions
+	{
+		LuaLibraryBase.NLFAddCallback CreateAndRegisterNamedFunction { get; set; }
```
YoshiRulz, line 5: Creating an interface for this is fine, but it shouldn't be necessary for this PR.

src/BizHawk.Client.Common/lua/IPrintingLibrary.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace BizHawk.Client.Common
+{
+	public interface IPrintingLibrary
+	{
+		void Log(params object[] outputs);
```
YoshiRulz, line 5: This should be in `LuaLibraryBase` with the other log function.

src/BizHawk.Client.Common/lua/LuaFile.cs
```diff
@@ -1,54 +1,59 @@
-﻿using NLua;
+﻿using System.Collections.Generic;
+using System.Linq;
+
+using NLua;
 
 namespace BizHawk.Client.Common
 {
+	public interface IKeepFileRunning : IDisposable { }
```
YoshiRulz, line 12: This marker interface needs documenting.

src/BizHawk.Client.Common/lua/LuaFile.cs
```diff
@@ -1,54 +1,59 @@
...
-		public RunState State { get; set; }
+		public RunState State { get; private set; }
+
+		public void AddDisposable(IDisposable disposable) => _disposables.Add(disposable);
+
+		public void RemoveDisposable(IDisposable disposable) => _disposables.Remove(disposable);
+
+		public bool ShouldKeepRunning()
+		{
+			return _disposables.Exists((d) => d is IKeepFileRunning)
```
YoshiRulz, line 58:
```suggestion
			return _disposables.OfType<IKeepFileRunning>().Any()
```

src/BizHawk.Client.Common/lua/LuaFile.cs
```diff
@@ -58,25 +63,40 @@ public void Stop()
 			}
 
 			State = RunState.Disabled;
+
+			foreach (NamedLuaFunction func in Functions
+				.Where(l => l.Event == NamedLuaFunction.EVENT_TYPE_ENGINESTOP)
```
YoshiRulz, line 72: Make a prop `private IEnumerable<NamedLuaFunction> ExitHandlers => Functions.Where(nlf => nlf.Event is NamedLuaFunction.EVENT_TYPE_ENGINESTOP);` which can then be used in this method and the one above.

src/BizHawk.Client.Common/lua/INamedLuaFunction.cs
```diff
@@ -1,25 +1,27 @@
-using BizHawk.Emulation.Common;
-
 namespace BizHawk.Client.Common
 {
 	public interface INamedLuaFunction
```
YoshiRulz: Wait, if `LuaFunctionList : IEnumerable<NamedLuaFunction>`, why does *this* interface exist?

src/BizHawk.Client.Common/lua/LuaFile.cs
```diff
@@ -58,25 +63,40 @@ public void Stop()
...
+			foreach (NamedLuaFunction func in Functions
+				.Where(l => l.Event == NamedLuaFunction.EVENT_TYPE_ENGINESTOP)
+				.ToList())
+			{
+				func.Call();
+			}
+			Functions.Clear();
+
+			foreach (IDisposable disposable in _disposables.ToList())
+				disposable.Dispose();
```
YoshiRulz, line 80: Braces XOR single-line.
Is the `ToList` call because the collection may be mutated by one of the objects' `Dispose` implementations?

src/BizHawk.Client.Common/lua/LuaLibraries.cs
```diff
@@ -0,0 +1,468 @@
...
+using NLua;
+using NLua.Native;
+
+using BizHawk.Common;
+using BizHawk.Common.StringExtensions;
+using BizHawk.Emulation.Common;
+
+namespace BizHawk.Client.Common
+{
+	public class LuaLibraries : ILuaLibraries
```
YoshiRulz, line 17: GitHub isn't detecting this as a rename :(

src/BizHawk.Client.Common/tools/Interfaces/IToolLoader.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+
+namespace BizHawk.Client.Common
+{
+	public interface IToolLoader
```
YoshiRulz, line 5: "Load" implies from disk. `IToolFormManager`/`IToolFormOpener`? Or `IToolManagerForApi` to match the sole impl. name?

src/BizHawk.Client.Common/IMainFormForTools.cs
```diff
@@ -0,0 +1,111 @@
+using BizHawk.Bizware.Graphics;
+using BizHawk.Emulation.Common;
+
+namespace BizHawk.Client.Common
+{
+	public interface IMainFormForTools : IDialogController
```
YoshiRulz, line 6: GitHub isn't detecting this as a rename either :(

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/FormsLuaLibrary.cs
```diff
@@ -58,6 +58,11 @@ private static void SetSize(Control control, int width, int height)
 		private static void SetText(Control control, string caption)
 			=> control.Text = caption ?? string.Empty;
 
+		public void Dispose()
+		{
+			DestroyAll();
```
YoshiRulz, line 63: Doesn't this change the behaviour around core reloads?

YoshiRulz (changes requested):
(I got down to `TAStudioLuaLibrary` and stopped since I'd already found a bunch of things that need addressing.)

src/BizHawk.Client.Common/Api/ApiManager.cs
SuuperW: It is not affected by call order. The order of things in the list doesn't matter. And if it did matter, it would be an existing problem, not introduced in this PR. (Since the order given by ReflectionCache is not defined.)

The only place it is called is in other static constructors. It is not mutated after static initialization.

I do agree that this solution isn't ideal, but I do not think it's any worse than using reflection to find the types. Finding a good solution to this is outside the scope of this PR.

EDIT: There was an issue here actually, with the static constructor not running. That has been fixed by moving it to `ToolManager`.

src/BizHawk.Client.Common/Api/ApiManager.cs
SuuperW: I disagree. I think the explicit type name is better.

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
SuuperW: It is moving because it has to. The property it accesses must be set from `Common` and the console library must be in `EmuHawk`. It also does not logically belong in the console library.
I will fix the code example.

src/BizHawk.Client.Common/lua/ILuaLibraries.cs
SuuperW: Why should those callback properties stay? I do not see any value in keeping such callback logic in `NamedLuaFunction`. They logically belong with the API code, and are clutter in `NamedLuaFunction`. I also do not see any difference between these two and any of the other special callbacks. (And there are a good number of them, including some for TAStudio that were never put in `NamedLuaFunction`.)

It could make sense to keep them if these two are combined, as a way of putting all the sets of `IsInInputOrMemoryCallback` in one place. Although I think that could use a better name, such as `IsInCallbackDuringFrame`. Then again, an even better solution might be to set it at the start and end of every frame (and call it `FrameIsRunning` or something like that) rather than waiting until a callback happens. Since the purpose appears to be preventing calls to certain API methods that should not run mid-frame.

EDIT: The above looks like it should have been a reply to the conversation above about `*Callback`. Text below is for this conversation.

I am not sure why `ILuaLibraries` exists. Your guess is the same thing I would guess. I can remove it if you want, but I think that change can just as well be done after/outside this PR.

src/BizHawk.Client.Common/lua/IRegisterFunctions.cs
SuuperW: It is necessary because the TAStudio Lua library implements this interface, and we cannot have `LuaLibraries` checking for that type now that it has moved to `Common`.

src/BizHawk.Client.Common/lua/IPrintingLibrary.cs
SuuperW: Why `LuaLibraryBase`? If you're thinking we can then remove the `LogCallback` property in the various API classes, that won't work. (For one, that's in the base API classes, not the Lua wrappers. For two, it is set differently for Lua Console, external tools, and tests.)
Anyway it cannot be done right now because the Lua log method references `LuaConsole`, an `EmuHawk` type.

src/BizHawk.Client.Common/lua/LuaFile.cs
SuuperW: I think it's more readable and direct as-is. This is the kind of thing `Exists` is meant for, there's no need to replace it with a filtering method followed by `Any` even if there happens to be a special filtering method that does what we want.

src/BizHawk.Client.Common/lua/LuaFile.cs
SuuperW: I thought good idea, went to implement, then realized that doesn't work because the method above uses `!=` instead of `==`. So no code duplication is avoided by using a property.

src/BizHawk.Client.Common/lua/INamedLuaFunction.cs
SuuperW: I do not know. Looks like a good thing to remove to clean up the code. The git log shows you created the interface, and a quick look makes me think it was something to do with Unix/Win32 but I don't think that separation exists anymore. e4a0175

I will remove it if you think that's a good idea.

src/BizHawk.Client.Common/lua/LuaFile.cs
SuuperW: > Braces XOR single-line.

Nah.

> Is the ToList call because the collection may be mutated by one of the objects' Dispose implementations?

Correct.

src/BizHawk.Client.Common/lua/LuaLibraries.cs
SuuperW: That is unfortunate, but at least it does if you look at the individual commit.

src/BizHawk.Client.Common/tools/Interfaces/IToolLoader.cs
SuuperW: `Loader` because the only methods in the interface are named "Load".

"Load" does not have this connotation in my mind. But if this is important to you, I would not object to renaming the interface and its methods to use "open" instead after this PR. (But don't name it `IToolManagerForApi`, the interface lacks many of the "managing" parts of the sole implementer.)

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/FormsLuaLibrary.cs
SuuperW: That's the idea. When a core is reloaded the `LuaLibraries` and it's Lua libraries are re-instantiated and all scripts are restarted. Leaving behind forms that are attached to code from a disposed `LuaLibraries` instance is a bad idea. Even leaving behind ones from a restarted script is problematic since the script will re-create any forms it needs, re-initialize things, and have a different local scope.

SuuperW:
> (I got down to `TAStudioLuaLibrary` and stopped since I'd already found a bunch of things that need addressing.)

If you haven't already, please look at the test code even if just to make sure we're on the same page about what behavior is expected. If you looked at it when I first opened this PR, that's good enough since the tested behavior hasn't changed.

SuuperW:
rebased

vadosnaprimer:
I'd test it but it has conflicts.

SuuperW:
Conflicts aren't trivial to fix, unfortunately. It could be a few days before I get around to it.

SuuperW:
Rebased. I also changed how restricted APIs are handled (now uses flags), but there should not be any behavior changes compared to before this force-push.

@YoshiRulz You added various calls to `Console.WriteLine` in LuaConsole.cs. I removed the one indicating failure to resume a script (meaning, exception resuming after `emu.frameadvance` or `emu.yield`) because I could not keep it in `LuaConsole` without additional logic. I also did your TODO for reporting the current script in the console when a script prints, and that will happen for exceptions too so you can see which scripts fail that way. If that's not enough, we could set up a temporary exception handler for `LuaLibraries` to use during `ResumeScripts` that writes to console.

SuuperW:
re:
> These specialised *Callback properties can stay. My suggestion is to change them to methods (like you've done with Callback.Invoke(...) --> Call(...)) but keep the logic centralised in this file.

The common thing they were all doing (setting `IsInInputOrMemoryCallback`) has been changed to setting flags that indicate which APIs are unavailable (see `ApiGroup`), and that information is passed into the constructor. The actual setting of the flags (`LuaLibraries.ProhibitedApis`) is moved to `LuaLibraries.Sandbox` because the `YIELDING` flag needs to be set even when not using a `NamedLuaFunction`.

vadosnaprimer:
Dang it says conflicts again.

vadosnaprimer:
So what's the bottom line on this one?

SuuperW:
What do you mean? OP lists the changes in behavior.

vadosnaprimer:
I mean the "discussion".

SuuperW:
As far as I'm concerned, everything is resolved since @YoshiRulz has not replied to my answers from nearly 2 months ago. There aren't any changes that I plan to make here.

PR #4582, opened 2025-12-12
Remove IMainFormForApiInit.

SuuperW:
@YoshiRulz

Doing this is also good for my Lua PR, since it currently still has LuaFormsLibrary assuming that the main form reference is a `IDialogParent`.


src/BizHawk.Client.EmuHawk/IMainFormForTools.cs
```diff
@@ -51,10 +54,14 @@ public interface IMainFormForTools : IDialogController
...
 
+		/// <remarks>only referenced from <see cref="LuaConsole"/></remarks>
+		/// <param name="forceWindowResize">Override <see cref="Common.Config.ResizeWithFramebuffer"/></param>
+		void FrameBufferResized(bool forceWindowResize = false);
+
 		/// <remarks>only referenced from <see cref="BasicBot"/></remarks>
 		bool LoadQuickSave(int slot, bool suppressOSD = false);
 
-		/// <remarks>referenced from <see cref="MultiDiskBundler"/> and RetroAchievements</remarks>
+		/// <remarks>only referenced from <see cref="MultiDiskBundler"/></remarks>
```
SuuperW: I reverted the entire file. Looking at it now, these changes mentioning RA shouldn't have been reverted.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaLibraries.cs
```diff
@@ -108,7 +104,15 @@ void EnumerateLuaFunctions(string name, Type type, LuaLibraryBase instance)
...
 					else if (instance is FormsLuaLibrary formsLib)
 					{
-						formsLib.MainForm = _mainForm;
+						// We want to not give tools/apis explicit references to Windows Forms.
+						// But this Lua API actually needs one, because it creates Windows Forms.
+						// If we ever have another frontend framework, this library should not be used. So we check its type.
+						if (mainForm is MainForm mainAsForm)
+						{
+							formsLib.DialogParent = mainAsForm;
+							formsLib.OwnerForm = mainAsForm;
```
YoshiRulz: These are one and the same semantically.
https://github.com/TASEmulators/BizHawk/blob/2363d5e791ba72a52f18cb7c7f624261f9282ffe/src/BizHawk.Client.Common/IDialogParent.cs#L5-L6
As a `Form`, MainForm is the owner of any windows that are created by `forms.newform`. As an `IDialogParent`, MainForm is the owner of the file picker dialog when `forms.openfile` is called. In either case the object is only being used for the window handle encoded within it.
And if we change UI framework, `FormsLuaLibrary` can cast to the relevant window handle type. This class shouldn't be concerned with that.

SuuperW: So it (`FormsLuaLibrary`) should only use a `IDialogParent` property, and cast that to `Form`?
I was imagining if we change UI framework, `FormsLuaLibrary` would not be used but another equivalent one would.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaLibraries.cs
```diff
@@ -201,10 +201,9 @@ void EnumerateLuaFunctions(string name, Type type, LuaLibraryBase instance)
 		public void Restart(
 			IEmulatorServiceProvider newServiceProvider,
 			Config config,
-			IEmulator emulator,
-			IGameInfo game)
+			ApiContainer apiContainer)
 		{
-			_apiContainer = ApiManager.RestartLua(newServiceProvider, LogToLuaConsole, _mainForm, _displayManager, _inputManager, _mainForm.MovieSession, _mainForm.Tools, config, emulator, game);
+			_apiContainer = apiContainer;
```
YoshiRulz, line 200: Now this is an interesting idea. I'm not sure if it's semantically sound, but the call order seems the same so it should be fine.

SuuperW: I am not convinced that the restart logic is correct, but yeah this should be equivalent to what is already there. And I see no reason why `LuaLibraries` would need to be responsible for constructing the API instances or `ApiContainer`.

Btw this is something my other PR does too.

src/BizHawk.Client.EmuHawk/tools/Lua/LuaLibraries.cs
SuuperW: Tools don't seem to have any way to reference the main form as `IDialogParent`. Rather than assume it is one, I have made LuaConsole pass itself to LuaLibraries for the IDialogParent reference.
Also changed `FormsLuaLibrary` to use it as `IWin32Window` instead of `Form`.

SuuperW:
I'd like to get this merged (and then the other Lua one). Is there any issue merging this as-is?

src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -216,12 +229,11 @@ public override void Restart()
...
 				Emulator.ServiceProvider,
-				MainForm,
-				DisplayManager,
-				InputManager,
+				MainFormForApi,
 				Config,
-				Emulator,
-				Game);
+				Tools,
+				this,
```
YoshiRulz: Should be `(IDialogParent) MainForm` to keep old behaviour

YoshiRulz:
That's my only remaining objection, otherwise I think this is an elegant simplification of https://github.com/TASEmulators/BizHawk/commit/f902c10b378c6d9f9e42f6d366517f1de9b65516.

PR #4583, opened 2025-12-13
TAStudio: better axis editing

SuuperW:
1. Fix a bunch of axis editing bugs in TAStudio, particularly when editing multiple frames at once.
2. User can now select multiple rows while in axis editing mode by shitf-clicking or ctrl-clicking columns other than the frame column.
3. Typing more digits than are allowed no longer deletes the typed value.
4. Backspacing when nothing is type no longer resets to neutral value.
5. Neutral value is displayed while in axis edit mode, instead of showing as blank.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
@@ -64,22 +64,19 @@ namespace BizHawk.Client.EmuHawk
 		private string _axisTypedValue = "";
 		private bool _didAxisType;
 		private int _axisEditYPos = -1;
-		private bool _exitAxisEditingOnMouseUp = true;
 		private int _axisRestoreId;
 
 		/// <summary>
 		/// Begin editing an axis value by dragging the mouse.
 		/// </summary>
 		/// <param name="yPos">The initial vertical position of the cursor.</param>
-		/// <param name="exitOnMouseUp">True to axis editing mode (not mouse edit mode) on mouse up even if no edit occured.</param>
-		private void BeginAxisMouseEdit(int yPos, bool exitOnMouseUp)
+		private void BeginAxisMouseEdit(int yPos)
 		{
 			Debug.Assert(AxisEditingMode, "Don't begin axis mouse edit outside of axis editing mode.");
 
 			_axisEditYPos = yPos;
 			_axisTypedValue = "";
 			_didAxisType = false;
-			_exitAxisEditingOnMouseUp = exitOnMouseUp;
 			_axisRestoreId = CurrentTasMovie.ChangeLog.MostRecentId;
 
 			CurrentTasMovie.ChangeLog.BeginNewBatch($"Axis mouse edit, frame {TasView.SelectedRows.First()}");
@@ -582,7 +579,7 @@ namespace BizHawk.Client.EmuHawk
 					}
 					else
 					{
-						BeginAxisMouseEdit(e.Y, true);
+						BeginAxisMouseEdit(e.Y);
 					}
 					return;
 				}
@@ -699,7 +696,7 @@ namespace BizHawk.Client.EmuHawk
 							else
 							{
 								AxisEditColumn = buttonName;
-								BeginAxisMouseEdit(e.Y, false);
+								BeginAxisMouseEdit(e.Y);
 							}
 
 							RefreshDialog();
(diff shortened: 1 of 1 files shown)


src/BizHawk.Client.EmuHawk/CustomControls/InputRoll/Cell.cs
```diff
@@ -106,7 +105,8 @@ public override void Add(Cell item)
 			var i = _list.BinarySearch(item);
 			if (i >= 0)
 			{
-				Debug.Assert(false, $"{nameof(CellList)}'s distinctness invariant was almost broken! CellList.Add({(item is null ? "null" : item.ToString())})");
+				// We can end up adding cells that are already selected pretty easily, by combining Alt multi-selection with Shift range-selection.
+				// System.Diagnostics.Debug.Assert(false, $"{nameof(CellList)}'s distinctness invariant was almost broken! CellList.Add({(item is null ? "null" : item.ToString())})");
```
YoshiRulz, line 109: 🤦
Fix the offending selection code. The usual selection actions respect the invariant.

SuuperW: The offending selection code would, if "fixed", simply check if the cell was already selected prior to selecting it. There is no other way to do it. That doesn't seem any better, and I don't see any reason to but that burden on users of `InputRoll` or `CellList`. In terms of performance, that could easily be worse since users of `InputRoll` aren't going to perform that check with a binary search.

SuuperW: And to be clear, this change isn't related to axis editing. But I wanted it to stop nagging me while I was testing things.

vadosnaprimer:
When testing Doom I realized I need a way to change the sign with a simple command/key. Currently that's inconsistent (tested on this PR): right after entering analog edit mode if you type minus, it will erase the numeric value and put the minus; but after you've also typed the numeric value, hitting minus will instead remove/add the sign.

Also now there's no way to keep changing the numeric value after you've typed all the available digits by typing more: you have to increment/decrement via other keys. Which means if you have autorestore on and you wanna find the optional value without leaving the edit mode, you can't type in arbitrary values repeatedly, only increment them via keys (mouse drag exits edit mode so it's not helpful either).

Also exiting analog edit mode after every drag-edit feels clunky because sometimes you wanna keep editing and seeing the results, or you simply can't drag far enough to get the value you need in one go, and double-clicking again every time makes it more annoying. I tried looking at that code but it was quite complicated to make it not exit the edit mode without breaking other things.

SuuperW:
> Currently that's inconsistent (tested on this PR): right after entering analog edit mode if you type minus, it will erase the numeric value and put the minus; but after you've also typed the numeric value, hitting minus will instead remove/add the sign.

I think this needs its own hotkey, then. If pressing minus right after entering axis editing mode flipped the sign, that would be inconsistent with the behavior of letting users type in a new value without backspacing the old one. If a user wants to type in 2, they'd double click then press '2'. If they want to type in -2, they'd need to double click, backspace (potentially multiple times), then press '-' and '2'. (Edit: I suppose they could, instead of backspacing, type '2' and THEN press minus. But I think that is generally a worse user experience.)

> Also now there's no way to keep changing the numeric value after you've typed all the available digits by typing more: you have to increment/decrement via other keys. Which means if you have autorestore on and you wanna find the optional value without leaving the edit mode, you can't type in arbitrary values repeatedly

I don't understand what you mean with this. There's never a need to use the increment/decrement hotkeys, or to leave edit mode: you can always backspace and type in a new digit. Perhaps it would be useful to make it so that typing a new digit replaces the last one ('105' -> type '6' -> '106'), but the old behavior of deleting everything was extremely annoying, and confusing, in my experience.

> Also exiting analog edit mode after every drag-edit feels clunky because sometimes you wanna keep editing and seeing the results, or you simply can't drag far enough to get the value you need in one go, and double-clicking again every time makes it more annoying. I tried looking at that code but it was quite complicated to make it not exit the edit mode without breaking other things.

That's just the way it's always been. I can make it so that releasing the mouse button never exits edit mode.

vadosnaprimer:
> you can always backspace

Oh I forgot!

> I can make it so that releasing the mouse button never exits edit mode.

I think that'd be good!

SuuperW:
New commit makes it so releasing left mouse never exits axis edit mode.
Force push to fix other issues that doing this revealed.

PR #4584, opened 2025-12-13
more sensible handling of our special hotkey default binding "conflicts"

SuuperW:
Previously I'd introduced code to automatically detect conflicts with our default hotkey bindings. I had made it ignore RA hotkeys, since some of those are bound to the same keys as other hotkeys but that didn't seem to be an actual issue.

I think this way of doing it is better, though: Instead of ignoring RA hotkeys, ignore the other group that has the same keys bound: TAStudio's analog edit hotkeys. I think this is better because MainForm.Hotkey.cs has special handling for analog editing mode, which prevents all other hotkeys from being triggered if analog edit mode is active. So it is actually impossible for these to conflict with anything else. But there is no such special handling for RA hotkeys. (Maybe there should be? I don't know.)

Someone familiar with the RA integration might have a different idea here.

### DIFF
--- a/src/BizHawk.Client.Common/config/Binding.cs
+++ b/src/BizHawk.Client.Common/config/Binding.cs
@@ -208,7 +208,8 @@ namespace BizHawk.Client.Common
 
 #if DEBUG
 			var bindings = dict.Values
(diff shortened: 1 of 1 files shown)


RetroEdit:
I had a different solution in #4516 but I wasn't necessarily fully happy with that. (Not sure either way about this PR -- haven't looked at it carefully yet.)

Issue #4591, opened 2025-12-21
Lua cannot unpause when user presses frame advance key


SuuperW:
### Summary
The frame advance key pauses, and this happens after Lua has run for the frame that was advanced.

### Repro
Run this Lua script:
```
while true do
	client.unpause()
	emu.frameadvance()
end
```

Observe that the frame advance key still causes emulation to pause, and this is true even when already paused: Lua cannot unpause on frame advance. A user frame advancing through a movie will want their Lua script to be able to conditionally unpause (e.g. to enable Lua-based seeking).

This is a unexpected behavior that has existed for a long time, but was unreported and perhaps unknown until #4347.

Related to this, is the question of whether the above Lua script should keep emulation unpaused when the user presses the pause button. After all, it would make sense if `client.unpause(); emu.frameadvance();` always triggered a frame advance.

An alternative solution for the particular use case that uncovered this issue might be to introduce a new Lua method that does what `emu.frameadvance` suggests in its name: actually trigger a frame advance. This would still need to be done in a way that does not pause emulation. And regardless of if we do that, I think Lua should be able to unpause after a frame advance.

RetroEdit:
For the pause button more broadly, I have occasionally found Lua pause interactions to be unintuitive, but having an "unpausable" script (where the only way to pause is to disable it and *then* pause) also seems like it might be somewhat annoying. I think the basic status quo should retain pause button responsiveness, but I'm not opposed to further new Lua API components that are able to override it.

SuuperW:
> having an "unpausable" script (where the only way to pause is to disable it and then pause) also seems like it might be somewhat annoying

Yeah, but supposedly anyone running such a script actually wants that. And my example script was just to demonstrate behavior, a typical Lua script would only unpause when some condition is met. I think it's more important to focus on what makes sense for people creating scripts, and on what is possible to do.

For scripts that seek only a few frames to get into some particular state (as it seems is the case for some camera hack scripts that apparently need to do this, but there are other reasons one might want to have a similar behavior) I think that it makes sense for users to want to be unable to interrupt the seek (except that stopping the script should always be an option). It might actually be the case that we need to support more than one kind of seeking?

I think that it might be helpful for Lua scripts to be able to know if the user is holding the frame advance key, as well as for Lua to have a `OnPauseToggle` event.

I made a Lua function that should be a reasonable short-seek:

```lua
-- Seek forward to a given frame.
[...]

PR #4592, opened 2025-12-21
Lua unpause and invisible seek support

SuuperW:
See issue #4591.

Invisible emulation also needed significant changes after Lua's `client.seekframe` was deprecated. 
The goal of invisible emulation, as is currently used, is to enable rendering a future frame as if it were the current one. To this end, all rendering, throttling, and FPS calculations are disabled during invisible emulation. Since we need the actual current frame to not render but do want to use throttling and FPS calculations, setting invisible emulation only really activates it once it is time to emulate the next fame. Conversely, we want to render the final Lua-seeked frame but not throttle or do FPS calculations for it. ~~`InvisibleEmulation` becomes `InvisibleEmulateNextFrame`, but I have not changed the name of the Lua function.~~ EDIT: Changed the name back, since it needs to be more complicated than that name suggests.

### DIFF
--- a/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
@@ -95,7 +95,14 @@ namespace BizHawk.Client.Common
 			=> "NLua+Lua";
 
 		[LuaMethodExample("client.invisibleemulation( true );")]
-		[LuaMethod("invisibleemulation", "Enters/exits turbo mode and disables/enables most emulator updates.")]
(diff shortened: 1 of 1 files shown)


SuuperW:
Ready for merge.

SuuperW:
Now with better TAStudio support.

This whole invisible emulation thing seems like a big convoluted hack, and just to avoid capturing dirty states in rewind history + savestates. Idk, maybe that's important for Sonic Advance but still it feels icky.

YoshiRulz:
Closed per https://github.com/TASEmulators/BizHawk/pull/4598#issuecomment-3833926990.

PR #4594, opened 2025-12-28
Update EmuHawkMono.sh

cD1rtX3:
[dev build for branch | USERNAME:BRANCHNAME](https://nightly.link/USERNAME/BizHawk/workflows/ci/BRANCHNAME?preview)

Added:

  * check against useless `"$0"`,
  * `export GTK_PREFIX=`, and
  * shellcheck directives against warnings;

replaced `printf` with `echo`; redirected status messages to `stderr`; added `--` safeguard for builtins when necessary; and removed `exec`.

Check if completed:
- [X] I, the commit author, have read the [licensing terms for contributors](https://github.com/TASEmulators/BizHawk/blob/master/contributing.md#copyrights-and-licensing) (last updated 2024-06-22) and am compliant


cD1rtX3:
I complied with what seems to be the style guide, but that isn't the standard style guide (notably, the use of double-quotes instead of single-quotes and the quoted `case`s).

Assets/EmuHawkMono.sh
```diff
@@ -1,35 +1,58 @@
 #!/bin/sh
-cd "$(dirname "$(realpath "$0")")"
-libpath=""
+
+case "$0" in
+	*"EmuHawkMono.sh");;
+	*"/bin/"*"sh")
+		# Very bad way to detect /path/to/shell
+		echo "I don't know where I am! Could you run me as \"/path/to/EmuHawkMono.sh\"?"
```
YoshiRulz: It should support being renamed via symlinking.

Assets/EmuHawkMono.sh
```diff
@@ -1,35 +1,58 @@
...
-	printf "Distro does not provide LSB release info API! (You've met with a terrible fate, haven't you?)\n"
+	echo "Distro does not provide LSB release info API! (You've met with a terrible fate, haven't you?)" >& 2
 fi
 if [ -z "$libpath" ]; then
-	printf "%s\n" "Unknown distro, assuming system-wide libraries are in /usr/lib..."
+	echo "Unknown distro, assuming system-wide libraries are in /usr/lib..." >& 2
 	libpath="/usr/lib"
 fi
+
+export GTK_DATA_PREFIX=
```
YoshiRulz:
```suggestion
export GTK_DATA_PREFIX=''
```

Assets/EmuHawkMono.sh
```diff
@@ -1,35 +1,58 @@
...
+	echo "(passing --mono-no-redirect is no longer necessary)" >& 2
 	shift
 fi
+# shellcheck disable=SC2009
 if (ps -C "mono" -o "cmd" --no-headers | grep -Fq "EmuHawk.exe"); then
-	printf "(it seems EmuHawk is already running, NOT capturing output)\n" >&2
-	exec mono EmuHawk.exe "$@"
+	echo "(it seems EmuHawk is already running, NOT capturing output)" >& 2
+	mono EmuHawk.exe "$@"
+	return 0
```
YoshiRulz: What's wrong with `exec`? And at least `return $?`.

Assets/EmuHawkMono.sh
```diff
@@ -1,35 +1,58 @@
...
+cd "$(dirname -- "$(realpath -- "$0")")" || ( echo "Can't navigate to \$0's path?" >& 2; exit 1 )
+
+libpath=
 if [ "$(command -v lsb_release)" ]; then
+	# shellcheck disable=SC2018,SC2019
 	case "$(lsb_release -i | head -n1 | cut -c17- | tr A-Z a-z)" in
 		"arch"|"artix"|"manjarolinux") libpath="/usr/lib";;
 		"fedora"|"gentoo"|"nobaralinux"|"opensuse") libpath="/usr/lib64";;
-		"nixos") libpath="/usr/lib"; printf "Running on NixOS? Why aren't you using the Nix expr?\n";;
+		"nixos") libpath="/usr/lib"; echo "Running on NixOS? Why aren't you using the Nix expr?" >& 2;;
```
YoshiRulz: This is not an error, this codepath is theoretically supported, though I have no idea if it works.

YoshiRulz: Also I'm in the habit of always using `printf` because of `echo`'s footguns.

YoshiRulz (changes requested):
I'm a bit confused by this PR. Did you just throw it into a linter and apply every suggestion?

Assets/EmuHawkMono.sh
cD1rtX3: Correct, it is not an error. However, status messages should always go to stderr, so not to conflate with actual output.

cD1rtX3: The footguns you mention only really happen with arbitrary variables. If one were to pass in `-n` from a variable, echo would get confused. However, when printing a simple string, it's better to use the simplest method.

Assets/EmuHawkMono.sh
cD1rtX3: Exec for a command is undefined behaviour in `sh`, which you're targeting. You are right about `return "$?"`, though.

Assets/EmuHawkMono.sh
cD1rtX3: Done.

Assets/EmuHawkMono.sh
cD1rtX3: Done.

Assets/EmuHawkMono.sh
cD1rtX3: Done.

cD1rtX3:
No, I didn't "just throw it into a linter", though I do have a linter in my IDE. I noticed some minor issues, and thought the simplest way to get them fixed was to simply open a PR.

Assets/EmuHawkMono.sh
YoshiRulz: > Exec for a command is undefined behaviour in `sh`, which you're targeting.

[Looks defined to me](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#exec), and I couldn't find anyone proscribing against it. Which shell doesn't support it?

YoshiRulz:
What's this style guide you're referring to?

Assets/EmuHawkMono.sh
YoshiRulz: Does this now support something like `ln -s /opt/bizhawk/EmuHawkMono.sh ~/.local/bin/emuhawk`?

cD1rtX3: Yes. It will just think you're running it with a shell if that executable's name ends in `sh`, but not `.sh`, and it is in a folder called `bin`. I don't know how else to better detect a shell, since bash for some reason uses its fullly-qualified path as `"$0"` when a script is sourced.

Assets/EmuHawkMono.sh
cD1rtX3: Huh. I thought I'd read that in the publication somewhere. Oh well.

Also, that link leads to the main page, not the specific page you likely wanted.

cD1rtX3:
I guess there isn't a strict “standard style guide”, just “what you'll generally see code written like online/in books”, thus being what more people are used to. If you mean *your* style guide, it's what I inferred from how you wrote the code in that file.

Assets/EmuHawkMono.sh
cD1rtX3: Also, it looks like `exec` will only look for non-builtins, though this behaviour can be mimicked by just using `./command`. It seems you can replace what I did with `exec -- # command` if you want to.

Assets/EmuHawkMono.sh
YoshiRulz: Oh, *that's* what this is for. I'm not sure I want to merge a hack just to detect a niche user error.

cD1rtX3: I guess that's fair. You could remove the hacky part, though you should probably keep the non-hacky part.

Assets/EmuHawkMono.sh
cD1rtX3: > […] though this behaviour can be mimicked by just using `./command`

not with mono 🤦🏼‍♀️. Instead, you could use `env mono # ...`.

cD1rtX3:
Why do we need to set `$LD_LIBRARY_PATH` to `/usr/lib`? Most apps get around just fine with it being empty, and if it's set, the system should search the default paths after the ones inside the path anyway (from my understanding). Was this catching us off-guard before?

Assets/EmuHawkMono.sh
cD1rtX3: Any updates?

YoshiRulz:
Can you revert the changes to whitespace and s/printf/echo/, and *explain what problems this solves*?

cD1rtX3:
I'll revert `echo` thing and squish the lines and redirections back together. On my local version, I've already reverted the `exec` thing (since that was just due to me misreading documentation) and the hack for detecting shells.

In the original, you use both `printf "message\\n"` and `printf "%s\\n" "message"`. Which one should I revert to?

YoshiRulz:
Use whichever was used before—'revert'.

cD1rtX3:
Do you have the file extensions for each of the ROM types we support? It seems that's the only way my system determines the actual ROM type, so knowing the extensions would give me a list of mimetypes. I'm currently just downloading pirated ROMs online and looking at their file extensions, but for many of the obscure consoles in the support list, I can't verify the given extension is non-standard; I can only know it is standard when my file explorer tells me it recognizes it. I also don't know whether the list would be different on other systems (seeing as ROM mimetypees aren't standard—the only standard type in my list is `application/octet-stream` since many ROMs don't actually have a unique mimetype).

Anyway, I've rolled back my changes to printing. However, wouldn't you want to at least standardize it?

cD1rtX3:
> Can you […] *explain what problems this solves*?

The original draft moves the status messages to stderr, uncomments something that says “TODO uncomment” (why was it commented in the first place?), adds the `export GTK_DATA_PREFIX=""` fix, covers a corner case where we don't actually know where we are, and suppresses a few linter warnings. It also made it mode readable in my opinion, but you are the project steward, and if it wasn't more readable to you, it wasn't more readable. The current draft is still on my system, but it's going to add a `--install` option which installs the program as a desktop file.

cD1rtX3:
The current script only checks the first argument for `--mono-no-redirect`, ignoring the others. That's also what I'm doing for `--install`, and it's most likely fine, but should I implement something more robust?

Also, should I scan for `--help` to tell both the user about `--install` and also list BizHawk's actual help?

cD1rtX3:
… apparently my system already provides a list of all mimetypes, which I can filter for ROMs…

cD1rtX3:
Where's the icon stored?

YoshiRulz:
I'm just going to close this. You still have unrelated changes mixed in, and I don't want a new flag added to the wrapper script. It already has too many responsibilities.
I've taken your change to direct the other warnings to stderr, and removed the warning for `--mono-no-redirect`.

The MIME type database is [hosted by IANA](https://www.iana.org/assignments/media-types/media-types.xhtml). I haven't bothered trying to match them up with our system IDs yet, since there's a lot of work to be done before we can stop relying on file extensions.
EmuHawk's icon is embedded into the assembly. I suggest you use `Icon=emuhawk-monort` to allow overrides via the icon theme, though I can't remember off the top of my head how you set a fallback for that.

Issue #3756, opened 2023-08-21
Auto-search on too many addresses hangs entire system


YoshiRulz:
Basically #1155 but for RAM Search. I'm thinking disable the button when there are >=65536 addresses—it's not like this removes functionality for those with powerful machines, the manual search button will still be available.

Meerkov:
Instead, what if you do it based on time? e.g. if it takes 15 seconds and still hasn't finished processing the frame, then cancel the auto search and pause the emu with a warning?

I've accidentally hanged my RAM search many times, and I just learned to suffer with it. But an escape hatch like this would be OK.

PR #4598, opened 2025-12-31
Lua: Show a future frame

SuuperW:
This is an alternative to the invisible emulation changes made in #4592. (If this is merged, the unpausing thing from that PR still needs to be done.)

The old seekframe and invisibleemulation API methods have been removed, in favor of `show_future`.

Tool updates do not run during future emulation. Tools used to update during invisible emulation, which caused problems. But Lua would not update during `seekframe` (except that input and memory callbacks would run), so this matches old Lua behavior.

It is possible that a user would want Lua to run during future emulation (or in old terms, invisible emulation) but we can't know that. If we want to support this, it should be something such that scripts must opt-in to running during future emulation since most things should not run there. Maybe a new type of event.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
+++ b/src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
@@ -145,7 +145,7 @@ namespace BizHawk.Client.Common
 		/// When the callback returns true, emulation will rewind to the real current frame and the just-run future frame will be displayed.
 		/// <br/>Pass null to disable future frame display.</param>
 		/// <param name="maxFrames">The maximum number of future frames to emulate. Useful to avoid freezing the client UI in case of accidentally never returning true from the callback.</param>
-		void ShowFuture(Func<int, bool> preFrameCallback, int maxFrames);
+		void ShowFuture(ShowFutureCallback/*?*/ preFrameCallback, short maxFrames);
 
 		void SpeedMode(int percent);
 
--- a/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
@@ -286,11 +286,17 @@ namespace BizHawk.Client.Common
 				"Unregister the callback with event.unregister____ to disable future frame display. " +
 				"No more than `maxFrames` future frames will be emulated. Useful to avoid freezing " +
 				"the client UI in case of accidentally never returning true from the callback.")]
-		public string ShowFuture(LuaFunction luaf, int maxFrames, string name = null)
+		public string ShowFuture(LuaFunction luaf, long maxFrames, string name = null)
 		{
+			if (maxFrames < 1 || maxFrames > short.MaxValue)
+			{
+				Log($"Invalid number of future frames ({maxFrames}); number must be positive and less than 2^15.");
+				return "";
+			}
+
 			INamedLuaFunction nlf = CreateAndRegisterNamedFunction(luaf, NamedLuaFunction.EVENT_TYPE_FUTURE, prohibitedApis: ApiGroup.PROHIBITED_MID_FRAME, name: name);
-			Func<int, bool> FutureCallback = (f) => nlf.Call(f) is [ bool r ] ? r : false;
-			APIs.EmuClient.ShowFuture(FutureCallback, maxFrames);
+			ShowFutureCallback FutureCallback = (f) => nlf.Call(f) is [ bool r ] ? r : false;
+			APIs.EmuClient.ShowFuture(FutureCallback, (short)maxFrames);
 			nlf.OnRemove += () => APIs.EmuClient.ShowFuture(null, 0);
 			return nlf.GuidStr;
 		}
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -1188,7 +1188,7 @@ namespace BizHawk.Client.EmuHawk
 
 		public event StateSavedEventHandler SavestateSaved;
 
-		public Func<int, bool>/*?*/ PreFutureFrameCallback { get; set; }
+		public ShowFutureCallback/*?*/ PreFutureFrameCallback { get; set; }
 
 		public int MaxFutureFrames { get; set; }
 
@@ -3081,7 +3081,7 @@ namespace BizHawk.Client.EmuHawk
 					statable.SaveStateBinary(new(state));
 
 					int frameCount = 0;
-					while (!PreFutureFrameCallback(frameCount) && frameCount < MaxFutureFrames)
+					while (!PreFutureFrameCallback((short)frameCount) && frameCount < MaxFutureFrames)
 					{
 						frameCount++;
 						MovieSession.HandleFrameBefore();
--- a/src/BizHawk.Client.Common/Api/Classes/EmuClientApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/EmuClientApi.cs
@@ -219,8 +219,14 @@ namespace BizHawk.Client.Common
 			}
 		}
 
-		public void ShowFuture(Func<int, bool> preFrameCallback, int maxFrames)
+		public void ShowFuture(ShowFutureCallback/*?*/ preFrameCallback, short maxFrames)
 		{
+			if (maxFrames < 1 && preFrameCallback != null)
+			{
+				_logCallback($"Invalid number of future frames ({maxFrames}); number must be positive.");
+				return;
+			}
+
 			_mainForm.PreFutureFrameCallback = preFrameCallback;
 			_mainForm.MaxFutureFrames = maxFrames;
 		}
--- a/src/BizHawk.Client.Common/EventTypes.cs
+++ b/src/BizHawk.Client.Common/EventTypes.cs
@@ -122,4 +122,11 @@ namespace BizHawk.Client.Common
 	/// <param name="sender">Object that raised the event</param>
 	/// <param name="e">Event arguments</param>
 	public delegate void StateSavedEventHandler(object sender, StateSavedEventArgs e);
+
+	/// <summary>
+	/// Represent a method that will control future emulation.
+	/// </summary>
+	/// <param name="futureFramesEmulated">The number of frames into the future that have been emulated. This is a short purely to help avoid accidentally passing in absurdly large numbers.</param>
+	/// <returns>True to display the current frame, false to continue emulating into the future.</returns>
+	public delegate bool ShowFutureCallback(short futureFramesEmulated);
 }
--- a/src/BizHawk.Client.Common/IMainFormForApi.cs
+++ b/src/BizHawk.Client.Common/IMainFormForApi.cs
@@ -35,7 +35,7 @@ namespace BizHawk.Client.Common
 		bool PauseAvi { get; set; }
 
 		/// <remarks>only referenced from <see cref="EmuClientApi"/></remarks>
-		public Func<int, bool>/*?*/ PreFutureFrameCallback { get; set; }
+		public ShowFutureCallback/*?*/ PreFutureFrameCallback { get; set; }
 
 		/// <remarks>only referenced from <see cref="EmuClientApi"/></remarks>
 		void ClearHolds();
--- a/src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
+++ b/src/BizHawk.Tests.Client.Common/FakeMainFormForApi.cs
@@ -27,7 +27,7 @@ namespace BizHawk.Tests.Client.Common
 
 		public bool PauseAvi { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
 		public int MaxFutureFrames { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
-		public Func<int, bool> PreFutureFrameCallback { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+		public ShowFutureCallback PreFutureFrameCallback { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
 
 #pragma warning disable CS0067 // Events are never used
 		public event BeforeQuickLoadEventHandler? QuicksaveLoad;


vadosnaprimer:
> This is an alternative to the invisible emulation changes made in #4592. (If this is merged, the unpausing thing from that PR still needs to be done.)

What if the other PR is merged first?

> It is possible that a user would want Lua to run during future emulation (or in old terms, invisible emulation) but we can't know that. If we want to support this, it should be something such that scripts must opt-in to running during future emulation since most things should not run there. Maybe a new type of event.

An event sounds good, tho I'd wait for a request to add it, because I can't imagine a case when it's needed.

vadosnaprimer:
Looked at the code and I absolutely prefer this approach, great idea! Just please mention in the docs that the function we pass will be automatically passed current frame as an argument (or whatever its argument is, but that's what I understood by looking at it).

Also mention that actual current frame remains the same, the only thing that changes is visuals that are taken from the target future frame... and possibly what it's used for, like we hack memory and then make that hacked frame appear. Because fake frame advancing now happens under the hood, so this info needs to be explicit.

SuuperW:
> What if the other PR is merged first?

It should not be merged at all. I think this one is better and since you agree (and no one has disagreed) I will go ahead and close the other. I can directly commit the unpause fix to master.

> An event sounds good, tho I'd wait for a request to add it, because I can't imagine a case when it's needed.

Agreed. Besides, the callback that already exists for checking if the desired future frame has been reached would likely be able to serve that purpose.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
```diff
@@ -71,12 +71,6 @@ public interface IEmuClientApi : IDisposable, IExternalApi
 
 		int GetWindowSize();
 
-		/// <summary>
-		/// Use with <see cref="SeekFrame(int)"/> for CamHack.
-		/// Refer to <c>MainForm.InvisibleEmulation</c> for the workflow details.
-		/// </summary>
-		void InvisibleEmulation(bool invisible);
```
YoshiRulz, line 78: Deprecate instead of removing.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
```diff
@@ -148,6 +136,13 @@ public interface IEmuClientApi : IDisposable, IExternalApi
 
 		void SetWindowSize(int size);
 
+		/// <summary>
+		/// Tell the client to display a frame from the future, instead of the current frame.
+		/// </summary>
+		/// <param name="preFrameCallback">This will be called before emulating each future frame. Emulation will stop when this function returns true, and the just-run frame will be displaed. Pass null to disable future frame display.</param>
+		/// <param name="maxFrames">The maximum number of future frames to emulate. Useful to avoid freezing the client UI in case of accidentally never returning true from the callback.</param>
+		void ShowFuture(Func<int, bool> preFrameCallback, int maxFrames);
```
YoshiRulz: Create a new delegate type.

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -94,18 +96,6 @@ public void FrameSkip(int numFrames)
 		public string GetLuaEngine()
 			=> "NLua+Lua";
 
-		[LuaMethodExample("client.invisibleemulation( true );")]
-		[LuaMethod("invisibleemulation", "Enters/exits turbo mode and disables/enables most emulator updates.")]
-		public void InvisibleEmulation(bool invisible)
```
YoshiRulz, line 101: For Lua too, deprecate instead of removing.

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -278,6 +268,20 @@ public int ScreenWidth()
 		public void SetWindowSize(int size)
 			=> APIs.EmuClient.SetWindowSize(size);
 
+		[LuaMethodExample("client.show_future(function(frame) return frame == 1 end, 1)")]
+		[LuaMethod(
+			name: "show_future",
+			description: "Tell the client to display a frame from the future, instead of the current frame. The given lua function " +
+				"will be called before each future frame is emulated until the function returns true, at which point future " +
+				"emulation will be stopped and the most recently emulated frame will be displayed.")]
+		public string ShowFuture(LuaFunction luaf, int maxFrames, string name = null)
```
YoshiRulz: This is like a specialised `event.onframestart`, so it should go in `event`.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1102,27 +1102,6 @@ private set
...
-		/// </list></description></item>
-		/// <item><description>Updating the screen</description></item>
-		/// <item><description>Loading the no-framebuffer state from RAM</description></item>
-		/// </list>
-		/// The most common use case is CamHack for Sonic games.
-		/// Accessing this from Lua allows to keep internal code hacks to minimum.
-		/// <list type="bullet">
-		/// <item><description><see cref="ClientLuaLibrary.InvisibleEmulation(bool)"/></description></item>
-		/// </list>
-		/// </summary>
```
YoshiRulz, line 1125: Keep docs, on `MaxFutureFrames` I guess?

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1207,6 +1186,10 @@ private set
 
 		public event StateSavedEventHandler SavestateSaved;
 
+		public Func<int, bool>/*?*/ PreFutureFrameCallback { get; set; }
+
+		public int MaxFutureFrames { get; set; }
```
YoshiRulz, line 1193: Should this be unsigned?

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3120,6 +3087,33 @@ private void StepRunLoop_Core(bool force = false)
...
+					int frameCount = 0;
+					while (!PreFutureFrameCallback(frameCount) && frameCount < MaxFutureFrames)
+					{
+						frameCount++;
+						MovieSession.HandleFrameBefore();
+						Emulator.FrameAdvance(InputManager.ControllerOutput, true, false);
+						CheatList.Pulse();
+						// No tools updates here. No existing tool (except Lua, but that gets the ShowFutureFrameCallback) needs to do anything.
+						// Maybe in the future we'll add a special update type, or add a callback for this.
+						// Note that other callbacks (e.g. memory hooks) are still being used.
```
YoshiRulz, line 3092: Is the GUI responsive during this? May need to `Refresh()`.

YoshiRulz (dismissed):

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
SuuperW: Why?

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
SuuperW: For `Func<int, bool>`?

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
SuuperW: No, it is not. It functions like a specialised `event.onframestart`, but conceptually it is not. It is not an event that sometimes happens and which Lua scripts may wish to be informed of. It is a callback that is only called when requested, and used to control the client.

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: These docs only apply to invisible emulation. They do not apply to the new system. If there is a lack of documentation for the new API, comment on what is missing there.

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: No. Nobody would want a value larger than int.MaxValue.
I do not recall considering what happens if a negative value is given, but I think the best thing to do in that case is either (a) not actually show a future frame, or (b) error.
Perhaps making it unsigned would fit with negative values being an error? But I'd rather not make it so easy to accidentally have a negative value get cast to >2 billion.

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: No, it is not. I do not know how user interactions with the UI would affect future emulation.
I'm not aware of any other blocking operation that leaves the GUI responsive. Attempting to do so may cause issues; even the latency introduced by `Refresh()` would degrade typical use of this feature.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
YoshiRulz: ApiHawk is intended for external use and so it should afford some semblance of stability.

Add `[Obsolete]` and make the implementation throw `NotSupportedException`.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
YoshiRulz: Yes. The parameter names will be used by IDEs when autocompleting a lambda expression.

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: If you make the API only accept non-negative integers then you won't have to worry about that possibility. [There is no implicit `int`-->`uint` cast](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/numeric-conversions).
You could also reduce the width (to `byte`) to prevent extremely large values, but really you should be checking the user-provided value ASAP and throwing `ArgumentOutOfRangeException`.

src/BizHawk.Client.EmuHawk/MainForm.cs
YoshiRulz: We've had similar problems with TAStudio (#2613) and RAM Search (#3756), and no-one has put forward an obvious best solution or even a good solution.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
SuuperW: Imo, making it always throw an exception is worse than deleting it entirely. Better to fail at compile time than at run-time. Or would this cause issues with running an external tool compiled for a previous version?

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: It doesn't matter that there's no implicit int -> uint cast. Someone will make an explicit cast without necessarily thinking about the implications. (Should they be aware? Yes. But I don't think making that change would help anyone be aware.)
Reducing it to `byte` might be a good idea, but I don't know a good way to determine what a hard upper limit should be. I wouldn't want an `ArgumentOutOfRangeException` preventing legitimate uses. (Still, 255 seems plenty high enough for what this feature is meant for. And low enough it won't force a user who accidentally triggers it to kill the process.)

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: The lack of a good solution in those cases makes me not want to do anything here. A good solution would likely be applicable to all of these situations and should not be custom-made for this one. Also the difficulty of making a good solution deters me from putting in the effort. Maybe another time in another PR.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
YoshiRulz: It makes it easier to update tools to the new version, yes. With `[Obsolete]`, it would warn or error at compile-time.

vadosnaprimer: Are there even any tools using this? One of the goals of the new approach is reducing the code clutter, so the more hacks we leave in the worse it is.

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
vadosnaprimer: Camhack scripts already don't work in 2.11 so we're not making things worse by fixing the old convoluted way. And we can ship the new scripts anyway.

src/BizHawk.Client.EmuHawk/MainForm.cs
vadosnaprimer: 1 byte might be not enough for insane sonic situations. Reminder: in genesis sonic you can't force the camera to an arbitrary position, you have to gradually scroll it, otherwise things break. So depending on level width (romhacks exist too) we may have to scroll the camera really far, taking many frames.

src/BizHawk.Client.EmuHawk/MainForm.cs
vadosnaprimer: GUI is kinda meant to not be responsive while it's doing invisible frame advance, so the process can't be interrupted by user accidentally, breaking the whole illusion.

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
YoshiRulz: There's no way of knowing which APIs are in use.
If it was nonfunctional in 2.11 as you said below, then a proper deprecation isn't as important.

vadosnaprimer: Yeah the only use I know of is sonic tasers that took my script and tweaked it to work with sonic 3. And they can't switch to new gpgx that has deep freezing (added specifically for them) because something broke in invisible emulation.

src/BizHawk.Client.EmuHawk/MainForm.cs
vadosnaprimer: > I do not recall considering what happens if a negative value is given, but I think the best thing to do in that case is either (a) not actually show a future frame, or (b) error. Perhaps making it unsigned would fit with negative values being an error? But I'd rather not make it so easy to accidentally have a negative value get cast to >2 billion.

I like the idea of an error on negative numbers. As for the limit, 65535 is probably a good one, so the type could just be a short (even signed short would give a decent limit).

vadosnaprimer:
> I can directly commit the unpause fix to master.

Any news on this one?

src/BizHawk.Client.Common/Api/Interfaces/IEmuClientApi.cs
```diff
@@ -139,7 +139,11 @@ public interface IEmuClientApi : IDisposable, IExternalApi
 		/// <summary>
 		/// Tell the client to display a frame from the future, instead of the current frame.
 		/// </summary>
-		/// <param name="preFrameCallback">This will be called before emulating each future frame. Emulation will stop when this function returns true, and the just-run frame will be displaed. Pass null to disable future frame display.</param>
+		/// <param name="preFrameCallback">This will be called before emulating each future frame.
+		/// The parameter is the number of frames into the future that have already been emulated.
+		/// Return false to emulate another future frame.
+		/// When the callback returns true, emulation will rewind to the real current frame and the just-run future frame will be displaed.
```
vadosnaprimer: displayed

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -271,9 +271,14 @@ public void SetWindowSize(int size)
...
 		[LuaMethod(
 			name: "show_future",
-			description: "Tell the client to display a frame from the future, instead of the current frame. The given lua function " +
-				"will be called before each future frame is emulated until the function returns true, at which point future " +
-				"emulation will be stopped and the most recently emulated frame will be displayed.")]
+			description: "Tell the client to display a frame from the future, instead of the current frame. " +
+				"The given lua function will be called before emulating each future frame. " +
+				"It can have 1 parameter, which is the number of frames into the future that have already been emulated. " +
+				"Return false to emulate another future frame. " +
+				"When the callback returns true, emulation will rewind to the real current frame and the just-run future frame will be displaed. " +
```
vadosnaprimer: displayed

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -271,9 +271,14 @@ public void SetWindowSize(int size)
...
+				"Unregister the callback with event.unregister____ to disable future frame dispaly. " +
```
vadosnaprimer: display

vadosnaprimer:
I tested this with sonic advance camhack in tastudio and it feels seamless! Video in the encode works as intended too.

Audio is unusable but it wasn't usable before either, and we don't exactly needed it anyway because normal encode's audio could be used. Fixing the audio may be tricky too because it can't have pauses in the dump anymore, so that bit would have to be fixed first. Low priority overall.

SuuperW:
> Any news on this one?

The "fix" I had intended to make at that time was somewhat problematic. I went with updating the documentation instead, telling users why the behavior is what it is and how to get what they want. That's still not totally ideal, since it won't work from a callback. But that is not related to this PR. Re-open #4591 if you want something done about that.

> Audio is unusable

I was wondering about that, it looked wrong. But I don't plan on fixing that since it was already broken.

src/BizHawk.Client.EmuHawk/MainForm.cs
SuuperW: It is now a `short`. For Lua, it is `long` since lua only has 1 integer type. Making the C# side be `short` results in an implicit cast so instead I made it check that it fits in a short.

vadosnaprimer:
conflicts appeared :(

SuuperW:
> conflicts appeared :(

And were resolved. I'll go ahead and merge it for you.

PR #4618, opened 2026-02-04
Resolve #4591 by documenting the behavior.

### LINKED ISSUE #4591: Lua cannot unpause when user presses frame advance key

SuuperW:
See #4591

### DIFF
--- a/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
@@ -304,7 +304,12 @@ namespace BizHawk.Client.Common
 		}
 
 		[LuaMethodExample("client.unpause( );")]
-		[LuaMethod("unpause", "Unpauses the emulator")]
+		[LuaMethod(
+			"unpause",
+			"Unpauses the emulator. Note that the user can pause again before the next frame,"
+				+ " either with the pause key or by releasing frame advance."
+				+ " If you want to force emulation to continue, put this and emu.yield (not frameadvance)"
+				+ " inside a loop that runs until the desired frame.")]
 		public void Unpause()
 			=> APIs.EmuClient.Unpause();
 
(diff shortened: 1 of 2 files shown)


vadosnaprimer (approved):

src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
```diff
@@ -304,7 +304,12 @@ public LuaTable TransformPoint(int x, int y) {
 		}
 
 		[LuaMethodExample("client.unpause( );")]
-		[LuaMethod("unpause", "Unpauses the emulator")]
+		[LuaMethod(
+			"unpause",
+			"Unpauses the emulator. Note that the user can pause again before the next frame, " +
+			"either with the pause key or by releasing frame advance. " +
+			"If you want to force emulation to continue, put this and emu.yield (not frameadvance) " +
+			"inside a loop that runs until the desired frame.")]
```
YoshiRulz: Put operators at start of line, e.g.
https://github.com/TASEmulators/BizHawk/blob/0b4d71db89805950237280d0b98f11d5e6862571/src/BizHawk.Client.Common/lua/CommonLibs/GuiLuaLibrary.cs#L128-L131

PR #4621, opened 2026-02-09
Make branch load lua event more useful

SuuperW:
I have a Lua script for tracking RNG, that includes logic that runs when rewinding. It also has some logic to handle branch loads, but this fails because Lua can't know a branch was loaded until after the regular Lua after-frame update. This PR fixes that by invoking the branch load event earlier.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
@@ -968,7 +968,7 @@ namespace BizHawk.Client.EmuHawk
 			return CurrentTasMovie.TasStateManager.GetStateClosestToFrame(frame > 0 ? frame - 1 : 0);
 		}
 
-		public void LoadState(KeyValuePair<int, Stream> state, bool discardApiHawkSurfaces = false)
+		public void LoadState(KeyValuePair<int, Stream> state, int? branchIndex = null)
 		{
 			StatableEmulator.LoadStateBinary(new BinaryReader(state.Value));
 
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.cs
```diff
@@ -977,9 +977,15 @@ public void LoadState(KeyValuePair<int, Stream> state, bool discardApiHawkSurfac
 				Emulator.ResetCounters();
 			}
 
+			if (branchIndex.HasValue)
+			{
+				BranchLoadedCallback?.Invoke(branchIndex.Value);
+			}
 			UpdateTools();
-			if (discardApiHawkSurfaces)
```
vadosnaprimer, line 981: What exactly was `discardApiHawkSurfaces` arg used for?

SuuperW: I don't really know. I just tested making it always call `DisplayManager.DiscardApiHawkSurfaces` and doing that doesn't seem to cause any issues. Neither does making it never call.
Unfortunately there are no comments in the display manager code indicating what this is actually for.

It's possible that it doesn't actually do anything, and the method is obsolete.

vadosnaprimer: Does git blame help?

YoshiRulz: Introduced in https://github.com/TASEmulators/BizHawk/commit/f0bc8fbd2e730231d703d6d4668e34073e47bf05

vadosnaprimer: @Morilli

Morilli: It's as I described in the commit message: When a loadstate happens and there is an immediate frame advance after (such as from seeking back/rewinding, a lua script could draw to the screen twice, once for both frames. So  the display needs to be cleared after the `UpdateTools` call if another frame advance happens after to not get double visuals on screen ([relevant discussion in the tasvideos discord](https://discord.com/channels/280806848909541376/1234179729767010305/1246554744776818759)).

The logic in here seems solid though, so this should be fine.

PR #4624, opened 2026-02-11
Change target frameworks to fix warnings and allow updating dependencies

Morilli:
Currently, `BizHawk.BizInvoke` target netframework 4.8 because it uses certain APIs that aren't available under netstandard2.0. Other projects depending on `BizHawk.BizInvoke` target netstandard2.0 though, which generates NU1702, which is currently suppressed.

If we want to update `System.Memory` (which is required at least for other dependencies), we start getting tons of `MSB3277` warnings because the framework version of the `System.Memory` dll has a higher assembly version than the netstandard one, resulting in unresolvable conflicts on build. Those are of course not fatal because the API surface is the same, so it could probably also be ignored.

However, considering that we're getting warnings that need to be suppressed simply because we're doing questionable things with target framework and dependencies I'd argue adjusting the `TargetFramework` property is the more correct resolution here, unless someone can convince me that targeting netstandard instead of net framework has any advantages.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -27,12 +27,12 @@
 		<PackageVersion Include="ppy.SDL2-CS" Version="1.0.630-alpha" /> <!-- last version with .NET Standard 2.0 support -->
 		<PackageVersion Include="Roslynator.Analyzers" Version="4.13.1" />
 		<PackageVersion Include="SharpCompress" Version="0.40.0" />
-		<PackageVersion Include="Silk.NET.OpenAL" Version="2.21.0" />
-		<PackageVersion Include="Silk.NET.OpenAL.Extensions.Creative" Version="2.21.0" />
-		<PackageVersion Include="Silk.NET.OpenAL.Extensions.Enumeration" Version="2.21.0" />
-		<PackageVersion Include="Silk.NET.OpenAL.Extensions.EXT" Version="2.21.0" />
-		<PackageVersion Include="Silk.NET.OpenGL" Version="2.21.0" />
-		<PackageVersion Include="Silk.NET.WGL.Extensions.NV" Version="2.21.0" />
+		<PackageVersion Include="Silk.NET.OpenAL" Version="2.23.0" />
+		<PackageVersion Include="Silk.NET.OpenAL.Extensions.Creative" Version="2.23.0" />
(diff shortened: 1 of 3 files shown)


YoshiRulz:
I think the only thing that having main solution libs target `netstandard2.0` does is that it prevents accidentally using APIs which wouldn't be available in .NET Core. We could use Microsoft.CodeAnalysis.BannedApiAnalyzers to block those instead.

edit: As a proxy for "what APIs are we using", here are all the imported namespaces in the 4 `net48` projects prior to this PR:
<details>
<summary>imports</summary>

```
using System;
using System.Buffers;
using System.Buffers.Binary;
using System.Collections;
using System.Collections.Concurrent;
using System.Collections.Generic;
using System.Collections.Immutable;
using System.Collections.ObjectModel;
using System.Collections.Specialized;
using System.ComponentModel;
using System.Data;
using System.Diagnostics;
using System.Diagnostics.CodeAnalysis;
using System.Drawing;
using System.Drawing.Drawing2D;
using System.Drawing.Imaging;
using System.Globalization;
using System.IO;
using System.IO.Compression;
using System.IO.Pipes;
using System.Linq;
using System.Net;
using System.Net.Http;
using System.Numerics;
using System.Reflection;
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;
using System.Runtime.Versioning;
using System.Security.AccessControl;
using System.Security.Principal;
using System.Text;
using System.Text.RegularExpressions;
using System.Threading;
using System.Threading.Tasks;
using System.Windows;
using System.Windows.Forms;
using System.Windows.Forms.Automation;
using System.Windows.Shell;
using System.Xml.Linq;
using System.Xml.XPath;

using Newtonsoft.Json;
using Newtonsoft.Json.Linq;
using NLua;
using NLua.Native;
using PcxFileTypePlugin.Quantize;
```
</details>

Pasting that into Client.Common (again, before this PR) results in only `System.Windows.Forms` and `System.Windows.Shell` being flagged, as expected, plus `PcxFileTypePlugin` which isn't referenced but is .NET Standard anyway.

Morilli:
Are there any relevant apis that only apply to net framework and not .NET (core)? If there are it's probably fine to disallow them, I'm not against using netstandard in places where it doesn't result in issues.

SuuperW:
Looks good to me.

Issue #4012, opened 2024-08-23
Update/add more default input mappings


TakuikaNinja:
This suggestion comes from the discussion in #3842. 
> The default controller mappings in general aren't updated often, so that might be doable here.

An especially glaring issue is that cores for computer systems (C64, ZX Spectrum, MSX, etc) lack default mappings for their keyboards. It would be a good idea to create sensible defaults for these, along with updating the rest of the mappings.

Asnivor:
I was sure I included default mappings for ZXS. Looks like they were all removed to fix this (or another) issue:
https://github.com/TASEmulators/BizHawk/commit/ec3a23259cf0f9528aef661b9df51846997485f4

If the key priority thing is now properly fixed, someone should be able to add the defaults back in maybe?

YoshiRulz:
Ah, I was looking for https://github.com/TASEmulators/BizHawk/issues/3257#issuecomment-1786291002. To summarise my opinions:
- I think we *shouldn't* have any default hotkey keybinds which are just a single key. (Maybe some standard things like F1, Alt alone, or Pause.) But I don't think it's a good idea to change those unless it's as part of a complete overhaul of the defaults.
	- edit 2025-09: The case against single-key binds is simply that it's too easy to accidentally trigger a hotkey (either by hitting the wrong key, or by beginning to type while the wrong window is focused) which then takes more effort to undo than it took to do. Save and load state are the main ones (see #2175).
		- If the hotkey priority isn't set to `InputPriority.INPUT` and you have a virtual Shift or Ctrl key bound to Shift or Ctrl, the number of hotkey actions you could accidentally trigger is much higher.
- Similarly, I think mapping most of the host keyboard to the virtual keyboard is a bad out-of-the-box experience. If only because the default keybinds *must not have conflicts*. I still wouldn't like it if all the hotkeys used modifiers though.
	- Earlier, I suggested we could have some sort of preset to populate the keyboard binds for those who wanted that immersive experience. See also #1636.
- Modifier keys are inherently different: If one is mapped to a virtual key (as was the default for C64 before the above commit), trying to input e.g. Shift+R on the *virtual* keyboard may instead trigger another binding (e.g. the Rewind hotkey). Thankfully we don't seem to have any virtual buttons bound to Ctrl or Shift for any system at the moment.
	- Perhaps if a modifier key alone is bound to anything, a message should appear at the bottom the keybinds dialog explaining how to use the bind priority toggle?

[...]

PR #4626, opened 2026-02-11
Bulk multibyte

SuuperW:
To enable bulk poke/peek access for 16 and 32 bit values, I have given `MemoryDomain` , `IMemoryApi`, and `MemoryLuaLibrary` some new functions.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
+++ b/src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
@@ -25,9 +25,16 @@ namespace BizHawk.Client.Common
 		int ReadS32(long addr, string domain = null);
 
 		uint ReadU8(long addr, string domain = null);
+		void BulkReadU8(long addr, Span<byte> dst, string domain = null);
+
 		uint ReadU16(long addr, string domain = null);
+		void BulkReadU16(long addr, Span<ushort> dst, string domain = null);
+
 		uint ReadU24(long addr, string domain = null);
+
 		uint ReadU32(long addr, string domain = null);
+		void BulkReadU32(long addr, Span<uint> dst, string domain = null);
+
 
 		void WriteByte(long addr, uint value, string domain = null);
 		void WriteByteRange(long addr, IReadOnlyList<byte> memoryblock, string domain = null);
@@ -39,8 +46,14 @@ namespace BizHawk.Client.Common
 		void WriteS32(long addr, int value, string domain = null);
 
 		void WriteU8(long addr, uint value, string domain = null);
+		void BulkWriteU8(long addr, ReadOnlySpan<byte> values, string domain = null);
+
 		void WriteU16(long addr, uint value, string domain = null);
+		void BulkWriteU16(long addr, ReadOnlySpan<ushort> values, string domain = null);
+
 		void WriteU24(long addr, uint value, string domain = null);
+
 		void WriteU32(long addr, uint value, string domain = null);
+		void BulkWriteU32(long addr, ReadOnlySpan<uint> values, string domain = null);
 	}
 }
--- a/src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
+++ b/src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
@@ -121,6 +121,43 @@ namespace BizHawk.Emulation.Common
 			}
 		}
 
+		public virtual void BulkPokeByte(long startAddress, ReadOnlySpan<byte> values)
+		{
+			long addr = startAddress;
+			using (this.EnterExit())
+			{
+				for (int i = 0; i < values.Length; i++, addr += sizeof(byte))
+				{
+					PokeByte(addr, values[i]);
+				}
+			}
+		}
+
+		public virtual void BulkPokeUshort(long startAddress, ReadOnlySpan<ushort> values, bool bigEndian)
+		{
+			long addr = startAddress;
+			using (this.EnterExit())
+			{
+				for (int i = 0; i < values.Length; i++, addr += sizeof(ushort))
+				{
+					PokeUshort(addr, values[i], bigEndian);
+				}
+			}
+		}
+
+		public virtual void BulkPokeUint(long startAddress, ReadOnlySpan<uint> values, bool bigEndian)
+		{
+			long addr = startAddress;
+			using (this.EnterExit())
+			{
+				for (int i = 0; i < values.Length; i++, addr += sizeof(uint))
+				{
+					PokeUint(addr, values[i], bigEndian);
+				}
+			}
+		}
+
+
 		public virtual void SendCheatToCore(int addr, byte value, int compare, int compare_type) { }
 
 		/// <summary>
--- a/src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
+++ b/src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
@@ -20,8 +20,6 @@ namespace BizHawk.Emulation.Common
 			Unknown,
 		}
 
-		private const string ERR_MSG_BUFFER_WRONG_SIZE = "Invalid length of values array";
-
 		public string Name { get; protected set; }
 
 		public long Size { get; protected set; }
@@ -87,53 +85,38 @@ namespace BizHawk.Emulation.Common
 			PokeByte(addr + 3, scratch[3]);
 		}
 
-		public virtual void BulkPeekByte(Range<long> addresses, byte[] values)
+		public virtual void BulkPeekByte(long startAddress, Span<byte> values)
 		{
-			if (addresses is null) throw new ArgumentNullException(paramName: nameof(addresses));
-			if (values is null) throw new ArgumentNullException(paramName: nameof(values));
-			if ((long) addresses.Count() > values.Length) throw new InvalidOperationException(ERR_MSG_BUFFER_WRONG_SIZE);
-
+			long addr = startAddress;
 			using (this.EnterExit())
 			{
-				for (var i = addresses.Start; i <= addresses.EndInclusive; i++)
+				for (int i = 0; i < values.Length; i++, addr += sizeof(byte))
 				{
-					values[i - addresses.Start] = PeekByte(i);
+					values[i] = PeekByte(addr);
 				}
 			}
 		}
 
-		public virtual void BulkPeekUshort(Range<long> addresses, bool bigEndian, ushort[] values)
+		public virtual void BulkPeekUshort(long startAddress, Span<ushort> values, bool bigEndian)
 		{
-			if (addresses is null) throw new ArgumentNullException(paramName: nameof(addresses));
-			if (values is null) throw new ArgumentNullException(paramName: nameof(values));
-
-			var start = addresses.Start;
-			var end = addresses.EndInclusive + 1;
-			if (values.LongLength * sizeof(ushort) < end - start) throw new InvalidOperationException(ERR_MSG_BUFFER_WRONG_SIZE);
-
+			long addr = startAddress;
 			using (this.EnterExit())
 			{
-				for (var i = 0; i < values.Length; i++, start += sizeof(ushort))
+				for (int i = 0; i < values.Length; i++, addr += sizeof(ushort))
 				{
-					values[i] = PeekUshort(start, bigEndian);
+					values[i] = PeekUshort(addr, bigEndian);
 				}
 			}
 		}
 
-		public virtual void BulkPeekUint(Range<long> addresses, bool bigEndian, uint[] values)
+		public virtual void BulkPeekUint(long startAddress, Span<uint> values, bool bigEndian)
 		{
-			if (addresses is null) throw new ArgumentNullException(paramName: nameof(addresses));
-			if (values is null) throw new ArgumentNullException(paramName: nameof(values));
-
-			var start = addresses.Start;
-			var end = addresses.EndInclusive + 1;
-			if (values.LongLength * sizeof(uint) < end - start) throw new InvalidOperationException(ERR_MSG_BUFFER_WRONG_SIZE);
-
+			long addr = startAddress;
 			using (this.EnterExit())
 			{
-				for (var i = 0; i < values.Length; i++, start += sizeof(uint))
+				for (int i = 0; i < values.Length; i++, addr += sizeof(uint))
 				{
-					values[i] = PeekUint(start, bigEndian);
+					values[i] = PeekUint(addr, bigEndian);
 				}
 			}
 		}
--- a/src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
+++ b/src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
@@ -22,8 +22,6 @@ namespace BizHawk.Emulation.Common
 
 		private const string ERR_MSG_BUFFER_WRONG_SIZE = "Invalid length of values array";
 
-		private const string ERR_MSG_UNALIGNED = "The API contract doesn't define what to do for unaligned reads and writes!";
-
 		public string Name { get; protected set; }
 
 		public long Size { get; protected set; }
@@ -93,7 +91,7 @@ namespace BizHawk.Emulation.Common
 		{
 			if (addresses is null) throw new ArgumentNullException(paramName: nameof(addresses));
 			if (values is null) throw new ArgumentNullException(paramName: nameof(values));
-			if ((long) addresses.Count() != values.Length) throw new InvalidOperationException(ERR_MSG_BUFFER_WRONG_SIZE);
+			if ((long) addresses.Count() > values.Length) throw new InvalidOperationException(ERR_MSG_BUFFER_WRONG_SIZE);
 
 			using (this.EnterExit())
 			{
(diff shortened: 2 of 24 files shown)


src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
```diff
@@ -33,6 +33,11 @@ public interface IMemoryApi : IExternalApi
 		void WriteByteRange(long addr, IReadOnlyList<byte> memoryblock, string domain = null);
 		void WriteFloat(long addr, float value, string domain = null);
 
+		ushort[] ReadU16Range(long addr, int count, string domain = null);
+		uint[] ReadU32Range(long addr, int count, string domain = null);
+		void WriteU16Range(long addr, Span<ushort> memoryblock, string domain = null);
+		void WriteU32Range(long addr, Span<uint> memoryblock, string domain = null);
```
YoshiRulz: `Write*Range` signatures are fine for `Span`, but then why do `Read*Range` allocate and return arrays?
See also #3472.
Not that you need these anyway, since the caller can just `MemoryMarshal.Cast` to/from `byte`.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
```diff
@@ -33,6 +33,11 @@ public interface IMemoryApi : IExternalApi
 		void WriteByteRange(long addr, IReadOnlyList<byte> memoryblock, string domain = null);
 		void WriteFloat(long addr, float value, string domain = null);
 
+		ushort[] ReadU16Range(long addr, int count, string domain = null);
```
YoshiRulz: Also these need to be explicit about endianness. The new Lua functions are labelled as little-endian, so put that in a doc comment at least.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
```diff
@@ -104,27 +101,46 @@ public virtual void BulkPeekByte(Range<long> addresses, byte[] values)
 			}
 		}
 
+		public virtual void BulkPokeByte(long startAddress, Span<byte> values)
```
YoshiRulz: ~~New method which isn't used?~~ At least implement it as a single `BulkPokeUint` call + 0..<4 `PokeByte` calls, ~~and change the array overload to call the `Span` overload.~~

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
```diff
@@ -104,27 +101,46 @@ public virtual void BulkPeekByte(Range<long> addresses, byte[] values)
...
 			if (values is null) throw new ArgumentNullException(paramName: nameof(values));
 
-			var start = addresses.Start;
-			var end = addresses.EndInclusive + 1;
-
-			if ((start & 1) != 0 || (end & 1) != 0)
-				throw new InvalidOperationException("The API contract doesn't define what to do for unaligned reads and writes!");
+			if (addresses.Count() > (uint)values.Length * sizeof(ushort)) throw new ArgumentException($"Length of {nameof(values)} must be at least {nameof(addresses)} count.", nameof(values));
 
-			if (values.LongLength * 2 != end - start)
```
YoshiRulz: Hard to tell, but there are at least 2 behavioural changes in this diff: oversized buffers are now acceptable, and unaligned reads are now acceptable. The latter especially I worry about.
(If you rebase now I think the diff will be a bit cleaner.)

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
```diff
@@ -104,27 +101,46 @@ public virtual void BulkPeekByte(Range<long> addresses, byte[] values)
...
-				throw new InvalidOperationException("Invalid length of values array");
+				long address = addresses.Start;
+				for (var i = 0; i < values.Length; i++, address += sizeof(ushort))
+					values[i] = PeekUshort(address, bigEndian);
 			}
+		}
+
+		public virtual void BulkPokeUshort(long startAddress, bool bigEndian, Span<ushort> values)
+		{
+			if (values == null) throw new ArgumentNullException(paramName: nameof(values));
```
YoshiRulz: A `ref struct` cannot be null... how is this not flagged

YoshiRulz (changes requested):
On top of cleaning up `IMemoryApi`, I would also like to clean up `Range<T>` (#3965) before adding new uses to `MemoryDomain`.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
SuuperW: I think it was just a lot simpler to do it that way, and the byte one returns an array IIRC so I was following that.
I knew you were wanting to update some of this stuff anyway, so I didn't think it very important.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
SuuperW: I find the whole endianness handling we have to be overly complicated, contradictory, and very confusing. Why do memory domains define an endianness, and then we don't use that defined endianness?

SuuperW:
Regarding all of your comments about being able to just convert between bulk byte and bulk int reads: No. The only reason I made this is to support #4555. The core MUST know what size we are reading, and the various sized reads cannot simply be re-interpreted as another.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
SuuperW: Why do you worry about unaligned reads? They have defined behavior and are already supported by the single 16 and 32-bit operations.

Why would accepting oversize buffers be bad? A tool may wish to re-use a buffer for multiple reads, for efficiency. (Although I suppose since we are using a `Span` here instead of array the user could very easily just change the size given.)

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
YoshiRulz: The domain has an endianness property so tools like the Hex Editor can show 'words' in the expected form. I believe it's also the default when an endianness argument isn't specified in an API call.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
YoshiRulz: The behaviour around unaligned access isn't documented, and existing cores might have assumed alignment is checked.

I'm on board with allowing oversized buffers specifically for arrays when the size (here, range) is passed in, I was just highlighting the change.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
SuuperW: The memory API does not appear to care about the domain's endianness, and does not even support specifying the endianness in poke or peek calls, only with `SetBigEndian`, and the default endianness is little regardless of the domain. (This is just from looking at the code.)
And the Hex Editor should not have to specify the endianness that is default. IMO, a domain should only support its default endianness and it should be up to the user to swap bytes if needed. (The memory API could be such a user, but I think even that really shouldn't handle endianness because it introduces unnecessary complexity for something that would be rarely used and is trivial for users who do want it to self-implement.) ...but that's really outside the scope of this PR.

`ReadU16Range` should not be explicit about endianness in the call signature because `ReadU16` is not explicit about endianness in the call signature. Both end up using `_isBigEndian` when accessing the domain.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
SuuperW: The behavior of unaligned access (even the definition of unaligned access, or if such a thing even exists) isn't documented at this level because it cannot be. It is defined by the individual cores. There are systems where unaligned access can happen internally, and we should not prohibit users from doing it manually.

If cores assume alignment is checked, they are already wrong. Alignment is never checked for non-bulk operations. And like as I said above, such an assumption would be incorrect in that is should not be assumed.

Of course, there is no way to perform unaligned access without external tools or Lua scripts. So there won't be any issues with say Hex Editor.

vadosnaprimer:
Conflicts appeared, but also it's unclear from the discussion if there's anything else that has to change still.

SuuperW:
I was going to wait for Yoshi to finish #3965 but I can go ahead and fix some stuff here in a few days.

vadosnaprimer:
> I was going to wait for Yoshi to finish #3965

That one looks like it will require 20 more years to get ~~done~~ abandoned forever.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
SuuperW: I did a little testing. Assigning `null` to a `Span<byte>` seems to be equivalent to assigning `Span<byte>.Empty`, and checking for `null` is similarly equivalent to checking `Length == 0`.

This is confusing imo, and not what I intended. Passing in a length zero span shouldn't be an issue, so I will remove those checks.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
SuuperW: Update on this: The new read methods return `IReadOnlyList`, so as to match the existing `ReadByteRange` method. I considered changing them all to accept a `Span` instead but that's too much work to do before release.
The existing `WriteByteRange` method was changed to accept a `Span` though, since this is good and easy to do and makes it match the new ones.

SuuperW:
Fixed bugs and added tests for the bulk API methods. That code is absolutely horrible, matching the existing bulk byte methods.

`IMemoryApi.WriteByteRange` was changed from accepting a `IReadOnlyList<byte>` to `Span<byte>` which is technically an API change but `byte[]` works for both without an explicit cast. That lets it use the new underlying `MemoryDomain.BulkPokeByte`, and it also what the new bulk poke API methods do. If we need the version that accepts a list, I can make overloads for them.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
YoshiRulz: Do not change `IReadOnlyList<byte>` to `Span<byte>` in `WriteByteRange`'s signature.

The added methods are redundant, but if you're going to add them, don't use `Span` there either.

src/BizHawk.Tests.Client.Common/Api/MemoryApiTests.cs
```diff
@@ -99,6 +99,198 @@ void TestCase(IReadOnlyList<byte> expected, Action<IMemoryApi> action, string me
...
+		public void TestBulkPeekUshort()
+		{
+			ushort[] memContents = new ushort[] { 0x0123, 0x4567, 0x89AB, 0xCDEF };
+			byte[] memBytes = new byte[8];
+			unsafe
+			{
+				fixed (byte* dst = memBytes)
+				{
+					ushort* dstUshort = (ushort*)dst;
+					for (int i = 0; i < memContents.Length; i++) dstUshort[i] = memContents[i];
```
YoshiRulz: ??? You do not need `unsafe` for this. `memContents.AsBytes().CopyTo(memBytes)`
edit: How is *this* not a build error? `$(AllowUnsafeBlocks)` doesn't seem to have any effect anymore, even when explicitly set to `false`?

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
SuuperW: > Do not change IReadOnlyList<byte> to Span<byte> in WriteByteRange's signature.

Why not? Because it changes the API for external tools?

> The added methods are redundant

PLEASE, quit saying this. I have already explained to you that writing two adjacent 8-bit integers is not always equivalent to writing a 16-bit integer.

> don't use Span there either.

Why? This isn't even an API change, and `Span` can be more efficient.

src/BizHawk.Emulation.Common/Base Implementations/MemoryDomain.cs
SuuperW: > New method which isn't used?

It's used in `MemoryApi.WriteByteRange`.

> change the array overload to call the Span overload.

What are you referring to? `MemoryDomain` does not contain any other overloads of `BulkPokeByte`

src/BizHawk.Tests.Client.Common/Api/MemoryApiTests.cs
SuuperW: Where are you getting `AsBytes` from? That method does not seem to exist.

Is `AllowUnsafeBlocks` set to false anywhere? I'm not seeing that either.

Morilli: You could use `memContents.BytesSpan().CopyTo(memBytes)` here.

SuuperW: Changed to using `BytesSpan` for copying.

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
YoshiRulz: >> The added methods are redundant
>
> PLEASE, quit saying this. I have already explained to you that writing two adjacent 8-bit integers is not always equivalent to writing a 16-bit integer.

I thought your comment below was in reference to cores' implementations.
*Poking* two adjacent 8-bit integers in *virtual/logical* memory *is* always equivalent to poking one 16-bit integer (assuming the endianness matches). If it's not, that can't be considered poking, and the core needs to be fixed.

>> Do not change IReadOnlyList to Span in WriteByteRange's signature.
>
> Why not? Because it changes the API for external tools?

>> don't use Span there either.
>
> Why? This isn't even an API change, and `Span` can be more efficient.

There's a reason we're not already using `Span` everywhere, and that's because it's not always obvious what design and signatures should be used. Which is because there are many related types, some struct and some class, each with their own quirks and limitations. I don't pretend to be an expert in them, and clearly neither are you since you used the mutable `Span` here with no intention of writing to it.
As I have said [before](https://github.com/TASEmulators/BizHawk/issues/4012#issuecomment-3304144913), making repeated changes to the interfaces that people use (in this case an API) is what you might call "churn" and should be avoided. Accepting a `Span` as input only to change it to e.g. `ReadOnlyMemory` later is annoying. Accepting a `Span` as input only to do a bunch of copying anyway is deceptive.

src/BizHawk.Tests.Client.Common/Api/MemoryApiTests.cs
YoshiRulz: To clarify:
https://github.com/TASEmulators/BizHawk/blob/16319296f6ebdf8dca8429b0271346fc85569624/src/BizHawk.Common/Extensions/CollectionExtensions.cs#L200-L202

And `$(AllowUnsafeBlocks)` isn't set for this project so it defaults to `false`, but as I said not even an explicit `false` has any effect. Maybe something else is overriding it? `Meziantou.Polyfill`?

Morilli: > Maybe something else is overriding it? Meziantou.Polyfill?

Looks like it: https://github.com/meziantou/Meziantou.Polyfill/blob/cf764cf2a905d7c45782051fc21b8b750aab8245/Meziantou.Polyfill/Meziantou.Polyfill.targets#L7-L8

src/BizHawk.Client.Common/Api/Interfaces/IMemoryApi.cs
SuuperW: > I thought your comment below was in reference to cores' implementations.

And how would a core who's implementation differentiates between two adjacent 8-bit writes and one 16-bit write be able to make that distinction for bulk operations if the only bulk operation hands everything over as bytes? Or would you want the MemoryApi methods to call `PokeUshort` in a loop? That would prevent the core from optimizing bulk operations.

> virtual/logical memory

Not all memory addresses refer to virtual or logical memory. melonDS has addresses which map to registers, and even has some that behave more like streams than traditional RAM. (Writing 1 then 0 is not the same as writing just 0, and reading a value can cause things to happen including causing the next read to be different.) Saying "change/poke the value at this address" is meaningless because there is no value at that address. The only thing you can do is write to that address or read from it. Unless you create a new API for handling these things, the existing memory API must support it.

> There's a reason we're not already using Span everywhere, and that's because it's not always obvious what design and signatures should be used.

And all of this somehow does not apply to `ReadOnlyList`? There is is somehow obvious?

> Accepting a Span as input only to do a bunch of copying anyway is deceptive.

I disagree. There's nothing indicating an intent to write to the span. But changing it to a `ReadOnlySpan` is a good idea anyway.

> As I have said https://github.com/TASEmulators/BizHawk/issues/4012#issuecomment-3304144913, making repeated changes to the interfaces that people use (in this case an API) is what you might call "churn" and should be avoided.

The comment you linked does not mention "churn" and I don't see how changing keybinds is related to this. I do agree changes should be kept to a minimum, but that doesn't mean we should not use a span for the NEW methods.

YoshiRulz: >> *Poking* two adjacent 8-bit integers in *virtual/logical* memory *is* always equivalent to poking one 16-bit integer (assuming the endianness matches). If it's not, that can't be considered poking, and the core needs to be fixed.
>
> Not all memory addresses refer to virtual or logical memory. melonDS has addresses which map to registers, and even has some that behave more like streams than traditional RAM. (Writing 1 then 0 is not the same as writing just 0, and reading a value can cause things to happen including causing the next read to be different.) Saying "change/poke the value at this address" is meaningless because there is no value at that address. The only thing you can do is write to that address or read from it. Unless you create a new API for handling these things, the existing memory API must support it.

What you're describing with register-like addresses on the system bus is clearly not poking. The domain(s) with that behaviour should be `Writable: false`, or maybe throw an exception with those specific addresses (not sure if there's precedence for that).

SuuperW:
Updated so that existing methods in `IMemoryApi` are not changed. New methods are added instead. Tests are updated accordingly, and I have also added some basic tests for the Lua API.

The failing builds are for .NET Core where MA0089 wants me to use an overload that does not exist in .NET Framework. Perhaps we should disable this rule. I don't know if it can be disabled only for .NET Core builds.

Issue #4629, opened 2026-02-15
Crash when rebooting emulator core


SuuperW:
The following Lua script crashes EmuHawk when rebooting the core:
```
event.onexit(function()
	memory.read_u8(0)
end)

while true do
	emu.frameadvance()
end
```

This happens because the script is not stopped until `LuaConsole.Restart`, at which point the old core instance has been disposed. (Of course, even if the exit handler read from the new core, that would not be correct either.)

### Repro
1. Open a ROM with a core that supports memory API
2. Start the above Lua script
3. Reload the ROM

### Output
No error message.

### Host env.
BizHawk 2.11 and current master

YoshiRulz:
X&ndash;Y problem.
Implementors of `IToolForm.Restart` are expected to use the updated values that have just been provided...
https://github.com/TASEmulators/BizHawk/blob/c8af4157a24ec3daf6385a4329dfa5a9244c4344/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs#L604-L608
...which includes `[RequiredService] IEmulator Emulator`. A separate `Stop` method would be redundant.
Some Lua callbacks, including `onexit`, only support a subset of the API. When a script attempts to call an unsupported function like peeking, we should terminate the callback (or the script, but for an `onexit` handler there's little difference) and print an error to the Lua Console.

SuuperW:
> Some Lua callbacks, including onexit, only support a subset of the API.

Source? Just because they happen to not work while the core is rebooting, doesn't mean they are not intended to work when a script is stopping. Memory reads and writes do work when a script is stopping, so long as the core hasn't been destroyed.

And saying they should not be used is unacceptable. Scripts that are used as cheats must be able to do RAM pokes that disable the cheat (restore original data or code) when they exit.

> Implementors of IToolForm.Restart are expected to use the updated values that have just been provided ... A separate Stop method would be redundant.

I think you do not understand the issue. Reading or poking with the newly created emulator core would be incorrect. Shutdown logic (the onexit handler) should not run on a just-created core, but on the core it was intended to run on. This is impossible to do if that core has been destroyed, thus the need for stopping the script prior to core disposal. Such shutdown logic may not be necessary in the case of a reboot, but as-is the user cannot handle that.

[...]

PR #4630, opened 2026-02-15
Stop tools before rebooting.

SuuperW:
See #4629

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/tools/Interfaces/IToolForm.cs
+++ b/src/BizHawk.Client.Common/tools/Interfaces/IToolForm.cs
@@ -28,6 +28,12 @@
 		void UpdateValues(ToolFormUpdateType type);
 
 		/// <summary>
+		/// Stop the tool, in preparation for a game or core change.
+		/// A tool should expect that either <see cref="Restart"/> or <see cref="Close"/> will be called after this.
+		/// </summary>
+		void Stop();
+
+		/// <summary>
 		/// Will be called anytime the dialog needs to be restarted, such as when a new ROM is loaded
 		/// The tool implementing this needs to account for a Game and Core change
 		/// </summary>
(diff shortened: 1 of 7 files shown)


YoshiRulz (changes requested):
For the avoidance of doubt, here's an elaboration of the comment I previously linked to. (The rest of the discussion regards Lua and is not relevant here.)

Implementors of `IToolForm.Restart` are expected to use the updated values that have just been provided:
https://github.com/TASEmulators/BizHawk/blob/c8af4157a24ec3daf6385a4329dfa5a9244c4344/src/BizHawk.Client.EmuHawk/tools/ToolManager.cs#L604-L608
If a tool needed to unhook from a core in some way, it can do that in for example the setter method of `[RequiredService] IEmulator _maybeCore`. A separate `Stop` method would be redundant.
You're trying to sculpt the tools' lifecycle, modelled on WinForms', into something that resembles a Lua script's lifecycle. That's what I mean by "the wrong approach".

SuuperW:
> If a tool needed to unhook from a core in some way, it can do that in for example the setter method of [RequiredService] IEmulator _maybeCore. A separate Stop method would be redundant.

Wouldn't the old core already be disposed by the time that setter is called? (This would introduce additional complications if we have multiple services, but that's really an issue with using reflection which is another thing I intend to eventually fix.)
`ToolManager.Restart` does `ServiceInjector.UpdateServices(_emulator.ServiceProvider, tool)` then immediately after calls the tool's `Restart` method. So anything that can be done from a service property's setter could also be done in `Restart` (assuming the tool hasn't lost its reference to the old service). If we could access the old core from `Restart`, then there would not be any need for a `Stop` method.

> You're trying to sculpt the tools' lifecycle, modelled on WinForms', into something that resembles a Lua script's lifecycle. 

No I'm not. The tool's lifecycle isn't relevant here. We care about the core being disposed, not the tool.

PR #4635, opened 2026-02-21
Add unit tests for `ArgParser` and improve output of `--help`

YoshiRulz:
From https://github.com/TASEmulators/BizHawk/commit/5c73d5b4178a6c637a7351f60fac39f058df995e#r177008277.

If this is merged, notify dotnet/command-line-api issue 2765.

### DIFF
--- a/src/BizHawk.Client.Common/ArgParser.cs
+++ b/src/BizHawk.Client.Common/ArgParser.cs
@@ -2,7 +2,10 @@
 
 using System.Collections.Generic;
 using System.CommandLine;
+using System.CommandLine.Help;
+using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
+using System.IO;
 using System.Linq;
 using System.Net.Sockets;
 
@@ -15,9 +18,30 @@ namespace BizHawk.Client.Common
 	/// <summary>Parses command-line flags into a <see cref="ParsedCLIFlags"/> struct.</summary>
 	public static class ArgParser
 	{
+		private sealed class HelpSedAction(HelpAction stockAction) : SynchronousCommandLineAction
+		{
+			public override int Invoke(ParseResult parseResult)
+			{
+				var outerOutput = parseResult.InvocationConfiguration.Output;
+				using StringWriter innerOutput = new();
+				parseResult.InvocationConfiguration.Output = innerOutput;
+				var result = stockAction.Invoke(parseResult);
+				var dollarZero = OSTailoredCode.IsUnixHost
+#if true
+					? "./EmuHawkMono.sh"
+#else //TODO for .NET Core: see https://github.com/dotnet/runtime/issues/101837
+					? Environment.GetCommandLineArgs()[0].SubstringAfterLast('/')
+#endif
+					: Environment.GetCommandLineArgs()[0].SubstringAfterLast('\\');
+				outerOutput.Write(innerOutput.ToString().Replace("EmuHawk", dollarZero)
+					.Replace("[<rom>...] [options]", "[option...] [rom]"));
+				return result;
+			}
+		}
+
 		private static readonly Argument<string[]> ArgumentRomFilePath = new("rom")
 		{
-			Description = "path; if specified, the file will be loaded the same way as it would be from `File` > `Open...`; this argument can and should be given LAST despite what it says at the top of --help",
+			Description = "path; if specified, the file will be loaded the same way as it would be from `File` > `Open...`",
 		};
 
 		private static readonly Option<string?> OptionAVDumpAudioSync = new("--audiosync")
@@ -134,6 +158,8 @@ namespace BizHawk.Client.Common
 				(string.IsNullOrEmpty(VersionInfo.CustomBuildString) ? "EmuHawk" : VersionInfo.CustomBuildString)
 			}, a multi-system emulator frontend\n{VersionInfo.GetEmuVersion()}");
 			root.Options.RemoveAll(option => option is VersionOption); // we have our own version command
+			var helpOption = root.Options.OfType<HelpOption>().First();
+			helpOption.Action = new HelpSedAction((HelpAction) helpOption.Action!);
 
 			// `--help` uses this order, so keep alphabetised by flag
 			root.Add(/* --audiosync */ OptionAVDumpAudioSync);
@@ -278,6 +304,13 @@ namespace BizHawk.Client.Common
 				userdataUnparsedPairs: userdataUnparsedPairs,
 				cmdRom: unmatchedArguments.LastOrDefault()); // `unmatchedArguments.Length` must be 0 or 1 at this point, but in case we change how that's handled, 'last' here preserves the behaviour from the old hand-rolled parser
 			return null;
+		}
+
+		internal static void RunHelpActionForUnitTest(TextWriter output)
+		{
+			var result = CommandLineParser.Parse(GetRootCommand(), [ "--help" ]);
+			result.InvocationConfiguration.Output = output;
+			result.Invoke();
 		}
 
 		public sealed class ArgParserException : Exception
--- a/src/BizHawk.Tests.Client.Common/ArgParserTests.cs
+++ b/src/BizHawk.Tests.Client.Common/ArgParserTests.cs
@@ -1,3 +1,6 @@
+using System.IO;
+using System.Linq;
+
 using BizHawk.Client.Common;
 
 namespace BizHawk.Tests.Client.Common
@@ -10,6 +13,16 @@ namespace BizHawk.Tests.Client.Common
 		[TestMethod]
 		public void TestConfigWithHelpOrVersion(params string[] args)
 			=> Assert.AreEqual(0, ArgParser.ParseArguments(out _, args, fromUnitTest: true));
+
+		[TestMethod]
+		public void TestHelpSaysPassFlagsFirst()
+		{
+			using StringWriter output = new();
+			ArgParser.RunHelpActionForUnitTest(output);
+			var outputLines = output.ToString().Split('\n');
+			var usageLine = outputLines[outputLines.Index().First(tuple => tuple.Item.Contains("Usage:")).Index + 1].ToUpperInvariant();
+			Assert.IsTrue(usageLine.IndexOf("OPTION") < usageLine.IndexOf("ROM"));
+		}
 
 		[DataRow("rom.nes", "--nonexistent")]
 		[DataRow("--nonexistent", "rom.nes")]


src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -177,34 +202,43 @@ private static void EnsureConsole()
 
 		/// <return>exit code, or <see langword="null"/> if should not exit</return>
 		/// <exception cref="ArgParserException">parsing failure, or invariant broken</exception>
-		public static int? ParseArguments(out ParsedCLIFlags parsed, string[] args)
+		public static int? ParseArguments(out ParsedCLIFlags parsed, string[] args, bool fromUnitTest = false)
```
SuuperW, line 205: Why did you add this parameter? None of your various uses are explained. The one making it always return 0 seems to make the unit test pointless.

src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -177,34 +202,43 @@ private static void EnsureConsole()
...
 				return 0;
 			}
 
+			var unmatchedArguments = result.GetValue(ArgumentRomFilePath) ?? [ ];
+			if (unmatchedArguments.Length >= 2)
+			{
+				var foundFlagLike = unmatchedArguments.FirstOrDefault(static s => s.StartsWith("--"));
+				throw new ArgParserException(foundFlagLike is null
+					? "Multiple rom paths provided. (Did you delete half of a flag? Or forget the \"--\" of one?)"
+					: $"Unrecognised flag(s): \"{foundFlagLike}\"");
```
SuuperW, line 239: This only prints one unrecognised flag.

src/BizHawk.Tests.Client.Common/ArgParserTests.cs
```diff
@@ -0,0 +1,40 @@
...
+			=> Assert.AreEqual(0, ArgParser.ParseArguments(out _, args, fromUnitTest: true));
+
+		[TestMethod]
+		public void TestHelpSaysPassFlagsFirst()
+		{
+			using StringWriter output = new();
+			ArgParser.RunHelpActionForUnitTest(output);
+			var outputLines = output.ToString().Split('\n');
+			var usageLine = outputLines[outputLines.Index().First(tuple => tuple.Item.Contains("Usage:")).Index + 1].ToUpperInvariant();
+			Assert.IsTrue(usageLine.IndexOf("OPTION") < usageLine.IndexOf("ROM"));
```
SuuperW, line 24: Why do you want this? The parser works with options and rom given in any order, even mixed.

src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -268,11 +302,17 @@ private static void EnsureConsole()
 				openExtToolDll: result.GetValue(OptionOpenExternalTool),
 				socketProtocol: result.GetValue(OptionSocketServerUseUDP) ? ProtocolType.Udp : ProtocolType.Tcp,
 				userdataUnparsedPairs: userdataUnparsedPairs,
-				cmdRom: result.GetValue(ArgumentRomFilePath)
-			);
+				cmdRom: unmatchedArguments.LastOrDefault()); // `unmatchedArguments.Length` must be 0 or 1 at this point, but in case we change how that's handled, 'last' here preserves the behaviour from the old hand-rolled parser
 			return null;
 		}
 
+		internal static void RunHelpActionForUnitTest(TextWriter output)
```
SuuperW, line 309: I think this would be better as a more generic method, that would support any future tests of other arguments.

SuuperW (commented):
The arg parser code would benefit from better documentation. Even the code before your changes has some parts that don't make any sense, like sending all the args to stderr. Even what does make sense after careful reading would still benefit from some comments. My being unfamiliar with `System.CommandLine` makes it worse.

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: I want to avoid writing to the console during unit tests so there's less noise to sift through.
I hardcoded an exit code of `0` for `--help` since that's what `Invoke` returns. Not that that's being asserted, only the text.

src/BizHawk.Tests.Client.Common/ArgParserTests.cs
YoshiRulz: See https://github.com/TASEmulators/BizHawk/commit/5c73d5b4178a6c637a7351f60fac39f058df995e#r177008277 for context. (GitHub is weird and won't show the thread until you expand the context of the first diff hunk.)

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: I could pass in `[ "--help" ]` from the unit test, but the only flags where we need to intercept and check what's being written to console are that and `--version`, and I don't see a reason to test `--version`.

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: Oops, fixed

YoshiRulz:
If you're asking about this line
https://github.com/TASEmulators/BizHawk/blob/2a9a11180a3bf2c1552a816c2a14c42a3e61da7f/src/BizHawk.Client.Common/ArgParser.cs#L208
it's so us command-line users can see the expanded version of what we've passed, in case it contains a variable or subshell, or a wrapper script has rewritten your flags. It's logged to stderr so it's not captured by a redirect/pipe.

edit: example
```
$> emuhawk-monort-local "$temp/Contra (USA).nes" 
(capturing output in /home/yoshi/Documents/BizHawk/output/EmuHawkMono_last*.txt)
libpng warning: iCCP: known incorrect sRGB profile
parsing command-line flags: --config=config.json /home/yoshi/Downloads/roms/debugging//Contra (USA).nes
```

src/BizHawk.Client.Common/ArgParser.cs
SuuperW: `result.Action is not null` is only true for `--help`? That's not clear at all. Maybe that's what the comment there means to say; if so it should be explicit. (Though that's not a comment from this PR.)

If the return value isn't being asserted, and `Invoke` would return it anyway, ... why bother hard-coding it in the test? If you're trying to avoid having stuff actually written to the console, I think that should be handled by changing the output stream. It looks like you're already changing the output stream for testing `--help` anyway.

The line saying what it is parsing should probably be commented out, it looks like it's just for debug.

src/BizHawk.Client.Common/ArgParser.cs
SuuperW: Would a more generic method for all tests not avoid the issue of writing to the console during unit tests? As mentioned above, I think you can and should get rid of the `fromUnitTest` parameter.

src/BizHawk.Tests.Client.Common/ArgParserTests.cs
SuuperW: I think I agree with Morilli that it doesn't need changing. Standard behavior of Unix programs isn't relevant for a program that primarily targets Windows, any my experience tells me the current behavior of accepting the path anywhere is standard behavior. The fact that the library handles `--` does not imply that it wants to be like Unix, it's just supporting a convention that some people will find useful when doing so does not harm any other users.

I would not be opposed to changing the help to state that the rom can be anywhere in command. But changing it from one correct-but-incomplete output to another correct-but-incomplete output is not useful.

SuuperW:
> it's so us command-line users can see the expanded version of what we've passed, in case it contains a variable or subshell, or a wrapper script has rewritten your flags. It's logged to stderr so it's not captured by a redirect/pipe.

I'm not convinced it belongs here. (you can see what your shell expands things to manually whenever you want, you don't need EmuHawk to do that for you, and I see no value in making the output captured by a redirect/pipe differ in the info given to the user normally)
But if you find it useful for everyday use, I won't make a fuss about this minor weird thing EmuHawk is doing. And if that were the only use of `fromUnitTest` then I think it would be acceptable.

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: Before https://github.com/TASEmulators/BizHawk/commit/2b25d09bcf4574cb1dbfb9e2d270f44d459596ba the condition was even more cryptic, hence the comments. At least now it's being tested for.

I'm hardcoding the value so that I don't have to call `Invoke` to get it, and I'm avoiding `Invoke` so it doesn't write to console *during the other unit tests*, including possible future tests. It's easy to miss, but `TestHelpSaysPassFlagsFirst` calls `RunHelpActionForUnitTest`, while the others call this method with `fromUnitTest: true`.

Printing the input is not just for debugging, see sibling thread.

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: I'm saying that a generalisation of this method would still only be applicable to `--help` ~~and `--version`~~ and actually just that, since it's a stripped-down version of `ParseArguments`, where all the logic is.

src/BizHawk.Tests.Client.Common/ArgParserTests.cs
YoshiRulz: Both orderings *are* accepted, see `TestWithNonexistent` below.

The help message could say
```
Usage:
  EmuHawk.exe [rom] [option...]
  OR
  EmuHawk.exe [option...] [rom]
```
but that seems redundant. So if I had to pick, I'm picking the conventional one every time.

src/BizHawk.Client.Common/ArgParser.cs
SuuperW: Hardcoding the expected return value based on the fact a unit test is being run should never happen.
Find another way to avoid writing to the console during unit tests.
`TestConfigWithHelpOrVersion` with `[DataRow("--config=config.json", "--help")]` is asserting a return value obtained by `fromUnitTest ? 0 : result.Invoke()`.

Even if this wasn't the case, even if you think this is OK because this line is only encountered when `--help` is passed and you know what `Invoke` would return, it makes the code confusing and difficult to maintain.

(I disagree that it was worse prior to that commit, but let's drop the discussion of the comment line since it's outside the scope of this PR.)

src/BizHawk.Tests.Client.Common/ArgParserTests.cs
SuuperW: The help message could say
```
Usage:
  EmuHawk.exe [rom] [option...]
  The ROM path may appear before or after any or all options.
```

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: `TestConfigWithHelpOrVersion` is, as the name suggests, checking that `--config`+`--help` doesn't throw (and isn't unrecognised, returning `null`); the exit code isn't really under test.
That branch could just as well read
```csharp
// means e.g. `./EmuHawkMono.sh --help` was passed, run whatever behaviour it normally has
if (!fromUnitTest)
{
	EnsureConsole();
	_ = result.Invoke();
}
return 0;
```
because the early-exit of `--help` is known at design-time to be a success state.

If you can think of a different way to programmatically control whether it writes to stdout that's of a comparable size and complexity to my solution, then please share it.

SuuperW: > If you can think of a different way to programmatically control whether it writes to stdout that's of a comparable size and complexity to my solution, then please share it.
 
 I already did. Use a more generic version of `RunHelpActionForUnitTest`.

PR #4637, opened 2026-02-23
Fixed FDS registers, and OAM Buffer

100thCoin:
A while back, I made changes to how APU registers are read. This is because APU registers are only readable if the 6502 address bus is within the range of $4000 to $401F. Unbeknownst to me, this change broke reads from the Famicom Disk System Registers. I have corrected that issue, and ran relevant tests.

Additionally, the primary purpose for me making this PR was to make a very small change to OAM evaluation, where writes to OAM2 become reads from OAM2. It was simply reading from OAM2[0] instead of using the OAM2 address, which I have corrected.

[Relevant test for this OAM2 behavior.](https://github.com/100thCoin/AccuracyCoin/blob/1f5b96c74f78003d05d193eb7270e3c7f23c7ba3/AccuracyCoin.asm#L2496) I'll admit that this isn't my most readable test. It just sets up OAM with a specific pattern, reads from $2004 on every ppu cycle of a specific scanline, (across a couple frames. This obviously cannot be done in a single frame) and then evaluates the results.

Previous results:
image

Error code 6 is pretty much saying that the data after those OAM2 writes become reads was incorrect.

After making this change:
image

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/PPU.run.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/PPU.run.cs
@@ -371,7 +371,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.NES
 										{
 											oam_index &= 0x1FC;
 										}
-										soam_index++;
+										// soam_index no longer increments now that Secondary OAM is full.
 										spr_true_count = 0;
 									}
 								}


100thCoin:
I have updated the test I wrote for this behavior and discovered another small inaccuracy in NesHawk, which I have fixed. The new version of the test changed the error codes (so now the error codes are unfortunately a bit less descriptive), but now the test checks for what happens when there are 8 or more objects on the scanline being evaluated.

The small change I made here is that after secondary OAM is filled, the secondary OAM address is no longer incremented when finding new objects in range of the scanline. I simply replaced the line incrementing the secondary OAM address with a comment stating that the secondary OAM address is no longer incremented in this situation.

PR #4638, opened 2026-02-24
Fix `events.onexit` callbacks accessing a disposed core and crashing

YoshiRulz:
Dead simple, and doesn't interfere with running `onexit` callbacks when individual scripts are stopped.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -3942,6 +3942,8 @@ namespace BizHawk.Client.EmuHawk
 				if (saveMovieResult == TryAgainResult.Canceled) return false;
 			}
 
+			if (Tools.Has<LuaConsole>()) Tools.LuaConsole.CallScriptExitCallbacks(); // important to do this here with the stale `IEmulator`, see https://github.com/TASEmulators/BizHawk/issues/4629
+
 			if (clearSram)
 			{
 				var path = Config.PathEntries.SaveRamAbsolutePath(Game, MovieSession.Movie);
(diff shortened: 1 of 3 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/LuaConsole.cs
```diff
@@ -568,6 +568,11 @@ public void CallStateLoadCallbacks(string userFriendlyStateName)
 		public void CallStateSaveCallbacks(string userFriendlyStateName)
 			=> LuaImp.CallSaveStateEvent(userFriendlyStateName);
 
+		public void CallScriptExitCallbacks()
```
SuuperW, line 571: I think the purpose of this needs to be documented. Calling the exit event without stopping the script is a weird thing to do, and that should be noted along with why it's done here.

Alternatively, you could do what I did and actually stop the scripts (renaming the method accordingly). Just keep a list of which ones were running so they can be started again in restart.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3942,6 +3942,8 @@ private bool CloseGame(bool clearSram = false)
 				if (saveMovieResult == TryAgainResult.Canceled) return false;
 			}
 
+			if (Tools.Has<LuaConsole>()) Tools.LuaConsole.CallScriptExitCallbacks(); // important to do this here with the stale `IEmulator`, see https://github.com/TASEmulators/BizHawk/issues/4629
```
SuuperW, line 3945: The emulator isn't stale yet, it is still current.

Issue #1466, opened 2019-01-28
Locked SaveRAM files are not handled gracefully


RetroEdit:
I've been trying out `File > Save RAM > Flush Save Ram` and found that if the save file is locked, the command silently fails with no notice.

I've also had an exception thrown upon exiting a game through `File > Close Rom` because the SaveRAM is locked.

It would be preferable in both cases if a dialog box appeared telling the user that the file is locked for editing, and allowing them to try again or skip saving.

I'm using BizHawk Version 2.3.1 (x64) on Windows 8.1.

zeromus:
Since I believe manual managing of saveram is the only way that it should be done and threaten to remove automatic saveram management one day without notice, the onus is on me to do a better job here with the manual management. Thanks for pointing this out

eToThePi-i:
@zeromus I suppose by "manual", you mean *automatic* save management? I think it would still be ideal if a "Would you like to save your game?" dialogue popped up before quitting a game (or at the worst, crashing, though in those cases it might be better to create a back up and possibly warn the player about save corruption).

zeromus:
No, I mean the opposite of automatic. That dialog will be a nuisance. It needs to appear when closing any game, not just the emulator. Everyone will turn it off. In short order it may as well not even be there. However a similar dialog is planned for the transitionary period, since nobody will be expecting this change of policy which is admittedly a huge change. 

If the emulator crashes, there's nothing we can do.

eToThePi-i:
@zeromus I don't think I understand. Isn't that what I just said? I see your point about most people turning the dialog off, but I think it's a nice default anyway, and wasn't really commenting about that particular aspect.

And yeah, I don't expect magic with crashes per se, but there are often caught exceptions when using BizHawk.

zeromus:
Yes that's what you said, and you called it automatic, which is the opposite of what it is.

Issue #2175, opened 2020-06-28
Mitigate accidental overwriting of SaveRAM with savestates


RetroEdit:
bramcn0745 brought this up in #2157, and I think this option would be extremely valuable. Loading savestates can be very destructive for casual users, as it can immediately overwrite a save file on disk without any warning. In my very early days, I have had this happen a few times, and it's easy to not even be aware of the dangers of loading savestates, or load a savestate without thinking.

One option is a warning like: "Loading a savestate will overwrite your save file. Are you sure you want to do this?" It could even have an option "Backup my save file". A warning might be overkill, but if it had an easy dialog option "don't warn me again", that might be a nice compromise.

Alternatively, BizHawk could create a backup SaveRAM file when loading a savestate for the first time in a session, and then any savestate loading mishaps are already recoverable. The more I think about it, the more this second option seems ideal for casual players and TASers alike, as it's a nice safety catch mechanism, but also not disruptive to power-users.

zeromus:
This has been litigated to death. It is standard practice for emulators. People need to learn it. This confusing as hell situation is known as "clobber's lament."

There will be no warning on loaded savestates. 

I have spent a long time thinking about this and dealing with the nuisance. I only consider two things acceptable for solving this, and they both need to be done eventually:

1. Make backups of saverams EVERY SINGLE TIME a state is flushed (unless in a movie session). The disk will be littered with saves. Users can deal with it. Users can learn how to clean up their mistakes from backups. Put careful thought into how it's organized; perhaps put them in a backups subdirectory. We can do this now.

2. Later, when we're feeling bold: users must choose to flush manually. #1 still applies--backups are always made first. The pressure for part 2 will be reduced by part 1. Moreover, for an interim learning period (or perhaps forever), backups will also be made directly at the point where the emulator would ordinarily automatically flush. So, as with part 1, people at least have the files needed to go in and clean up after themselves.

nattthebear:
Can we just close this, please?  A user who is freely interleaving saveram saves and savestate saves is going to teach themselves a life lesson sooner or later regardless of how many coats of paint we put on the dung pile.  Then there's the problem of systems and games where failing to include saveram in the savestate produces an incoherent save that can crash the emulated game on load...

RetroEdit:
>Then there's the problem of systems and games where failing to include saveram in the savestate produces an incoherent save that can crash the emulated game on load...
[...]

PR #4644, opened 2026-02-25
Enable AutoSaveRAM by default

YoshiRulz:
Are there any reasons not to do this? HDDs aren't widespread anymore, the user can still turn it off if they want, and I never did figure out what [this handholding argument](https://github.com/TASEmulators/BizHawk/issues/1466#issuecomment-458188027) was about.

### DIFF
--- a/src/BizHawk.Client.Common/config/Config.cs
+++ b/src/BizHawk.Client.Common/config/Config.cs
@@ -188,12 +188,12 @@ namespace BizHawk.Client.Common
 		/// <summary>
 		/// Whether to make AutoSave files at periodic intervals
 		/// </summary>
-		public bool AutosaveSaveRAM { get; set; }
+		public bool AutosaveSaveRAM { get; set; } = true;
 
 		/// <summary>
 		/// Intervals at which to make AutoSave files
 		/// </summary>
-		public int FlushSaveRamFrames { get; set; }
+		public int FlushSaveRamFrames { get; set; } = 5/*min*/ * 60/*s/min*/ * 60/*fr/s*/;
 
 		public bool TurboSeek { get; set; }
 


SuuperW:
In my opinion, it should only be on by default if it does not result in overwriting the SaveRAM file in the following situation:

1) Start a movie
2) Stop the movie
3) Continue playing with the movie's SRAM loaded

As of right now, it does overwrite in that situation. So I am opposed to auto save.

Related, I think we should not automatically make SaveRAM files for movies regardless of this setting. Imo, SRAM should never be written to disk while a movie is (or has been) loaded, except when the user does "Save Save Ram As...", which is a feature that does not even exist right now.

RetroEdit:
To me, this has limited usefulness, but also limited harm given SaveRAM already saves automatically.

This is seemingly only relevant for crashes(?), since SaveRAM typically saves when a game or the emulator closes. It being [tied to emulator frame increments](https://github.com/TASEmulators/BizHawk/blob/f588942b79ae79e62d9d86f7a001c9d3ce973453/src/BizHawk.Client.EmuHawk/MainForm.cs#L3014-L3036) instead of real time is a bit odd to me, but I guess it's fine because SaveRAM will generally only changes notably after some frames have elapsed.

Regarding "Save Save Ram As...", it would probably be a good addition; however, I don't have a particular problem with the current way movie SaveRAM works and occasionally find it useful, particularly as it saves to a separate SaveRAM file.

SuuperW:
> it saves to a separate SaveRAM file

It only does this until the movie is stopped. After that, it will overwrite the non-movie file.

And I personally dislike the extra files, since it's just a bunch of clutter and I hardly ever want any of them.

RetroEdit:
> I never did figure out what [this handholding argument](https://github.com/TASEmulators/BizHawk/issues/1466#issuecomment-458188027) was about.

Not sure what you mean by hand-holding, but zeromus's position is explained in more depth here: #2175

The basic argument seems to be implicit Save RAM flushes create confusion and unintentional file clobbers for users who begin using savestates or movies. Thus, it would be better if users learned to always save intentionally.

Not sure I totally buy the "fix" zeromus proposes, but I do think it's a genuine problem.

For illustration: say a user been playing a game through BizHawk for some sessions and the implicit saving has been working just fine. Trouble happens if they begin using savestates and particularly movies. The "two-step closing" of movie stopping leaving behind a game instance will mercilessly overwrite the preexisting save. This is a dubious counterintuitive behavior; it means by default there is no way for a preexisting non-movie save to coexist with movie recording.

YoshiRulz:
Are there any reasons that don't involve movies?
I skimmed zeromus' comments on that other thread and they don't mention AutoSaveRAM, only savestates. I don't plan to change the behaviour around savestates at the moment.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -769,6 +769,8 @@ _argParser.SocketAddress is var (socketIP, socketPort)
 					}
 				}
 			}
+			//TODO should we be calling `OnRomChanged` from this ctor? it handles this and a bunch of other things also done here
```
SuuperW, line 772: If a ROM gets loaded, it should be getting called by the method that loads the ROM, e.g. `LoadRom` through `LoadRomInternal`.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -3897,6 +3901,7 @@ private void OnRomChanged()
 			UpdateCoreStatusBarButton();
 			UpdateDumpInfo();
 			SetMainformMovieInfo();
+			_trackingAutoSaveRAM = Config.AutosaveSaveRAM && MovieSession.Movie is null;
```
SuuperW, line 3904: This does not work, because the movie has not been set yet at this point. If you open a ROM and then TAStudio right after starting EmuHawk, `_trackingAutoSaveRAM` will be true. (assuming the config option is on) (... so why is `SetMainformMovieInfo` being called here??)

This should probably also be a value that just tracks if a movie was used, since a user may wish to change the auto save setting after opening a ROM.

SuuperW (commented):
RE: savestates
I think that auto save after loading a savestate could be confusing and lead to unwanted loss of .SaveRAM files. The loaded savestate might be from a movie, or a different play session. I don't think there's a good way to detect this even if you know what the user wants, and I also don't think simply turning off auto save after any savestate load is a good idea. I think this problem is a good argument for never having auto save.

But there's really two "auto save" things. This PR is changing the autosave that will only get used as a backup in case EmuHawk crashes. The other auto save is flushing SRAM on ROM close or core reset. I don't see any reason why these two should differ in terms of when they are active aside from the user setting.

PR #4648, opened 2026-03-02
Skip copying melondDS default MAC address

### LINKED ISSUE #4646: DS(i) mac address is always overridden
[//]: # "Also, please don't waste your time writing until you've checked for similar Issues. Remember to check closed Issues too!"

### Summary
The Mac address of the imported firmware is always overridden by Melonds's default mac address

For a personal dump verification, you can use MelonDS 1.1, which won't override your firmware's mac address.

### Repro
1. Set your ds dumps (bios, firmware, nand for the dsi)
2. Disable firmware override
3. If you want to check without a game, on the dsi you can set firmware start up as "manual boot"
4. Check your mac address. It'll always be a predefined one, not the one you dumped

### Host env.
- BizHawk 2.11
- BizHawk_devbuild_e2dc99ea

Real96:
Fixes #4646


waterbox/melon/BizConsoleCreator.cpp
```diff
@@ -63,6 +63,17 @@ static std::unique_ptr<T> CreateBiosImage(u8* biosData, u32 biosLength, std::opt
 	return std::move(bios);
 }
 
+static bool isValidMacAddress(const melonDS::MacAddress& mac)
+{
+    // 48bit MAC address v1–v5: 00:09:BF:XX:XX:XX
+    const bool isV1toV5 = mac[0] == 0x00 && mac[1] == 0x09 && mac[2] == 0xBF;
+
+    // 48bit MAC address v6–v7: 00:16:56:XX:XX:XX
+    const bool isV6toV7 = mac[0] == 0x00 && mac[1] == 0x16 && mac[2] == 0x56;
```
CasualPokePlayer: These checks appear to be non-exhastive.

For DSi, `A4 C0 E1` is valid.
For DS Lite, `00 25 A0` is valid.
For iQue DS Lite, `00 1D BC` is valid.

In theory, there's potentially a ton of other valid MAC addresses, as Nintendo has registered a ton of MAC prefixes. It might be better to just make sure it's not all 0xFFs (not uncommon to find with "no-intro" matching dumps which 0xFF wifi area) along with not being 03:09:BF (broadcast channel, not valid)

CasualPokePlayer (changes requested):

YoshiRulz:
You forgot to update the comments, so now they're non-sequiturs.

waterbox/melon/BizConsoleCreator.cpp
Real96: Done

Real96:
> You forgot to update the comments, so now they're non-sequiturs.

Done

PR #4661, opened 2026-03-19
implement a multitrack recording/editing feature

SuuperW:
This is primarily intended to be a better implementation of #4218/#4224. It could also be considered an implementation of #2510 and #2349.

temp

When "allow editing hidden columns" is disabled (and it is enabled by default, so as to not surprise users with a potentially confusing and new feature), none of TAStudio's various methods of editing input will affect the inputs in hidden columns. The only exception is truncating, which will delete frames if they match the default input values. ("empty") When multiple input rolls exist, only the active one is considered. ("active one" being the one most recently clicked, which should always be the one that has focus [if any have focus])

The columns menu has been removed, and replaced with a new right-click menu that appears when the user right-clicks on a column header.

Inserting frames will always add frames to the input log, and deleting frames will never remove frames from the input log. The behavior of delete may be expected to remove frames in certain situations, but given that having empty frames at the end doesn't do much I decided it's best for now to keep the implementation simple. And I am completely ignoring auto-fire and auto-hold, because (1) I cannot decide exactly how they should behave, (2) they are already somewhat broken or at least do not behave as I would expect when using undo or painting past the end of the movie, and (3) that would be a significant amount of work that might not even be useful to anyone.

I have not considered how best to handle binding markers to input when editing hidden columns is disabled, beyond deciding that combining these features is probably a bad idea. A message box will warn users who set these settings.

There are some new tests for the new behavior, plus a couple for issues I fixed along the way. (a) truncating a TasMovie by 1 frame did not work (b) multiple calls to `Bk2Movie.GetInputState` would return the same instance, invalidating the result of prior calls [though I am not aware of any previous uses of `GetInputState` where this results in a real issue]

Additional testing is still required, particularly to ensure undo and redo work in all situations. ~~And I have not tested performance in long movies.~~ With optimization long movies mostly work well now.

Potentially desirable features that have not been implemented:
1) Automatically create input rolls that are split by player number. Details on how the UI for this should look are TBD. Should it happen automatically?
2) Currently even "follow cursor" applies only to the active one, if scroll sync is disabled; should it apply to all?
3) Other user experience improvements? Suggestions welcome.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/inputAdapters/MultitrackAdapter.cs
+++ b/src/BizHawk.Client.Common/inputAdapters/MultitrackAdapter.cs
@@ -1,0 +1,58 @@
+﻿#nullable enable
+
+using System.Collections.Generic;
+using BizHawk.Emulation.Common;
+
+namespace BizHawk.Client.Common
+{
+	/// <summary>
+	/// Used to enable recording a subset of a controller's buttons, while keeping the existing inputs for the other buttons.
+	/// Also used to allow TAStudio to clear (or otherwise edit) a subset of input columns.
+	/// </summary>
+	internal class MultitrackAdapter : IController
+	{
+		/// <summary>
+		/// Input states in this definition will come from <see cref="ActiveSource"/>. All others will come from <see cref="BackingSource"/>.
+		/// </summary>
+		public ControllerDefinition ActiveDefinition { get; set; }
+
+		public IController ActiveSource { get; set; }
+
+		public IController BackingSource { get; set; }
+
+		public ControllerDefinition Definition => BackingSource.Definition;
+
+		public MultitrackAdapter(IController active, IController backing, ControllerDefinition activeDefinition)
+		{
+			ActiveSource = active;
+			BackingSource = backing;
+			ActiveDefinition = activeDefinition;
+		}
+
+		public bool IsPressed(string button)
+		{
+			if (ActiveDefinition.BoolButtons.Contains(button))
+			{
+				return ActiveSource.IsPressed(button);
+			}
+			else
+			{
+				return BackingSource.IsPressed(button);
+			}
+		}
+		public int AxisValue(string name)
+		{
+			if (ActiveDefinition.Axes.ContainsKey(name))
+			{
+				return ActiveSource.AxisValue(name);
+			}
+			else
+			{
+				return BackingSource.AxisValue(name);
+			}
+		}
+
+		public IReadOnlyCollection<(string Name, int Strength)> GetHapticsSnapshot() => throw new NotImplementedException(); // don't use this
+		public void SetHapticChannelStrength(string name, int strength) => throw new NotImplementedException(); // don't use this
+	}
+}
--- a/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.cs
+++ b/src/BizHawk.Client.Common/movie/bk2/Bk2Movie.cs
@@ -5,7 +5,16 @@ namespace BizHawk.Client.Common
 {
 	public partial class Bk2Movie : BasicMovieInfo, IMovie
 	{
-		private Bk2Controller _adapter;
+		private IController _defaultValueController;
+		protected IController DefaultValueController
+		{
+			get
+			{
+				// LogKey isn't available at construction time, so we have to create this instance when it is accessed.
+				_defaultValueController ??= new Bk2Controller(Session.MovieController.Definition, LogKey);
+				return _defaultValueController;
+			}
+		}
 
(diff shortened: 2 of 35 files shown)


vadosnaprimer:
Please post some screenshots showcasing the UI changes while I'm getting to it.

SuuperW:
There are no UI changes, except for a new checkbox in the settings dialog. There is still only 1 input roll,  and there is no automatic splitting based on player 1/player 2 inputs. With what I've done so far, users will have to toggle the visibility of columns to control which inputs are being edited. There is no way to see inputs in a column at the same time that those inputs are not being affected by clear/insert/etc.

SuuperW:
Updated to now support having multiple input rolls. Right-click on a column header for a little menu that includes the option to make a new input roll.

Make sure you open the settings and turn off "Allow editing hidden columns" (in the new Input Roll tab) to test editing only one input roll's visible columns.

DmytroM1998:
I tried this out in a DOSBox-X game. No bugs have been spotted. 
It certainly improves TAS production!

vadosnaprimer:
Please fix the conflicts.

SuuperW:
Fixed conflicts, fixed some bugs, optimized for long movies.
I also added a tooltip for the option "Allow editing hidden columns (with clear, record, paste, etc.)":
> When disabled, inputs that are not visible on the active input roll cannot be edited.
> This allows you to, for example, insert and delete frames on a roll with player 1's inputs
without affecting player 2's inputs, by hiding all columns for player 2.
> You can hide columns and create/delete input rolls by right-clicking on a column header.
> Because markers are not associated with any input column, it is recommended that you turn
off "Bind markers to input" if you uncheck this box. Otherwise markers may move unexpectedly.
> This feature is experimental but should be ready for use.

Is it OK to put that (last line) there?

vadosnaprimer:
> Is it OK to put that (last line) there?

Yeah.

vadosnaprimer:
I should have tested this...

> The columns menu has been removed, and replaced with a new right-click menu that appears when the user right-clicks on a column header.

I see why it got removed, but I suggest a solution that doesn't reduce convenience for people who won't use this feature. When there's only 1 roll, use the old Columns menu to control it. When there are several, disable the old menu and rely on context menus instead. If context menu mirrors the functionality of the old menu when there's only 1 roll, that seems fine to me.

I'm trying the "new input roll" menu on NES and it adds it with hidden column flags inverted: power and reset are shown but everything else is hidden. For GBA too, looks like a universal problem.

SuuperW:
> If context menu mirrors the functionality of the old menu when there's only 1 roll, that seems fine to me.

It does exactly the same thing. The only reduction of convenience is 1 additional click (right-click -> "columns", as opposed to just "columns").

> I'm trying the "new input roll" menu on NES and it adds it with hidden column flags inverted: power and reset are shown but everything else is hidden. For GBA too, looks like a universal problem.

That is intentional. It's so that if you hide Player 2 from the input roll then make a new one, the new one will have Player 2 on it. If all columns were visible in the existing input roll(s), all columns will also be visible in the new one.

vadosnaprimer:
> The only reduction of convenience is 1 additional click (right-click -> "columns", as opposed to just "columns").

90% of the users won't be using secondary rolls, there's benefit in making it more clunky for them. With systems that have many buttons that may affect the game, one may have to use the columns many many times during the session, which quickly adds to the annoyance without adding to convenience.

> It's so that if you hide Player 2 from the input roll then make a new one, the new one will have Player 2 on it. If all columns were visible in the existing input roll(s), all columns will also be visible in the new one.

Is there a way to read the same list we use when hiding columns initially, and keep those columns hidden when a new roll is added? Manual changes inverting sounds sensible, but those columns hidden by default are very rarely needed.

SuuperW:
> 90% of the users won't be using secondary rolls, there's benefit in making it more clunky for them. With systems that have many buttons that may affect the game, one may have to use the columns many many times during the session, which quickly adds to the annoyance without adding to convenience.

I rarely change visibility of columns after setting up a project, and didn't know it was a common thing to do. That still seems like there could be a better solution than what we have/had...

I will try re-adding the menu and making it operate on the currently selected input roll (or on the 1 single roll if there is just 1).

> Is there a way to read the same list we use when hiding columns initially,

It isn't a plain list, but I can probably make it work with a little refactoring.

vadosnaprimer:
> I rarely change visibility of columns after setting up a project, and didn't know it was a common thing to do. That still seems like there could be a better solution than what we have/had...

Normally you'd keep the keyboard keys hidden but then you have to figure out which ones you will need in the tas, and there's no nice way around adding them back one by one when you realize you do need those few (or many), and you can't solve it by simply showing the entire keyboard.

PR #4668, opened 2026-03-25
fix: moving markers in TAStudio by alt-dragging would focus the menu

SuuperW:
Someone more knowledgeable should test the Virtual Pad.

Holding alt allows the user to drag markers in TAStudio. When doing this, the menu bar would be focused after releasing alt. This fixes that by having `Input` not totally swallow input events, so the alt-handling code can know that a click happened. Instead they are passed through with a new `HostInputType` which `InputManager` checks for.

There was some disabled code in `Input.cs` that would have (if enabled) released all mouse buttons when a tool form is focused. Per the comment, doing this "will interfere with the virtualpad". I do not know in what way it was interfering. The updated code does allow mouse buttons to be released while a tool form is focused (but does not automatically release them when the tool form is focused). I do not know in what way it was interfering. I don't regularly use the virtual pad, but a basic test did not reveal any issues. The comment is over 10 years old so perhaps the issue no longer exists.

### DIFF
--- a/src/BizHawk.Bizware.Input/HostInputType.cs
+++ b/src/BizHawk.Bizware.Input/HostInputType.cs
@@ -9,5 +9,6 @@ namespace BizHawk.Bizware.Input
 		Mouse = 1,
 		Keyboard = 2,
 		Pad = 4,
+		Ignored = 8,
 	}
 }
--- a/src/BizHawk.Client.EmuHawk/Input/Input.cs
+++ b/src/BizHawk.Client.EmuHawk/Input/Input.cs
@@ -174,6 +174,31 @@ namespace BizHawk.Client.EmuHawk
 			}
 		}
 
+		private void EnqueueNewEvents(bool prohibitInput = false)
+		{
+			if (_newEvents.Count != 0)
+			{
+				//WHAT!? WE SHOULD NOT BE SO NAIVELY TOUCHING MAINFORM FROM THE INPUTTHREAD. ITS BUSY RUNNING.
+				AllowInput allowInput = prohibitInput ? AllowInput.None : MainFormInputAllowedCallback();
+
+				foreach (var ie in _newEvents)
+				{
+					if (ie.EventType == InputEventType.Press && ShouldSwallow(allowInput, ie.Source))
+					{
+						EnqueueEvent(new InputEvent() {
+							EventType = ie.EventType,
+							LogicalButton = ie.LogicalButton,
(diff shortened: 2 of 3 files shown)


src/BizHawk.Client.EmuHawk/Input/Input.cs
```diff
@@ -174,6 +174,31 @@ private void EnqueueEvent(InputEvent ie)
 			}
 		}
 
+		private void EnqueueNewEvents(bool prohibitInput = false)
```
YoshiRulz, line 177: Making this a local method at the end of `UpdateThreadProc` should reduce diff

src/BizHawk.Client.EmuHawk/Input/Input.cs
```diff
@@ -205,8 +230,6 @@ private void UpdateThreadProc()
 				//this block is going to massively modify data structures that the binding method uses, so we have to lock it all
 				lock (this)
 				{
-					_newEvents.Clear();
```
YoshiRulz, line 208: Changed from clearing at start to clearing at end—I assume that's fine since this method never returns and is never called again if it throws?

src/BizHawk.Bizware.Input/HostInputType.cs
```diff
@@ -9,5 +9,6 @@ public enum HostInputType
 		Mouse = 1,
 		Keyboard = 2,
 		Pad = 4,
+		Ignored = 8,
```
YoshiRulz, line 12: Isn't this conceptually the same as `None`? `None` doesn't seem to be used anyway.

src/BizHawk.Client.EmuHawk/Input/Input.cs
SuuperW: I did not think about what would happen if it throws, but yeah it looks like it would not ever be called again.
There's probably no harm in keeping the clear in `UpdateThreadProc` though.

src/BizHawk.Bizware.Input/HostInputType.cs
SuuperW: I thought it better to use a value with a more appropriate name than `None`.

vadosnaprimer:
I'm not noticing the change compared to master. Just like in 2.11 I hold Alt, underlines appear on menus, I move the marker, release Alt, and underlines stay there.

SuuperW:
Just tested in 2.11 and the bug this fixes isn't present there. In master when I use alt-drag to move a marker, the File menu is highlighted and pressing buttons navigates the menu.
This bug was introduced by #4480.

Issue #333, opened 2015-06-22
[Request] Allow binding keyboard to emulated analog that isn't full magnitude


GoogleCodeExporter:
```
When loading N64 games, the  sensitivity slider under the "analog controls" tab 
in the controller menu only seems to affect joypad bindings. The keyboard keys 
configured under "normal controls" seem to be always set at 100%, which makes 
precise aiming in first person nigh impossible when using a keyboard. 
A lower value such as 65% would be ideal for precise aiming while still 
allowing the character to run at full speed.

I request that the slider under analog controls start to affect the keyboard 
bindings as well, so if you set to 20% pressing up on the keyboard would move 
the analog range by only 20%. 
```

Original issue reported on code.google.com by `spergmei...@gmail.com` on 1 Dec 2014 at 7:29

Attachments:
- [n64 analog.jpg](https://storage.googleapis.com/google-code-attachments/bizhawk/issue-333/comment-0/n64%20analog.jpg)

vgturtle127:
@adelikat 
Is this practical? If you are using a keyboard isn't the sensitivity mostly pointless? I would imagine setting a different sensitivity for a button would make it difficult to setup, and would have to be configured on a per-game basis...

YoshiRulz:
Dolphin does this, there's a percentage numbox below the nunchuck analog binds that's used to scale the emulated input:
![Dolphin analog UI](https://i.imgur.com/vSLNqou.png)

UI suggestion (call it something more helpful than `A-`):
![controller config window mockup](https://i.imgur.com/lM6cOQX.png)

Issue #3414, opened 2022-10-14
[Feature Request] Allow binding of multiple axes to Analog Controls


Sophilautia:
A feature I really appreciate at the moment is being able to multiply bind controls to both DirectInput and XInput at the same time. This is convenient because it allows me to send my config.ini to a friend, and all their controls will be set up to work with an Xbox or Playstation controller without any further configuration required. I do some stream events with people who don't frequently use BizHawk, so I like to make sure the setup required for my participants is as small as possible, so this feature helps a lot.

But currently, there doesn't seem to be any way to bind multiple axes to Analog stick controls. I have to choose between XInput or DirectInput, which means I can't prepare it for someone beforehand without knowing whether they use an XInput controller or DirectInput controller. If we could have some way to multiply bind Analog controls like we do for buttons, that would be a huge help for me.

YoshiRulz:
Would also be resolved by Steam Input-style implementation of #2244.

PR #4670, opened 2026-03-27
Adjust analog binds to also allow button binds

### LINKED ISSUE #4597: Can't map keyboard to analog on 3DS
### Summary
Can't bind keyboard keys to analog controls when using Encore to play 3DS Games

Image

### Repro
1. start any 3DS game that requires the analog stick without a controller
2. Try to bind a keyboard key to the analog stick

Morilli:
[dev build for branch | USERNAME:BRANCHNAME](https://nightly.link/Morilli/BizHawk/workflows/ci/analog-button-binds?preview)

- closes #4597

Initial version of an analog bind panel that allows binding buttons instead of physical axes to axis controls, to allow controlling cores that have axes controls like joysticks with the keyboard.
image

It's a bit rough around the edges at the moment; I had to hack `InputWidget` a bit because it doesn't like being without a `CompositeWidget`, but it works pretty well as is. This doesn't allow multiple binds (#3414) at the moment but I'm not sure we need that? Should be as easy as toggling `Auto-Tab` regardless.

The sensitivity setting is also fully applied to button values, so with sensitivity of 50% pressing the `Axis+` button will only result in a 50% value being sent to the core.

Unless there are major concerns about design or code I'd like to get this in because it's been missing for a long time.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/config/ControllerConfig/AnalogBindControl.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/config/ControllerConfig/AnalogBindControl.Designer.cs
@@ -30,15 +30,15 @@
 		{
 			this.components = new System.ComponentModel.Container();
 			this.textBox1 = new System.Windows.Forms.TextBox();
+			this.labelAxisName = new BizHawk.WinForms.Controls.LocLabelEx();
 			this.trackBarSensitivity = new System.Windows.Forms.TrackBar();
+			this.labelSensitivity = new BizHawk.WinForms.Controls.LocLabelEx();
 			this.timer1 = new System.Windows.Forms.Timer(this.components);
 			this.buttonBind = new System.Windows.Forms.Button();
 			this.trackBarDeadzone = new System.Windows.Forms.TrackBar();
+			this.labelDeadzone = new BizHawk.WinForms.Controls.LocLabelEx();
 			this.buttonFlip = new System.Windows.Forms.Button();
 			this.buttonUnbind = new System.Windows.Forms.Button();
-			this.labelDeadzone = new BizHawk.WinForms.Controls.LocLabelEx();
-			this.labelSensitivity = new BizHawk.WinForms.Controls.LocLabelEx();
-			this.labelAxisName = new BizHawk.WinForms.Controls.LocLabelEx();
 			this.iwPositiveButton = new BizHawk.Client.EmuHawk.InputWidget();
 			this.iwNegativeButton = new BizHawk.Client.EmuHawk.InputWidget();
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Nice!!!

Morilli:
Fixes #333 too I suppose.

src/BizHawk.Client.EmuHawk/config/ControllerConfig/AnalogBindControl.cs
```diff
@@ -16,16 +16,20 @@ public AnalogBindControl(string buttonName, AnalogBind bind)
...
 			TrackBarSensitivity_ValueChanged(null, EventArgs.Empty);
 			TrackBarDeadzone_ValueChanged(null, EventArgs.Empty);
 			textBox1.Text = bind.Value;
+			iwPositiveButton.Text = bind.ButtonBindPositive;
+			iwNegativeButton.Text = bind.ButtonBindNegative;
 		}
 
 		public string ButtonName { get; }
-		public AnalogBind Bind => _bind;
+		public AnalogBind Bind => new(_bind.Value, _bind.Mult, _bind.Deadzone, iwPositiveButton.Text, iwNegativeButton.Text);
```
SuuperW, line 32: Why is this getter changed to create a new instance when accessed, instead of returning `_bind`?

Morilli: Because `_bind` doesn't get changed when the button bind texts change, and I thought this was the easiest way to do it.

YoshiRulz:
```csharp
iwPositiveButton.TextChanged = (_, _) => _bind.ButtonBindPositive = iwPositiveButton.Text;
iwNegativeButton.TextChanged = (_, _) => _bind.ButtonBindNegative = iwNegativeButton.Text;
```
?

Morilli: Probably possible, although generally I think it's better to construct the AnalogBind when it's requested instead of constantly updating the stored one whenever any text or value changes. I don't really care either way though.

Morilli:
> Auto Tab should at least advance from + to - within an axis, if not between axes. Otherwise LGTM.

That would be nice, although I'm not sure this is easily doable right now because the auto-tab functionality relies on the `InputCompositeWidget` to handle it and that one is... annoying to construct at the very least (no winforms editor support so it needs to be positioned and scaled manually, needs effective modifier keys for some reason so these need to be passed from somewhere...), and if we want to allow multiple binds then auto-tab can't work anyway.

PR #4677, opened 2026-04-12
Improve compatibility with NVDA screen reader

### LINKED ISSUE #4663: screen reader accessibility for UI
The UI is currently not accessible with screen readers like Narrator or NVDA, could this please be looked into. I am not sure if this app works with keyboard alone for stuf like the alt menu and similar, as no speech via screen reader.

YoshiRulz:
[dev build for branch | Lethal-Lawnmower:master](https://nightly.link/Lethal-Lawnmower/BizHawk/workflows/ci/master?preview)

### DIFF
--- a/docs/ACCESSIBILITY_README.md
+++ b/docs/ACCESSIBILITY_README.md
@@ -1,0 +1,72 @@
+# Accessible BizHawk
+
+An accessibility fork of [BizHawk](https://github.com/TASEmulators/BizHawk), the multi-system emulator developed by the TASVideos community.
+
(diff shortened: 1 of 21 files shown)


YoshiRulz:
@Lethal-Lawnmower Please rebase, since I cherry-picked some of your changes already and it's made a merge conflict. It would be nice to have it split into individual commits too.

If the `MenuItem` stuff is necessary, manually recreating every Form's menu is not a sustainable solution. We could have something like `FormBase` do it automatically at runtime, or we could do the replacement at design-time.

PR #4678, opened 2026-04-12
[Feature] Use librashader for .slangp (RetroArch) shader support

### LINKED ISSUE #4198: Adding librashader same as ares
Hi Bizhawk Team

in case you didnt noted, there are https://github.com/SnowflakePowered/librashader that https://github.com/ares-emulator/ares use to support RetroArch slang Shaders, so it will be nice if BizHawk add it too

thanks

mr-ke:

[dev build for branch | mr-ke:feature/librashader](https://nightly.link/mr-ke/BizHawk/workflows/ci/feature%2Flibrashader?preview)

## Summary

This PR integrates [librashader](https://github.com/libretro/librashader) as a modern shader filter backend for BizHawk, replacing the legacy Cg/GLSL shader system with support for modern slangp and glslp shader presets.

### New Files
- `BizHawk.Client.Common/DisplayManager/Filters/Librashader.cs` - Native bindings for librashader.dll
- `BizHawk.Client.Common/DisplayManager/Filters/LibrashaderFilter.cs` - Filter implementation for the display pipeline

### Modified Files
- `BizHawk.Client.Common/DisplayManager/DisplayManagerBase.cs`
  - Added `LibrashaderFilter` field and `RefreshLibrashader()` method
  - Integrated librashader into `BuildDefaultChain()` (TargetDisplayFilter=4)
  - Added disposal of filter resources

- `BizHawk.Client.EmuHawk/config/DisplayConfig.cs`
  - Added new RadioButton for librashader selection
  - Split file filters: `.cgp` for legacy UserFilter, `.slangp/.glslp` for librashader
  - Auto-select appropriate filter type based on file extension

- `BizHawk.Client.EmuHawk/config/DisplayConfig.Designer.cs`
  - Added UI controls for librashader option

- `BizHawk.Client.EmuHawk/MainForm.Events.cs`
  - Added `RefreshLibrashader()` call after DisplayConfig changes

## Features
- Support for modern slangp and glslp shader presets via librashader
- Automatic shader chain initialization and management
- Proper OpenGL context integration
- Debug logging via `Util.DebugWriteLine()` (DEBUG builds only)

## Requirements
- `librashader.dll` must be placed in `dll\` directory
- OpenGL display method must be selected

## Testing
Tested with various slangp shader presets (VHS, CRT effects) on NES and other cores.

Check if completed:


vadosnaprimer:
WOW

What's the workflow to use of an arbitrary libretro shader with this? For example https://github.com/libretro/slang-shaders/blob/master/crt/crt-royale.slangp

mr-ke:
> WOW
> 
> What's the workflow to use of an arbitrary libretro shader with this? For example https://github.com/libretro/slang-shaders/blob/master/crt/crt-royale.slangp

Hi, good question, I guess you could follow below steps:
1. git clone https://github.com/libretro/slang-shaders.git
2. copy crt/crt-royale.slangp (preset file) and its depend *.slang into the Folder you want
3. In menu of Config->Display->display method tab->OpenGL (have not port direct3D part yet)
4. In menu of Config->Display->librashader->select, choose your presets you store as upon step
5. Start to play

src/BizHawk.Client.Common/DisplayManager/Filters/Librashader.cs
```diff
@@ -0,0 +1,157 @@
...
+			}
+
+			Util.DebugWriteLine("[librashader] Successfully loaded librashader.dll");
+			_loaded = true;
+			LoadFunctions();
+			return true;
+		}
+
+		[DllImport("kernel32.dll", SetLastError = true)]
+		private static extern IntPtr LoadLibrary(string lpFileName);
```
YoshiRulz: Use a direct `[DllImport]` if possible ([example](https://github.com/TASEmulators/BizHawk/blob/master/src/BizHawk.Common/checksums/LibBizHash.cs)), or else use our `[BizImport]` abstraction ([example](https://github.com/TASEmulators/BizHawk/blob/master/src/BizHawk.Client.EmuHawk/RetroAchievements/LibRCheevos.cs)).

src/BizHawk.Client.EmuHawk/config/DisplayConfig.cs
```diff
@@ -373,37 +381,44 @@ private void RefreshState()
 
 		private void BtnSelectUserFilter_Click(object sender, EventArgs e)
 		{
+			var fsFilter = rbLibrashader.Checked ? LibrashaderFSFilterSet : CgpFSFilterSet;
```
YoshiRulz: If this is updating the radio button selection based on what file is loaded, why not have the filter show all 3 kinds of files?

src/BizHawk.Client.EmuHawk/config/DisplayConfig.cs
```diff
@@ -373,37 +381,44 @@ private void RefreshState()
...
-					var filter = new RetroShaderChain(_gl, cgp, Path.GetDirectoryName(choice));
-					ok = filter.Available;
-					errors = filter.Errors;
-				}
-				catch {}
-				if (!ok)
+				rbLibrashader.Checked = true;
+			}
+			else
+			{
```
YoshiRulz, line 404: `return;` from the `if` block and remove the `else`

YoshiRulz (commented):
Thank you!

src/BizHawk.Client.EmuHawk/config/DisplayConfig.cs
mr-ke: Hi,
    Sharp spot, the "else block" is more or less the backwards compatible equivalent when old cgp filter is select, if a empty file with .cgp extension is select, original implementation will as early as possible thow out a ExceptionBox. Are we going to clean it up?

src/BizHawk.Client.EmuHawk/config/DisplayConfig.cs
mr-ke: Hi,
    The handling to toggle radio button again after file been selected seems make user a little confused.
    The intention of existing design as below:
    If we select User(Cgp) radio button, then only *.cgp will be availabe to select.
    If we select Librashader radio button,  then `[*.slangp, *.glslp]` will be availabe to select.
    So the handling after we select `**` conflict type file (If you arbitrarily use `*.*` `**` should be removed, do I read it correctly??

YoshiRulz: You've removed the part that changed the radio button selection, so this is no longer relevant.

src/BizHawk.Client.EmuHawk/config/DisplayConfig.cs
YoshiRulz: I simply wanted you to flip the conditional so the diff was smaller: 32dc37e2a33aacaa6240c99b1bba8f384b7ff724

src/BizHawk.Client.Common/DisplayManager/Filters/LibrashaderFilter.cs
```diff
@@ -0,0 +1,284 @@
...
+		private static GL GetSilkGL(IGL_OpenGL igl)
+		{
+			var glField = typeof(IGL_OpenGL).GetField("GL", BindingFlags.NonPublic | BindingFlags.Instance);
+			return glField?.GetValue(igl) as GL;
+		}
+
+		private static uint GetTexID(ITexture2D texture)
+		{
+			if (texture == null || TexIDField == null) return 0;
+			return (uint)TexIDField.GetValue(texture);
```
YoshiRulz: I think this is equivalent to `return (texture as OpenGLTexture2D)?.TexID ?? default;`. Did you use reflection because `OpenGLTexture2D` is `internal`?

src/BizHawk.Client.Common/DisplayManager/Filters/LibrashaderFilter.cs
```diff
@@ -0,0 +1,284 @@
...
+			{
+				return SDL2OpenGLContext.GetGLProcAddress(procName);
+			}
+			catch
+			{
+				return IntPtr.Zero;
+			}
+		}
+
+		private static string PtrToStringUTF8(IntPtr ptr)
```
YoshiRulz: We already have `Mershul.PtrToStringUTF8`, which isn't identical to this, but I think it does the same thing.

src/BizHawk.Client.EmuHawk/config/DisplayConfig.Designer.cs
```diff
@@ -207,6 +209,17 @@ private void InitializeComponent()
 			this.rbUser.Text = "User";
```
YoshiRulz: Since the new option is also a user shader, and will be the preferred way to add shaders, we should probably rename this old option to something like "Legacy".

src/BizHawk.Client.Common/DisplayManager/Filters/LibrashaderFilter.cs
mr-ke: > I think this is equivalent to `return (texture as OpenGLTexture2D)?.TexID ?? default;`. Did you use reflection because `OpenGLTexture2D` is `internal`?

If we do not use reflection to access, then `InternalsVisibleTo Include="BizHawk.Client.Common" `is needed for project BizHawk.Bizware.Graphics, maybe isolate this changeset without change expose charastic of BizHawk.Bizware.Graphics would be a option as better, what do you say?
Same reason apply to drawFbo

src/BizHawk.Client.Common/DisplayManager/Filters/LibrashaderFilter.cs
mr-ke: > We already have `Mershul.PtrToStringUTF8`, which isn't identical to this, but I think it does the same thing.

Do you refer to native Marshal.PtrToStringUTF8 ?? But BizHawk.Client.Common.csproj tell TargetFramework is net48, so does it support this Marshal.PtrToStringUTF8?

SuuperW: No, he does in fact mean `BizHawk.Common.Mershul`

src/BizHawk.Client.Common/DisplayManager/Filters/LibrashaderFilter.cs
YoshiRulz: @CasualPokePlayer https://github.com/TASEmulators/BizHawk/pull/4678/commits/dce8e0740dd12582aafbc0b455682a279506fde8 y/n

CasualPokePlayer: Don't put this librashader handling in `BizHawk.Client.Common` in the first place, put it in `BizHawk.Bizware.Graphics`. This should outright be part of the IGL so it can handle the IGL impl specific details (`BizHawk.Client.Common` should only be calling IGL APIs, it's not supposed to handle raw OpenGL commands).

This impl using raw OpenGL commands also means the feature is locked to the OpenGL display method, rather than also being available to the (default!) Direct3D11 display method (which librashader can support just fine).

mr-ke: > Don't put this librashader handling in `BizHawk.Client.Common` in the first place, put it in `BizHawk.Bizware.Graphics`. This should outright be part of the IGL so it can handle the IGL impl specific details (`BizHawk.Client.Common` should only be calling IGL APIs, it's not supposed to handle raw OpenGL commands).
> 
> This impl using raw OpenGL commands also means the feature is locked to the OpenGL display method, rather than also being available to the (default!) Direct3D11 display method (which librashader can support just fine).

Well, you are spot on the view of what should  be in BizHawk.Client.Common and what should be in BizHawk.Bizware.Graphics, in the meanwhile, I've made a progress on https://github.com/mr-ke/BizHawk/commits/feature/librashader_d3d11, which I also make librashader works on Bizhawk default d3d11 mode, do you consider it a idea we consist d3d11 together or we need to firstly put these architecture adjustment at first place for OpenGL part then consist d3d11 afterwards.
Looking forward to hear all your ideas in mind.

YoshiRulz: If you got D3D working, you can include that in this PR too.

mr-ke: Here it goes with the joint work d3d & openGL (Also abstract common part into LibrashaderFilterBase.cs).
I've done a instrument on filter_chain_create (shader compile at boot) for both d3d & openGL as below:
Preset: vhs_and_crt_godot.slangp (complex enough)
**OpenGL 140ms (Cold start)
OpenGL 55ms (2nd start)**
D3D11  1600ms (Cold start)
D3D11  55ms (2nd start)    2026/4/21 updated!!!
That's make sense OpenGL's GLSL is much faster than D3D's HLSL compiler.
Also OpenGL's shader cache works along with D3D's shader cache not work.  (You can inspect at %LOCALAPPDATA%\librashader)
I've take a while to check option I set for D3D11, **no obvious basis prove we use wrong option for D3D11** but librashader.db.1 can not generated to accelerate D3D11's shader compile.

**Updated 2026/4/21** I've identified why shader cache is not generated as expected in D3D11 mode, it is due to a concurrent issue in librashader.dll, after carefull inspect the rust code, the cache issue can be work around by  attach **D3D11_CREATE_DEVICE_SINGLETHREADED** flag to D3D device created.
Please help me check if https://github.com/TASEmulators/BizHawk/pull/4678/commits/0e698fd424350a0833e0115898290b28deabf145 really matter on other cases.
I understood that **EncoreGLTextureProvider** and **MelonDSGLTextureProvider** use GL Interop to share the GL texture to D3D, but it still have fallback methods, do I read it correctly?

@YoshiRulz  @CasualPokePlayer

CasualPokePlayer: GL interop is used as a fast method to transfer the GL texture to the display method. The fallback method is slow. This slowness doesn't matter too much at 1x internal resolution, it ends up mattering a lot if you decide to bump up the internal resolution (which I imagine ends up massively overlapping with the userbase that would use these shaders).

Even then, if there's some concurrency occurring within librashader it does not make sense to make the D3D11 device single threaded, if you're saying there's threading in librashader then the single threaded flag should not be set.

mr-ke: > GL interop is used as a fast method to transfer the GL texture to the display method. The fallback method is slow. This slowness doesn't matter too much at 1x internal resolution, it ends up mattering a lot if you decide to bump up the internal resolution (which I imagine ends up massively overlapping with the userbase that would use these shaders).
> 
> Even then, if there's some concurrency occurring within librashader it does not make sense to make the D3D11 device single threaded, if you're saying there's threading in librashader then the single threaded flag should not be set.

Don't make D3D resource create as single thread is fine since essentially the cause of shader cache not correctly handled comes from a barely perceptible issue at librashader https://github.com/SnowflakePowered/librashader/issues/200#issue-4327724968.
I kindly asked here to explore if single thread FLAG can be uniformly be process, if it is so sensitive for some reason like some core must use GL texture, and the only feasibility is use **"GL Interop"**, then it is not good idea.
Kindly review if other part fit all we want then I can still put efforts further.

@CasualPokePlayer  @YoshiRulz

chyyran:
just fyi @mr-ke I've merged in a fix for serializing writes to the cache in a process, that should be released with librashader 0.11 later next week

PR #4681, opened 2026-04-15
Lua tastudio branches

### LINKED ISSUE #4073: Allow saving new branches via Lua?
Surprisingly, as far as I can tell, there's not a way to create a new branch via Lua past slot 10. To me, there should be an explicit function that creates a new branch like the `+` button in TAStudio.

If there is a way, it's not well-documented. `savestate.save` is seemingly ignored by TAStudio. `savestate.saveslot` can save up to 10 branches, but anything outside of the range 1 to 10 is rejected.

And then there is [`tastudio.setbranchtext(string text, [int? index = nil])`](https://github.com/TASEmulators/BizHawk/blob/d513632cf96fe6ff98d7c36a776a6c4969963b35/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs#L368) which "adds the given message to the existing branch, or to the branch that will be created next if branch index is not specified". For a better API, queued naming seems like it should be deprecated in favor of a parameter in a branch creation function.

YoshiRulz:
[dev build for branch | TASEmulators:lua-tastudio-branches](https://nightly.link/TASEmulators/BizHawk/workflows/ci/lua-tastudio-branches?preview)

resolves #4073

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -448,6 +448,19 @@ namespace BizHawk.Client.EmuHawk
 			}
 		}
 
+		[LuaMethodExample("""
+			tastudio.set_branch_text_by_id("97021544-2454-4483-824f-47f75e7fcb6a", "success at frame "..emu.framecount());
+		""")]
+		[LuaMethod(
+			name: "set_branch_text_by_id",
+			description: "Loads the branch with the given ID.")]
+		public void SetBranchTextByID(string id, string/*?*/ text)
+		{
+			var found = GetBranchObjByID(id: id, callerFunctionName: "set_branch_text_by_id");
+			if (found is not null) found.UserText = text;
+			// else already logged error
+		}
+
 		[LuaMethodExample("local nltasget = tastudio.getbranches( );")]
 		[LuaMethod("getbranches", "Returns a list of the current tastudio branches.  Each entry will have the Id, Frame, and Text properties of the branch")]
 		[return: LuaZeroIndexed]
--- a/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
@@ -514,18 +514,21 @@ namespace BizHawk.Client.EmuHawk
 			description: "Loads the branch with the given ID.")]
 		public void LoadBranchByID(string id)
 		{
+			var found = GetBranchObjByID(id: id, callerFunctionName: "load_branch_by_id");
+			if (found is not null) Tastudio.LoadBranch(found);
+			// else already logged error
+		}
+
+		private TasBranch/*?*/ GetBranchObjByID(string id, string callerFunctionName)
+		{
 			if (!Guid.TryParseExact(id, format: "D", out var parsed))
 			{
-				Log($"[tastudio.load_branch_by_id] not a valid UUID: {id}");
-				return;
+				Log($"[tastudio.{callerFunctionName}] not a valid UUID: {id}");
+				return null;
 			}
 			var found = Tastudio.CurrentTasMovie.Branches.FirstOrDefault(b => b.Uuid == parsed);
-			if (found is null)
-			{
-				Log($"[tastudio.load_branch_by_id] no such branch: {id}");
-				return;
-			}
-			Tastudio.LoadBranch(found);
+			if (found is null) Log($"[tastudio.{callerFunctionName}] no such branch: {id}");
+			return found;
 		}
 
 		[LuaMethodExample("tastudio.loadbranch(0)")]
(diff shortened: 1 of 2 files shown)


src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -524,6 +568,45 @@ public void LoadBranch(int index)
 			}
 		}
 
+		[LuaMethodExample("""
+			local branch_id = tastudio.create_branch("automated attempt");
+		""")]
+		[LuaMethod(
+			name: "create_branch",
+			description: "Creates a new branch at the end of the list, and returns its UUID.")]
+		public string CreateBranch(string/*?*/ text)
```
YoshiRulz, line 577:
```suggestion
		public string CreateBranch(string/*?*/ text = null)
```

SuuperW:
I think these would be good to have.

You could reduce code duplication with the new methods, though. It looks like all but the last line of load/delete/set are identical?

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -447,6 +447,19 @@ public void SetBranchText(string text, int? index = null)
 			}
 		}
 
+		[LuaMethodExample("""
+			tastudio.set_branch_text_by_id("97021544-2454-4483-824f-47f75e7fcb6a", "success");
+		""")]
```
SuuperW, line 453: I think "success" is a strange branch text to use as the example here. Something more obviously a branch's text (even "branch text" or "branch name") would be more appropriate.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -505,6 +518,31 @@ public LuaTable GetBranchInput(string branchId, int frame)
...
+			var found = GetBranchObjByID(id);
+			if (found is not null) Tastudio.LoadBranch(found);
+			// else already logged error
+		}
+
+		private TasBranch/*?*/ GetBranchObjByID(string id)
+		{
+			if (!Guid.TryParseExact(id, format: "D", out var parsed))
+			{
+				Log($"not a valid UUID: {id}");
```
SuuperW: Something indicating that this message is about a branch operation would be good. Seeing this pop up in the output when running a Lua script, the user (even if they are the script author) might have no idea what this is about or what a UUID is. "not a valid branch id: {id}" would be better, I think.

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -524,6 +562,36 @@ public void LoadBranch(int index)
 			}
 		}
 
+		[LuaMethodExample("""
+			local branch_id = tastudio.create_branch("automated attempt");
+		""")]
+		[LuaMethod(
+			name: "create_branch",
+			description: "Creates a new branch at the end of the list, and returns its UUID.")]
```
SuuperW: Similarly here I think UUID should just be ID. (and the description for delete)

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
```diff
@@ -524,6 +562,36 @@ public void LoadBranch(int index)
...
+		[LuaMethod(
+			name: "create_branch",
+			description: "Creates a new branch at the end of the list, and returns its UUID.")]
+		public string CreateBranch(string/*?*/ text = null)
+		{
+			Tastudio.BookMarkControl.Branch();
+			var index = Tastudio.CurrentTasMovie.Branches.Count - 1;
+			var branch = Tastudio.CurrentTasMovie.Branches[index];
+			branch.UserText = text;
+			Tastudio.BranchSavedCallback?.Invoke(index);
```
SuuperW, line 578: I'd prefer changing the `Branch` method to return the instance, instead of grabbing it by index here.

Also, I think the `Branch` method should handle calling `BranchSavedCallback`. It looks like from reading related code that the callback isn't called if you branch with the context menu item, so you'd be fixing that too.

SuuperW (commented):
(and there might be stuff to update regarding the recent addition of LuaCATS documentation, but I haven't looked at that)

src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs
YoshiRulz: The other existing callsites use the index.

The callback can't be moved to the end of `Branch` because of this:
https://github.com/TASEmulators/BizHawk/blob/a2e8eed8b6aa50d2a9784c23540e0129e99cc210/src/BizHawk.Client.EmuHawk/tools/TAStudio/BookmarksBranchesBox.cs#L227-L229

SuuperW: > The other existing callsites use the index.

So? They don't grab the branch, and updating `Branch` to return the branch will not require changing any existing uses.

> The callback can't be moved to the end of Branch because of this:

Because of what? The branch not having text at the end of `Branch`? It should be easy enough to place a call to `EditBranchTextPopUp` inside `Branch` depending on a parameter.

SuuperW:
Is there a reason you decided to keep "UUID" in the error message?

YoshiRulz:
https://github.com/TASEmulators/BizHawk/blob/dc7c820e20dc95617c532c29fa3ec07036a2a886/src/BizHawk.Client.EmuHawk/tools/Lua/Libraries/TAStudioLuaLibrary.cs#L537-L539
That is an accurate description of the failure.

SuuperW:
But at a level of detail most users will not understand. You changed the "UUID" in the Lua documentation to just "ID", so I thought this one might be an oversight.

PR #4689, opened 2026-04-17
Add `#nullable` annotations to Lua libs

kalimag:
Add `#nullable` annotations to Lua libraries for #3755 and #4688

TODO:
- [x] `CommLuaLibrary` (this one is a mess)
- [x] `FormsLuaLibrary`
- [x] `GuiLuaLibrary`
- [x] `MovieLuaLibrary`
- [x] `TAStudioLuaLibrary`
- [ ] Add `#nullable` to Lua libs that currently don't have any reference types in their API surface?

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/SaveStateApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/SaveStateApi.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System.IO;
 
 namespace BizHawk.Client.Common
--- a/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
+++ b/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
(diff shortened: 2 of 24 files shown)


SuuperW:
Most of the things that you added `required` for are not actually set via object initializers, since most Lua libraries are constructed through reflection.
I'm not sure if this is really a problem with using `required`. (It's a problem with reflection, but that isn't what this PR is about.)

kalimag:
> Most of the things that you added `required` for are not actually set via object initializers

Yeah, ultimately it's just about picking a way to shut up the analyzer. IMO `required` is the least bad option here, at least it expresses intent even if the requirement isn't actually compiler-enforced.

kalimag:
There is an annoying mismatch between C# and Lua semantics here:

In C# (and more importantly, BizHawk's Lua bindings), `Foo(string? bar)` and `Foo(string? bar = null)` are different. But in Lua, `foo(nil)` and `foo()` are normally the same thing, and there is no way to express the difference in LuaCATS annotations.

This is an issue in cases where the last non-optional parameter in a function is nullable, e.g. [`client.addcheat`](https://github.com/kalimag/BizHawk/blob/bab5e7dd2b67cfb8ea830749a69c697586298041/src/BizHawk.Client.Common/lua/CommonLibs/ClientLuaLibrary.cs#L421-L428). `client.addcheat(nil)` works (but is a no-op), while `client.addcheat()` throws an error, but LLS will warn on neither of them.

I think I'll remove/omit the `?` on such parameters if it doesn't make sense to pass `nil`, even if the method accepts `nil` without throwing an error.

YoshiRulz:
If `client.addcheat(nil)` is a no-op, `client.addcheat()` should be as well i.e. the parameter should have a default value of `null`. (IMO both should throw.)

YoshiRulz (changes requested):
There's a lot of changes to things other than nullability annotations in here...
We don't use `is { } nonNullObj`.
You can't make any of those member props `required` because all the subclasses of `LuaLibraryBase` are instantiated by reflection.

kalimag:
>You can't make any of those member props `required` because all the subclasses of `LuaLibraryBase` are instantiated by reflection.

What's your preferred alternative?

SuuperW:
> You can't make any of those member props required because all the subclasses of LuaLibraryBase are instantiated by reflection.

Not all. Only the ones in `Common` (which is most).
I think it's fine, though. The thing to fix is getting rid of the reflection use, but that's outside the scope of this PR. At least `required` will be correct as soon as someone tries to construct any of the classes in the normal non-reflection way.

Issue #4005, opened 2024-08-17
Make all `Form`s inherit `FormBase` instead


YoshiRulz:
This was always intended to become a universal base class. The goal was to inject anything into every form if that was ever necessary, but it's since become clear that proper input handling involves passing the focused window's inputs (or a subset thereof) to the root, so this would have multiple benefits.

(TODO put a nice checklist here)

PR #4690, opened 2026-04-18
refactor: migrate first batch of Forms to FormBase

mvanhorn:
Migrated 7 forms from `Form` to `FormBase` as tracked in #4005:

- `ArchiveChooser`
- `FFmpegWriterForm`
- `GifWriterForm`
- `JmdForm`
- `SynclessRecordingTools`
- `AutofireConfig`
- `ControllerConfig`

Each change is a one-line base class swap. The remaining forms from the checklist can follow in subsequent PRs.

Partial fix for #4005

### DIFF
--- a/src/BizHawk.Client.EmuHawk/AVOut/FFmpegWriterForm.Designer.cs
+++ b/src/BizHawk.Client.EmuHawk/AVOut/FFmpegWriterForm.Designer.cs
@@ -136,7 +136,6 @@
             this.Name = "FFmpegWriterForm";
             this.ShowIcon = false;
             this.StartPosition = System.Windows.Forms.FormStartPosition.CenterParent;
-            this.Text = "Choose Video Format";
             this.ResumeLayout(false);
             this.PerformLayout();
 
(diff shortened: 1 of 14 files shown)


YoshiRulz (dismissed):
Thanks for stepping up, but just changing the supertype isn't enough, you at least need to override `WindowTitleStatic` (which is not `abstract` for technical reasons). These will all throw when opened because of that, so clearly you haven't done any testing either. The main thing to test would be the Alt+key "mnemonics".

mvanhorn:
@YoshiRulz you're right, and thank you for the direct feedback. The supertype change alone would have thrown on every one of these forms - I missed that `WindowTitleStatic` is virtual-not-abstract and that FormBase's `Text` setter throws outside DesignMode.

Pushed 59a6fcc which for each of the seven forms:
- adds `protected override string WindowTitleStatic => "<title>";` using the exact title that was in the Designer
- removes `this.Text = "<title>";` from the Designer

Followed the pattern from existing FormBase-derived forms (CDL, HexEditor, GBAGPUView).

Honest note on testing: I'm on macOS without a working .NET build, so I cannot open these forms or drive the Alt+key mnemonics from here. If the pattern checks out for you and you'd like, I'm happy to leave this PR paused until I can set up a Windows env to verify, or close it and resubmit from there. Totally understand if you'd rather not merge something I can't demonstrate running.

Issue #3976, opened 2024-07-23
Add `userdata` to tastudio greenzone states


vadosnaprimer:
The `userdata` lua library currently only works with regular savestates but not tastudio greenzone (which uses memory savestates iirc?), which makes it problematic to restore lua values from past tastudio frames for things that can't be calculated on the fly for every given frame.

I have a counter that increments on a memory hook, and I need to store it in savestates in some way so I could go back without it infinitely incrementing on every advance and rewind, and I need to it also go back and show me values it had back then.

YoshiRulz:
Are you able to use a custom column for the same purpose?

vadosnaprimer:
As a theoretic workaround maybe, but it'd be too different from the intended workflow (custom columns are for things that change all the time and you need to see the dynamics), and doesn't make sense to just generally use one instead of the other.

PR #4709, opened 2026-04-24
Add cheated flag to `PagedStateManager` and show in blue in TAStudio

### LINKED ISSUE #4570: TAStudio must clear state cache after a hex edit / poke / freeze
Mix between feature request and a bug report. The video here explains it better than a lengthy report:

https://www.youtube.com/watch?v=UjnguAbnapQ

Basically, modifying a value in memory should invalidate all cached states that have been stored ahead, just like changing an input would. This ensures the modification is properly stored if a re-record is made thereafter.

YoshiRulz:


SuuperW:
OK well Idk why I didn't think of using an event for this.

This implementation is not correct, though. It can incorrectly indicate cheat state in either direction:
1) Emulate so that you have some clean states.
2) Rewind to frame 0.
3) Poke.
4) Load a clean state.
5) Emulate to make more states; they will be marked as cheated.

OR

1) Make a branch with a cheated state.
2) Edit an input, so that all states are clean.
3) Load the branch.
4) Emulate to make more states; they will be marked as clean.

EDIT: Also it looks like this won't be able to handle situations were an external tool loads a state using the `IStatable` emulator service, or Lua script loads a state with `memorysavestate`. (Actually I can test this real quick: using the `GBA/SonicAdvance_CamHack.lua` script does indeed result in incorrectly marking states as cheated.)

EDIT 2: I noticed your title says they will show as blue. Idk what went wrong, but I am not seeing any blue.
image

YoshiRulz:
The colour config handling must fill in zeroes for a missing value instead of pulling from the defaults. From a clean config it should work.

I'm not sure how to move forward on the correctness issues. A call to `IStatable.LoadStateBinary` wouldn't go via the frontend at all, right? So maybe this flag actually needs to be part of each core, like `IInputPollable.IsLagFrame`, in order to be persisted properly.

SuuperW:
> So maybe this flag actually needs to be part of each core, like IInputPollable.IsLagFrame, in order to be persisted properly.

This was my initial thought, which is why I did not think this feature could realistically be implemented.

But I had an idea that maybe can work: wrap the `IStatable` service in a new `IStatable` implementation that is controlled by the frontend and essentially adds a header to the core's states. The header can include a cheated flag and potentially user data too, addressing #3976. I do not know how much refactoring will be needed to ensure the raw, unwrapped core's `IStatable` is ever accessed by tools/API classes.

Issue #4714, opened 2026-04-26
DOSBox-X Core leaks memory on big games


SergioMartin86:
Every time you reload a DOSBox-X game, some of its memory ends up leaked -- especially on those big 2GB hdd games. 

Reproduce:

1- Pick a big (2GB) .hdd image to open with DOSBox-X
2- Open it
3- Close the ROM
4- Repeat 2 and 3 many times

A 2GB leak will remain no matter how many times you close it. I suspect the HDD file remains loaded in some kind of file cache. 

<img width="917" height="22" alt="Image" src="https://github.com/user-attachments/assets/c53c1ff4-7071-41fb-b47b-4837b6b2f2ef" />

YoshiRulz:
Is the memory freed when you open a different rom?

CasualPokePlayer:
Repeating what I said on Discord, I would suspect this is just a matter of the managed memory buffer that has the HDD not being "freed" immediately because the GC does not free memory immediately (this memory probably ends up getting freed if you close the ROM and wait around doing nothing for like 10 minutes or so)

PR #4715, opened 2026-04-26
Partial fix to DOSBox-X Mem leak

SergioMartin86:
This PR adds some measures to reduce leakage on the DOSBox-X core, as described in #4714 

- Timely deload of read only HDD files right after init (no longer needed)
- Freeing of the RW HDD file buffer upon closure (added deInit() function)

There's one more leak remaining we'll need to find eventually. Not sure it's DosBox or Bizhawk related

### DIFF
--- a/src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
+++ b/src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
@@ -641,9 +641,6 @@ namespace BizHawk.Emulation.Cores.Computers.DOS
 				}
 			}
 
-			// Freeing up core's own resources
-			_libDOSBox.deInit();
-
 			base.Dispose();
 		}
 	}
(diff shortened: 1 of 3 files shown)


SergioMartin86:
This fix is ready for merge, btw

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.cs
```diff
@@ -632,6 +641,9 @@ public override void Dispose()
 				}
 			}
 
+			// Freeing up core's own resources
+			_libDOSBox.deInit();
```
CasualPokePlayer: This is pointless in practice. Any memory allocated within the waterbox should be destroyed by the waterboxhost. You're not allocating with a normal heap, you're allocating within the waterboxhost's memory block, which by the view of the host OS is already allocated fully by the waterboxhost and should be released by the waterboxhost. There's nothing in the core code to do here (the OS won't recognize free'ing in core code as actually freeing the memory, the memory is still allocated by the waterboxhost).

SergioMartin86: Right, I forgot about this

PR #4716, opened 2026-04-27
Revert `FileWriter`/`FileWriteResult`

YoshiRulz:
per https://github.com/TASEmulators/BizHawk/pull/4400#issuecomment-4324502314

The idea is for a2e8eed8b6aa50d2a9784c23540e0129e99cc210 + this changeset to have the same file state as [the ~200 commits less #4400] rebased onto 7b07976237a5339496c158de9ea768e0fa8cc908 edit: which I've also [started on here](https://github.com/TASEmulators/BizHawk/tree/without-filewriter).


SuuperW:
I expect there were a significant number of conflicts to resolve here. I think the risk of new bugs due to incorrect conflict resolution plus the re-introduction of the issues fixed by that PR (such as closing when the user says to cancel due to unsaved changes) are worse than the potential for still undiscovered regressions in the PR you want to revert.

And there isn't any new reason to expect more testing or review for it outside release, since the time it was merged.

PR #4718, opened 2026-04-30
Handle mouse buttons when the active form changes

### LINKED ISSUE #4717: First mouse input press ignored when main window is out-of-focus (BizHawk 2.11.1 dev)
This makes TASing DS with TAStudio slightly more annoying, because any time you click on the TAStudio window, you have to click twice to do mouse inputs again in recording mode.

Bisect shows this is a regression from 3eb9cb9d765bfca829ea66bb514fb125bca90377 ( @SuuperW )

So a regression from 2.11 as well.

(Tested on Windows 11)

SuuperW:
Before [3eb9cb9](https://github.com/TASEmulators/BizHawk/commit/3eb9cb9d765bfca829ea66bb514fb125bca90377) the state of mouse buttons was not checked by `Input` when the active form did not "want mouse" (meaning, clicking on that form should not result in controller inputs being pressed or hotkeys being triggered). After, they are checked and so no non-"ignored" mouse input event would be generated if you click on the main form when TAStudio was active.

This PR fixes that by having `Input` detect when the active form changes, and then clears its mouse state and generates new mouse events. This only happens when moving from a non-mouse-wanting form to a mouse-wanting form, which seems consistent with the behavior of 2.11 but might not be ideal. This is because clicking on TAStudio while the main form is active will result in a mouse input most of the time; `Input` sees the mouse button being pressed before `Form.ActiveForm` changes. If we clear mouse state and generate new events going the other way too then any controller inputs bound to the mouse button will be quickly released after `ActiveForm` changes, but the inputs will still happen. But even with this issue, this is an improvement over 2.11. In 2.11 if you click on TAStudio while the main form is active, any controller input bound to the mouse button will be pressed and will not be released when the mouse button is released. Now, it will be released when the mouse button is released.

I also had to change call sites of `Input.ControlInputFocus`, which previously was being called in forms' `OnActivated` and `OnDeactivate` methods. This doesn't make any sense to me, since the call to `Input.ControlInputFocus` in `OnDeactivate` is not required; mouse inputs won't happen when the form is active because `Input` checks which form is active. EDIT: It had to be changed because if the form does not remain in the wanting-mouse state, `Input` may see it does not want mouse (because `ControlInputFocus` had not yet been called in `OnActivated`) and thus not produce new mouse events when the active form switches. It's a race condition.

Please test, @RetroEdit.

Fixes #4717 

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/Input/Input.cs
+++ b/src/BizHawk.Client.EmuHawk/Input/Input.cs
@@ -23,6 +23,8 @@ namespace BizHawk.Client.EmuHawk
 		}
 
 		private readonly HashSet<Control> _wantingMouseFocus = new HashSet<Control>();
+		private bool _wantsMouse;
+		private Control/*?*/ _lastControl;
 
 #pragma warning disable CA2211 // public field
 		public static Input Instance;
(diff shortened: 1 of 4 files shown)


RetroEdit:
Seems to have resolved the issue.

PR #4722, opened 2026-05-03
Remove the old digital-->analog controls from the N64 cores

YoshiRulz:
https://github.com/TASEmulators/BizHawk/commit/24d99733d00fe1b2f5a5a6597c043483f6fb83ce + this completes the migration

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64Input.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64Input.cs
@@ -34,47 +34,10 @@ namespace BizHawk.Emulation.Cores.Nintendo.N64
 			ThisFrameInputPolled = false;
 		}
 
-		private const sbyte _maxAnalogX = 127;
-		private const sbyte _minAnalogX = -128;
-		private const sbyte _maxAnalogY = 127;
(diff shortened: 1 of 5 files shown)


Morilli:
We should be able to remove all references to this now, see https://github.com/TASEmulators/BizHawk/commit/a42251124d5f501047f9eb14e035d5a3e60d0246.

src/BizHawk.Emulation.Cores/Consoles/Nintendo/N64/N64Input.cs
```diff
@@ -34,47 +34,10 @@ public void ShiftInputPolledBools()
 			ThisFrameInputPolled = false;
 		}
 
-		private const sbyte _maxAnalogX = 127;
-		private const sbyte _minAnalogX = -128;
-		private const sbyte _maxAnalogY = 127;
-		private const sbyte _minAnalogY = -128;
-
 		/// <param name="i">player number (from 1)</param>
 		public static (sbyte X, sbyte Y) GetStickValues(IController Controller, int i)
```
Morilli, line 38: I don't really like that this function exists in the first place, at least not in this file that's part of the mupen core, but I don't really care too much because ideally this entire folder will be replaced soon anyway.

Morilli (approved):
Looks good.

PR #4743, opened 2026-05-20
Turbo fix

SuuperW:
I decided to go with the option of turning TAStudio's `_seekingTo` into a property backed by `MainForm.PauseOnFrame`. I avoided the issue of using MainForm's end-of-seek logic by having it update tools at the end of a turbo seek, in the normal place (which happens before end-of-seek logic) instead of in the end-of-seek logic. Cancelling a seek by setting `PauseOnFrame` to `null` no longer updates tools, because (a) this isn't needed anymore when a seek ends; and (b) that logic was unable to have the core render a frame, so it only did part of what needs to happen when a seek is cancelled.

In order to properly cancel a turbo seek, we need to emulate one more frame. So TAStudio's `StopSeeking` has additional logic for that now.

Clicking the icon at the bottom of the main form  to "stop seeking" does not actually set `PauseOnFrame`, it just pauses. So issues still exist there. I decided against fixing that because it's a separate issue and isn't trivial to fix. (setting `PauseOnFrame` there would break the pause at end of movie feature for regular movies)

Pausing during a turbo seek will not result in rendering or tools updating. This maybe should be changed, but that also is a separate issue.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
+++ b/src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
@@ -108,6 +108,14 @@ namespace BizHawk.Client.EmuHawk
 			_shouldMoveGreenArrow = true;
 			if (_seekingTo == -1) return;
 
+			if (Config.TurboSeek && Emulator.Frame < _seekingTo)
+			{
+				// Turbo seek does various things, including telling the core to skip rendering most frames.
+				// Thus, we need the seek to end naturally.
+				_seekingTo = Emulator.Frame + 1;
+				return;
+			}
+
 			if (WasRecording && !skipRecModeCheck)
 			{
 				TastudioRecordMode();
@@ -116,7 +124,6 @@ namespace BizHawk.Client.EmuHawk
 
 			_seekingByEdit = false;
 			_seekingTo = -1;
-			MainForm.PauseOnFrame = null; // This being unset is how MainForm knows we are not seeking, and controls TurboSeek.
 			if (_pauseAfterSeeking)
 			{
 				MainForm.PauseEmulator();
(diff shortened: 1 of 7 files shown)


src/BizHawk.Client.EmuHawk/tools/TAStudio/TAStudio.ListView.cs
```diff
@@ -106,7 +106,15 @@ private void BeginAxisMouseEdit(int yPos)
...
 		{
 			_shouldMoveGreenArrow = true;
-			if (_seekingTo == -1) return;
+			if (SeekingTo == -1) return;
+
+			if (Config.TurboSeek && Emulator.Frame < SeekingTo)
+			{
+				// Turbo seek does various things, including telling the core to skip rendering most frames.
+				// Thus, we need the seek to end naturally.
+				SeekingTo = Emulator.Frame + 1;
```
Morilli: I understand the intention behind this but it doesn't seem to really work when the emulator is paused. Canceling a seek will just look like it didn't work because the seek will only be finished on the next frame. And emulating a frame in response to a cancel seek request doesn't seem right either. It's probably easier to just not do this.

SuuperW: Yeah, this isn't good when paused. I feel like pausing during a turbo seek should also do 1 frame advance with rendering+full tool updates, and if we did that then this thing would be redundant. And modifying pause behavior like that is not for this PR and maybe someone who actually uses turbo seek (not me) should do it.

Morilli (approved):
Looks fine to me as far as I can see. There may be some implications to not calling `UpdateAfter` when a seek is canceled, but honestly I don't know which and it probably doesn't matter too much. I like the consistency in the logic in `StepRunLoop_Core`  so if anything this may make more sense than before now.

SuuperW:
In addition to a canceled turbo seek potentially not having the correct frame rendered, some logic in update after may depend on update before (I have a Lua script like this) having happened. So I don't think we should be doing just that one part of updating.

PR #4748, opened 2026-05-25
[DOSBox-x] Adding key-activated alternative mouse resolution

SergioMartin86:

[dev build for branch | TASEmulators:dosboxMouseAltResolution](https://nightly.link/TASEmulators/BizHawk/workflows/ci/dosboxMouseAltResolution?preview)

Some DOS games operate with different resolutions during gameplay. One example is Syndicate which uses 800x600 during menuing:

image

and 800x500(+50 offset) during a mission:

image

This means no single mouse absolute resolution can be used to perfectly map it to both. That's why I am introducing a way to specify an alternative resolution (with offsets) that can be (de)activated with a key during gameplay. 

Check if completed:

### DIFF
--- a/src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.ISettable.cs
+++ b/src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.ISettable.cs
@@ -277,6 +277,27 @@ namespace BizHawk.Emulation.Cores.Computers.DOS
 			[DefaultValue(LibDOSBox.SVGA_MAX_HEIGHT)]
 			public uint MouseAbsoluteScreenHeight { get; set; }
 
+			[DisplayName("Mouse Internal Resolution (Left)")]
+			[Description("Indicates the left position (in pixels) of the internal mouse resolution. Useful for games (e.g., Syndicate) that implement an internal resolution that is different from the screen resolution. Must be activated manually during gameplay.")]
+			[DefaultValue(0)]
+			public uint MouseInternalResolutionLeft { get; set; }
+
+			[DisplayName("Mouse Internal Resolution (Right)")]
+			[Description("Indicates the right position (in pixels) of the internal mouse resolution. Useful for games (e.g., Syndicate) that implement an internal resolution that is different from the screen resolution. Must be activated manually during gameplay.")]
+			[DefaultValue(LibDOSBox.SVGA_MAX_WIDTH)]
+			public uint MouseInternalResolutionRight { get; set; }
+
+			[DisplayName("Mouse Internal Resolution (Top)")]
+			[Description("Indicates the top position (in pixels) of the internal mouse resolution. Useful for games (e.g., Syndicate) that implement an internal resolution that is different from the screen resolution. Must be activated manually during gameplay.")]
+			[DefaultValue(0)]
+			public uint MouseInternalResolutionTop { get; set; }
+
+			[DisplayName("Mouse Internal Resolution (Bottom)")]
+			[Description("Indicates the bottom position (in pixels) of the internal mouse resolution. Useful for games (e.g., Syndicate) that implement an internal resolution that is different from the screen resolution. Must be activated manually during gameplay.")]
+			[DefaultValue(LibDOSBox.SVGA_MAX_HEIGHT)]
+			public uint MouseInternalResolutionBottom { get; set; }
+
+
 			[DisplayName("Mount Formatted Hard Disk Drive")]
 			[Description("Determines whether to mount an empty writable formatted hard disk in drive C:. The hard disk will be fully located in memory so make sure you have enough RAM available. Its contents can be exported to the host filesystem.\n\nThis value will be ignored if a hard disk image (.hdd) is provided.")]
 			[DefaultValue(HardDiskOptions.None)]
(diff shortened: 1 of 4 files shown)


src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.ISettable.cs
```diff
@@ -277,6 +277,27 @@ public class SyncSettings
...
+			[Description("Indicates the top position (in pixels) of the internal mouse resolution. Useful for games (e.g., Syndicate) that implement an internal resolution that is different from the screen resolution. Must be activated manually during gameplay.")]
+			[DefaultValue(0)]
+			public uint MouseInternalResolutionTop { get; set; }
+
+			[DisplayName("Mouse Internal Resolution (Bottom)")]
+			[Description("Indicates the bottom position (in pixels) of the internal mouse resolution. Useful for games (e.g., Syndicate) that implement an internal resolution that is different from the screen resolution. Must be activated manually during gameplay.")]
+			[DefaultValue(LibDOSBox.SVGA_MAX_HEIGHT)]
+			public uint MouseInternalResolutionBottom { get; set; }
+
+
```
vadosnaprimer, line 300: Do these represent **offsets** from the edge of the screen? What to do if the resolution change is in the opposite direction?

SuuperW:
If you draw on the emu core surface with Lua, is the location of such a draw correctly offset? Should it be offset the same as mouse?

This seems like something that should be possible to handle automatically, rather than asking the user to specify a translation and manually activate/deactivate it.

src/BizHawk.Emulation.Cores/Computers/DOS/DOSBox.ISettable.cs
SergioMartin86: > Do these represent offsets from the edge of the screen?

These are the absolute positions of the top/left/right/bottom coordinates of the inner resolution, taking top-left as (0,0) and increasing towards the bottom-right

> What to do if the resolution change is in the opposite direction?

Always specify the maximum resolution as main screen size, and the smaller resolution as inner one. So if the game goes from small to big, simply de-activate the inner resolution toggle to come back to the big mouse resolution.

vadosnaprimer: What is there are more internal resolutions?

SergioMartin86: yeah, this PR only supports two resolutions. If there were more, well, something else would need to be done in that case

SergioMartin86:
> If you draw on the emu core surface with Lua, is the location of such a draw correctly offset? Should it be offset the same as mouse?

> This seems like something that should be possible to handle automatically, rather than asking the user to specify a translation and manually activate/deactivate it.

Sorry, I meant to quote your comments, not edit it. Restoring it to normal and putting the answers here:

> If you draw on the emu core surface with Lua, is the location of such a draw correctly offset? Should it be offset the same as mouse?
 
 I'm not sure I understand the question.
 
> This seems like something that should be possible to handle automatically, rather than asking the user to specify a translation and manually activate/deactivate it.
 
 It'd be nice for it to be automatic, but this seems to be an entire "sub-resolution" management done by software by the game. It subtracts 50px from the bottom and top, leading to an inner resolution of 700x600 + 50x0 offset, which probably cannot be achieved natively. Hence, they have to convert the 800x600 mouse coordinates into the inner one, and this PR inverts the conversion to allow the Bizhawk pointer to coincide with that of the game during a mission.
 
 Maybe you could try the game out and figure if there's a smarter solution to this?

vadosnaprimer:
What's your TAS workflow with this game? Why do you need host's mouse pointer to match guest's? Sounds like live recording is involved.

SergioMartin86:
> What's your TAS workflow with this game? Why do you need host's mouse pointer to match guest's? Sounds like live recording is involved.

Exactly, the host pointer matches the guest's. So I can do precise mouse inputs on frame advance. 

I don't insert inputs while unpaused, but frame-by-frame recording is much easier than on Windows games where the deltas are taken.

vadosnaprimer:
What if this is instead solved on the GUI side, maybe even with lua? Then you'd be able to have infinite resolutions and offset the pointer accordingly. If via lua, then some in-game variable could even be used to automate it fully.

SergioMartin86:
Alright! I'll give it a try with Lua

vadosnaprimer:
That way you'll only need 2 new functions, since it's just 2 offsets, and I suspect in reality only 1 will have to be used per game.

SergioMartin86:
Closing, as something Lua-based like this (provided by Suuper), can do the job:

```
local nameY = "Mouse Position Y"
local nameX = "Mouse Position X"
local offsetBy = 100

function adjustMouse()
    input = joypad.getimmediate()
    tastudio.submitanalogchange(emu.framecount(), nameY, input[nameY] + offsetBy)
    tastudio.submitanalogchange(emu.framecount(), nameX, input[nameX])
    tastudio.applyinputchanges()
end

event.onframestart(adjustMouse)
```

vadosnaprimer:
I feel such a script is worth shipping, to help users know it's possible.

PR #4758, opened 2026-06-02
libretro: Implement opengl rendering

Morilli:
[dev build for branch | USERNAME:BRANCHNAME](https://nightly.link/Morilli/BizHawk/workflows/ci/libretro-opengl?preview)

This implements OpenGL rendering support for libretro, which is required for certain cores (I tested [craft](https://github.com/libretro/craft)) and because libretro cores are generally very well implemented we get a hard crash if we gracefully deinitialize the core when we can't support it.

The main thing the cores need is for the frontend to handle an opengl context, pass it a framebuffer to use and handle reading pixel data itself after the core is done rendering. There's a lot of boilerplate and passing of functions from one place to another so a lot of the logic in here is needlessly complex.

I don't like the logic too much as most of it feels or is hacky, but it works (as far as I can tell). I wanted to add some more generic utility functions for handling this type of OpenGL rendering method, but because of the current project dependency structure it looks like garbage now.

Check if completed:


zeromus:
Don't care anymore. I only dropped by to say that doing this for n64 (didn't we?) was an enormous mess, and nowadays, I'd consider using ANGLE instead so we can just plumb a single graphics API through bizhawk and spin up a compatible prebuilt-for-system ANGLE. That would solve a lot of difficulties and necessities-of-keeping-gl-parts-in-bizhawk problems.  You get better compatibility too because system-provided opengl is a wasteland of game hacks whereas ANGLE is necessarily focused on compatibility.

There could be a few ways of doing this: 
1. manage ANGLE in bizhawk and send proc addresses yourself
2. initialize ANGLE in bizhawk and send just minimal angle context through to core; let it get its own proc addresses from ANGLE somehow. 
3. Can cores manage their own ANGLES and then communicate opaque stuff back into frontend to be ripped apart? (are such opaque things available?) You'd need to tell the core which driver to use to match bizhawk frontend

Don't be tempted to change opengl to be bizhawk's sole renderer, though. It's too lousy for that (see compatibility note above)

nattthebear (commented):
When I'm dying in a retirement home we'll finally be able to use OGL4 on all devices.  Or maybe not.

The PR doesn't seem bad from what I can tell, although I'm a bit rusty on all fronts.

CasualPokePlayer:
> I only dropped by to say that doing this for n64 (didn't we?) was an enormous mess

N64 itself basically never used the newer GetGLTexture() API, nor even what was present beforehand. There was some decayed mess of code that tried to get a GL texture from a 1 sized video buffer that wasn't even used for N64 (although comments indicated it was intended to do so). I scrapped it (as we need an API that returns an actual standard video buffer for screenshots/video dumping) and implemented it as a separate GetGLTexture() function, used for the 3DS and DS cores (for GL renderers).

Even then, N64 was basically its own isolated thing with GL stuff. Any GL contexts were created and managed in the unmanaged side of the core, and they were made on a separate thread anyways so it did not interfere with any potential GL context the frontend may have, and they always just did the old readback method (the fast GetGLTexture() has not been implemented, although that's probably something to more look at just for the newer core port, but I digress).

GetGLTexture() works on both OpenGL and D3D11 renderers anyways, so far. GL interop is used to copy the emu GL texture over to a D3D11 texture. This works in practice on most GPUs (main exception is just some ancient Intel GPU drivers, which are blacklisted, which are forced to only use GetVideoBuffer() which must readback). Future renderers could also implement such interop.

> I don't like the logic too much as most of it feels or is hacky, but it works (as far as I can tell). I wanted to add some more generic utility functions for handling this type of OpenGL rendering method, but because of the current project dependency structure it looks like garbage now.

My vision of GL cores were they were to implement any GL texture/FBO management entirely on the unmanaged side. This is mainly under the assumption that GL cores would have GL stuff on the unmanaged side anyways, so a BizHawk implementation could use such (and in practice, this is heavily preferred when screen control is at play, which also implies core specific GL management to handle such / using upstream screen control functions in tandem with GL management). So say something like `OpenGLFBOWithTexture` would just be entirely implemented on the unmanaged side. Managed side would only hand over callbacks for context creation and proc address obtaining which the unmanaged side can use, which works in practice for GL cores.

Libretro is more an "exception" just because it is both a GL core but also not actually a real core and thus not a GL core. We just have a bridge, and we only have unmanaged code in the first place as a way to workaround libco being an asshole. In theory, it should be easy to just move this over to the unmanaged side easily (using like [glad](https://github.com/dav1dde/glad), but using managed side code I guess works (although this really isn't intended, and the project dependency was made understanding the managed side woudn't be doing this)

Morilli:
> So say something like OpenGLFBOWithTexture would just be entirely implemented on the unmanaged side ... the project dependency was made understanding the managed side woudn't be doing this

Yeah the dependency chain is not helping because nothing from `Bizware.Graphics` is visible from `Emulation.Cores`, but I think it makes sense to expose OpenGL-specific internals if the cores need to use them, ideally by making the implementing classes directly available. I don't see the advantage of having each core (or rather, the bizhawk-specific unmanaged api for the core) implement these utility functions when we could just have them available from the frontend.

I hope to be able to use these functions for mupen as well as that core too only requests the `GLProcAddr` function and a uint for a framebuffer to be rendered to.

Morilli:
Adding a project reference from `Emulation.Cores` to `Bizware.Graphics` allows doing this: https://github.com/Morilli/BizHawk/tree/libretro-opengl-2

PR #4761, opened 2026-06-03
Service COM during frame throttle so screen readers stay responsive

### LINKED ISSUE #4663: screen reader accessibility for UI

RealAmethyst:
EmuHawk's frame throttle uses Thread.Sleep, which does not dispatch incoming COM calls. Screen readers
  (NVDA, Narrator, etc.) read the window via UI Automation / MSAA — COM calls marshaled to the STA UI
  thread — so during every frame's sleep the thread stops answering them. The result: all screen readers
  become very laggy whenever EmuHawk is the foreground window, even though the app itself stays responsive.

  This replaces the two Thread.Sleep calls in Throttle.cs with a small ResponsiveSleep that waits the same
  duration via CoWaitForMultipleHandles(COWAIT_DISPATCH_CALLS, ...), which dispatches COM calls during the
  wait (COM only — no window input/paint — so there's no reentrancy into the emulation loop). The Unix
  path keeps Thread.Sleep, since CoWaitForMultipleHandles is Windows-only.

### DIFF
--- a/src/BizHawk.Client.EmuHawk/Throttle.cs
+++ b/src/BizHawk.Client.EmuHawk/Throttle.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Threading;
 
 using BizHawk.Client.Common;
@@ -34,8 +35,8 @@ namespace BizHawk.Client.EmuHawk
 				skipNextFrame = false;
 				framesToSkip = 0;
 
-				//keep from burning CPU
-				Thread.Sleep(15);
+				//keep from burning CPU (but dispatch COM so screen-reader accessibility stays serviced)
+				ResponsiveSleep(15);
 				return;
 			}
 
@@ -277,6 +278,34 @@ namespace BizHawk.Client.EmuHawk
 			return rv;
 		}
 
+		[DllImport("ole32.dll")]
+		private static extern int CoWaitForMultipleHandles(uint dwFlags, uint dwTimeout, int cHandles, IntPtr[] pHandles, out uint lpdwindex);
+
+		[DllImport("kernel32.dll", SetLastError = true)]
+		private static extern IntPtr CreateEventW(IntPtr lpEventAttributes, bool bManualReset, bool bInitialState, IntPtr lpName);
+
+		private const uint COWAIT_DISPATCH_CALLS = 0x8;
+		private static readonly IntPtr[] _responsiveSleepHandles = { CreateEventW(IntPtr.Zero, true, false, IntPtr.Zero) };
+
+		/// <summary>
+		/// Sleeps ~<paramref name="ms"/> ms while still dispatching incoming COM calls. Screen readers'
+		/// UI Automation / MSAA queries are COM calls marshaled to this STA UI thread; a plain Thread.Sleep
+		/// does NOT service them, so during each frame's sleep the thread stops answering screen readers and
+		/// ALL of them (NVDA, Narrator, ...) go laggy while EmuHawk is foreground — even though the app stays
+		/// responsive to its own input. COWAIT_DISPATCH_CALLS dispatches only COM calls (not window
+		/// input/paint), keeping accessibility responsive without reentering the emulation loop.
+		/// </summary>
+		private static void ResponsiveSleep(int ms)
+		{
+			if (ms < 1) ms = 1;
+			if (OSTailoredCode.IsUnixHost)
+			{
+				Thread.Sleep(ms); // CoWaitForMultipleHandles is Windows-only
+				return;
+			}
+			_ = CoWaitForMultipleHandles(COWAIT_DISPATCH_CALLS, (uint)ms, 1, _responsiveSleepHandles, out _);
+		}
+
 		private void SpeedThrottle(Sound sound, bool paused)
 		{
 			AutoFrameSkip_BeforeThrottle();
@@ -326,7 +355,7 @@ namespace BizHawk.Client.EmuHawk
 							break;
 					}
 
-					Thread.Sleep(Math.Max(sleepTime, 1));
+					ResponsiveSleep(Math.Max(sleepTime, 1));
 				}
 				else if (sleepTime > 0) // spin for <1 millisecond waits
 				{


src/BizHawk.Client.EmuHawk/Throttle.cs
```diff
@@ -34,8 +35,8 @@ public void Step(Config config, Sound sound, bool allowSleep, int forceFrameSkip
 				skipNextFrame = false;
 				framesToSkip = 0;
 
-				//keep from burning CPU
-				Thread.Sleep(15);
+				//keep from burning CPU (but dispatch COM so screen-reader accessibility stays serviced)
+				ResponsiveSleep(15);
```
YoshiRulz, line 39: Is touching this call-site necessary, that is, does the lag appear when emulation is paused?

src/BizHawk.Client.EmuHawk/Throttle.cs
```diff
@@ -277,6 +278,34 @@ private int AutoFrameSkip_GetSkipAmount(int min, int max)
...
 		}
 
+		[DllImport("ole32.dll")]
+		private static extern int CoWaitForMultipleHandles(uint dwFlags, uint dwTimeout, int cHandles, IntPtr[] pHandles, out uint lpdwindex);
+
+		[DllImport("kernel32.dll", SetLastError = true)]
+		private static extern IntPtr CreateEventW(IntPtr lpEventAttributes, bool bManualReset, bool bInitialState, IntPtr lpName);
+
+		private const uint COWAIT_DISPATCH_CALLS = 0x8;
+		private static readonly IntPtr[] _responsiveSleepHandles = { CreateEventW(IntPtr.Zero, true, false, IntPtr.Zero) };
```
YoshiRulz, line 288: I suspect this will need `OSTailoredCode.IsUnixHost ? [ default ] :` to avoid crashing on startup.

src/BizHawk.Client.EmuHawk/Throttle.cs
```diff
@@ -277,6 +278,34 @@ private int AutoFrameSkip_GetSkipAmount(int min, int max)
...
+		/// Sleeps ~<paramref name="ms"/> ms while still dispatching incoming COM calls. Screen readers'
+		/// UI Automation / MSAA queries are COM calls marshaled to this STA UI thread; a plain Thread.Sleep
+		/// does NOT service them, so during each frame's sleep the thread stops answering screen readers and
+		/// ALL of them (NVDA, Narrator, ...) go laggy while EmuHawk is foreground — even though the app stays
+		/// responsive to its own input. COWAIT_DISPATCH_CALLS dispatches only COM calls (not window
+		/// input/paint), keeping accessibility responsive without reentering the emulation loop.
+		/// </summary>
+		private static void ResponsiveSleep(int ms)
+		{
+			if (ms < 1) ms = 1;
```
YoshiRulz, line 300: Both callers already pass in >= 1, so this line can be removed.

src/BizHawk.Client.EmuHawk/Throttle.cs
RealAmethyst: I never tested when active, so yes. When opening the emulator it starts lagging right away.

Morilli:
I don't know how a screenreader is typically used, but I started NVDA to see if I can notice anything, and I don't see what you mean when you say "laggy". If anything, there should be a <15ms delay in response time which should be negligible.

RealAmethyst:
so without this fix it's like 10 seconds of lag if not more, this fixes that, this also happens on every other screen reader I tested. If you don't get that lag can you please inform me of what screen reader version you use?

CasualPokePlayer:
I would be wary of trying to "pump COM messages" while sleeping. While technically the STA thread contract dictates this should be done, in practice [this ends up causing more issues](https://github.com/CasualPokePlayer/GSE/commit/ded343e05cc54f7becc01b1f0c7bbf63ed7ee072). This "fix" would also be incomplete, as this throttling is only applicable assuming Clock Throttle. Audio Throttle has sleep calls elsewhere and VSync Throttle / VSync Enabled are just impossible to "fix" here. I would also be skeptical about "very laggy" claims, the sleep calls are normally very short and message pumping occurs between them. Are you sure you didn't end up setting BizHawk to say 1% speed? (which case there will be big sleep calls, although in this case the UI would also just be laggy in the first place / this is a known issue)

Also, `COWAIT_DISPATCH_CALLS` would do nothing in this case (or rather it would be ignored), our thread is an STA thread, not an ASTA thread (ASTA is its own thing with UWP UI, it is not applicable to WinForms nor most applications). So in practice you are getting `COWAIT_DEFAULT` (0) behavior.

RealAmethyst:
Whenever I would open the emulator and try to use the UI it would be very lagy, this is also not just me, but every other blind user I've talked to has the exact same issue, this is before doing anything, not even loading a rom, as the UI is just slow with screenr eaders in general. It also does not matter what screen reader you use, same issue.

YoshiRulz:
Clearly there are more differences between Morilli's and my dev environments and a "real" user's environment, and it just hasn't occurred to any of us what those differences are. Do you use your PC with the screen off? Is the lag dependent on the display method (D3D vs GDI+)? Does Windows itself change if it's installed with a11y features enabled?

RealAmethyst:
I tried checking D3D vs GDI and no difference on my end, I have a monitor connected but it's not on most of the time, installing windows with accessibility stuff shouldn't change anything as it comes with the OS by default

BlindGuyNW:
Just chiming in here as a fellow blind user to confirm that the lag is not imaginary :) IT's specifically just ambient as soon as you open the app with NVDA running. The slightly confusing thing is that the lag only occurs in screen reader response time, not in the app performance itself as far as I am aware.

CasualPokePlayer:
To be clear, does this lag also occur for Audio Throttle and/or VSync Throttle and/or Vsync Enabled?

RealAmethyst:
I have no idea, as trying to turn those settings on is a pain due to screen reader lag, this happens without us changing any settings, just an program launch

RealAmethyst:
Tested more, the app itself is fast, no lag, games themselves has no lag, the screen reader itself lags, even when I try to do screen reader specific commands, when the emulator is in forground, it will lag, this is for all screen readers.

nattthebear:
I'm willing to believe this.  Well, I'll believe anything.  But beyond that, COM is hard to use, screen readers are a bit of a mess, and EmuHawk (especially MainForm) is a bit of a mess, so sure, I could see this working and fixing all of your problems.

But there must be more to the story, no?  I'm hearing that we get 10+ second lags in screen readers.  But even without this change, MainForm is pumping events ~60 times a second, so 600+ times in that 10 second period.

Whatever the explanation, it seems like our slower cores are completely doomed.  If a core is only running at 65-70fps unthrottled, out of a 60fps target, then it doesn't have much of any sleep time or COM service time to work with.

CasualPokePlayer:
>> To be clear, does this lag also occur for Audio Throttle and/or VSync Throttle and/or Vsync Enabled?
>
> I have no idea, as trying to turn those settings on is a pain due to screen reader lag, this happens without us changing any settings, just an program launch

Could you test those settings with the changes in this PR then?

RealAmethyst:
Sure, I can test it. Where are those settings located?

YoshiRulz:
`Config` > `Speed/Skip`

PR #4762, opened 2026-06-03
Add a speech Lua library for screen reader and brail output

RealAmethyst:
Adds a universal `speech` Lua library so that lua scripts can utilize screen readers and brail directly, opening games up to accessibility scripting for the blind. Methods: speech.say(text, interrupt),
   speech.output(text, interrupt) (speak + braille), speech.braille(text), speech.stop().

   Backed by prism (https://github.com/ethindp/prism, MPL-2.0), which routes to the active screen reader
   (NVDA, JAWS, Narrator/OneCore, ...) with a SAPI fallback, via P/Invoke to its C API. The native prism.dll
   (win-x64) is added under Assets/dll. Initialisation is lazy and failures are non-fatal
   (calls become no-ops), so a missing DLL can't crash EmuHawk


YoshiRulz:
(There is [a .NET bindings package](https://www.nuget.org/packages/prismatoid) for this library, but it's .NET Core only.) I think, despite your intentions, that this will crash on Linux and possibly Windows < Win10. It will have to go through one of our helpers like `DynamicLibraryImportResolver` instead of a simple `[DllImport]`.

src/BizHawk.Client.Common/lua/CommonLibs/SpeechLuaLibrary.cs
```diff
@@ -0,0 +1,150 @@
...
+		public SpeechLuaLibrary(ILuaLibraries luaLibsImpl, ApiContainer apiContainer, Action<string> logOutputCallback)
+			: base(luaLibsImpl, apiContainer, logOutputCallback) {}
+
+		public override string Name => "speech";
+
+		[LuaMethodExample("speech.say(\"Charizard, 142 of 142 HP\");")]
+		[LuaMethod("say", "Speaks the given text through the active screen reader (or SAPI fallback). When interrupt is true (the default), any current speech is cancelled first.")]
+		public void Say(string text, bool interrupt = true)
+		{
+			if (!PrismSpeech.Speak(text ?? string.Empty, interrupt, out var error)) Log($"speech.say failed: {error}");
```
YoshiRulz, line 22: Is it reasonable to announce an empty (or all-whitespace) string?

YoshiRulz (commented):
I'm also wondering how to deal with attribution, linking messages to the script that produced them. With `console.writeline`, we've just accepted that the script author decides whether to include an identifying message, but we are now including the filename when logging to stdout. Maybe the first time a script tries to announce something, we could prepend "script &lt;scriptname&gt; has begun using TTS"?

src/BizHawk.Client.Common/lua/CommonLibs/SpeechLuaLibrary.cs
RealAmethyst: I mean there is no reason why speech should announce  empty stuff.

SuuperW: I occasionally use `print` to see some information and it ends up being actually empty, usually due to a bug in my script. In this case I can still see an empty line once the next print happens. It'd be much more frustrating if the `print` call did nothing, and I imagine it will also be very frustrating if a speech call that accidentally contains a blank string is ignored.

RealAmethyst: I actually didn't think about that, and yeah you're right, that would be frustrating, so having it still speak it is the best.

YoshiRulz: > In this case I can still see an empty line once the next print happens.

Explain how that applies to TTS.

RetroEdit: I mean, I would be curious how TTS handles the empty string, but SuuperW's "debug empty log messages" use-case seems relevant whether you can see the screen or not.

PR #4768, opened 2026-06-06
fix(achievements): remediate HTTP infinite loop on rate limit hit

wescopeland:

[dev build for branch | wescopeland:fix/rcheevos-rate-limit-loop](https://nightly.link/wescopeland/BizHawk/workflows/ci/rcheevos-rate-limit-loop?preview)

On hitting a rate limit to RetroAchievements' server, BizHawk enters a retry loop where `?r=ping` requests slam our back-end non-stop.

`HttpGet` and `HttpPost` both return `null` for any non-2xx response. This is interpreted as "likely a timeout" and `ShouldRetry` becomes `true`. A 429 triggers an immediate retry. Each retry also returns a 429, and the cycle becomes an infinite loop.

This PR caps transient retries at 3 per request. I did not add exponential backoff or honoring of our `Retry-After` header. I've verified these changes locally against a test script.

Screenshot 2026-06-06 at 9 11 24 AM

Screenshot 2026-06-06 at 9 12 09 AM

Check if completed:


CasualPokePlayer:
> This PR caps transient retries at 3 per request. I did not add exponential backoff or honoring of our Retry-After header.

I imagine properly fixing this could come within the HTTP loop. That already should generally give plenty of time to backoff (it's literally a 10 second throttle per iteration), although I guess it has the minor issue of basically spamming multiple requests at the same time if such end up queued in the 10 second interval (and it's not simple to just rate limit them individually since it's all async). The loop probably should be modified to wait until all requests are done if there is a retriable event available, then try to wait until the latest `Retry-After` header available (or just another 10 seconds perhaps if none available or it has already passed), then proceed to retry operations. Granted, I imagine this might end up hitting the same rate limiting issue generally since again it's going to spam multiple requests at once, some probably getting through, but once it hits a rate limit it has to wait again, and so on until the amount of requests remaining is 0.

This PR looks like an OK quick fix in the meantime (granted, keep in mind existing users won't get this until likely months later whenever the next release happens, which is not exactly great for you guys).

wescopeland:
@CasualPokePlayer We need this fixed. What is the path forward for this? Shall I close this PR and defer a separate fix to you?

CasualPokePlayer:
If you could, could you fix the merge conflict? I'll merge this for now after that and maybe improve it later when I have time.

wescopeland:
Sure thing - I've cleaned that up in latest.

PR #4774, opened 2026-06-17
make savestates for markers optional, per marker

### LINKED ISSUE #3892: [Feature Request] TAStudio's non-save state marker, marker.
I only realized this whilst using Encore and then thought about it.
Basically a way to have a main marker that acts as a state (so what Hawk currently has) and then secondary markers to be like where something works, and you want to check around it and it goes back to that main marker.
This would also presumably reduce the filesize of tasproj files if used effectively by the TASer.

Like for an example on what I mean by "where something works" when I TAS Tekken I place markers on where something connects the opposing player, and I want to know that it can be there, and check around that frame instead of having that as a state, I can have that as an idea of "Yes, I can confirm that they get hit here, but I want to know what happens if I do this instead".

SuuperW:
A new column has been added to the marker list, which controls if TAStudio will keep a savestate for each specific marker. By default all markers will have states, but this can be changed with a new settings. Clicking on the savestate cell for a particular marker will toggle if it keeps a state.

For markers not keeping a state, the cell will be blank. For markers wanting a state but not having one, the cell will have `!`. For markers wanting a state and actually having a state, the cell will have `✔`.

There might be a better way to design the user interface. Suggestions welcome.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/movie/tasproj/TasBranch.cs
+++ b/src/BizHawk.Client.Common/movie/tasproj/TasBranch.cs
@@ -252,17 +252,7 @@ namespace BizHawk.Client.Common
 				});
 
 				b.Markers = new TasMovieMarkerList(movie);
-				bl.GetLump(nmarkers, abort: false, tr =>
-				{
-					string line;
-					while ((line = tr.ReadLine()) != null)
-					{
-						if (!string.IsNullOrWhiteSpace(line))
-						{
-							b.Markers.Add(new TasMovieMarker(line));
-						}
-					}
-				});
+				bl.GetLump(nmarkers, abort: false, b.Markers.LoadFromFile);
 
 				bl.GetLump(nusertext, abort: false, tr =>
 				{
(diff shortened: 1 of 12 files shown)


vadosnaprimer:
I think it's better to indicate explicitly which markers have states disabled, for example with a cross. Then ones wanting a state could be empty and ones with it could have the checkmark. `!` is kinda ambiguous, and cross is very clear.

SuuperW:
Crosses are sometimes used to "check" things, so I'm not sure that's better. I feel like if a user clicks a box and it toggle between a cross and blank, the cross state would feel more like the enabled state. Though I don't think I've seen crosses used like that anywhere recently, maybe it is rare enough today.

Maybe instead the `!` could change to a `-`, making it more like a three-state checkbox. Or do both, so the box is never blank?

vadosnaprimer:
I meant a red cross. That never means anything good :D

`-` is probably fine.

SuuperW:
Red cross. And dash.
image

`Color.OrangeRed` because `Red` looks bad, especially when the row is selected.

vadosnaprimer (approved):

SuuperW:
Fixed bugs and updated the tooltip text for the new setting because I re-read it and didn't like it.

PR #4780, opened 2026-06-23
Easy drag selection in TAStudio

### LINKED ISSUE #3553: TAStudio click and drag row selection UX idea
To paraphrase adelikat: It should remember the row under the cursor on mousedown and select from there to the cursor, i.e. moving the cursor away from the start enlarges the selection, and moving the cursor back shrinks it.

SuuperW:
Based on branch show-hover because this starting a selection while unpaused following cursor doesn't work right, and the new behavior just makes that problem more confusing.

Check if completed:


vadosnaprimer (approved):
Works.

PR #4782, opened 2026-06-24
Add the WM_CLASS property to the BizHawk window

Leopardly:

[dev build for branch | Leopardly:linux-wm-class](https://nightly.link/Leopardly/BizHawk/workflows/ci/linux-wm-class?preview)

Adds a set of calls to XLIB to set a WM_CLASS on Linux for BizHawk; runs once but has to be run in the main loop as the application needs to be fully loaded and rendering (If there is a better place to put this, please do let me know!). Also added support for defining the WM_CLASS to use via a CLI argument.

There were 2 issues essentially:
- Using custom .desktop s to launch Bizhawk with the roms and Lua scripts I want (for things like Kaizo) - the system has no way to correlate the running Bizhawk instance with what was launched so it falls back to using the embedded icon, rather than the icon of the .desktop file.
- Attempting to set rules with KDE's built in "Window Rules" gives you a warning that no WM_CLASS is set, meaning things like "Remove window borders" rules etc have to rely on less reliable pickups like window title.

I'll admit the first case (which is basically my justification for the CLI arg) is niche and I'm not against dropping it entirely; it just makes having say "Run X in Bizhawk" as a shortcut much easier to uniquely integrate rather than running instances all looking the same.

CI suites ran; I think that's all I needed to do?

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/ArgParser.cs
+++ b/src/BizHawk.Client.Common/ArgParser.cs
@@ -117,6 +117,11 @@ namespace BizHawk.Client.Common
 			Description = "pairs in the format `k1:v1;k2:v2` (mind your shell escape sequences); if the value is `true`/`false` it's interpreted as a boolean, if it's a valid 32-bit signed integer e.g. `-1234` it's interpreted as such, if it's a valid 32-bit float e.g. `12.34` it's interpreted as such, else it's interpreted as a string",
 		};
 
+		private static readonly Option<string?> OptionWMClass = new ("--wmclass")
+		{
+			Description = "set a custom WM_CLASS for this Bizhawk initiation, Linux only."
+		};
+
 		static ArgParser()
 		{
 			OptionAVDumpFrameList.Description = $"comma-separated list of integers, indices of frames which should be included in the A/V dump (encoding); implies `{OptionAVDumpEndAtFrame.Name}=<end>` where `<end>` is the highest frame listed";
@@ -162,6 +167,7 @@ namespace BizHawk.Client.Common
 			root.Add(/* --url-post */ OptionHTTPClientURIPOST);
 			root.Add(/* --userdata */ OptionUserdataUnparsedPairs);
 			root.Add(/* --version */ OptionQueryAppVersion);
+			root.Add(/* --wmclass */ OptionWMClass);
 
 			return root;
 		}
@@ -268,7 +274,8 @@ namespace BizHawk.Client.Common
 				openExtToolDll: result.GetValue(OptionOpenExternalTool),
 				socketProtocol: result.GetValue(OptionSocketServerUseUDP) ? ProtocolType.Udp : ProtocolType.Tcp,
 				userdataUnparsedPairs: userdataUnparsedPairs,
-				cmdRom: result.GetValue(ArgumentRomFilePath)
+				cmdRom: result.GetValue(ArgumentRomFilePath),
+				wmClassName: result.GetValue(OptionWMClass)
 			);
 			return null;
 		}
--- a/src/BizHawk.Client.Common/ArgParser.cs
+++ b/src/BizHawk.Client.Common/ArgParser.cs
@@ -119,7 +119,7 @@ namespace BizHawk.Client.Common
 
 		private static readonly Option<string?> OptionWMClass = new ("--wmclass")
 		{
-			Description = "set a custom WM_CLASS for this Bizhawk initiation, Linux only."
+			Description = "set a custom WM_CLASS for this Bizhawk initiation, Linux only.",
 		};
 
 		static ArgParser()
--- a/src/BizHawk.Client.EmuHawk/MainForm.cs
+++ b/src/BizHawk.Client.EmuHawk/MainForm.cs
@@ -914,6 +914,8 @@ namespace BizHawk.Client.EmuHawk
 		/// </summary>
 		private bool _skipNextAltRelease = true;
 
+		private bool _runWMSetOnce = true;
+
 		public int ProgramRunLoop()
 		{
 			// needs to be done late, after the log console snaps on top
@@ -1014,6 +1016,13 @@ namespace BizHawk.Client.EmuHawk
 				StepRunLoop_Core();
 				Render();
 				StepRunLoop_Throttle();
+
+				//Code to set WM_CLASS only need be run once, but must be run after the application has rendered at least once
+				if(_runWMSetOnce && OSTailoredCode.CurrentOS==OSTailoredCode.DistinctOS.Linux)
+				{
+					OSTailoredCode.SetWMClass(_x11Display,"BizHawk");
+					_runWMSetOnce = false;
+				}
 
 				// HACK: RAIntegration might peek at memory during messages
 				// we need this to allow memory access here, otherwise it will deadlock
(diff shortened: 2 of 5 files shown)


src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1015,6 +1017,14 @@ public int ProgramRunLoop()
 				Render();
 				StepRunLoop_Throttle();
 
+				//Code to set WM_CLASS only need be run once, but must be run after the application has rendered at least once
+				if(_runWMSetOnce && OSTailoredCode.CurrentOS==OSTailoredCode.DistinctOS.Linux)
```
YoshiRulz: There are a few possible places for init code, mainly the constructor and the `Load` handler `MainForm_Load`.

src/BizHawk.Client.EmuHawk/MainForm.cs
```diff
@@ -1015,6 +1017,14 @@ public int ProgramRunLoop()
 				Render();
 				StepRunLoop_Throttle();
 
+				//Code to set WM_CLASS only need be run once, but must be run after the application has rendered at least once
+				if(_runWMSetOnce && OSTailoredCode.CurrentOS==OSTailoredCode.DistinctOS.Linux)
+				{
+					string wmclass = _argParser.wmClassName is not null ? _argParser.wmClassName : "BizHawk";
+					OSTailoredCode.SetWMClass(_x11Display,wmclass);
```
YoshiRulz: I think this is meant to be a lazy-initialised variable:
https://github.com/TASEmulators/BizHawk/blob/bd7626eb657d755e57fb71f3040a8a5a5352210e/src/BizHawk.Client.EmuHawk/MainForm.cs#L4866-L4868
So you should do the same here.

src/BizHawk.Common/OSTailoredCode.cs
```diff
@@ -292,5 +292,35 @@ public static string SimpleSubshell(string cmd, string args, string noOutputMsg)
 			if (stdout.EndOfStream) throw new Exception($"{noOutputMsg} ({cmd} wrote nothing to stdout)");
 			return stdout.ReadLine()!;
 		}
+
+		public static void SetWMClass(IntPtr x11Display, string className)
```
YoshiRulz: Move to `XlibImports`

src/BizHawk.Common/OSTailoredCode.cs
```diff
@@ -292,5 +292,35 @@ public static string SimpleSubshell(string cmd, string args, string noOutputMsg)
...
+		{
+			//Setup the ClassHint
+			var wmSet = new XlibImports.XClassHint{res_name = Marshal.StringToCoTaskMemAnsi("BizHawk"), res_class = Marshal.StringToCoTaskMemAnsi(className)};
```
YoshiRulz: Can you explain why the same string needs to be passed twice?

src/BizHawk.Client.Common/ArgParser.cs
```diff
@@ -117,6 +117,11 @@ public static class ArgParser
 			Description = "pairs in the format `k1:v1;k2:v2` (mind your shell escape sequences); if the value is `true`/`false` it's interpreted as a boolean, if it's a valid 32-bit signed integer e.g. `-1234` it's interpreted as such, if it's a valid 32-bit float e.g. `12.34` it's interpreted as such, else it's interpreted as a string",
 		};
 
+		private static readonly Option<string?> OptionWMClass = new ("--wmclass")
+		{
+			Description = "set a custom WM_CLASS for this Bizhawk initiation, Linux only.",
```
YoshiRulz, line 122: Why is this a CLI flag? When should a user be changing it? If it's for vendoring, we already have `VersionInfo.CustomBuildString`.

YoshiRulz (changes requested):
There are a bunch of lint problems too, but you can ignore those until we finalise the design.

src/BizHawk.Common/OSTailoredCode.cs
Leopardly: The WM_CLASS property is made up of two parts (name and class) - both need to be provided - they can be different/unique however it is somewhat common to see application name used in both fields (as I have here as the default behaviour).

src/BizHawk.Client.Common/ArgParser.cs
Leopardly: My initial reasoning for looking into adding WM_CLASS was to help improve BizHawk's integration with the DE (I use KDE). By defining a WM_CLASS we can hint StartupWMClass for example in .desktop files.
In allowing WM_CLASS to be set in a CLI arg it is then feasible to distinguish between different installs of Bizhawk (for example,  a Kaizo Ironmon setup and a regular Bizhawk install)

src/BizHawk.Client.EmuHawk/MainForm.cs
Leopardly: Moved into MainForm_Load

src/BizHawk.Client.EmuHawk/MainForm.cs
Leopardly: Added check+initialise step and unload to method moved to XlibImports

src/BizHawk.Common/OSTailoredCode.cs
YoshiRulz: Incidentally, I saw this in Mono's issue tracker: https://gitlab.winehq.org/mono/mono/-/work_items/32

src/BizHawk.Client.Common/ArgParser.cs
YoshiRulz: I also use KDE. In what way does this improve integration? That is, what problems were you running into that led you to make these changes?

src/BizHawk.Common/OSTailoredCode.cs
Leopardly: Interesting, I can't say I'd seen that happening (although I'm just on whatever mono is in Cachy's repos) - Picking through Keepass's implementation it looks like their is a neater way to be picking up the Window frame to assign WM_CLASS to - so I may take a look at that next week (away all weekend)

src/BizHawk.Client.Common/ArgParser.cs
Leopardly: 2 issues essentially:

- Using custom .desktop s to launch Bizhawk with the roms and Lua scripts I want (for things like Kaizo) - the system has no way to correlate the running Bizhawk instance with what was launched so it falls back to using the embedded icon, rather than the icon of the .desktop file.
- Attempting to set rules with KDE's built in "Window Rules" gives you a warning that no WM_CLASS is set, meaning things like "Remove window borders" rules etc have to rely on less reliable pickups like window title.

I'll admit the first case (which is basically my justification for the CLI arg) is niche and I'm not against dropping it entirely; it just makes having say "Run X in Bizhawk" as a shortcut much easier to uniquely integrate rather than running instances all looking the same.

YoshiRulz: Hmm I should really get around to [making a textbox to set the window title thing](https://gitlab.com/YoshiRulz/yoshis-hawk-thoughts/-/work_items/76). Then you could pull from that, and your `.desktop` files could pass `--config`. Or maybe it could use the rom name and you wouldn't need any extra flags?
edit: fbdb12812e599cbfca87d76148af7286c60a33a6

PR #4784, opened 2026-06-25
Remove OpenGLProvider and add project reference instead

Morilli:
[dev build for branch | USERNAME:BRANCHNAME](https://nightly.link/Morilli/BizHawk/workflows/ci/remove-openglprovider?preview)

I think this makes more logical sense and makes it easier to add new classes and functions in the future, and will also help if we ever want to use hardware contexts for other renderers (vulkan, direct3d) and would have to do the same frontend graphics integration there. See also #4758 for an example where this helps implementation.

Check if completed:


src/BizHawk.Emulation.Cores/Computers/AmstradCPC/Media/Disk/FloppyDisk.cs
```diff
@@ -1,11 +1,10 @@
...
-
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
-using BizHawk.Common.CollectionExtensions;
+using BizHawk.Common;
 using BizHawk.Common.StringExtensions;
+using CollectionExtensions = BizHawk.Common.CollectionExtensions.CollectionExtensions;
```
YoshiRulz, line 7: Was this to avoid a name collision or something?

Morilli: Yeah, `BizHawk.Common.CollectionExtensions` and `System.Collections.Generic.CollectionExtensions`.

YoshiRulz: Cause is an explicit (non-Source Generator) backport... BY MS https://github.com/dotnet/runtime/blob/c369fd627a26b606f36e6b3cbef6f25f3c00bbac/src/libraries/Microsoft.Extensions.DependencyModel/ref/Microsoft.Extensions.DependencyModel.cs#L221-L223 -_-
which we depend on via Silk.NET, hence no collisions before adding the dep on Bizware.Graphics.
edit: https://github.com/dotnet/runtime/issues/129881

Issue #4052, opened 2024-09-24
Arm64 port discussion


InfamousKnight:
I know this repo is against wine, and I am too.

So the only solution i can think of is using mkbundle to make it x86_64 build so you won't have to install mono.

The only other thing needed is box64. If you install mono on an arm64 machine, it will just install mono arm64 which won't work. It will load the program, but no roms will load.

So far I have managed to get bizhawk running on winlator for android, but it's very not stable. I have ways to run linux on android.

YoshiRulz:
As I've told you before (https://github.com/TASEmulators/BizHawk/issues/2731#issuecomment-1473030238), it should "just work" if you have an [AArch64 copy of Mono](https://www.mono-project.com/docs/about-mono/supported-platforms/), and the same for any native libraries used by a core.

edit: As a workaround, you may also have some success with https://fex-emu.com/ (on macOS, https://mac.getutm.app/).
edit edit: Nix `''exec '${pkgs.pkgsCross.*.stdenv.hostPlatform.emulator pkgs}' '${pkgs.pkgsCross.*.hello}' ''`

InfamousKnight:
Oh, I didn't know it was that simple.. sorry

InfamousKnight:
c22f2ab4f5df93436a7513a54aaafb3fd5de6120

Retrying this with the same setup, and it crashes at startup with this message:

```
System.TypeInitializationException: The type initializer for 'BizHawk.Bizware.Graphics.ImGui2DRenderer' threw an exception. ---> System.DllNotFoundException: cimgui assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) BizHawk.Bizware.Graphics.ImGui2DRenderer.ImDrawListSharedData_ImDrawListSharedData()
  at BizHawk.Bizware.Graphics.ImGui2DRenderer..cctor () [0x00000] in <6943343e38fc431da513df9dd525dff5>:0 
   --- End of inner exception stack trace ---
  at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_generic_class_init(intptr)
  at BizHawk.Bizware.Graphics.IGLExtensions.Create2DRenderer (BizHawk.Bizware.Graphics.IGL gl, BizHawk.Bizware.Graphics.ImGuiResourceCache resourceCache) [0x0000a] in <6943343e38fc431da513df9dd525dff5>:0 
[...]

Issue #3697, opened 2023-07-01
Mac Ports Stopping At An Older Version of BizHawk


MarioGuyWasTakenAgain:
1. Why did you stop making Mac ports?
2. Will you ever make ones for the newest versions?

Sappharad:
I covered this in the forum thread for the Mac builds. (Away right now, don’t have the link handy)

The technology used by BizHawk isn’t available for 64-bit macOS so when Apple dropped 32-bit support everything stopped being possible. 

There are some pieces that are possible now but the main versions are not compatible and a lot of effort is required to make some of the dependencies work. That’s the short answer.

Sappharad:
Now that I'm sitting in front of a computer, I can type out a more detailed response. Thanks @YoshiRulz for adding links to my reply above this one. I missed some pieces that I explained in the posts linked above, like forgetting that I had done a few 64-bit builds although that stuff broke again around macOS 10.15. Writing this up might help anyone interested in looking into things understand what would be involved to get it working again going forwards.

In short, BizHawk in a .NET App that relies on "Windows Forms" (or WinForms for short) as its UI technology. WinForms was designed for Microsoft Windows. The Mono project was created to run .NET apps on Linux and eventually Mac. They created a version of WinForms that could run on Linux and Mac many years ago. When Microsoft acquired Xamarin (the company that did most of the development work on Mono), Microsoft decided that the best path forward was official support for Linux and Mac and so they created .NET Core which is now the main .NET going forwards. That involved building new technologies that would work well cross platform and WinForms was not one of the things that made sense for them to support on Linux and Mac because it was originally intended for Windows. When Apple dropped 32-bit support, the old WinForms implementation from Mono needed that and broke. Now Microsoft has something called MAUI that can be used across Windows, Mac, Android and iOS because it was designed to be multi-platform from the beginning. But MAUI is completely different from WinForms and not just something you can upgrade to. There are other UI options as well from third parties that work better as multi-platform, but the problem is the entire UI for BizHawk would need to be rewritten in one of these and I don't think anyone has the time or motivation to do that.

[...]

PR #4785, opened 2026-06-25
Experimental macOS (x86_64) port -- Mono + XQuartz, unlocks waterbox (DS) on macOS

DarthMDev:
**Draft / RFC — experimental, not asking for a support commitment.**

Revives a macOS port by running the existing **x86_64** build under **Mono + XQuartz**
(natively on Intel, via **Rosetta 2** on Apple silicon). Verified working: EmuHawk GUI,
**Lua**, **GB/GBC (Gambatte)**, **GBA (mGBA)**, and **Nintendo DS (melonDS)** — Pokémon
SoulSilver boots and runs at full speed from a real save. All changes are OS-gated; no
behavioural change on Windows/Linux.

**Highlights**
- waterboxhost on macOS: file-backed (not shm) block handle so guest code can be mapped
  executable; SIGBUS handling for dirty-detection; a macOS `%gs`/TSD trampoline
  (`interop_macos.s`) since the GS base can't be repointed under Rosetta. All 26
  `cargo test --target x86_64-apple-darwin` pass.
- EmuHawk: `OSTailoredCode.DllExtension`, `MemoryBlock` MAP_ANON fix, SDL2 GLX-not-EGL on
  macOS, and host-GL reported unavailable (XQuartz GL is broken/2.1) so cores fall back to
  software rendering.
- `EmuHawkMonoMacOS.sh` launcher, `Dist/stage-macos-dylibs.sh`, `build-release-macos.sh`,
  and a README section with setup + an experimental/maintainer warning.

**Scope/limitations**: source-only (native dylibs are build artifacts; build with the
included scripts + Homebrew). Software (GdiPlus) video only -- no hardware GL / HD
upscaling. Requires macOS 14+ (Homebrew dep floor).

Refs #1430, #3697, #4052. Discussion: https://github.com/DarthMDev/BizHawk/discussions/1

image


Assets/EmuHawkMonoMacOS.sh
```diff
@@ -0,0 +1,51 @@
...
+# Launcher for EmuHawk on macOS via Mono (x86_64, under Rosetta 2 on Apple Silicon).
+# Mono's System.Windows.Forms is X11-based, so an X server (XQuartz) must be running.
+# See also EmuHawkMono.sh (the Linux equivalent this is modelled on).
+cd "$(dirname "$(realpath "$0")")"
+
+# Native libs ship in ./dll; also let the loader find Homebrew-provided deps
+# (SDL2, OpenAL, Lua, zstd, libgdiplus, ...) under /usr/local, and XQuartz's X11
+# libs under /opt/X11/lib (Mono's WinForms dlopens libX11 etc. from there).
+export DYLD_LIBRARY_PATH="$PWD/dll:$PWD:/usr/local/lib:/opt/X11/lib${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}"
+export DYLD_FALLBACK_LIBRARY_PATH="$PWD/dll:$PWD:/usr/local/lib:/opt/X11/lib:/usr/lib${DYLD_FALLBACK_LIBRARY_PATH:+:$DYLD_FALLBACK_LIBRARY_PATH}"
```
YoshiRulz: Can you merge the two launch scripts by putting macOS-specific commands into a `case` branch?

Assets/EmuHawkMonoMacOS.sh
```diff
@@ -0,0 +1,51 @@
...
+if ! command -v Xquartz >/dev/null 2>&1 && [ ! -d "/Applications/Utilities/XQuartz.app" ] && [ ! -d "/opt/X11" ]; then
+	printf "%s\n" "XQuartz does not appear to be installed; Mono WinForms needs an X11 server. Install with: brew install --cask xquartz" >&2
+fi
+
+# Prefer the x86_64 Mono from Homebrew at /usr/local. Invoke it DIRECTLY (not via `arch`):
+# /usr/bin/arch is SIP-protected and strips DYLD_* from the environment, which breaks our
+# library paths. An x86_64-only binary auto-runs under Rosetta 2 when exec'd directly.
+mono_bin="mono"
+if [ -x "/usr/local/bin/mono" ]; then
+	mono_bin="/usr/local/bin/mono"
```
YoshiRulz: Is this not equivalent to `export PATH="/usr/local/bin:$PATH"`? Generally I'd want `PATH` to be respected on macOS as well as Linux; for example if you were to install Mono with Nix instead of Homebrew.

Assets/EmuHawkMonoMacOS.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+# Force the GdiPlus (software) display method: XQuartz's OpenGL is Apple's legacy GLX bridge,
+# which is only GL 2.1 and crashes (CGLSetCurrentContext) under Rosetta. Without this, EmuHawk
```
YoshiRulz: Is the OpenGL API level the problem, or is it a Rosetta bug?
edit: see thread on `OpenGLProvider`

Dist/stage-macos-dylibs.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
+# Stage the macOS (x86_64 / Rosetta) native-library glue into the EmuHawk output dir.
+#
+# Run this once after building (Dist/BuildRelease.sh) and before launching with
+# Assets/EmuHawkMonoMacOS.sh. It symlinks the Homebrew-provided dependencies (SDL deps,
```
YoshiRulz: `Dist/Package.sh`?

Dist/stage-macos-dylibs.sh
```diff
@@ -0,0 +1,65 @@
...
+# OpenAL, Lua, zstd, SQLite, the X11 client stack) and creates the Linux-soname aliases
+# (libX11.so.6, etc.) that BizHawk's P/Invokes and Mono expect, all pointing at the SAME
```
YoshiRulz: Can you use `LD_LIBRARY_PATH` or equivalent instead of making symlinks in the output? We obviously can't ship symlinks.

Also Mono has a name mapping feature for dynamic libraries, but I don't think I've ever gotten it working.

src/BizHawk.Bizware.Graphics/OpenGL/SDL2OpenGLContext.cs
```diff
@@ -23,19 +23,23 @@ static SDL2OpenGLContext()
 		{
 			if (OSTailoredCode.IsUnixHost)
```
YoshiRulz:
```csharp
if (OSTailoredCode.CurrentOS is OSTailoredCode.DistinctOS.macOS) {
...
} else if (OSTailoredCode.IsUnixHost) {
...
}
```

src/BizHawk.Bizware.Input/KeyMouseInput/X11KeyMouseInput.cs
```diff
@@ -23,9 +23,9 @@ internal sealed class X11KeyMouseInput : IKeyMouseInput
 
 		public X11KeyMouseInput()
 		{
-			if (OSTailoredCode.CurrentOS != OSTailoredCode.DistinctOS.Linux)
+			if (OSTailoredCode.CurrentOS != OSTailoredCode.DistinctOS.Linux && OSTailoredCode.CurrentOS != OSTailoredCode.DistinctOS.macOS)
 			{
-				throw new NotSupportedException("X11 is Linux only");
+				throw new NotSupportedException("X11 is Linux/macOS only");
 			}
```
YoshiRulz: Honestly just remove this check. If you're on a more exotic OS with X11, it may just workTM, and if not or you're on Windows there's another check right below this.

src/BizHawk.Client.EmuHawk/GraphicsImplementations/OpenGLProvider.cs
```diff
@@ -9,7 +10,11 @@ namespace BizHawk.Client.EmuHawk
 	public class OpenGLProvider : IOpenGLProvider
 	{
 		public bool SupportsGLVersion(int major, int minor)
-			=> OpenGLVersion.SupportsVersion(major, minor);
+			// On macOS the only GL available (XQuartz/GLX) is the legacy Apple bridge, which is
+			// capped at GL 2.1 and aborts under Rosetta; probing it crashes. Report no host GL so
+			// cores (melonDS, N64, ...) fall back to their software renderers.
+			=> OSTailoredCode.CurrentOS != OSTailoredCode.DistinctOS.macOS
```
YoshiRulz, line 16: Would this be correct, or are even the lower API levels unusable under Rosetta?
```csharp
=> OSTailoredCode.CurrentOS is OSTailoredCode.DistinctOS.macOS
	? new Version(major, minor) <= new Version(2, 1)
	: OpenGLVersion.SupportsVersion(major, minor);
```

src/BizHawk.Common/MemoryBlock/MemoryBlockLinuxPal.cs
```diff
@@ -17,9 +17,13 @@ internal sealed class MemoryBlockLinuxPal : IMemoryBlockPal
 		/// <exception cref="InvalidOperationException">
 		/// failed to mmap
 		/// </exception>
+		// MAP_PRIVATE is 0x02 everywhere, but MAP_ANON differs: 0x20 on Linux/BSD, 0x1000 on macOS.
+		private static readonly int MapPrivateAnon
+			= 0x02 | (OSTailoredCode.CurrentOS == OSTailoredCode.DistinctOS.macOS ? 0x1000 : 0x20);
```
YoshiRulz:
```csharp
private const int MAP_PRIVATE = 0x2;
private static readonly int MAP_ANON = OSTailoredCode.CurrentOS is OSTailoredCode.DistinctOS.macOS ? 0x1000 : 0x20;
...
mmap(..., MAP_PRIVATE | MAP_ANON, ...)
```
(and put the consts at the top, not in the middle of a doc comment :P)

src/BizHawk.Common/OSTailoredCode.cs
```diff
@@ -12,6 +12,12 @@ public static class OSTailoredCode
 		public static readonly DistinctOS CurrentOS;
 		public static readonly bool IsUnixHost;
 
+		/// <summary>
+		/// The host OS' conventional file extension for dynamically-linked libraries, including the leading dot
+		/// (<c>".dll"</c> on Windows, <c>".dylib"</c> on macOS, <c>".so"</c> on Linux/BSD).
+		/// </summary>
+		public static readonly string DllExtension;
```
YoshiRulz: Name?
I can't find the prop/method which returns this in .NET Core, so maybe there isn't one, and I was misremembering Nix `stdenv.hostPlatform.extensions.sharedLibrary`.

.gitignore
```diff
@@ -48,3 +48,5 @@ UpgradeLog.htm
 /packages
 
 launchSettings.json
+Assets/EmuHawkMono_laststderr.txt
+Assets/EmuHawkMono_laststdout.txt
```
YoshiRulz: You must have ran a broken version of the launch script to create logs here.

src/BizHawk.Emulation.Cores/Waterbox/WaterboxHost.cs
```diff
@@ -75,7 +75,9 @@ public sealed class WaterboxHost : IMonitor, IImportResolver, IStatable, IDispos
 		static WaterboxHost()
 		{
 			NativeImpl = BizInvoker.GetInvoker<WaterboxHostNative>(
-				new DynamicLibraryImportResolver(OSTailoredCode.IsUnixHost ? "libwaterboxhost.so" : "waterboxhost.dll", hasLimitedLifetime: false),
+				new DynamicLibraryImportResolver(
+					(OSTailoredCode.IsUnixHost ? "libwaterboxhost" : "waterboxhost") + OSTailoredCode.DllExtension,
```
YoshiRulz: I'd expect most DLIR usages to be similarly broken. Can we have its ctor handle the logic instead of each individual caller? That would bring its behaviour closer to `[DllImport]` too.

YoshiRulz (commented):
HYPE

src/BizHawk.Client.EmuHawk/GraphicsImplementations/OpenGLProvider.cs
DarthMDev: Even ≤2.1 is unusable here: XQuartz's libGL aborts under Rosetta as soon as a context is created (the CGLSetCurrentContext crash), and nothing in BizHawk requests <3.2 (display wants 3.2, melonDS 3.2/4.3), so advertising 2.1 would only risk a crash with no functional gain;hence false. On a native Intel Mac (no Rosetta) 2.1 might actually work; I can gate the <=2.1 form on sysctl.proc_translated if you want, though it wouldn't change behaviour since no core asks for <3.2.

Assets/EmuHawkMonoMacOS.sh
DarthMDev: Both. XQuartz's GLX tops out at GL 2.1, and under Rosetta its libGL also aborts on init (CGLSetCurrentContext) ;a bug in Apple's GLX→CGL bridge under translation. EmuHawk needs ≥3.2 so it can't use it either way. --gdi is so the display path (IGL_OpenGL) doesn't attempt GL init and crash; OpenGLProvider covers the core GL path. Cleaner alternative if preferred: default DispMethod to GdiPlus on macOS in config rather than forcing the flag.

Dist/stage-macos-dylibs.sh
DarthMDev: The launcher already uses DYLD_LIBRARY_PATH/DYLD_FALLBACK_LIBRARY_PATH, so most deps need no symlink. The symlinks are only for genuine name mismatches a search path can't fix: code/Mono want libX11.so.6, liblua54.dylib, libe_sqlite3.dylib but Homebrew/XQuartz ship libX11.6.dylib, liblua5.4.dylib, libsqlite3.dylib. The shippable fix for those is Mono dllmap (*.dll.config <dllmap> entries); I'll try to get it working. The current symlinks are generated locally by the script (not shipped in the repo/release); a real release Package.sh should bundle the dylibs with corrected install_names.

Dist/stage-macos-dylibs.sh
DarthMDev: Agreed; once the symlink/dllmap approach in the thread is settled I'll fold this into Package.sh instead of a standalone script.

src/BizHawk.Emulation.Cores/Waterbox/WaterboxHost.cs
DarthMDev: Scoped to the macOS-touched caller here; happy to migrate the other ad-hoc callers (some irregular, e.g. libmgba.dll.so, libmupen64plus.so.2) in a follow-up or in this PR if you prefer.

DarthMDev:
Do we also want to wait for #4784 to be merged? And see how that plays out before this is merged?

Assets/EmuHawkMonoMacOS.sh
YoshiRulz: Two options for changing the default config:
https://github.com/TASEmulators/BizHawk/blob/0abf9aadf5ffbc758cbbb43f36d6b55537a0f15c/src/BizHawk.Client.Common/config/Config.cs#L267
```csharp
=> HostCapabilityDetector.HasD3D11
	? EDispMethod.D3D11
	: OSTailoredCode.CurrentOS is OSTailoredCode.DistinctOS.macOS ? EDispMethod.GDIPlus : EDispMethod.OpenGL;
```
```csharp
=> HostCapabilityDetector.HasD3D11
	? EDispMethod.D3D11
	: HostCapabilityDetector.HasOpenGL ? EDispMethod.OpenGL : EDispMethod.GDIPlus;
```
It sounds like the latter wouldn't be feasible.

DarthMDev:
ill fix those ci errors as soon as i can

YoshiRulz:
It's just an unused import, no rush

Dist/stage-macos-dylibs.sh
DarthMDev: dllmap turned out fragile here (Mono's own gdiplus dllmap loads the wrong libgdiplus via DYLD ordering), and it can't cover the dlopen-based loaders. The symlinks aren't shipped; they're generated on the user's machine by the setup step (to be folded into Package.sh), purely bridging name mismatches (libX11.so.6 ↔ libX11.6.dylib, etc.). I'm open to a better idea, but dllmap isn't reliable.

src/BizHawk.Emulation.Cores/Waterbox/WaterboxHost.cs
DarthMDev: Centralised in PlatformFileName and migrated the callers that follow [DllImport] naming (waterboxhost, encore). The others have non-conforming names (lib-prefix on Windows, .dll.so, versioned sonames) ; unifying them means renaming shipped blobs on every platform, which is a separate cross-platform cleanup I've kept out of this macOS PR to avoid Win/Linux churn.

Dist/stage-macos-dylibs.sh
YoshiRulz: I'm open to shipping a couple stable/slow-evolving libraries, but not dozens, so you'll have to use symlinks after all, and that will have to be done in `EmuHawkMono.sh` (`Package.sh` runs in CI as part of the packaging process). For reference, I [did a similar thing](https://github.com/TASEmulators/BizHawk/blob/0abf9aadf5ffbc758cbbb43f36d6b55537a0f15c/Dist/launch-scripts.nix#L55-L70) for NixOS.

PR #4786, opened 2026-06-27
Set TAStudio column sizes based on font size

### LINKED ISSUE #3048: TAStudio UI doesn't scale correctly on high-dpi monitors on Windows 10

### Summary
I use a 2560x1440 monitor and Windows 10. I'm running mamehawk but I'm told this issue also happens on other versions of BizHawk, because it's just a TAStudio issue.

By default, TAStudio looks like this:
image

but the font is too small for me to read on my monitor, so I increased the font size under config > set font. If I change the font size to something higher, it breaks the UI because the columns stay the same width but the font gets bigger.
image

### Host environment
- MameHawk (from this TASVideos thread: http://tasvideos.org/forum/viewtopic.php?t=21778)
- Windows 10 64-bit

SuuperW:
Fixes #3048 and other causes of column sizes being messed up.

All button columns with a single-character mnemonic use the same size, since that preserves the uniform look of the input roll for most buttons. Bigger columns are sized based on a measurement of text length (column text and max analog value length), including button columns.

I considered using `JsonIgnore` on `RollColumn.VerticalWidth` and `HorizontalHeight` since we now resize on loading a movie, but decided against that due to Lua columns. Lua columns can be resized now by re-running `tastudio.addcolum` though. (column text can also be updated too now because why not)

The cursor icon looks weird with large fonts. I don't know if it should just be centered, or if it should be scaled up to match the font size.

Warning: Running this branch will mess up your config file, so make a backup. All non-TAStudio `InputRoll` controls will be unusable on the master branch with the updated config file. (The `Width` property of `RollColumn` objects will be missing, resulting in it being 0 after deserialization when you go back to the master branch.)

Check if completed:


vadosnaprimer:
> Bigger columns are sized based on a measurement of text length (column text and max analog value length)

Min value may be bigger due to negative sign, or if some insane arcade machine has negative range that's outright bigger than positive.

> The cursor icon looks weird with large fonts. I don't know if it should just be centered, or if it should be scaled up to match the font size.

Scaling sounds better, but I'd need screenshots to know for sure.

SuuperW:
Fixed the minus sign thing. Idk why is was ever excluded, but this use of `MaxDigits` is the only use so it should be fine to change it to `MaxCharacters`.

Current frame cursor with font size 20:
image
Vertically centered:
image
Larger triangle manually drawn to the same scale factor as cell height:
image

Depending on how the triangle is drawn the big version might look slightly different. The space to the left of the triangle probably shouldn't be scaled as much.

EDIT: Actually the space to the left of the arrow can be used by branch number text.
image
image
So, width on this column should increase with font size.

It already looks bad when an arrow overlaps a 2-digit branch number, with the default font. I wonder if something should be done about that.

If I stand on the other side of the room, I can barely tell there's anything in the cursor column in the first two pictures but the text is still clear. So making the arrow bigger is probably good.

vadosnaprimer:
Yeah scaling the arrow and the column is the way to go (also those icons for states/markers).

SuuperW:
image

Scale by drawing new triangles when font changes.

Maybe should also delete the unused images. There are also some other tastudio images that were already unused.

vadosnaprimer:
Probably fine to nuke, just list them here explicitly.

src/BizHawk.Client.EmuHawk/Properties/Resources.cs
```diff
@@ -189,15 +190,6 @@ internal static class Resources
...
 		internal static readonly Bitmap TAStudio = ReadEmbeddedBitmap("TAStudio");
-		internal static readonly Bitmap icon_anchor = ReadEmbeddedBitmap("tastudio.icon_anchor");
-		internal static readonly Bitmap icon_anchor_lag = ReadEmbeddedBitmap("tastudio.icon_anchor_lag");
-		internal static readonly Bitmap icon_marker = ReadEmbeddedBitmap("tastudio.icon_marker");
-		internal static readonly Bitmap ts_h_arrow_blue = ReadEmbeddedBitmap("tastudio.ts_h_arrow_blue");
-		internal static readonly Bitmap ts_h_arrow_green = ReadEmbeddedBitmap("tastudio.ts_h_arrow_green");
-		internal static readonly Bitmap ts_h_arrow_green_blue = ReadEmbeddedBitmap("tastudio.ts_h_arrow_green_blue");
-		internal static readonly Bitmap ts_v_arrow_blue = ReadEmbeddedBitmap("tastudio.ts_v_arrow_blue");
-		internal static readonly Bitmap ts_v_arrow_green = ReadEmbeddedBitmap("tastudio.ts_v_arrow_green");
-		internal static readonly Bitmap ts_v_arrow_green_blue = ReadEmbeddedBitmap("tastudio.ts_v_arrow_green_blue");
```
YoshiRulz, line 200: Remove the images at the same time

SuuperW: ok

YoshiRulz: If you're removing all the unused files, these lines can go too:
https://github.com/TASEmulators/BizHawk/blob/b30494f03bc3056ab5a3c1a53d513cf229c3aed5/src/BizHawk.Client.EmuHawk/BizHawk.Client.EmuHawk.csproj#L41-L42

SuuperW: They did go. The went in the commit that removed those files, which is the last commit of this PR.

vadosnaprimer (approved):
Tested on Windows, works great!

PR #4800, opened 2026-07-23
[TAStudio] Add "Update+rename branch" button

### LINKED ISSUE #2254: TAStudio Feature Request: Branch Update and Change Label Button
I find myself making new branches all the time because it's faster than updating a branch and then separately changing the label. I would appreciate a single command or button to change a label and update the branch. Thank you!

SuuperW:
Seems to work fine without any modifications to the branch undo logic: you can undo the update with text to get the old branch back with its old text.

Icon is the update icon with a flipped pencil over it.
image

Looks a little strange to have the pencil the other way, but non-flipped pencil would cover the arrows of the update icon:
image

Check if completed:


vadosnaprimer:
What about

625422565-085d4185-aeee-41dc-9260-eff54d05c1ea

SuuperW:
I think if the arrows are flipped for this one they should also get flipped in the image for regular update.

vadosnaprimer:
I guess that's the lesser evil.

SuuperW:
I went back in the git history and found a higher-resolution reboot icon and was able to modify and downscale that into something that I think is acceptable.

image

vadosnaprimer (approved):

Issue #4769, opened 2026-06-09
Autofire controls override joypad.set


SuuperW:
### Summary
Autofire inputs override buttons being unheld by Lua's `joypad.set`.

### Repro
1. Open a ROM
2. In Config -> Controllers... -> Autofire Controls, bind a button to a keyboard key. For example, bind the "A" button to the "D" key.
3. Open the Lua Console
4. Ensure the game is paused
5. In the Lua Console's input box, set the button to false with joypad.set. For example, `joypad.set({A = false})`
6. Select the main form
7. Ensure input display is turned on (View -> Display Input)
8. Press the key configured for autofire. ("D" in this example)

### Output
Observe that the autofire button is held.

### Host env.
Bizhawk 2.11.1 and master branch, Windows 11

Morilli:
I assume this is because `joypad.set` only overrides the normal user controller, but the autofire one is separate and gets `or`d together with normal user input.

However the question here is: Is that even incorrect behavior? *Should* `joypad.set` be able to unset a button that is being activated by autofire? And should that apply to user-controller autofire, automatic autofire, or both?

SuuperW:
Autofire that is activated by a user input the same way regular buttons are activated by user input I think should definitely not be able to override `joypad.set`. In both cases the user is holding a physical button to perform some input, and I see no reason to handle them differently.

But what is automatic autofire?

Morilli:
> But what is automatic autofire?

Everything that is sticky that isn't from user input, aka what TAStudio uses, the virtual pad, or (currently) joypad.setanalog.

In general I'm just wondering what the expected priority of different input systems is and how they should interact with each other.

SuuperW:
For TAStudio's autofire, I think that should be the same as user holding a autofire button. But I hardly ever use it. And I never use virtual pad. Without knowing how they are actually used I would expect `joypad.set` to override both. A Lua script seems much more flexible than anything else, and someone who doesn't want buttons overridden would probably not specify their state in `joypad.set`.

Issue #4766, opened 2026-06-06
Lua unable to control DOSBox mouse


SergioMartin86:
I was trying to imitate this PR https://github.com/TASEmulators/BizHawk/pull/4748 using LUA so I can adjust the mouse position according to the game's situation, but encountered a bug.

### Summary

This LUA script breaks Bizhawk (DOSBox-x) mouse inputs:

```
while true do
	input = joypad.get()
	--print(input)
	newInput = {};
    newInput["Mouse Position Y"] = input["Mouse Position Y"] + 100
    joypad.setanalog(newInput)

	emu.frameadvance();
end
```

The expected behaviour is that it takes the mouse's Y position, and adds 100 to it, so that the in-game pointer is adjusted. Instead it breaks the input, making the Y position get stuck in the default forever.

### Repro

1. Open DOSBox core (any game will do) with mouse activated
2. Move the mouse around and see the inputs move accordingly
3. Now activate the LUA script above
4. Mouse Position Y freezes out (bug)
5.  Deactivate the LUA script
6. Bug still persists

### Output

<img width="699" height="449" alt="Image" src="https://github.com/user-attachments/assets/657a5f66-dc3a-4895-8e61-0b9d0a911781" />

Mouse position X moves freely, but Y is stuck at default.

### Host env.

Bizhawk 2.11.1; Windows 11, Intel/NVIDIA

YoshiRulz:
Probably a duplicate of #4535.

SergioMartin86:
See:

https://github.com/TASEmulators/BizHawk/pull/4748

For the correct way to achieve this using Lua.

Closing because as I don't consider it a bug, when a more appropriate way exists.

PR #4804, opened 2026-07-27
Input stuff

SuuperW:
This PR changes both `JoypadApi.SetAnalog` overloads, so that they behave similarly to the `Set` methods for buttons. This means (1) using `OverrideAdapter` instead of the sticky controller [fixes the Lua issue observed in #4766], (2) changing the way analog values are unset in the dictionary overload is by passing a dictionary without the relevant keys [allowing it to set neutral axis values], and (3) for Lua, throwing an exception if the input value cannot be converted to an int [instead of silently failing].

The old behavior is still available in a `[Obosolete]` overload. External tools wishing to make use of this fix will need to be updated for the new method signature. The C# method signature for Lua is identical, so the fixed version uses an updated name `set_analog`. Lua scripts using `setanalog` wishing to make use of this fix will need to be updated.

The documentation for the set methods are updated to include "as if the inputs came from the user" (same as the new SetAnalog). This implies that #4769 is indented behavior; auto-fire will obviously work when the button is not otherwise held. Sticky holds and autofires will also interact with buttons held by `joypad.set`. But this implication is probably not very obvious and better documentation may still be beneficial. Alternatively, we might want to change the behavior.

It also fixes the behavior of the set methods when called from the frame start event, adds tests, and cleans up some code a little bit.

Check if completed:

### DIFF
--- a/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
@@ -74,16 +74,23 @@ namespace BizHawk.Client.Common
 			}
 		}
 
-		public void SetAnalog(IReadOnlyDictionary<string, int?> controls, int? controller = null)
+		public void SetAnalog(IReadOnlyDictionary<string, int> controls, int? controller = null)
 		{
-			foreach (var (k, v) in controls) SetAnalog(k, v, controller);
+			// If a controller is specified, we need to iterate over unique button names. If not, we iterate over
+			// ALL button names with P{controller} prefixes
+			foreach (var axis in _inputManager.ActiveController.ToAxisControlNameList(controller))
+			{
+				SetAnalog(axis, controls.TryGetValue(axis, out var state) ? state : null, controller);
+			}
 		}
 
 		public void SetAnalog(string control, int? value = null, int? controller = null)
 		{
 			try
 			{
-				_inputManager.StickyHoldController.SetAxisHold(controller == null ? control : $"P{controller} {control}", value);
+				var axisToSet = controller == null ? control : $"P{controller} {control}";
+				if (value == null) _inputManager.OverrideAdapter.UnSetAxis(axisToSet);
+				else _inputManager.OverrideAdapter.SetAxis(axisToSet, value.Value);
 			}
 			catch
 			{
--- a/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
+++ b/src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
@@ -46,8 +46,8 @@ namespace BizHawk.Client.Common
 				LogCallback($"invalid mnemonic string: {inputLogEntry}");
 				return;
 			}
-			foreach (var button in controller.Definition.BoolButtons) _inputManager.ButtonOverrideAdapter.SetButton(button, controller.IsPressed(button));
-			foreach (var axis in controller.Definition.Axes.Keys) _inputManager.ButtonOverrideAdapter.SetAxis(axis, controller.AxisValue(axis));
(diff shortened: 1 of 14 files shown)


perfectchud:
I've tested this fix and it works perfectly. My mouse movements no longer override lua script's X/Y positions, and I don't have to manually unbind Touch X and Touch Y whenever I want to run a lua script that does touch inputs.
There is one issue that persists: my accidental left clicks still trigger unwanted inputs that would interfere with intended result of my lua script. But here's a workaround: force the script to do joypad.set({ ["Touch"] = false }) whenever the script isn't doing any Touch inputs.

src/BizHawk.Client.Common/Api/Classes/JoypadApi.cs
```diff
@@ -65,38 +65,32 @@ public void Set(string button, bool? state = null, int? controller = null)
...
+				else _inputManager.OverrideAdapter.SetButton(buttonToSet, state.Value);
 			}
 			catch
 			{
 				// ignored
 			}
 		}
 
-		public void SetAnalog(IReadOnlyDictionary<string, int?> controls, int? controller = null)
+		public void SetAnalog(IReadOnlyDictionary<string, int> controls, int? controller = null)
```
YoshiRulz, line 94: The implementation looks like it would work with `int?`, so can you leave the signature as-is until next release? Better yet, have both as overloads and mark the nullable one `[Obsolete]`.

src/BizHawk.Client.Common/lua/CommonLibs/JoypadLuaLibrary.cs
```diff
@@ -45,13 +45,13 @@ public void Set(LuaTable buttons, int? controller = null)
 		}
 
 		[LuaMethodExample("joypad.setanalog( { [ \"Tilt X\" ] = -63, [ \"Tilt Y\" ] = 127 } );")]
-		[LuaMethod("setanalog", "Sets the given analog controls to their provided values as autoholds. Set axes to the empty string to clear individual holds.")]
+		[LuaMethod("setanalog", "Sets the given analog controls to their provided values for the current frame, as if the inputs came from the user. Any analog inputs previously set but missing from the given table will be unset.")]
```
YoshiRulz: This also needs a deprecation cycle. Might need to resort to renaming it.

src/BizHawk.Tests.Client.Common/Api/JoypadApiTests.cs
```diff
@@ -0,0 +1,225 @@
...
+				inputContext.source.MakeReleaseEvent(button, 0);
+				inputContext.BasicInputProcessing();
+			}
+		}
+
+		private static void AssertDictMatches(IReadOnlyDictionary<string, object> expected, IReadOnlyDictionary<string, object> actual)
+		{
+			foreach (var kvp in expected)
+			{
+				Assert.IsTrue(actual.TryGetValue(kvp.Key, out object value), $"Expected to find key {kvp.Key} but did not.");
```
YoshiRulz, line 60:
```suggestion
				Assert.IsTrue(actual.TryGetValue(kvp.Key, out var value), $"Expected to find key {kvp.Key} but did not.");
```

src/BizHawk.Client.Common/lua/CommonLibs/JoypadLuaLibrary.cs
SuuperW: New method could be named `set_analog` since that's a good name change anyway.

src/BizHawk.Tests.Client.Common/Api/JoypadApiTests.cs
SuuperW: `var` is generally a bad idea, since it provides less information to the reader. Code is easier to read and understand when type names are visible.

SuuperW:
Force push, to keep old behavior in `[Obsolete]` methods.

PR #4812, opened 2026-08-10
Updated Vs. System Mapper

100thCoin:
Prevented array index out of bounds error, and corrected nametable mirroring.
 - If you attempted to read from VRAM address $3000 through $3EFF, instead of reading from a mirror of the nametable, bizhawk would crash.
- This crash can be reproduced in VS Super Mario Bros. using the attached .bk2 file.
[Crash.zip](https://github.com/user-attachments/files/30912359/Crash.zip)

Prevented the Vs. System from reading from $2004.
 - The RGB PPU found inside the Vs. System never had any revisions capable of reading from OAM DATA.

The RGB PPU would also be incapable of reading from Palette RAM, though I have not made that change.

Check if completed:

- [ ] I have run any relevant test suites.

### DIFF
--- a/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/Boards/VS_M99.cs
+++ b/src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/Boards/VS_M99.cs
@@ -66,7 +66,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.NES
 			}
 			else
 			{
-				addr -= 0x2000;
+				addr &= 0x0FFF;
 				if (addr<0x800)
 				{
 					return NES.CIRAM[addr];
@@ -87,7 +87,7 @@ namespace BizHawk.Emulation.Cores.Nintendo.NES
 			}
 			else
 			{
-				addr -= 0x2000;
+				addr &= 0x0FFF;
 				if (addr < 0x800)
 				{
 					NES.CIRAM[addr] = value;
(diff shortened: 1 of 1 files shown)


src/BizHawk.Emulation.Cores/Consoles/Nintendo/NES/Boards/VS_M99.cs
```diff
@@ -99,6 +99,30 @@ public override void WritePpu(int addr, byte value)
 			}
 		}
 
+		public override byte ReadReg2xxx(int addr)
+		{
+			if ((addr & 7) == 4)
```
YoshiRulz:
```suggestion
			if ((addr & 0b111) is 0b100)
```

100thCoin: The same should probably be done for the PeekReg2xxx equivalent.

100thCoin:
Is there any particular reason why you unchecked the "I have run any relevant test suites." box? I'm not aware of any test roms covering the nametable mirroring, though I did check the resulting behavior with a tracelogger to make sure the changes to $2004 reads were working as expected.

YoshiRulz:
There are no relevant test suites. (I looked into adding some, starting with your AccuracyCoin testrom, but struggled to get the compiler chain to work in Nix.)

YoshiRulz (approved):

PR #4847, opened 2026-09-19
IPC is now its own HostInputType and doesn't get swallowed

SetsukiFR:
_This is my very first pull request on GitHub ever. I apologize if I didn't respect an etiquette or did this incorrectly._

### Goal
Make IPC inputs function even when EmuHawk doesn't have focus, without requiring setting the 'listen to input even when in background' setting.

I made this originally for a custom "TwitchPlays" plugin. What I needed was for EmuHawk to listen to input from another process running, while in the background, and not listening to keyboard inputs (keyboard was being used by the streamer during the live).

### How
I added a new HostInputType for IPC, and separated the way these inputs are processed from regular keyboard controls.

### Potential issues

1. My main doubt is in `Input.cs ShouldSwallow()`, where I made IPC being swallowed under NO circumstances.

2. I wasn't sure wether to 

> set the value of HostInputType.IPC to 8
> replace the value of HostInputType.Ignore from 8 to 16

or, what I did

> set the value of HostInputType.IPC to 16

### No AI
No AI tools have been used for this (plus it's pretty simple)I hope this'll come in handy!
Thank you so much for your work and have a nice day!


src/BizHawk.Client.EmuHawk/Input/Input.cs
```diff
@@ -315,6 +320,9 @@ private void UpdateThreadProc()
 
 		private static bool ShouldSwallow(AllowInput allowInput, HostInputType inputFocus)
 		{
+			if (inputFocus == HostInputType.IPC)
+				return false;
+
 			return allowInput == AllowInput.None || (allowInput == AllowInput.OnlyController && inputFocus != HostInputType.Pad);
```
YoshiRulz: `(allowInput is AllowInput.None && inputFocus is not HostInputType.IPC) || ...`

SetsukiFR: Thank you very much ; I wasn't sure about all the ins and outs of the ShouldSwallow function, as well as allowInput. I have tested your change for my purposes and it still works perfectly.
Your code has been added in commit [57a5f1b](https://github.com/TASEmulators/BizHawk/pull/4847/commits/57a5f1b774abaca34ebc2ec9a706bf78e98e63b7)

